PR #1, opened 2016-09-02
First pass at django based API

kaitlin:
Includes the basic scaffolding for a django rest framework-based API. 

Currently it's just one endpoint with a scant model and a bit of mocked data is at
http://localhost:8000/api/v1/awards/

@cmc333333 would be grateful for you to take a look, it's been a while since I had to set up a Django project so my conventions may be a bit out of date :)


README.md
```diff
@@ -1 +1,118 @@
...
+1. Download the correct PostgreSQL installer for your operating system from [EnterpriseDB](http://www.enterprisedb.com/products-services-training/pgdownload) (we recommend PostgreSQL 9.4.x).
+2. Run the installer. As you proceed through the installation wizard, note your choices for port number, username, and password. You will need those when creating the broker's configuration file.
+
+More complete install documentation is available on the PostgreSQL [wiki](https://wiki.postgresql.org/wiki/Detailed_installation_guides).
+
+**Note:** If you're setting up the broker on Mac OSX, we recommend using [homebrew](http://brew.sh) to install PostgreSQL.
+
+### Install Python and Create Virtual Environment
+
+The broker's backend components currently run on Python 3.x. These instructions will walk you through the process of installing Python and creating a Python-based virtual environment to house the API. A virtual environment will isolate the broker software and its libraries from those running on your local system and prevent potential conflicts.
```
cmc333333: "The broker" is mentioned a few times in this doc, which I don't _think_ you want?

kaitlin: derp, nope. I cribbed heavily. Will update.

README.md
```diff
@@ -1 +1,118 @@
...
+    Operations to perform:
+    Apply all migrations: admin, auth, awards, contenttypes, sessions
+    Running migrations:
+    Applying awards.0001_initial... OK
+
+Once you've done that, you're ready to start the development server
+
+    $ python manage.py runserver
+
+This will run the application at `0.0.0.0:8000` by default. You can change that by passing the host and port number in to the `runserver` command.
```
cmc333333: Minor, but `127.0.0.1` is slightly different than `0.0.0.0` (the latter can be accessed from outside localhost, which is important for virtual machines). This should be the former.

usaspending_api/awards/tests.py
```diff
@@ -0,0 +1,17 @@
+from django.test import TestCase, Client
+
+class AwardTests(TestCase):
+
+    #load the test fixture
+    fixtures = ['awards']
```
cmc333333: Super minor, but I've never been a fan of loading fixtures from a file -- it requires the test readers look in multiple places to understand what's being tested. 

Out of scope for this PR, but something I'd recommend looking at: the sooner you get `py.test` in, the sooner you'll see it's benefits. It's completely compatible with unittest-style stuff, but I don't think you'll want to go back.

usaspending_api/awards/tests.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def setUp(self):
+        self.c = Client()
```
cmc333333: I believe django's `TestCase` comes with a `self.client`

usaspending_api/local_settings.example.py
```diff
@@ -0,0 +1,9 @@
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': 'usaspending_api',
+        'USER': 'usaspending_api',
+        'PASSWORD': '',
+    }
+}
```
cmc333333: Not a big issue, but something to consider earlier on -- I get the sense that the `local_settings` approach is now frowned on in favor of environment variables (in part due to the [12-factor push](https://12factor.net/)). In the database scenario, we tend to use [`dj-database-url`](https://github.com/kennethreitz/dj-database-url) on cloud.gov projects; it effectively expands a single database string into these four fields.

README.md
kaitlin: good catch

usaspending_api/awards/tests.py
kaitlin: cool. I will work on it this afternoon. Thanks!

usaspending_api/awards/tests.py
kaitlin: so it does!

usaspending_api/local_settings.example.py
kaitlin: yeah I had the vague feeling as I was writing it that it was going to date me. I'll work on updating this for this PR.

usaspending_api/settings.py
```diff
@@ -0,0 +1,135 @@
...
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/1.10/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = '&=pi7xiahctpu@=nsp9!9qec6k(jcn##)b@u&s*sx9%kp+m66o'
```
cmc333333: This'll throw warnings when/if you integrate [`bandit`](https://github.com/openstack/bandit) in the future. One approach that works relatively well is to generate a new secret key (e.g. via `django.utils.crypto.get_random_string`) whenever the app is restarted. The downside to this approach is that your user will get logged out.

usaspending_api/settings.py
```diff
@@ -0,0 +1,135 @@
...
+    'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'django.contrib.messages.middleware.MessageMiddleware',
+    'django.middleware.clickjacking.XFrameOptionsMiddleware',
+]
+
+ROOT_URLCONF = 'usaspending_api.urls'
+
+TEMPLATES = [
+    {
+        'BACKEND': 'django.template.backends.django.DjangoTemplates',
```
cmc333333: No doubt a style thing, but I _think_ the community's moving towards jinja2 for templates. If you prefer them, now'd be a good time to set the standard.

usaspending_api/local_settings.example.py
kaitlin: how do people generally manage this locally? autoenv or virtualenvwrapper post activate hooks?

cmc333333:
@kaitlin all good stuff! I threw in random comments, but I think this is good to merge whenever you're comfortable.

cmc333333:
Oh also, probably out of scope for this PR, but setting up automated tests, pep8 testing, and the like would all be good to do upfront, too

requirements.txt
```diff
@@ -0,0 +1,7 @@
+Django==1.10.1
+django-debug-toolbar==1.5
+django-filter==0.14.0
+djangorestframework==3.4.6
+Markdown==2.6.6
+psycopg2==2.6.2
+sqlparse==0.2.1
```
cmc333333: :clap: :clap: :clap: for pinning your dependencies

usaspending_api/local_settings.example.py
cmc333333: I think autoenv is a good approach, but I'm definitely on poor footing here and don't practice what I'm preaching 😄

usaspending_api/settings.py
kaitlin: updated

usaspending_api/settings.py
kaitlin: yeah I think I'll leave for now since the only templates we're using are the ones from DRF.

PR #8, opened 2016-09-07
Aj ds 330 first reflection

alexjajabah:
@afrasier

### DIFF
--- a/reflected_models.py
+++ b/reflected_models.py
@@ -9,7 +9,6 @@ from __future__ import unicode_literals
 
 from django.db import models
 
-#aj comment
 class AdtAgency(models.Model):
     adt_agency_id = models.AutoField(primary_key=True)
     agency_id = models.IntegerField(blank=True, null=True)
@@ -970,7 +969,7 @@ class AdtUsers(models.Model):
         managed = False
         db_table = 'adt_users'
 
-
+#AJ START
 class Agency(models.Model):
     agency_id = models.AutoField(primary_key=True)
     location = models.ForeignKey('Location', models.DO_NOTHING)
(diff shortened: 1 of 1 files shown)


reflected_models.py
```diff
@@ -970,25 +969,26 @@ class Meta:
         managed = False
         db_table = 'adt_users'
 
-
+#AJ 09/07/2016...moved foreign key to the top
```
afrasier: For future commits, try to always put a space between your # and the text of the comment. That's the typical convention in Python. i.e. instead of ```#AJ``` do ```# AJ```

reflected_models.py
```diff
@@ -2229,109 +2269,111 @@ class SubmissionProcess(models.Model):
...
+#    update_user_id = models.CharField(max_length=50, blank=True, null=True)
+#
+#    class Meta:
+#        managed = False
+#        db_table = 'test_treasury_appropriation_account_2'
+
+#AJ 09/07/2016...For treasury_account_identifier, changed AutoField to IntegerField
 class TreasuryAppropriationAccount(models.Model):
-    treasury_account_identifier = models.AutoField(primary_key=True)
+    treasury_account_identifier = models.IntegerField(primary_key=True)
```
afrasier: Are the other AutoFields fine, or are Bernard and Anand looking into those?

reflected_models.py
alexjajabah: Sure will do for next time

reflected_models.py
alexjajabah: Yes we looked at others one, and that's the only one

PR #9, opened 2016-09-07
Aj ds 330 first reflection

alexjajabah:
@afrasier


usaspending_api/financial_activities/models.py
```diff
@@ -1,3 +1,185 @@
 from django.db import models
 
 # Create your models here.
+
+# AJ 09/07/2016...Added 5 tables below for financial_activities
+
+class TreasuryAppropriationAccount(models.Model):
```
afrasier: The top two models here are already available in /accounts/ - we should probably just use:

``` from usaspending_api.accounts.models import TreasuryAppropriationAccount ``` 

And likewise for Appropriation account balances.

PR #16, opened 2016-09-21
Updates to Award model changes

kaitlin:
This PR includes:

- updates to the award related models
- updates to the contract loader
- fixes tests
- requires migrations to be run!!


afrasier:
Looks like the tests in `usaspending_api/financial_activities/test_financial_activities.py` are failing - is this something you want me to fix after merge or do you want to fix it on this PR?

kaitlin:
@afrasier So, the fixture loading actually works with the changes, but model mommy is doing this weird thing where it ignores AutoFields (such as the one in `RefProgramActivity`) and tries to insert a program activity object without a primary key. If I change `ref_program_activity_id` to `IntegerField` the tests pass. I think it should be fine if I leave the `primary_key` attribute in there. 

However, for some reason there's a migration conflict again on the PR test. Looking into that now.

kaitlin:
Ok @afrasier there was a migration merge that needed to happen in the awards app after I pulled in your upstream changes. Everything passes now so please feel free to give it a code review.

PR #17, opened 2016-09-26
Ds 339 django etl aj

playmakervi:
Updated AwardAction model
Updated Award Object model
Updated Procurement model
Updated Location model
Updated Legal Entity model

@afrasier


usaspending_api/awards/models.py
```diff
@@ -88,8 +88,11 @@ class Award(models.Model):
         ('L', 'Loan'),
     )
 
-    award_identifier = models.CharField(unique=True, max_length=50)
```
afrasier: Why was this removed? I believe we were going to have this field be either the piid/fain/uri based upon the award type rather than having separate fields.

PR #19, opened 2016-09-28
DS 339 Django ETL

afrasier:
This PR completes the story DS 339.

ETL works for File A, File B, File C, File D2

ETL tests are implemented by using a spoofed database cursor which overrides the execute method and returns appropriate testing data. You can easily extend this by adding another entry to the dictionary for the SQL query you want to hijack. (For the time being, until the ETL is replaced in the future via API from the broker)

This also pulls in several commits from Alex and Bernard.

Tagging @kaitlin for review into master. If you don’t have time let me know and I’ll self review. :)


usaspending_api/awards/serializers.py
```diff
@@ -1,6 +1,7 @@
 from rest_framework import serializers
 from usaspending_api.awards.models import *
 from usaspending_api.accounts.serializers import AppropriationAccountBalancesSerializer
+from usaspending_api.references.serializers import *
```
kaitlin: I personally like see the imported classes listed explicitly, just for readability. But it's just a style thing.

usaspending_api/references/.idea/misc.xml
```diff
@@ -0,0 +1,14 @@
...
+    <OptionsSetting value="true" id="Remove" />
+    <OptionsSetting value="true" id="Checkout" />
+    <OptionsSetting value="true" id="Update" />
+    <OptionsSetting value="true" id="Status" />
+    <OptionsSetting value="true" id="Edit" />
+    <ConfirmationsSetting value="0" id="Add" />
+    <ConfirmationsSetting value="0" id="Remove" />
+  </component>
+  <component name="ProjectRootManager" version="2" project-jdk-name="Python 3.5.2 (C:\Users\542127\AppData\Local\Programs\Python\Python35-32\python.exe)" project-jdk-type="Python SDK" />
+</project>
```
kaitlin: I think this whole `.idea` folder needs to be removed

usaspending_api/references/management/commands/load_agencies.py
```diff
@@ -0,0 +1,43 @@
+from django.core.management.base import BaseCommand, CommandError
+from usaspending_api.references.models import Agency, SubtierAgency
```
kaitlin: I think we deleted `Subtieragency` so this import would throw an error

kaitlin: I guess this whole loader needs to be updated. I'm ok with this not working atm, though since it's not really in scope of the sprint.

usaspending_api/submissions/management/commands/load_submission.py
```diff
@@ -0,0 +1,409 @@
...
+from datetime import datetime
+import logging
+import django
+import os
+
+from usaspending_api.submissions.models import *
+from usaspending_api.accounts.models import *
+from usaspending_api.financial_activities.models import *
+from usaspending_api.awards.models import *
+from usaspending_api.references.models import *
```
kaitlin: Same comment as above, I think it's helpful to be explicit with the imports.

usaspending_api/submissions/management/commands/load_submission.py
```diff
@@ -0,0 +1,409 @@
...
+                # self.logger.info('Submission id ' + str(submission_id) + ' already exists. Records will be updated.')
+                update = True
+        except ObjectDoesNotExist:
+            submission_attributes = SubmissionAttributes()
+
+        # Update and save submission attributes
+        # Create our value map - specific data to load
+        value_map = {
+            'user_id': 1,
+            'update_date': timezone.now()
```
kaitlin: I think @playmakervi and @alexjajabah went through and updated the `create_date` and `update_fields` to use Django's auto date fields. If the models using this dict already have that, we shouldn't need to set the create or update explicitly - Django will do it on `model.save()` for us.

usaspending_api/submissions/management/commands/load_submission.py
```diff
@@ -0,0 +1,409 @@
...
+            except ObjectDoesNotExist:
+                treasury_account = TreasuryAppropriationAccount()
+
+                field_map = {
+                    'tas_rendering_label': 'tas',
+                    'allocation_transfer_agency_id': 'allocation_transfer_agency',
+                    'responsible_agency_id': 'agency_identifier'
+                }
+
+                value_map = {
```
kaitlin: Same comment as above

usaspending_api/submissions/management/commands/load_submission.py
```diff
@@ -0,0 +1,409 @@
...
+            load_data_into_model(afd_trans, row, value_map=value_map, save=True)
+
+        # File D2
+        db_cursor.execute('SELECT * FROM award_financial_assistance WHERE submission_id = %s', [submission_id])
+        award_financial_assistance_data = dictfetchall(db_cursor)
+        # self.logger.info('Acquired award financial assistance data for ' + str(submission_id) + ', there are ' + str(len(award_financial_assistance_data)) + ' rows.')
+
+        for row in award_financial_assistance_data:
+            # Create LegalEntity
+            legal_entity_location_field_map = {
```
kaitlin: if this dict doesn't include any variables as values, should it exist outside the for loop?

usaspending_api/submissions/management/commands/load_submission.py
```diff
@@ -0,0 +1,409 @@
...
+        # File D1
+        db_cursor.execute('SELECT * FROM award_procurement WHERE submission_id = %s', [submission_id])
+        procurement_data = dictfetchall(db_cursor)
+        # self.logger.info('Acquired award procurement data for ' + str(submission_id) + ', there are ' + str(len(procurement_data)) + ' rows.')
+
+        for row in procurement_data:
+            pass
+
+
+# Loads data into a model instance
```
kaitlin: noice docstring

usaspending_api/submissions/test_load_submission.py
```diff
@@ -0,0 +1,43 @@
...
+
+import pytest
+import os
+
+from model_mommy import mommy
+
+
+class LoadSubmissionTest(TestCase):
+    def setUp(self):
+        SubmissionAttributes.objects.all().delete()
```
kaitlin: Don't these get deleted when the test db gets trashed at the end of the test run?

usaspending_api/references/.idea/misc.xml
afrasier: It should be, I thought I took care of it. I think it's because I added it to the git ignore in the same commit I deleted them.

usaspending_api/references/management/commands/load_agencies.py
afrasier: Ah, yeah, this didn't throw any errors because there's no test for it I believe.

usaspending_api/submissions/management/commands/load_submission.py
afrasier: Wasn't aware of that, I'll check and update.

usaspending_api/submissions/test_load_submission.py
afrasier: I wanted to make sure everything was gone before loading to ensure that it has the appropriate number for the assert statements

usaspending_api/submissions/management/commands/load_submission.py
afrasier: It should have been, my bad.

PR #20, opened 2016-09-29
Ds 339 django etl kpj

kjacksonkearneyco:
Added sub_award and place_of_performance objects to award models.py. Updated model.py by removing create_user_id and update_user_id from all of the object models and updated create_date and update_date to auto populate Django fields in all of the object models.

@afrasier


usaspending_api/.idea/misc.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
afrasier: All these .idea files should not be committed. Add .idea/* to the .gitignore file and remove these from the PR

usaspending_api/awards/models.py
```diff
@@ -271,3 +275,46 @@ class FinancialAssistanceAward(AwardAction):
     class Meta:
         managed = True
         db_table = 'financial_assistance_award'
+
+class SubAward(models.Model):
```
afrasier: @kaitlin Did we want to have a separate sub-award model, or just use the regular award model and have it link to other awards as a parent?

usaspending_api/awards/models.py
```diff
@@ -271,3 +275,46 @@ class FinancialAssistanceAward(AwardAction):
...
+    create_date = models.DateTimeField(auto_now_add=True, blank=True, null=True)
+    update_date = models.DateTimeField(auto_now=True, null=True)
+
+# AJ 09/07/2016...changed managed to TRUE
+    class Meta:
+        managed = True
+        db_table = 'sub_award'
+
+class PlaceOfPerformanceRelationship(models.Model):
+    place_of_performance_relationship_id = models.AutoField(primary_key=True)
```
afrasier: Is it necessary for this to be its own model? Wouldn't it be simpler to have place_of_performance be a foreignkey field in the Award model pointing to a location?

usaspending_api/awards/models.py
kaitlin: hmmm, it seems like it should be able to use the regular award model.

PR #25, opened 2016-10-04
Ds 346 terselabel update

playmakervi:
Updated the following with the 60 character label and prefixed derived data elements with "drv":

financial_activities/model.py
 awards/model.py
 accounts/model.py

Removed white spaces based on test of python code.

@afrasier


usaspending_api/awards/models.py
```diff
@@ -5,53 +5,53 @@
 from django.db.models import F, Sum
 
 
-# Model Objects
+# Model Objects added white spaces
 class FinancialAccountsByAwards(models.Model):
     financial_accounts_by_awards_id = models.AutoField(primary_key=True)
-    appropriation_account_balances = models.ForeignKey(AppropriationAccountBalances, models.CASCADE)
+    appropriation_account_balances = models.ForeignKey(AppropriationAccountBalances, models.DO_NOTHING)
```
afrasier: Same question - why not cascade?

PR #27, opened 2016-10-06
Ds 355 routing parameters

afrasier:
@kaitlin review at your leisure


usaspending_api/awards/urls.py
```diff
@@ -1,7 +1,23 @@
...
+    url(r'^piid/(?P<piid>\w+)', views.AwardList.as_view())
+]
+
+award_summary_id_patterns = [
+    url(r'^$', views.AwardListSummary.as_view()),
+    url(r'^uri/(?P<uri>.+)', views.AwardListSummary.as_view()),
+    url(r'^fain/(?P<fain>.+)', views.AwardListSummary.as_view()),
+    url(r'^piid/(?P<piid>.+)', views.AwardListSummary.as_view()),
+    url(r'^fy/(?P<fy>[0-9]{4})', views.AwardListSummary.as_view()),
+    url(r'^agency/(?P<agency>.{4})', views.AwardListSummary.as_view())
```
kaitlin: So are there different endpoints for filtering on FY and agency? The way the story is written I would have expected to be able to filter on both in the same endpoint.

usaspending_api/submissions/management/commands/load_submission.py
```diff
@@ -268,16 +268,15 @@ def handle(self, *args, **options):
...
-                "awarding_agency_id": "awarding_agency_code",
-                "funding_agency_id": "funding_agency_code",
                 "description": "award_description",
                 "type": "assistance_type"
             }
 
             award_value_map = {
                 "period_of_performance_start_date": format_date(row['period_of_performance_star']),
                 "period_of_performance_current_end_date": format_date(row['period_of_performance_curr']),
+                "awarding_agency": Agency.objects.filter(pk=row["awarding_agency_code"]).first(),
```
kaitlin: I think the agency primary key was changed from CGAC to an autoincrementing ID. So this filter would need to change to querying by CGAC. Although, looking at the updates to the model, it seems like it was changed to an autoincrementing id but the CGAC field was not added to replace it. Also the comment was not updated, meaning it's suuuper confusing :). I can push a commit to update the agency model in a few minutes.

afrasier: This has been updated!

usaspending_api/awards/urls.py
afrasier: I've changed this so you can specify url/?fy=2016 or ?agency=FPDScode or ?agency=FPDScode&fy=2016 on any award summary request

PR #35, opened 2016-10-18
DS 365 D2 Implementation

afrasier:


afrasier:
@kaitlin at your liesure

usaspending_api/awards/serializers.py
```diff
@@ -22,11 +22,27 @@ class Meta:
...
+        model = FinancialAssistanceAward
+        fields = '__all__'
+
+
 class AwardSerializer(serializers.ModelSerializer):
 
     recipient = LegalEntitySerializer(read_only=True)
     awarding_agency = AgencySerializer(read_only=True)
     funding_agency = AgencySerializer(read_only=True)
+    procurement_set = ProcurementSerializer(many=True, read_only=True)
```
kaitlin: should there be logic in the serializer that only shows procurement if it's a procurement award and vice versa for financial assistance? Or if these are null will the key not show up?

usaspending_api/submissions/management/commands/load_submission.py
```diff
@@ -289,6 +289,15 @@ def handle(self, *args, **options):
...
             award, created = Award.objects.get_or_create(**award)
 
+            fad_value_map = {
+                "award": award,
+                "submission": submission_attributes
+
+            }
+
+            financial_assistance_data = load_data_into_model(FinancialAssistanceAward(), row, value_map=fad_value_map, as_dict=True)
+            fad, created = FinancialAssistanceAward.objects.get_or_create(**financial_assistance_data)
```
kaitlin: This could potentially make more financial assistance award objects than we want because it's checking for a match on all of the fields. I think we might need a way to update a record if this matches the award ID and mod number, rather than creating a new record.

afrasier: We can add award ID and mod # as unique together, which should prevent any accidental duplication. Additionally, I'll update this to check for an existing pair manually beforehand.

Shouldn't all data be new data if the submission ID hasn't been in the data store before? Or is it possible one line of D2 is updated by a later line in the same D2 from Broker?

usaspending_api/awards/serializers.py
afrasier: If it is empty it will be an empty list, but I can add a check to only drop it in if it exists.

usaspending_api/submissions/management/commands/load_submission.py
kaitlin: 🤔 Hmmm so, it could definitely exist from the usaspending loader, because that data is updated more frequently and on an ongoing/rolling basis, whereas the broker submissions are quarterly. Not sure if it will appear in a previous broker submission but since both loaders use the same tables, I think my comment still stands.

usaspending_api/awards/serializers.py
afrasier: So since this is just on summary awards it will need to have both listed in the serializer, but on the specific view for procurement and financial assistance awards it won't have such a set.

PR #38, opened 2016-10-21
Added LoadTas for Tas CSV file

alexjajabah:
Added LoadTas for Tas CSV file


alexjajabah:
@afrasier. I added a test case for tas loader

PR #39, opened 2016-10-25
Arbitrary Query Filters

afrasier:
* Added post method to awards/summary/
* You can now specify any valid field as a get or post field parameter
* POST requests can now have any combination of filters via OR and AND. User friendly documentation forthcoming


usaspending_api/awards/views.py
```diff
@@ -33,69 +34,70 @@ def get(self, request, uri=None, piid=None, fain=None, format=None):
 
 
 class AwardListSummary(APIView):
+    def post(self, request, format=None):
+        fg = FilterGenerator()
+        body_unicode = request.body.decode('utf-8')
+        body = json.loads(body_unicode)
```
kaitlin: Can you add exception handling for a poorly formed request object?

usaspending_api/common/api_request_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    # 'filters' set in that object of filters to combine
+    # The combination method for filters at the root level is 'AND'
+    # Available operations are equals, less_than, greater_than, contains, in, less_than_or_equal, greather_than_or_equal, range
+    # Note that contains is always case insensitive
+    def create_from_post(self, parameters):
+        return self.create_q_from_filter_list(parameters.get('filters', []))
+
+    def create_q_from_filter_list(self, filter_list, combine_method='AND'):
+        q_object = Q()
+        for filt in filter_list:
```
kaitlin: what are your thoughts on how we should sanitize the input for different filters? Perhaps for now we could raise a 400 somewhere in this block and in the next sprint try and surface some detailed info about which filter or parameter value was malformed.

afrasier: Yeah, sanity checks for operators are pretty easy (i.e. for __range your value must have a list of length 2) but as far as a malformed field name that can be a bit trickier. For now I'll implement the value/operator validations and 400 if the fieldname is somehow bad.

afrasier:
@kaitlin I've added a validator for post requests that will spit out 400 responses with an error message if the request is malformed or uses an invalid operator, or uses a bad value for certain operators. Added tests for those (though I need to up the robustness of the tests in the future)

If they specify a field that doesn't exist (or something related, like a foreign key lookup) it will spit out the standard django error about it

Let me know what else needs doing, or go ahead and approve+merge :) Thanks! 👍

PR #40, opened 2016-10-26
Ds 393 cfda data loader

playmakervi:
Added CFDA Loader
Added CFDA Test


playmakervi:
@afrasier

afrasier:
@playmakervi Looks like there's a conflicting migration - the fields 'archived_date' and 'published_date' are trying to be set to a DateField and TextField simultaneously, I can fix this for you, I just need to know which one it should actually be

playmakervi:
@afrasier It should be TextField.

PR #42, opened 2016-10-30
Threaded data loader

afrasier:
For small files the overhead for spawning threads means it's slightly slower, but for larger files it overtakes on efficiency.

I've currently updated the contracts_loader to use the new system and can do so for the other loaders in future sprints.


usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -19,53 +18,28 @@ def add_arguments(self, parser):
...
+            "federal_action_obligation": "dollarsobligated",
+            "description": "descriptionofcontractrequirement",
+            "modification_number": "modnumber"
+        }
+
+        value_map = {
+            "award": lambda row: Award.objects.get_or_create(piid=row['piid'], type='C')[0],
+            "recipient": lambda row: LegalEntity.objects.get_or_create(recipient_name=row['dunsnumber'])[0],
+            "awarding_agency": lambda row: Agency.objects.get(subtier_code=self.get_agency_code(row['maj_agency_cat'])),
+            "action_date": lambda row: datetime.strptime(row['signeddate'], '%m/%d/%Y').strftime('%Y-%m-%d'),
```
kaitlin: maybe just call `convert_date` from line 38?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+        self.pre_row_function = pre_row_function
+        self.post_row_function = post_row_function
+        self.post_process_function = post_process_function
+        self.fields = [field.name for field in self.model_class._meta.get_fields()]
+
+    # Loads data from a file using parameters set during creation of the loader
+    # The filepath parameter should be the string location of the file for use with open()
+    def load_from_file(self, filepath):
+        start_time = time.time()
+        print('Started processing file ' + filepath + ' at ' + str(start_time))
```
kaitlin: can you use the logger instead of `print`?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+
+class DataLoaderThread(threading.Thread):
+    def __init__(self, name, data_queue, loader):
+        threading.Thread.__init__(self)
+        self.name = name
+        self.data_queue = data_queue
+        self.loader = loader
+
+    def run(self):
+        print("Starting " + self.name)
```
kaitlin: Use the logger instead

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+        self.process_data()
+        print("Exiting " + self.name)
```
kaitlin: Use the logger instead

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+                    # If it's a long label, we need to look it up to map to data
+                    if model_collision_field in settings.LONG_TO_TERSE_LABELS:
+                        data_collision_field = settings.LONG_TO_TERSE_LABELS[model_collision_field]
+                    query_object = Q({model_collision_field: row[data_collision_field]})
+                    collision_instance = self.loader.model_class.objects.filter(query_object).first()
+
+                    if collision_instance is not None:
+                        behavior = self.loader.collision_behavior
+                        if behavior is "delete":  # Delete the row from the data store and load new row
+                            collision_instnace.delete()
```
kaitlin: typo in variable name. Are there tests for each collision_instance value?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+                        if behavior is "update":  # Update the row in the data store with new values from the CSV
+                            update = True
+                        if behavior is "skip":  # Skip the row
+                            continue
+                        if behavior is "skip_and_complain":  # Log a warning and skip the row
+                            print('Hit a collision on row %s' % (row))
```
kaitlin: Use the logger instead

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+                                                              self.loader.value_map,
+                                                              row)
+                except Exception as e:
+                    print(e)
+                # If we have a post row function, run it before saving
+                if self.loader.post_row_function is not None:
+                    self.loader.post_row_function(row=row, instance=model_instance)
+
+                model_instance.save()
+            time.sleep(1)
```
kaitlin: what's this for?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+
+        for thread in pool:
+            thread.join()
+
+        if self.post_process_function is not None:
+            self.post_process_function()
+
+        end_time = time.time()
+        elapsed = end_time - start_time
+        print('Finished processing at ' + str(end_time) + '; Elapsed: ' + str(elapsed))
```
kaitlin: Use the logger instead

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.exit_flag = True
+
+        for thread in pool:
+            thread.join()
+
+        if self.post_process_function is not None:
+            self.post_process_function()
+
+        end_time = time.time()
```
kaitlin: it seems like these should only be executed if a debug or profiling tag is turned on somewhere

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+                            update = True
+                        if behavior is "skip":  # Skip the row
+                            continue
+                        if behavior is "skip_and_complain":  # Log a warning and skip the row
+                            print('Hit a collision on row %s' % (row))
+                        if behavior is "die":  # Raise an exception and cease execution
+                            raise Exception("Hit collision on row %s" % (row))
+
+                if not update:
+                    model_instance = self.loader.model_class()
```
kaitlin: it seems like there's an implicit behavior here: overwrite, instead of update. Should that be explicit or should we note somewhere that it's a default in certain cases?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+    #                           update - Update the row in the data store with new values from the CSV
+    #                           skip - Skip the row
+    #                           skip_and_complain - Log a warning and skip the row
+    #                           die - Raise an exception and cease execution
+    #   post_row_function - A function to call after every row. Should support parameters of row and instance
+    #                       i.e. def do_this_after_a_row(row=None, instance=None)
+    #   pre_row_function - Like post_row_function, but before the model class is updated
+    #   post_process_function - A function to call when all rows have been processed, uses the same
+    #                           function parameters as post_row_function
+    def __init__(self, model_class, threads=10, field_map={}, value_map={}, collision_field=None, collision_behavior='delete', pre_row_function=None, post_row_function=None, post_process_function=None):
```
kaitlin: it seems like `update` should be the default for `collision_behavior`. Any particular reason it's `delete`?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+                            continue
+                        if behavior is "skip_and_complain":  # Log a warning and skip the row
+                            print('Hit a collision on row %s' % (row))
+                        if behavior is "die":  # Raise an exception and cease execution
+                            raise Exception("Hit collision on row %s" % (row))
+
+                if not update:
+                    model_instance = self.loader.model_class()
+                else:
+                    model_instance = collision_instance
```
kaitlin: it seems like this is also a kind of implicit behavior, maybe `keep` ? When does this get executed if `delete` is the default in the function parameter?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+                    self.loader.pre_row_function(row=row, instance=model_instance)
+
+                try:
+                    processed_row = self.load_data_into_model(model_instance,
+                                                              self.loader.fields,
+                                                              self.loader.field_map,
+                                                              self.loader.value_map,
+                                                              row)
+                except Exception as e:
+                    print(e)
```
kaitlin: Use the logger instead

using_the_api.md
```diff
@@ -0,0 +1,344 @@
+# The DATA Act Data Store Application Programming Interface (API)
```
kaitlin: Let's just call this The USAspending API since nobody outside of this team knows what the data store is :)

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+        self.collision_behavior = collision_behavior
+        self.pre_row_function = pre_row_function
+        self.post_row_function = post_row_function
+        self.post_process_function = post_process_function
+        self.fields = [field.name for field in self.model_class._meta.get_fields()]
+
+    # Loads data from a file using parameters set during creation of the loader
+    # The filepath parameter should be the string location of the file for use with open()
+    def load_from_file(self, filepath):
+        start_time = time.time()
```
kaitlin: like below, this benchmarking seems like something that should only execute if a debug is on or some other flag is set.

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+            thread = DataLoaderThread("Thread-" + str(len(pool)), row_queue, self)
+            thread.start()
+            pool.append(thread)
+
+        with open(filepath) as csvfile:
+            reader = csv.DictReader(csvfile)
+            for row in reader:
+                row_queue.put(row)
+
+        while not row_queue.empty():
```
kaitlin: Is this to just pause here while the threads are working?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+class DataLoaderThread(threading.Thread):
+    def __init__(self, name, data_queue, loader):
+        threading.Thread.__init__(self)
+        self.name = name
+        self.data_queue = data_queue
+        self.loader = loader
+
+    def run(self):
```
kaitlin: Is `run` a function that all objects that inherit from Thread are supposed to have? And does it get called when you call `start` on the thread? (totally asking out of my own ignorance -- haven't needed to do much thread programming).

usaspending_api/common/threaded_data_loader.py
afrasier: No reason - I'll switch it to update

usaspending_api/common/threaded_data_loader.py
afrasier: Yeah this was leftover, I meant to remove it.

usaspending_api/common/threaded_data_loader.py
afrasier: Yes, otherwise it will signal all threads to exit immediately which would not be so good :)

usaspending_api/common/threaded_data_loader.py
afrasier: Yup! It's a pretty common set-up across most languages. Classes implementing threading usually have a run() function which is executed upon the thread object being .start()ed

usaspending_api/common/threaded_data_loader.py
afrasier: Not yet - I am working on that. I have to make some tiny data sets to validate that values are actually updated.

usaspending_api/common/threaded_data_loader.py
afrasier: The only time update is true is when the collision method is 'update' and there was a collision - there's no actual overwriting here. Either the model_instance to load data into is the collision_instance (in the case of an update) or it's brand new. Every other case either: skips the row, deletes the collision, or throws an error and dies. 

(Although this did make me notice I skipped a 'continue' statement)

If this is still unclear let me know. Since we're switching to the default collision_method being update, it should be a bit clearer

usaspending_api/common/threaded_data_loader.py
afrasier: See above comment.

usaspending_api/common/threaded_data_loader.py
afrasier: To prevent thread lock when accessing self.load.exit_flag (without it you might run into a situation where a single thread is constantly checking that variable and holding priority)

PR #47, opened 2016-11-04
Proper multiprocessing

afrasier:
* python doesn’t do parallel threading
* switched to multiprocessing


usaspending_api/common/threaded_data_loader.py
```diff
@@ -34,10 +36,13 @@ class ThreadedDataLoader():
     #   pre_row_function - Like post_row_function, but before the model class is updated
     #   post_process_function - A function to call when all rows have been processed, uses the same
     #                           function parameters as post_row_function
-    def __init__(self, model_class, threads=10, field_map={}, value_map={}, collision_field=None, collision_behavior='update', pre_row_function=None, post_row_function=None, post_process_function=None):
+    def __init__(self, model_class, processes=-1, field_map={}, value_map={}, collision_field=None, collision_behavior='update', pre_row_function=None, post_row_function=None, post_process_function=None):
```
cmc333333: Small detail, but would it make sense to use `None` instead? Magic numbers are always confusing.

Also, unrelated to this PR, but using objects like `{}` and `[]` as default parameters leads to some unexpected problems. Those are initialized _once_ (at function creation time, not call time), which means they are shared across function calls.

usaspending_api/common/threaded_data_loader.py
```diff
@@ -52,117 +57,129 @@ def __init__(self, model_class, threads=10, field_map={}, value_map={}, collisio
...
+            "collision_field": self.collision_field,
+            "collision_behavior": self.collision_behavior,
+            "logger": self.logger,
+            "pre_row_function": self.pre_row_function,
+            "post_row_function": self.post_row_function,
+            "fields": self.fields.copy()
+        }
+
+        # Spawn our processes
+        db.connections.close_all()
```
cmc333333: Curious: why do we need to close all the DB connections? Are they being shared across processes? Are they not thread safe?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -52,117 +57,129 @@ def __init__(self, model_class, threads=10, field_map={}, value_map={}, collisio
...
         pool = []
-        while len(pool) < self.threads:
-            thread = DataLoaderThread("Thread-" + str(len(pool)), row_queue, self)
-            thread.start()
-            pool.append(thread)
+        while len(pool) < self.processes:
+            process = DataLoaderThread("Process-" + str(len(pool)), self.model_class, row_queue, self.field_map.copy(), self.value_map.copy(), references)
+            pool.append(process)
```
cmc333333: Small detail, but this might be a wee bit simpler:
```python
pool = [DataLoaderThread('Process-' + str(i), self.model_class, row_queue, ...)
         for i in range(self.processes)]
```

usaspending_api/common/threaded_data_loader.py
```diff
@@ -52,117 +57,129 @@ def __init__(self, model_class, threads=10, field_map={}, value_map={}, collisio
...
-        while not row_queue.empty():
-            pass
+        for i in range(self.processes):
+            row_queue.put(None)
 
-        self.exit_flag = True
-
-        for thread in pool:
-            thread.join()
+        row_queue.join()
```
cmc333333: Something important to test is what happens when an exception is raised in the subprocess and if that's the response you want. That'll be domain specific.

usaspending_api/common/threaded_data_loader.py
afrasier: Yes, I could use None instead. Never thought to use it for that before, good point.

It's perfectly fine if those objects are shared across function calls because they are later copied for each process, and are never modified as-is.

usaspending_api/common/threaded_data_loader.py
afrasier: When you spawn a new process Django copies the current database connection if it exists for that process to use. That means if you have 10 processes they're all trying to use the same connection which can lead to some deadlock/loss of performance. By closing all connections before spawning the processes, you ensure each process gets their own connection.

usaspending_api/common/threaded_data_loader.py
afrasier: Simpler for someone who knows python well, but I thought since the DataLoaderThread object has a long batch of arguments that it would be easier to read and understand for novices if it was broken up explicitly.

If it bothers you, I can change it.

usaspending_api/common/threaded_data_loader.py
afrasier: Yes, I am working on a set of tests so that I can implement some extra exception handling, it will be added in the future.

usaspending_api/common/threaded_data_loader.py
cmc333333: Re using `{}` and `[]` in the constructors -- that's why it hasn't bitten yet, but it's something to keep in mind for the future. It's easy to imagine doing something seemingly innocuous like

```python
# file1.py
loader = ThreadedDataLoader(some_model_class)
loader.field_map['another_field'] = "some value"
do_something_with(loader)

# file2.py
loader = ThreadedDataLoader(some_other_model)
assert loader.field_map['another_field'] == "some value"  # wait what?
```

usaspending_api/common/threaded_data_loader.py
cmc333333: Gotcha, yeah the list comprehension isn't the important part, I think the `while` guard is. Here's an alternative:

```python
for in in range(self.processes):
    pool.append(DataLoaderThread('Process-' + str(i), ...))
```

PR #49, opened 2016-11-08
Ds 398 add verbose names to object models

playmakervi:
Added verbose_name label to models.py in awards and references folders.

Adjusted decidmal_places from 0 to 2.


playmakervi:
@afrasier

PR #50, opened 2016-11-08
Ds 419 update tasloader to utilize threaded data loader

playmakervi:
Updated TASloader to utilize multiprocessing data loading.


playmakervi:
@afrasier

PR #54, opened 2016-11-17
Ds 421 426 extra op search

afrasier:
* Added 'range_intersect' operation and documentation
* Added 'value_format' filter parameter and documentation
* Added autocomplete handler code and documentation (although a bit sparse, mind you)
* Added autocomplete endpoint for award summary
* Fixed an issue where FY objects weren't having a timezone
* Updated using_the_api.md to be shiny and up to date - with a bonus table of contents 🎉

### DIFF
--- a/usaspending_api/common/api_request_utils.py
+++ b/usaspending_api/common/api_request_utils.py
@@ -1,7 +1,8 @@
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django.contrib.postgres.search import SearchVector
 from django.db.models import Q
-from datetime import date
+from django.utils import timezone
+from datetime import date, time, datetime
 
 
 # This class represents a fiscal year
@@ -10,10 +11,11 @@ from datetime import date
 class FiscalYear():
     def __init__(self, fy):
         self.fy = fy
+        tz = time(0, 0, 1, tzinfo=timezone.utc)
         # FY start previous year on Oct 1st. i.e. FY 2017 starts 10-1-2016
-        self.fy_start_date = date(int(fy)-1, 10, 1)
+        self.fy_start_date = datetime.combine(date(int(fy)-1, 10, 1), tz)
         # FY ends current FY year on Sept 30th i.e. FY 2017 ends 9-30-2017
-        self.fy_end_date = date(int(fy), 9, 30)
+        self.fy_end_date = datetime.combine(date(int(fy), 9, 30), tz)
 
     # Creates a filter object using date field, will return a Q object such that
     # Q(date_field__gte=start_date) & Q(date_field__lte=end_date)
@@ -44,7 +46,8 @@ class FilterGenerator():
         'search': '__search',
 
         # Special operations follow
-        'fy': 'fy'
+        'fy': 'fy',
+        'range_intersect': 'range_intersect'
     }
 
     # Creating the class requires a filter map - this maps one parameter filter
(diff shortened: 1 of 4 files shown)


using_the_api.md
```diff
@@ -179,15 +192,36 @@ The structure of the post request allows for a flexible and complex query with b
...
-    * `fy` - Evaluates if the field (generally should be a datetime field) falls within the federal fiscal year specified as `value`. `value` should be a 4-digit integer specifying the fiscal year. An example of a fiscal year is FY 2017 which runs from October 1st 2016 to September 30th 2017.
+    * `fy` - Evaluates if the field (generally should be a datetime field) falls within the federal fiscal year specified as `value`. `value` should be a 4-digit integer specifying the fiscal year. An example of a fiscal year is FY 2017 which runs from October 1st 2016 to September 30th 2017. Does not need `value_format` as it is assumed.
     ```
     {
       "field": "date_signed",
       "operation": "fy",
       "value": 2017
     }
     ```
+    * `range_intersect` - Evaluates if the range defined by a two-field list intersects with the range defined
```
kaitlin: Can you be more explicit about the use case? I.e. finding overlap between the fiscal year range and the period of performance range? I think it will help people grok it.

usaspending_api/common/api_request_utils.py
```diff
@@ -171,9 +188,84 @@ def validate_post_request(self, request):
...
+
+        if mode == "contains":
+            mode = "__icontains"
+        elif mode == "startswith":
+            mode = "__istartswith"
+
+        for field in fields:
+            q_args = {}
+            q_args[field + mode] = value
+            value_dict[field] = list(set(data_set.filter(Q(**q_args)).values_list(field, flat=True)))  # Why this weirdness? To ensure we eliminate duplicates
```
kaitlin: Should this use the now built in search filter for postgress (https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/search/) instead of the `contains` and `startswith` keywords? It seems like this could get prohibitively slow as we add a lot more data and we'll want to use a pre-indexed `ts_vector` field or something equivalent.

afrasier: It can, and we already support full text search; but I do not think that it supports functionality such as 'startswith'; although I am not sure about the usefulness of that in general. If you are ok with losing that ability I can move it to using the already implemented search functionality.

kaitlin: Hmm, yeah I don't know if we even need that. I'm ok with getting rid of it.

afrasier: I mentioned it in slack but I'll mention it here for posterity - one benefit to using 'contains' or 'startswith' vs. 'search' is that it allows partial matches - that is, 'Tex' would match 'Texas'; whereas when using 'search' it only matches whole words. For autocomplete I'd wager we'd want to do partial matches. So I'm not sure we'd want to switch to 'search'

kevinli-work:
Can you clarify the docs for `range_intersect`? It might be helpful to write a human description of what you're searching for in the two provided examples.

afrasier:
Is the new description still unclear? 
```
An example of where this is useful is when a contract spans multiple fiscal years, to evaluate whether it overlaps with any one particular fiscal year - that is, the range defined by `period_of_performance_start` to `period_of_performance_end` intersects with the fiscal year.
```
Or do you mean I need to explain what it means for two ranges of numbers to intersect?

PR #65, opened 2016-12-01
Unique value functionality  + documentation

afrasier:
Adding unique value functionality

* Just throw in a list of fields into `unqiue_values`
* Added some docs

### DIFF
--- a/usaspending_api/common/api_request_utils.py
+++ b/usaspending_api/common/api_request_utils.py
@@ -1,6 +1,6 @@
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django.contrib.postgres.search import SearchVector
-from django.db.models import Q
+from django.db.models import Q, Count
 from django.utils import timezone
 from datetime import date, time, datetime
 
@@ -217,6 +217,33 @@ class FilterGenerator():
         q_case[fields[0] + "__lte"] = values[1]  # f1 <= r2
         q_case[fields[1] + "__gte"] = values[0]  # f2 >= r1
         return Q(**q_case)
+
+
+# Handles unique value requests
+class UniqueValueHandler():
+    @staticmethod
+    # Data set to use (should be filtered already)
+    # Fields to find unique values for
+    # Returns a dictionary containing fields, values and their counts
+    # {
+    #   "recipient__name": {
+    #       "Jon": 5,
+    #       "Joe": 2
+    #   }
+    # }
+    def get_values_and_counts(data_set, fields):
+        data_set = data_set.all()  # Do this because we don't want to get finnicky with annotations
+        response_object = {}
+        if fields:
+            for field in fields:
+                response_object[field] = {}
+                unique_values = data_set.values(field).distinct()
+                for value in unique_values:
+                    q_kwargs = {}
+                    q_kwargs[field] = value[field]
+                    print(q_kwargs)
+                    response_object[field][value[field]] = data_set.filter(**q_kwargs).count()
+        return response_object
 
 
 # Handles autocomplete requests
--- a/usaspending_api/common/api_request_utils.py
+++ b/usaspending_api/common/api_request_utils.py
@@ -17,8 +17,10 @@ class FiscalYear():
         # FY ends current FY year on Sept 30th i.e. FY 2017 ends 9-30-2017
         self.fy_end_date = datetime.combine(date(int(fy), 9, 30), tz)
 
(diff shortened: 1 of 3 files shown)


kaitlin:
@bsweger can you take a look?

bsweger:
@kaitlin @afrasier absolutely! will look first thing tomorrow!

using_the_api.md
```diff
@@ -99,6 +100,33 @@ The structure of the post request allows for a flexible and complex query with b
 
 * `page` - If your request requires pagination, this parameter specifies the page of results to return. Default: 1
 * `limit` - The maximum length of a page in the response. Default: 100
+* `unique_values` - A list of fields for which you would like to know the unique values and how many items have that value. An example response with that value would be:
```
bsweger: Thanks for keeping the docs up to date! You hint at this towards the end of the unique values description, but do you think it's worth explicitly stating that `unique_value` tallies things after filters are applied? 

If someone tries to `exclude` a field and do a `unique_value` on that same field, is that allowable?

usaspending_api/common/api_request_utils.py
```diff
@@ -219,6 +219,33 @@ def range_intersect(self, fields, values):
...
+        data_set = data_set.all()  # Do this because we don't want to get finnicky with annotations
+        response_object = {}
+        if fields:
+            for field in fields:
+                response_object[field] = {}
+                unique_values = data_set.values(field).distinct()
+                for value in unique_values:
+                    q_kwargs = {}
+                    q_kwargs[field] = value[field]
+                    print(q_kwargs)
```
bsweger: Is this a leftover debug statement?

usaspending_api/common/api_request_utils.py
```diff
@@ -219,6 +219,33 @@ def range_intersect(self, fields, values):
...
+    #   "recipient__name": {
+    #       "Jon": 5,
+    #       "Joe": 2
+    #   }
+    # }
+    def get_values_and_counts(data_set, fields):
+        data_set = data_set.all()  # Do this because we don't want to get finnicky with annotations
+        response_object = {}
+        if fields:
+            for field in fields:
```
bsweger: Maybe this is what you're alluding too in the annotations comment above...forgive the question since I'm new to the codebase and am still refreshing my good old Django ORM skills, but does it make sense to issue a count unique query for reach field in the incoming fields list?

Again, that's an honest question as I get up to speed on the project!

usaspending_api/common/api_request_utils.py
```diff
@@ -219,6 +219,33 @@ def range_intersect(self, fields, values):
         return Q(**q_case)
 
 
+# Handles unique value requests
+class UniqueValueHandler():
+    @staticmethod
+    # Data set to use (should be filtered already)
```
bsweger: This is great info and made the code review a lot easier--thanks!

usaspending_api/common/api_request_utils.py
afrasier: So I tried to use a method to annotate the data with the counts all at once, but it didn't function properly. What I had tried was: 
```data_set.value(field).annotate(count=Count(field))``` but it didn't seem to create the count as expect. If you have more experience with that aspect of Django I'd appreciate the help, as it should cut down on a few of the count calls. Couldn't find anything googling around.

File: usaspending_api/common/api_request_utils.py (hunk 2 above)
Comment (afrasier) [inline usaspending_api/common/api_request_utils.py]: Ach! I'm embarrassed, I usually get all those cleared out. I guess I was excited to be done.

**Comment by afrasier** [comment]
Re: Doc strings. Probably because those comments are created a line or two at a time as I think/write stuff. Easy enough to convert into a doc string, if you think that makes the most sense I can migrate them as I come across them.

File: using_the_api.md [hunk 5]
```diff
@@ -100,7 +100,7 @@ The structure of the post request allows for a flexible and complex query with b
 
 * `page` - If your request requires pagination, this parameter specifies the page of results to return. Default: 1
 * `limit` - The maximum length of a page in the response. Default: 100
-* `unique_values` - A list of fields for which you would like to know the unique values and how many items have that value. An example response with that value would be:
+* `unique_values` - A list of fields for which you would like to know the unique values and how many items have that value. These are processed _after_ the filters. An example response with that value would be:
```
bsweger: 👍

bsweger:
Thanks for updates and fancy docstrings @afrasier!

PR #68, opened 2016-12-05
Automated endpoint testing

afrasier:
* Added a new test file which will iterate over tests stored in a JSON file. This allows developers to quickly add new endpoint tests using a standard format to ensure all functionality and features are properly tested. See `usaspending_api/data/testing_data/endpoint_testing_data.json` for examples. (This still needs a few more tests added for other filters, which I am working on)

* Added a new fixture which is a subset of SBA data (roughly 20 records) that will populate all tables with data (i.e. a perfect testing database!). you can find it in `usaspending_api/common/fixtures/endpoint_fixture_db.json`

* Combined `agencies` `program activities` `object class` and `country codes` fixtures into a single fixture, found at `usaspending_api/references/fixtures/reference_fixture.json` and updated `fixtures.md` to reflect this change. This will make spinning up a new instance easier.

* Updated tests that used the old fixtures to use the new fixtures


afrasier:
@bsweger or @kaitlin 
Yay, back over 95% coverage

usaspending_api/common/tests/test_endpoints.py
```diff
@@ -0,0 +1,58 @@
...
+            elif method == "GET":
+                response = self.client.get(url, format='json')
+
+            # Check if the status code is correct
+            self.assertEqual(response.status_code, status_code)
+            # Check if the response object is correct
+            # because the native assertDictEqual cares about order (and we don't)
+            # we will convert our data to sets
+            self.assertEqual(set(response.data), set(response_object))
+            print("\tPASSED")
```
bsweger: Do we need this output?

PR #70, opened 2016-12-08
add POST options to all endpoints

bsweger:
**update: no longer in WIP status** 

@afrasier I wanted to put this out as a WIP before applying the new POST stuff to every endpoint. So far, this PR:

* Moves the POST options that were in the awards summary endpoint to `api_request_utils.py`.
* Applies those options to both the awards summary and awards endpoints

Would you mind checking this approach to see if it's what you had in mind?


bsweger:
@afrasier Thanks for taking a look--I added the page metadata back per your notes!

PR #71, opened 2016-12-09
Fancy location endpoint

afrasier:
* Added /references/locations/ endpoint that accepts normal POST requests
* Added /references/locations/geocomplete endpoint that accepts new Geographical hierarchy autocomplete requests (see _using_the_api.md_ as usual for docs)

### DIFF
--- a/usaspending_api/common/api_request_utils.py
+++ b/usaspending_api/common/api_request_utils.py
@@ -6,6 +6,8 @@ from django.contrib.postgres.search import SearchVector
 from django.db.models import Q
 from django.utils import timezone
 
+from usaspending_api.references.models import Location, RefCountryCode
+
 
 class FiscalYear():
     """Represents a federal fiscal year."""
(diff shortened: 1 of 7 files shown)


usaspending_api/common/api_request_utils.py
```diff
@@ -309,6 +311,79 @@ def validate(body):
...
+
+    def build_response(self):
+        # Array of search fields, and their 'parent' fields and types
+        search_fields = self.search_fields
+        value = self.request_body.get("value", None)
+        mode = self.request_body.get("mode", "contains")
+
+        if mode == "contains":
+            mode = "__icontains"
+        elif mode == "startswith":
```
afrasier: If you'll look at line 362 it defaults to "contains" if mode is not specified :)

usaspending_api/common/api_request_utils.py
```diff
@@ -309,6 +311,79 @@ def validate(body):
...
+            "type": "STATE",
+            "parent": "location_country_code__country_name"
+          },
+          "location_state_name": {
+            "type": "STATE",
+            "parent": "location_country_code__country_name"
+          },
+          "location_city_name": {
+            "type": "CITY",
+            "parent": "location_county_name"
```
afrasier: It doesn't inform how it currently works, no. 
I don't think multi-county cities will impact anything bizarrely, it will just function off of whatever county is specified by the Broker, or is found in RefCityCountyCodes

bsweger:
Thanks for the clarifications, @afrasier!

PR #77, opened 2016-12-13
Nested serializer field limiting

afrasier:
* Made all serializers `LimitableSerializers`
* Added `SerializerMethodField()` as all nested serializers to allow for passing cascading field inclusion/exclusion lists
* Added test cases to `endpoint_testing_data.json` to test the serializers


afrasier:
That's already the method implemented inside of `LimitableSerializer`.
https://github.com/fedspendingtransparency/usaspending-api/blob/multi-depth-limitable-serializer/usaspending_api/common/serializers.py#L25-L35

Passing context will also require the use of `SerializerMethodField` due to how the Serializers are initialized (to the best of my knowledge, anyway)

bsweger:
@afrasier Do you think it makes sense to explore accessing the request request within `LimitableSerializer`? That way we don't have pass `fields` and `exclude` when we instantiate the serializer.

For example, assuming that the serializer has access to the request, its `init` could do something like:
```python
if request.query_params.get('fields'):
    # handle field limiting for GET
if request.data.get('fields'):
    # handle field limiting for POST
```
Not quite related but towards the goal of reducing the need to roll our own, did you encounter drf-toolbox in your travels? This bit about Controlling Fields sounds like what we're trying to do: http://drf-toolbox.readthedocs.io/en/latest/nested.html#controlling-fields

afrasier:
The request isn't passed to the serializer, so we'd run into the same issue of needing to instantiate them and pass in the request in the same way we pass in the fields at the moment. Additionally, because the code is abstracted into `LimitableSerializer` we'd end up having to parse the entire list of fields each time rather than parsing a subset; which, while probably a negligible detraction from performance in most cases seems redundant.

The drf-toolbox example you would work if we were statically defining our fields; but I can't seem to find an example of someone using that library to dynamically change the fields, nor fields for higher depth nesting. (For example, `award` -> `recipient` -> `location`)

bsweger:
Re: getting the request in the serializer...I gather that the request context is made available to the serializer by the view automatically if we take advantage of the [framework's generic views](http://www.django-rest-framework.org/api-guide/generic-views/).

Now we're straying far from story at hand, I know, but I want to make sure we're 1) taking advantage of the conveniences built into the framework which will 2) make the code base more maintainable in the long run.

As for drf-toolbox, I haven't put a ton of thought into creating the field lists dynamically, but does it make sense to extend the part of that library that sets up the controlling fields rather than write our own code to traverse the model relationships?

afrasier:
From what I can see of the request is made available to the views, not the serializer. Thus one would need to pass the request to the serializer, and then we're back at square one in comparison to passing an array of fields. 

I'd say we are actually taking advantages of conveniences built into the framework, i.e. serializer mixins and serializer method fields for data acquisition and restructuring. Do the change in this PR seem unmaintainable?

I looked into the source of drf-toolbox and the functionality we want would require not insignificant modifications. I would think it would be more maintainable to have our own code that handles this functionality rather than forking and modifying a separate library, of whose functionality we would only be using a small portion. Additionally, this library was last updated as far as I can see in 2014, so I am not sure of it's compatibility with newer versions of Django and DRF

cmc333333:
@bsweger asked for a second opinion, so I did a little research. Here's what I sent her (forgot she was OOO today):

Okay, so the idea is that the webargs get passed to the top level serializer somewhere outside of this PR's code, and that each serializer passes those arguments down to subserializers through these "method fields". I agree that that seems cumbersome and that it'd be nice if there were a global context to query down in the guts of the serializers rather than passing the info along. I also agree that using GenericViews is a good idea; I'm less confident that it would resolve this problem. Some quick research:
1. GenericViews pass along request data into the serializer's "context" variable: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/generics.py#L136. You can access it from the root serializer via `self.context['request']`. I'm having trouble figuring out if the root serializer passes its context to its children, but leaning towards no based on this code.
2. Alternatively, you can combine these new serializers with GenericViews, but you'd need to override `get_serializer` to inspect self.request:
https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/generics.py#L104 . This seems like reinventing the wheel from 1, however.
3. You _might_ be able to modify the queryset to achieve the same goal (perhaps with https://docs.djangoproject.com/en/1.9/ref/models/querysets/#values or https://docs.djangoproject.com/en/1.9/ref/models/querysets/#only). If you can encode the filtering in your queryset, you don't need to pass that filtering info through the serializer construction
4. The use of a "MethodField" feels wonky and verbose. I think there's a way to replace just that component by e.g. overriding `get_fields` to set appropriate context from the parent serializer: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/serializers.py#L377
5. Not a fully formed idea, but there's a `SkipField` exception that can be thrown at various points. I wonder if that'd be a clean way to limit the fields.

I think that GenericViews would give you a few more crumbs to go on (like the context in the root serializer), but wouldn't directly (as far as I can tell) solve the problem (the request still needs to be passed down the hierarchy of serializers).

afrasier:
1. To the best of my knowledge the root serializer does not pass its context to children, thus requiring the use of `SerializerMethodField`
2. I'm not sure how necessary it would be to override that method on a GenericView, as you can set `serializer_class` on the class level. However, we'd have to override most of the accessing methods to pass extra arguments to the serializer from what I understand of GenericViews.
3. I'm not sure that this works. Even if it did work on the root serializer I doubt deferred fields would work on nested serializers.
```
>>> from usaspending_api.awards.models import Award
>>> from usaspending_api.awards.serializers import AwardSerializer
>>> m = AwardSerializer(Award.objects.only("type").first())
>>> m.data
{'uri': None, 'date_signed': '2016-09-19', 'create_date': '2016-12-05T18:27:09.934000Z', 'id': 1, 'awarding_agency': 567, 'total_outlay': None, 'description': 'FY 2016 B2BRP', 'fain': 'SBAHQ16V0028', 'funding_agency': 567, 'certified_date': None, 'period_of_performance_current_end_date': '2017-06-30', 'latest_submission': 977, 'type': '05', 'data_source': 'DBR', 'period_of_performance_start_date': '2016-09-01', 'parent_award_id': None, 'total_obligation': None, 'piid': None, 'update_date': '2016-12-05T18:27:09.934000Z', 'last_modified_date': None, 'recipient': 927146407, 'place_of_performance': 2}
```
4. I haven't found any way to access the parent serializer, if there was a way I would simply access the field limitations in `__init__` on `LimitableSerializer`; If you know how to access the parent serializer from the child I'd love to know how.
5. I'm not sure when these are raised - I think during updating an object from a serializer which we don't do.

cmc333333:
For 1: I agree that it doesn't look like Serializers will pass that context down. I'm saying that GenericViews get you one step further, but don't solve the problem.

For 2: Setting the `serializer_class` alone wouldn't work as current implementation is expecting extra parameters on instantiation. So, to use these serializers with GenericViews, you'd need to override `get_serializer`. Again, that seems like reinventing the wheel per 1.

For 3: Agreed that it's not obvious if there's a way to handle it at the queryset level. Hypothetically it'd be possible to configure the initial queryset to include the nested field filters using `only()`, but you'd need to inspect the queryset at serialization time (`only()` doesn't remove fields from the requested models). `values()` is a bit more straightforward in terms of filtering fields, but it's not returning models, so it'd require more rework.

For 4: Sorry, I'm suggesting `get_fields` in the _parent_, so that it sets some context on the child serializers. Data'd still be flowing down from parent to child, it'd just be happening automatically (without needing MethodFields). Something like:
```python
def get_fields(self):
    fields = super(LimitableSerializers, self).get_fields()
    # maybe a step to filter out fields by self.context here
    for field in fields:
        field.context = self.context
    return fields
```

For 5: I think `SkipField` might be useful, but it's a very different strategy (it'd require logic which is processed per field rather than per serializer).

In terms of direction, I'd recommend combining 1 and 4: use GenericViews to kick start passing the request into the root serializer and then set that context on each child serializer when enumerating the root's fields. I've not tried any of this, but that seems feasible looking at DRF's code. It'd bring you the benefits of GenericViews elsewhere along with a simpler interface for passing the serialization parameters around.

afrasier:
I do like the idea of 1 & 4; however, I think we have gotten a little off track of the goal for this PR. This was an evaluation of code rather than approach; if there were opinions or concerns regarding the approach that would be taken with this it would have been best to discuss during sprint planning or at the developer meeting the day after. The story as written is to allow users to specify field exclusions on nested serializers. As it stands, the code is (in my opinion) easily maintainable and understandable by others, as well as fully functional and completes the story as written. If we want to shift our approach for all API views this would need to be another story in a later sprint. Currently, this has now exceeded my estimated hours for this story. If we do not want to merge this code in, we should update @kaitlin and meet and come to a consensus on an approach and then write the appropriate story (or stories) to capture that requirement.

bsweger:
@cmc333333 thanks for weighing in, and @afrasier thanks for your patience as we all work through this. I agree that we should have discussed before writing the code and apologize for not being at the last sprint review. I've also learned a lot more about DRF between then and now. 

I feel pretty strongly that we should invest the extra time for the 1 & 4 approach. My experience working on the broker code base says that taking the time to align the code base with the framework at hand sooner rather than later is a good investment.

kaitlin:
FWIW I am ok with not completing this story in the current sprint if it means doing it the right way ™️ . I leave it to the technical folks on this thread to come to consensus on what that is. 😃

PR #81, opened 2016-12-20
Mixin complex filters

bsweger:
This PR is a first stab at building our views via a combination of Django Rest Framework viewsets and mixins instead of `APIView`. I wanted to get this code out here now, because we have some upcoming stories that may work better on this foundation (for example, the story to view an individual award).

This is best reviewed changeset by changeset--see the commit messages for more details on the thinking.

If folks think this is a reasonable approach, the next step is to update the remaining endpoints/views.


afrasier:
Looks good, merge at your leisure. Leaving it open in case you have some small changes to add.
Is the functionality for checking the request directly current used, or is that just groundwork?

bsweger:
@afrasier Thanks for taking a look so quickly! To your question about the serializer, the `award/summary` endpoint is hitting the code that's not explicitly passing the fields/exclude request parameters to the serializer.

The view's list method is currently building a response ([here](https://github.com/fedspendingtransparency/usaspending-api/blob/mixin-complex-filters/usaspending_api/common/mixins.py#L207)) by using `get_serializer_class()` to pass the serializer class to the `build_response` method in `ResponseMetadatasetMixin.` 

`build_response` then has to pass the request in as context when instantiating the serializer: https://github.com/fedspendingtransparency/usaspending-api/blob/mixin-complex-filters/usaspending_api/common/mixins.py#L207

I'd like to get rid of `ResponseMetadatasetMixin` altogether. I think if we craft a DRF pagination method that suits our purposes, we'll no longer have to manually add metadata to the response. In that glorious future, the endpoint views could look something like:

```python
class SomeModelsViewset(FilterQuerysetMixin, viewsets.ReadOnlyModelViewSet):
    serializer_class = SomeModelSerializer
    paginator = instantiate SomePaginationClass
    def get_queryset(self):
        queryset = SomeModel.objects.all()
        filtered_queryset = self.filter_records(self.request, queryset=queryset, filter_map_stuff)
        ordered_queryset = self.order_records(self.request, queryset=filtered_queryset)
        # some pagination magic I haven't played with yet
```

In the super glorious future, we might also be able to do filtering and ordering in a more framework compliant way and get rid of the `FilterQuerysetMixin` too. Even if we keep our custom filters, I think there's a way to hook that stuff into methods on the generic view. Need to do more research on that.

PR #86, opened 2016-12-29
standardize on DRF patterns

bsweger:
As a prelude to continuing to add features to the USAspending API, standardize our view handling around common Django Rest Framework patterns to:
*  reduce the amount of code needed to create endpoints and
* make the code base more maintainable over time

There's still some work to do here, notably reducing the custom code we use to create the response metadata (perhaps by using DRF's built-in pagination) and using routers instead of manually mapping request types to ViewSet methods. However, this should be a reasonable start towards leveraging the features of DRF.

This is best reviewed commit by commit.

A few background notes:
* the `unique_values` metadata has been removed from the expected response in integration tests because that feature is now handled by our aggregate endpoint (added as part of a different story)
* we have a story on the board for fix up our awards endpoints for requesting a single award, so I didn't worry too much about the `<piid>` `<fain>`, etc. endpoints.

### DIFF
--- a/usaspending_api/accounts/views.py
+++ b/usaspending_api/accounts/views.py
@@ -25,7 +25,6 @@ class TreasuryAppropriationAccountBalancesViewSet(FilterQuerysetMixin,
                                                   DetailViewSet):
     """Handle requests for appropriation account balance information."""
     serializer_class = AppropriationAccountBalancesSerializer
-    queryset = AppropriationAccountBalances.objects.all()
 
     def get_queryset(self):
         queryset = AppropriationAccountBalances.objects.all()
--- a/usaspending_api/accounts/views.py
+++ b/usaspending_api/accounts/views.py
@@ -1,99 +1,34 @@
-import json
-
-from rest_framework import status
-from rest_framework.views import APIView
-from rest_framework.response import Response
-
 from usaspending_api.accounts.models import TreasuryAppropriationAccount
 from usaspending_api.accounts.serializers import TreasuryAppropriationAccountSerializer
 from usaspending_api.accounts.models import AppropriationAccountBalances
 from usaspending_api.accounts.serializers import AppropriationAccountBalancesSerializer
-from usaspending_api.common.api_request_utils import FilterGenerator, ResponsePaginator, DataQueryHandler
+from usaspending_api.common.mixins import FilterQuerysetMixin, ResponseMetadatasetMixin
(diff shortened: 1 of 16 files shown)


usaspending_api/accounts/views.py
```diff
@@ -1,99 +1,34 @@
...
+
+
+class TreasuryAppropriationAccountBalancesViewSet(FilterQuerysetMixin,
+                                                  ResponseMetadatasetMixin,
+                                                  DetailViewSet):
+    """Handle requests for appropriation account balance information."""
+    serializer_class = AppropriationAccountBalancesSerializer
+    queryset = AppropriationAccountBalances.objects.all()
+
+    def get_queryset(self):
```
cmc333333: Curious: doesn't this wipe out the `queryset` on line 28?

usaspending_api/common/api_request_utils.py
```diff
@@ -492,9 +492,9 @@ def build_response(self):
 class ResponsePaginator:
     @staticmethod
     def get_paged_data(data_set, page=1, page_limit=100, request_parameters={}):
-        if 'limit' in request_parameters:
+        if request_parameters.get('limit') is not None:
```
cmc333333: I don't understand this change -- what did the previous version miss?

bsweger: Yeah, you're right...the previous version should still work and is probably more readable. I just got swept away by the change from a standard dict back to the QueryDict.

usaspending_api/accounts/views.py
bsweger: Good catch--line 28 was old code I forgot to remove.

usaspending_api/common/mixins.py
```diff
@@ -185,8 +185,8 @@ def build_response(self, request, *args, **kwargs):
         queryset = kwargs.get('queryset')
 
         # workaround to handle both GET and POST requests
-        params = dict(self.request.query_params)
-        params.update(dict(self.request.data))
+        params = self.request.query_params.copy()  # copy() creates mutable copy of a QueryDict
```
afrasier: Just noting this creates a shallow copy of the dict contents which I don't think _will_ matter but _may_

bsweger: :thinking: Thanks for the heads-up...this is behaving as expected (and the QueryDict object doesn't have a deepcopy method), so will leave this for now if that's okay by you.

PR #91, opened 2017-01-09
Ds 493 contract loader robusting

afrasier:
* Updated award action models to have better character lengths 
* Updated loadcontracts with a _lot_ of extra field processing. Most fields require us to grab the first part of a string such as "N: No actions", so we call multiple splits on it. Additionally, some fields require only the first character (Fields that commonly use "False" "True" "T" "F" "X" "No" "Yes" "N" "Y"). Since these fields are max_length=1 we just grab the first character.
* Added some processing for fields that need to use enumerations defined in the schema
* Added logic to search for country codes when an improper code is specified (i.e. when the country code is 'SOUTH KOREA' instead of 'KOR')
* Fixed a few bugs in the threaded data loader, and added the ability to change the logger used
* Updated recipient object to base uniqueness off of the duns instead of the arbitrary autoincrement PK (this was an oversight in a much earlier change)


afrasier:
@bsweger Tagging you for review. One this is merged and the migrations have been run, it should be possible to load the sliced contract data onto the development environment.

usaspending_api/references/models.py
```diff
@@ -247,7 +250,7 @@ class LegalEntity(DataSourceTrackedModel):
     class Meta:
         managed = True
         db_table = 'legal_entity'
-        unique_together = (('recipient_name'),)
+        unique_together = (('recipient_unique_id'),)
```
bsweger: For creating a unique index on a single field, you can just specify `unique=True` when defining field, right? Any reason to put this down in `Meta`?

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -26,25 +27,227 @@ def handle(self, *args, **options):
...
+            "other_statutory_authority": "otherstatutoryauthority",
             "modification_number": "modnumber"
         }
 
         value_map = {
             "data_source": "USA",
-            "award": lambda row: Award.objects.get_or_create(piid=row['piid'], type='C')[0],
-            "recipient": lambda row: LegalEntity.objects.get_or_create(recipient_name=row['dunsnumber'])[0],
-            "awarding_agency": lambda row: Agency.objects.get(subtier_code=self.get_agency_code(row['maj_agency_cat'])),
+            "award": lambda row: self.create_or_get_award(row),
```
bsweger: You already know this, but just flagging that we'll have to circle back and update this once the PR for DS-542 is merged.

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -26,25 +27,227 @@ def handle(self, *args, **options):
...
+            "awarding_agency": lambda row: self.get_agency(row),
```
bsweger: general question about all of the cleanup you're doing here: since award transaction data that we're getting from the broker is coming from the same source systems that feed the current usaspending, will we also need to apply the cleanup to the D1/D2 broker files?

Likely not something we need to worry about immediately, but just planting a seed, since the broker doesn't really validate D1/D2 details, and I'm sure where the data quality breakdown is occurring.

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -26,25 +27,227 @@ def handle(self, *args, **options):
...
             "action_date": lambda row: self.convert_date(row['signeddate']),
             "last_modified_date": lambda row: self.convert_date(row['last_modified_date']),
             "gfe_gfp": lambda row: row['gfe_gfp'].split(":")[0],
+            "multiple_or_single_award_idv": lambda row: row['multipleorsingleawardidc'].split(' ')[0].split(':')[0],
```
bsweger: Since the `split(' ')[0].split(':')[0]` pattern is used to clean up so many of the fields, would it be more readable just to declare a function that can be re-used instead of doing all of the lambdas?

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -26,25 +27,227 @@ def handle(self, *args, **options):
...
+        }
+
+        le = LegalEntity.objects.filter(recipient_unique_id=row['dunsnumber']).first()
+        if not le:
+            le = LegalEntity.objects.create(**recipient_dict)
+
+        return le
+
+    def parse_first_character(self, flag):
+        if len(flag) > 0:
```
bsweger: I'm sure you've thought this through, but I assume that this will cause the field in question to be set to `None` (and thus `NULL` in the db) if `len(flag) > 0` is not true. If that's the way it's supposed to work, what do you think about being explicit and returning `None` in that case (for ease of reading)?

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -26,25 +27,227 @@ def handle(self, *args, **options):
...
+
+    def create_or_get_recipient(self, row):
+        # First, create the locations
+        location_dict = {}
+        pop_dict = {}
+
+        loc_country_code = row["vendorcountrycode"].split(":")[0]
+
+        loc_country_code = RefCountryCode.objects.filter(Q(**{"country_code": loc_country_code}) | Q(**{"country_name__iexact": loc_country_code})).first()
+        if not loc_country_code:
```
bsweger: Thanks for going above and beyond the call of duty when dealing with the location mess. Of all the data cleaning you've done, this strikes me as one thing we'll need to make sure we capture somewhere when doing story [DS-590](https://federal-spending-transparency.atlassian.net/browse/DS-590).

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -26,25 +27,227 @@ def handle(self, *args, **options):
...
+            "black_american_owned_business": self.parse_first_character(row['baobflag']),
+            "hispanic_american_owned_business": self.parse_first_character(row['baobflag']),
+            "native_american_owned_business": self.parse_first_character(row['naobflag']),
+            "subcontinent_asian_asian_indian_american_owned_business": self.parse_first_character(row['saaobflag']),
+            "us_local_government": self.parse_first_character(row['islocalgovernmentowned']),
+            "division_name": self.parse_first_character(row['divisionname']),
+            "division_number": self.parse_first_character(row['divisionnumberorofficecode'])
+        }
+
+        le = LegalEntity.objects.filter(recipient_unique_id=row['dunsnumber']).first()
```
bsweger: Do we need a mechanism to update an existing DUNS entry (a la `create_or_update`)? I assume that the flags that describe legal entities can change over time?

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -26,25 +27,227 @@ def handle(self, *args, **options):
...
+        if loc_country_code.country_code == "USA":
+            location_dict = {
+                "location_country_code": loc_country_code,
+                "location_zip5": row["zipcode"][:5],
+                "location_zip_last4": row["zipcode"].replace("-", "")[5:],
+                "location_address_line1": row["streetaddress"],
+                "location_address_line2": row["streetaddress2"],
+                "location_address_line3": row["streetaddress3"],
+                "location_state_code": row["state"],
+                "location_city_name": row["city"],
```
bsweger: Admittedly I don't know the data universe you've been mucking around in, so a few questions:
1. Is there a reason we don't use city code and county code when looking up and creating the recipient location?
1. Do we need a mechanism for updating an existing location object?

bsweger:
@afrasier oh yes--a final question/observation: we really need some robust unit tests around the ETL. Because we really need to get something loaded, we could possibly get away adding these in the next sprint? Given the state of the data, I don't feel comfortable going waiting too long to add them.

usaspending_api/references/models.py
afrasier: I wanted to leave it open to have combination of fields determine uniqueness, namely `recipient_unique_id` and some form of date

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: I think everything coming from broker already conforms to the data structure we expect? It's worth looking into.

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: I was having to change them around a lot. Also, due to the way the function passing works, you'd end up having to do something like 

`multiple_or_single_award_idv`: lambda row: self.parse_with_splits("multipleorsingleawardidv")`

Which is probably still slightly more readable, so this is kind of just a relic of hammering my head against it for a while.
Let's put it down for refactoring when I make the changes for DS-542?

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: Yeah, for clarification of anyone coming behind to read:
Sometimes "vendorcountrycode" isn't a country code at all, sometimes it is a name. Like "SOUTH KOREA" instead of "KOR". However, our reference table stores countries in patterns like "KOREA, SOUTH" or "BAHAMAS, THE". Thus, I chain together filters on each word to be able to locate the country code by name if there is no direct match to name or country code.

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: 1. It's not currently set up as a FK system (it probably should be, like how country_code is), and we don't currently have City codes loaded (I think). That is, the location model needs some work. . . 
2.  I don't think so, locations are unique via the combination of all of their parts. If one of the parts changes, it is a new location, is it not? Or we can just say "screw it" and geocode everything and just store lat/lng :)

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: Yes. Going back to the topic about recipient uniqueness flags, we need to add some chronological elements to it.

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: Sure, I can add that when I make changes for DS-542 updates as with the refactor

afrasier:
Yes, I agree. We need some killer tests for the ETL. It shouldn't be too difficult to add, but I don't think we can fit it into this sprint. Something like taking the first 10 rows from the historical data, running it through the loader, and making sure it all got put in the right spot. This will get annoying to upkeep if the schema keeps changing field names, though. So that's something to keep in mind.

usaspending_api/awards/management/commands/loadcontracts.py
bsweger: Thanks for the clarification. I'll add a story for the city and county codes, in case we want to circle back and update the model. I think the answer to #2 depends on how we handle #1. Agreed that it makes sense for locations to be unique via a combination of their parts. I raised the question because it doesn't look like we're using all the parts (not using the codes especially threw me off...like, what if a city name changes but the code is still the same...didn't realize we weren't getting those yet)

bsweger:
@afrasier Ok, I'll make a story for unit tests and bring it to everyone's attention at backlog grooming tomorrow. Kaitlin can make the call of whether or not it's worth delaying other features in the sprint to get the unit tests in. When I say unit tests, I don't even think we need to use records per se...I really like the mocking approach that @cmc333333 brought to the broker. It's a good way to test all the edge cases from within the testing code itself w/o having to maintain a set of external files or records.

I'm going to make this as approved, merge it, and them merge it into my branch for 542. Thanks again for all of this work!

PR #94, opened 2017-01-11
Augment award summary

bsweger:
The main purpose of this PR is to make sure we fully populate Award objects when we create them in the data loads: [DS-542](https://federal-spending-transparency.atlassian.net/browse/DS-542).
To do this, the award "summary" object get/create method is now attached to the Award object to ensure that it's accessible by all ETL scripts.

The initial Award creation doesn't include much information. The details are filled in by the Award model's `update_from_mod` method, which is invoked whenever an award transaction record is saved.

Note that this PR doesn't include the addition keys that we'll eventually need to consider when creating distinct awards--we'll have to purge and recreate the data once that's done.

The PR also:
* adds some missing data to the AwardAction model
* updates the submission load script to populate those new fields
* beefs up the contract piece of the submission loader to include place of performance and recipient information

See notes on the individual commits for more details.


usaspending_api/awards/models.py
```diff
@@ -156,15 +144,68 @@ def __get_type_description(self):
...
+        self.recipient = transaction_latest.recipient
         self.total_obligation = transaction_set.aggregate(total_obs=Sum(F('federal_action_obligation')))['total_obs']
+        # what txn-level fields do we sum to get the award's total outlay?
+        # self.total_outlay = ??
+        if hasattr(transaction_latest, 'assistance_type'):
+            # this is a financial assistance award, so use assistance_type field
+            self.type = transaction_latest.assistance_type
+        else:
+            # this is a contract
+            self.type = 'C'
```
afrasier: 'C' isn't actually a type of award. This comes in from I believe file D. It should be one of the choices in AWARD_TYPES (an integer, or 'U') the other character types are leftovers from the first pass at the summary object. We probably need to change this so it uses the appropriate integer type  rather than 'C'

PR #95, opened 2017-01-11
Write some unittests for functions related to contract loading

cmc333333:
@bsweger kindly asked that I write some unit tests in py.test style to demonstrate the value of both. This PR (which is best reviewed commit by commit) hopefully does that, by writing tests and refactoring (in sequence) a few chunks of the logic in the contract loading admin command. Several edge cases for each chunk are tested.

This doesn't make use of pytests' monkeypatch test fixture (which is great for code with lots of dependencies,) but I can probably find a use case if desired.

### DIFF
--- a/usaspending_api/awards/management/commands/loadcontracts.py
+++ b/usaspending_api/awards/management/commands/loadcontracts.py
@@ -255,7 +255,7 @@ def get_or_create_location(row):
             location_foreign_city_name=row["city"],
         )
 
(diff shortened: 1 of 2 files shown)


afrasier:
@bsweger Merge at your leisure once you have a look

bsweger:
@cmc333333 This is terrific--thank you so much. If we have any monkey patching questions, we know where to find you!

PR #96, opened 2017-01-11
More loadcontracts updates

afrasier:
* Added process termination to threaded data loader to clear up some
hanging thread issues

*Updated load contracts to handle new field requirements as well as


usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -94,15 +98,13 @@ def handle(self, *args, **options):
         loader = ThreadedDataLoader(Procurement, field_map=field_map, value_map=value_map)
         loader.load_from_file(options['file'][0])
 
-    def get_agency(self, row):
-        agency = Agency.objects.filter(subtier_code=self.get_agency_code(row['maj_agency_cat'])).first()
+    def get_agency(self, agency_string):
+        agency = Agency.objects.filter(subtier_code=self.get_agency_code(agency_string)).first()
```
bsweger: I see this pattern of `first()` in our ETL a lot. At some point, do you think it makes sense to do a `get()` instead so that we can capture MultipleObject errors and log if we multiple records in a reference table? Or are we ok relying on unique indexes at the db level to protect us from these types of integrity issues?

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -94,15 +98,13 @@ def handle(self, *args, **options):
...
         if not agency:
-            print("Missing agency: " + row['maj_agency_cat'])
+            print("Missing agency: " + agency_string)
```
bsweger: Should we use the logging system here instead of print statement?

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -228,34 +227,92 @@ def fetch_country_code(vendor_country_code):
...
             query_set = query_set.filter(country_name__icontains=word)
         country_code = query_set.first()
-    
+
     return country_code
 
 
-def get_or_create_location(row):
-    country_code = fetch_country_code(row['vendorcountrycode'])
+'''
```
bsweger: Is this a leftover debug/note to self?

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -228,34 +227,92 @@ def fetch_country_code(vendor_country_code):
...
+
+
+def get_or_create_location(row, mode="vendor"):
+    place_of_performance_mapping = {
+        "city": row["placeofperformancecity"],
+        "congressionaldistrict": row["placeofperformancecongressionaldistrict"][2:],  # Need to strip the state off the front
+        "country": row["placeofperformancecountrycode"],
+        "zipcode": row["placeofperformancezipcode"].replace("-", ""),  # Either ZIP5, or ZIP5+4, sometimes with hypens
+        "state_code": row["pop_state_code"].split(":")[0]  # Format is VA: VIRGINIA, so we need to grab the first bit
```
bsweger: Thanks for leaving this formatting comments inline--our future colleagues will surely thank you for this

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -228,34 +227,92 @@ def fetch_country_code(vendor_country_code):
...
-        "location_address_line3": row["streetaddress3"],
     }
 
     if country_code.country_code == "USA":
         location_dict.update(
-            location_zip5=row["zipcode"][:5],
-            location_zip_last4=row["zipcode"].replace("-", "")[5:],
-            location_state_code=row["state"],
-            location_city_name=row["city"],
+            location_zip5=mapping["zipcode"][:5],
```
bsweger: Your handling of zip codes here is more robust than the POP stuff I added to the submission loader. I'll go back and fix it up... in the meantime, just curious: does the location search endpoint look at all of the various zip code fields in the location model?

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -228,34 +227,92 @@ def fetch_country_code(vendor_country_code):
...
+            location_zip_last4=mapping["zipcode"][5:],
+            location_state_code=mapping["state_code"],
+            location_city_name=mapping["city"],
+            location_congressional_code=mapping["congressionaldistrict"]
         )
     else:
         location_dict.update(
-            location_foreign_postal_code=row["zipcode"],
-            location_foreign_province=row["state"],
+            location_foreign_postal_code=mapping["zipcode"],
```
bsweger: Some of this logic (e.g., setting foreign postal code to zip code when appropriate) also applies to the submission ETL. At what point does it make sense to centralize some of it to ensure that things remain consistent regardless of data source? I suspect we're already out of sync on some things.

usaspending_api/awards/models.py
```diff
@@ -212,14 +212,15 @@ def get_or_create_summary_award(piid=None, fain=None, uri=None, parent_award_id=
...
-            piid, fain, uri, parent_award_id))
+             piid, fain, uri, parent_award_id))
 
     class Meta:
         db_table = 'awards'
 
 
 class AwardAction(DataSourceTrackedModel):
     award = models.ForeignKey(Award, models.CASCADE, related_name="actions")
+    usaspending_unique_transaction_id = models.CharField(max_length=256, blank=True, null=True)
```
bsweger: Good idea. Is there any value in naming this something more generic (e.g., `source_unique_transaction_id`) and also using it to capture the surrogate key of the applicable broker table when running the submission ETL?

usaspending_api/common/threaded_data_loader.py
```diff
@@ -96,6 +96,9 @@ def load_from_file(self, filepath, encoding='utf-8'):
 
         row_queue.join()
 
+        for process in pool:
```
bsweger: This is just for my own curiosity--why this change?

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: I think for some things, .first() makes sense. (Here, because agencies are guaranteed to be unique). The main reason is .filter().first() will return None if there are no results, whereas .get() will throw an exception. We'd just need a slightly more robust  try/catch block all over that place that might be worth putting somewhere it could be shared.

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: Yep, my bad. I will fix it.

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: Yeah, I thought I removed it, will do.

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: It looks at location_zip5 for the hierarchical location search. (Most people don't even know their +4 :) )

usaspending_api/awards/management/commands/loadcontracts.py
bsweger: Yeah, agreed that if we're guaranteeing uniqueness at the db level, it might be overkill to put in try/catch all over the place. Will keep it as a watch point going forward, thinking more about where it makes sense.

usaspending_api/awards/management/commands/loadcontracts.py
afrasier: Yeah, we should definitely centralize as much of the logic as is possible and makes sense

usaspending_api/awards/models.py
afrasier: Sure, we can do that.

afrasier: Actually, would it be useful to add it to `DataSourceTrackedModel` so it applies to anything we're tracking the source of?

usaspending_api/common/threaded_data_loader.py
afrasier: To ensure all the processes close. If any of the processes hang or don't close for some reason, you can be left with some phantom processes that, while they aren't using local resources, are still maintaining a connection to the DB, which could cause you to fill max_connections with effectively ghosts.

usaspending_api/awards/management/commands/loadcontracts.py
bsweger: Ok, well then I better fix that up on the submission loader. Thanks!

usaspending_api/awards/models.py
bsweger: Maaaybe. You know what--let's hold this thought and come back to it once we have some data loaded.

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -279,8 +277,8 @@ def get_or_create_location(row, mode="vendor"):
     if mode == "vendor":
         location_dict.update(
             location_address_line1=mapping["street1"],
-            location_address_line2=mapping["street3"],
-            location_address_line3=mapping["street2"],
+            location_address_line2=mapping["street2"],
```
bsweger: Good catch--sorry I missed that during the earlier review.

PR #98, opened 2017-01-12
Ds 544 awarding agency top sub office tier

afrasier:
* Created new tiered agency models
* Updated agency model to combine three tiered agency models
* Updated agency loader
* Updated various tests and testing fixtures
* Updated reference fixture
* Updated load_submission and loadcontracts to use new agency structure


usaspending_api/references/models.py
```diff
@@ -51,31 +48,60 @@ class Meta:
...
+    name = models.CharField(max_length=150, blank=True, null=True, verbose_name="Top-Tier Agency Name")
+
+    class Meta:
+        managed = True
+        db_table = 'toptier_agency'
+
+
+class SubtierAgency(models.Model):
+    subtier_agency_id = models.AutoField(primary_key=True)
+    # will equal fpds_code if a top level department
```
bsweger: I'm still confused about this. So, the "awarding agency" fields in incoming FPDS data are analagous to what we call sub-tier agencies in our world. If that's right, can we make the comment say something that effect for the sanity of our future selves?

usaspending_api/references/test_agency_loader.py
```diff
@@ -22,13 +22,13 @@ def test_department(self):
         Make sure an instance of a department is properly created
         """
 
-        department = Agency.objects.get(cgac_code='002', fpds_code='0000', subtier_code='0000')
+        department = Agency.objects.get(toptier_agency__cgac_code='002', toptier_agency__fpds_code='0000', subtier_agency__subtier_code='0000')
```
bsweger: Forgive me, I'm still grokking the agencies...are FPDS codes the same thing as CGAC codes? If so, why do we always to lookups on fpds codes? That's legacy info, yes?

usaspending_api/references/models.py
```diff
@@ -30,9 +30,6 @@ class RefCountryCode(models.Model):
     create_date = models.DateTimeField(auto_now_add=True, blank=True, null=True)
     update_date = models.DateTimeField(auto_now=True, null=True)
 
-    def __str__(self):
```
afrasier: I don't know why this was taken out. I will put it back . . .

usaspending_api/awards/management/commands/loadcontracts.py
```diff
@@ -99,7 +99,7 @@ def handle(self, *args, **options):
         loader.load_from_file(options['file'][0])
 
     def get_agency(self, agency_string):
-        agency = Agency.objects.filter(subtier_code=self.get_agency_code(agency_string)).first()
+        agency = Agency.objects.filter(subtier_agency__subtier_code=self.get_agency_code(agency_string)).first()
```
afrasier: I do believe they are, they are based off of cgac_code if I am not mistaken, so a subtier inside of something with a cgac_code of 13 will always start with 13. i.e. 1341 is NIST and 1300 is DOC

usaspending_api/references/test_agency_loader.py
afrasier: This is just a test, I don't think anyone would bother looking up by FPDS codes.

PR #99, opened 2017-01-12
Tweaking column names

afrasier:
Previously we were using maj_ fields to get the agency, which would
only get us information on the level of top tier departments. This
change will get us the granularity of subtier agencies as expected.


catherinedevlin:
Makes sense!  Merging.

PR #101, opened 2017-01-13
Extract code useful for financial assistance loading into helpers.py …

catherinedevlin:
WORK IN PROGRESS 

From helpers.py, they can be used by contracts loader or the financial assistance loader.

Also some formatting changes (using isort and yapf) and refactoring of colon-splitting into a single function.

### DIFF
--- a/usaspending_api/awards/management/commands/helpers.py
+++ b/usaspending_api/awards/management/commands/helpers.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import logging
 
 from django.db.models import Q
 
@@ -9,16 +10,9 @@ from usaspending_api.references.models import Agency, Location, RefCountryCode
 def up2colon(input_string):
     'Takes the part of a string before `:`, if any.'
 
-    return input_string.split(':')[0].strip()
-
-
-def get_agency(self, agency_string):
-    agency_code = up2colon(agency_string)
-    agency = Agency.objects.filter(
-        subtier_agency__subtier_code=agency_code).first()
-    if not agency:
-        self.logger.error("Missing agency: " + agency_string)
-    return agency
+    if input_string:
+        return input_string.split(':')[0].strip()
+    return ''
 
 
 def convert_date(date):
--- a/usaspending_api/awards/management/commands/helpers.py
+++ b/usaspending_api/awards/management/commands/helpers.py
@@ -1,0 +1,108 @@
+from datetime import datetime
+
+from django.db.models import Q
+
+from usaspending_api.awards.models import Award
+from usaspending_api.references.models import Agency, Location, RefCountryCode
+
+
+def up2colon(input_string):
+    'Takes the part of a string before `:`, if any.'
+
+    return input_string.split(':')[0].strip()
+
+
+def get_agency(self, agency_string):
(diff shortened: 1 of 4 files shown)


usaspending_api/awards/management/commands/helpers.py
```diff
@@ -0,0 +1,108 @@
...
+        for word in code_str.split():
+            query_set = query_set.filter(country_name__icontains=word)
+        country_code = query_set.first()
+
+    return country_code
+
+
+def get_or_create_award(row):
+    piid = row.get("piid", None)
+    parent_award_id = row.get("idvpiid", None)
```
afrasier: This stuff is specific to Procurement/Contracts. May needs to add some things to grab fain/uri or a separate method for financial assistance

usaspending_api/awards/management/commands/helpers.py
```diff
@@ -0,0 +1,108 @@
...
+from usaspending_api.references.models import Agency, Location, RefCountryCode
+
+
+def up2colon(input_string):
+    'Takes the part of a string before `:`, if any.'
+
+    return input_string.split(':')[0].strip()
+
+
+def get_agency(self, agency_string):
```
afrasier: Whoops, missed this. 'self' needs to be removed here.

usaspending_api/awards/management/commands/helpers.py
```diff
@@ -0,0 +1,108 @@
+from datetime import datetime
+
+from django.db.models import Q
+
+from usaspending_api.awards.models import Award
+from usaspending_api.references.models import Agency, Location, RefCountryCode
+
+
+def up2colon(input_string):
+    'Takes the part of a string before `:`, if any.'
```
afrasier: This throws an error if `input_string` is None, might want to check that and just return None if input_string is none

I'm not exactly sure how/when/why the input is ever None, I think it has something to do with csvreader, but while testing this did come up.

usaspending_api/awards/management/commands/helpers.py
```diff
@@ -0,0 +1,108 @@
...
+
+    return input_string.split(':')[0].strip()
+
+
+def get_agency(self, agency_string):
+    agency_code = up2colon(agency_string)
+    agency = Agency.objects.filter(
+        subtier_agency__subtier_code=agency_code).first()
+    if not agency:
+        self.logger.error("Missing agency: " + agency_string)
```
afrasier: self won't work here, also logger is not defined. Might want to pass in the logger or grab it when the function is called

usaspending_api/awards/management/commands/helpers.py
catherinedevlin: That's right, I'd forgotten about that.  I'll move this back to loadcontracts after all.

usaspending_api/awards/management/commands/helpers.py
catherinedevlin: Good point.  Here's a fix.

catherinedevlin:
I ended up working from a different branch and creating a pull request from that instead: https://github.com/fedspendingtransparency/usaspending-api/pull/106.  Thanks for these comments, they were incorporated over there.

PR #102, opened 2017-01-13
Null awards pre-filtering

afrasier:
Adding managers to award object.
To get awards that aren’t “generated null awards”, use Award.nonempty

Award.objects = gets all awards, including the empty ones
Award.nonempty = gets all awards, but filters out the empty ones

### DIFF
--- a/usaspending_api/awards/models.py
+++ b/usaspending_api/awards/models.py
@@ -84,6 +84,17 @@ class FinancialAccountsByAwardsTransactionObligations(DataSourceTrackedModel):
         db_table = 'financial_accounts_by_awards_transaction_obligations'
 
 
+class AwardManager(models.Manager):
+    def get_queryset(self):
+        q_kwargs = {
+            "type": "U",
+            "total_obligation__isnull": True,
+            "date_signed__isnull": True,
+            "recipient__isnull": True
+        }
+        return super(AwardManager, self).get_queryset().filter(~Q(**q_kwargs))
+
+
 class Award(DataSourceTrackedModel):
 
     AWARD_TYPES = (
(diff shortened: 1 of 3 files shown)


usaspending_api/awards/models.py
```diff
@@ -84,6 +84,17 @@ class Meta:
         db_table = 'financial_accounts_by_awards_transaction_obligations'
 
 
+class AwardManager(models.Manager):
+    def get_queryset(self):
+        q_kwargs = {
+            "type": "U",
+            "total_obligation__isnull": True,
```
bsweger: Just curious--where did the criteria to include null total_obligation, date_signed, and recipient come from?

bsweger:
@afrasier Sorry, should have asked on the original review--do we need tests for this?

usaspending_api/awards/models.py
afrasier: A generated award will have these set to null, but will also receive no transactions. Thus, these will remain null. This finds those awards and throws them out. As soon as one of those awards gets a transaction (i.e. it is no longer empty), these will be updated via update_from_mod and the award will no longer match these criteria

afrasier:
It's actually being tested by all the stuff in test_endpoints.py, but we can add something more specific if you'd prefer, since test_endpoints.py does quite a lot.

usaspending_api/awards/models.py
bsweger: Thanks for clarifying. That all makes sense, but I can see how this might look arbitrary to someone not familiar with how all of this works. What do you think about putting some of your explanation into a comment?

bsweger:
@afrasier if it's possible to put in a quick, specific test to ensure that no empty award records come back, that would be great

PR #106, opened 2017-01-18
Initial financial assistance load

catherinedevlin:
Several questions remain about the relationships between fields from the "other financial assistance" downloads at https://www.usaspending.gov and our schema, but this should be ready for preliminary review.


usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+        # Create a new submission attributes object for this timestamp
+
+        subattr = SubmissionAttributes()
+        subattr.usaspending_update = datetime.now()
+        subattr.save()
+
+        # field_map is used to simply translate a column name in our schema
+        # to the field name in the input source
+        field_map = {
+            "award_id": "federal_award_id",
```
afrasier: The `award` field on the FinancialAssistance object is a fk to a summary award that contains it, does `federal_award_id` correspond to what is called `idv` or `parent` award id?

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+            "federal_action_obligation": "fed_funding_amount",
+            "non_federal_funding_amount": "non_fed_funding_amount",
+            "cfda_number": "cfda_program_num",
+            "cfda_title": "cfda_program_title",
+            "uri": "unique_transaction_id",
```
afrasier: `unique_transaction_id` is actually a USASpending internal transaction id. In contract loader, I store this in a field called `usaspending_unique_transaction_id` on the procurement object. It might be worth refactoring that into the base AwardAction if it is shared among the two. This field allows you not to re-load records you've already seen in USAspending

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+            "face_value_loan_guarantee": "face_loan_guran",
+            "original_loan_subsidy_cost": "orig_sub_guran",  # ??
+            "award_description": "project_description",  # ??
+        }
+
+        # TODO: csv contains `exec1_amount`... `exec5_amount` and
+        # `exec1_fullname`... `exec5_fullname`.  We seem to be ignoring those.
```
afrasier: We are indeed ignoring these for now - I don't think we have models for 'highlycompensatedofficers' yet

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # TODO: What are all the drv_ fields in financial_assistance_award?
```
afrasier: These exist on a bunch of models for "derived" fields, I don't think we're using them yet. You'd have to ask Bernard or Alex about them since they made them. Not sure where they came from.

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # progsrc - program source?
+        # I’ve also got a “progsrc_acnt_code”, “progsrc_agen_code”, “progsrc_subacnt_code”, which I’m guessing stands for Program Source, and corresponds to… Funding Agency as opposed to Awarding Agency?
+        # none of those really look like enough to identify an agency
+        # how to identify the funding agency?
```
afrasier: Not sure. For data questions, I'd ask Kristine Sweet as she's our guru on the data

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+
+        value_map = {
+            "data_source": "USA",
+            "submitted_type": "C",  # ?? For CSV?
+            "recipient": self.get_or_create_recipient,
+            "award": self.get_or_create_award,
```
afrasier: Get or create award here is going to override your "award_id" from earlier anyway, as the threaded data loader prefers `value_map` over `field_map` :)

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+            "last_modified_date":
+            lambda row: h.convert_date(row['last_modified_date']),
+            "transaction_number":
+            lambda row: self.parse_first_character(row['transactionnumber']),
+            "solicitation_identifier":
+            lambda row: self.parse_first_character(row['solicitationid']),
+            "submission": subattr
+        }
+
+        loader = ThreadedDataLoader(
```
afrasier: A good addition here is to set `collision_field="usaspending_unique_transaction_id", collision_behavior="skip"` when you add it to the model, this way you don't reload data that is already there (unless you specifically want to!), in which case you can use `collision_behavior="delete"`

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+            FinancialAssistanceAward, field_map=field_map, value_map=value_map)
+        loader.load_from_file(options['file'][0])
+
+    def get_or_create_award(self, row):
+        fain = row.get("federal_award_id", None)
+        award = Award.get_or_create_summary_award(fain=fain)
+        return award
+        # TODO: really don't know how to detect parent awards
```
afrasier: I'll have to take a look at some example data, let's have a Slack chat sometime to clear up any ambiguous/unknown column mapping

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+            # or indian_tribe_federally_recognized ?
+            'Other nonprofit': 'nonprofit_organization',
+            'Private higher education': 'educational_institution',
+            'Individual': 'individual',
+            'Profit organization': 'for_profit_organization',
+            'Small business': 'small_business',
+            # should the more specific contract flags for small businesses also set this to Y?
+            'All other': '',
+        }
+        flag = mappings.get(type_name, "")
```
afrasier: Are these guaranteed to have the same capitalization/wording each time? (I hope so...)

usaspending_api/awards/management/commands/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+        Q(country_code=code_str) | Q(country_name__iexact=code_str)).first()
+    if not country_code:
+        # We don't have an exact match on the name or the code, so we need to
+        # chain filter on the name
+        query_set = RefCountryCode.objects
+        for word in code_str.split():
+            query_set = query_set.filter(country_name__icontains=word)
+        country_code = query_set.first()
+
+    return country_code
```
catherinedevlin: Makes sense.  I'll just be aware that calling a string method on it might error, but I'm not doing any of that.  And our loaders are populating the database with a blank in those cases.  (I guess that relates a little to the comment above, because we could fill with NULLs instead.)

usaspending_api/awards/management/commands/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+
+from usaspending_api.references.models import Location, RefCountryCode
+
+
+def up2colon(input_string):
+    'Takes the part of a string before `:`, if any.'
+
+    if input_string:
+        return input_string.split(':')[0].strip()
+    return ''
```
catherinedevlin: loadcontracts and loadfinancialassistance use a lot of `row.get["something", ""]`, so at least in those cases, we've decided to save a `None` as a blank string instead of a `NULL` in the database.   I don't know which way we want to go, but we should probably be as consistent as we can.

usaspending_api/awards/management/commands/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+    if country_code.country_code == "USA":
+        location_dict.update(
+            location_zip5=location_dict["location_zip"][:5],
+            location_zip_last4=location_dict["location_zip"][5:])
+        location_dict.pop("location_zip")
+    else:
+        location_dict.update(
+            location_foreign_postal_code=location_dict.pop("location_zip"),
+            location_foreign_province=location_dict.pop("location_state_code"))
+        if "location_city_name" in location_dict:
```
catherinedevlin: Oh, good idea, thanks.

usaspending_api/awards/management/commands/loadfinancialassistance.py
```diff
@@ -0,0 +1,189 @@
...
+
+import usaspending_api.awards.management.commands.helpers as h
+from usaspending_api.awards.models import Award, FinancialAssistanceAward
+from usaspending_api.common.threaded_data_loader import ThreadedDataLoader
+from usaspending_api.references.models import Agency, LegalEntity, Location
+from usaspending_api.submissions.models import SubmissionAttributes
+"""
+
+No PIID in the download from https://www.usaspending.gov/DownloadCenter/Pages/dataarchives.aspx ?
+
```
catherinedevlin: Hmm, the CSVs I've been using always had a FAIN, but I guess I'd better be ready to use either.  Thanks.

usaspending_api/awards/management/commands/loadfinancialassistance.py
catherinedevlin: Darn.  I was guessing at federal_award_id being FAIN because I thought we really needed a FAIN, and there doesn't seem to be any other field for it.  But knowing that we can get by with a URI helps.

afrasier: I think FAINs are actually a recent addition. Depending on the FY you look at it might decide to show up.

usaspending_api/awards/management/commands/loadfinancialassistance.py
catherinedevlin: Great, thanks.

usaspending_api/awards/management/commands/loadfinancialassistance.py
catherinedevlin: Ah, looks like you (or somebody) already had it in the `AwardAction` - great!

usaspending_api/awards/management/commands/loadfinancialassistance.py
catherinedevlin: No idea... I can lowercase it to be on the safe side, good idea.  What I don't know yet is how much variability there might be - whitespace?  Punctuation?  Alternate terms?  Word order?  Could be scary.

usaspending_api/awards/management/commands/loadfinancialassistance.py
catherinedevlin: Turns out that I've got a lot of records with no `fain` and blank `uri`.  I don't know what to use to call `Award.get_or_create_summary_award` with in that case.  If multiple FA records can be tied together into a single award, I don't see a field in the CSV to help me do that.  On the other hand, if each row in my source csv is going to have its own award, I could just generate and use a GUID for each row.

bsweger: The lack of `fain` and `uri` is something to raise at standup or in the channel. As you've seen, the entire financial assistance award load process is based on the assumption that one of those values will be present, so we'll have to get some SME guidance on what to do in those cases.

PR #116, opened 2017-01-23
Agency autocomplete update

afrasier:
* Agency autocomplete endpoint
* Includes upgrades to the autocomplete endpoint to allow returning
matching objects, as well as introducing a “limit” function

### DIFF
--- a/usaspending_api/common/api_request_utils.py
+++ b/usaspending_api/common/api_request_utils.py
@@ -8,6 +8,7 @@ from django.utils import timezone
 
 from usaspending_api.common.exceptions import InvalidParameterException
 from usaspending_api.references.models import Location
+from usaspending_api.awards.models import Award
 
 
 class FiscalYear():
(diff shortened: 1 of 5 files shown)


usaspending_api/common/api_request_utils.py
```diff
@@ -289,35 +290,71 @@ class AutoCompleteHandler():
     @staticmethod
     # Data set to be searched for the value, and which fields to look in
     # Mode is either "contains" or "startswith"
-    def get_values_and_counts(data_set, fields, value, mode="contains"):
+    def get_values_and_counts(data_set, filter_matched_ids, pk_name):
         value_dict = {}
         count_dict = {}
 
+        for field in filter_matched_ids.keys():
+            q_args = {}
```
catherinedevlin: This `q_args = {}; q_args[]` every time through the loop makes it look as if maybe (?) the intent is to keep building up the dict on each pass through the loop, in which case it won't work right for multiple fields b/c the dict is reset on each iteration.  I can't tell if that's the intent, though.  If I'm right, `q_args = {}` should go up above the beginning of the loop.  If I'm wrong, replacing both lines with `q_args = {pk_name + "__in": filter_matched_ids[field]}` would be slightly more concise and not look like a potential bug.

usaspending_api/common/api_request_utils.py
```diff
@@ -289,35 +290,71 @@ class AutoCompleteHandler():
     @staticmethod
     # Data set to be searched for the value, and which fields to look in
     # Mode is either "contains" or "startswith"
-    def get_values_and_counts(data_set, fields, value, mode="contains"):
+    def get_values_and_counts(data_set, filter_matched_ids, pk_name):
```
catherinedevlin: The comment above still refers to a `mode` parameter

usaspending_api/common/api_request_utils.py
catherinedevlin: Looking at it more, I'm pretty sure the intent is for `q_args` to only have one key each time through, so this isn't a bug, just an optional code style suggestion.

afrasier: Right, I left it that way because I was working on another feature that I ended up taking out - this style was leftover from it.

PR #119, opened 2017-01-25
Txn endpoint

bsweger:
- [x] Back-end code review
- [x] Request/response review from @kevinli-work (refer to the updated [summary endpoints piece](https://github.com/fedspendingtransparency/usaspending-api/blob/txn-endpoint/using_the_api.md#summary-endpoints-and-methods) of the API document

This PR adds two endpoints to the API, both with the goal of supporting the spending over time chart on the front-end:

* `/transactions/`: A first pass at a transaction endpoint (until we update our AwardAction model, this endpoint is based on the Procurement model and, thus, returns only contract data)
* `/transactions/total`: The same aggregate functionality currently on the awards endpoint, now on award transaction data


bsweger:
@afrasier Sort of--I tried to mock up the structure of the response to be something close to what we've discussed in the channel. So the fields that are relevant to both awards types are presented in the top level of the response, and the fields that pertain to contracts only are nested in `contract_data`:

My thinking is that when we stand up the `Transaction` table (aka the old `AwardAction` abstract model), this endpoint will pull from there and fill in the `contract_data` and `assistance_data` parts of the request from those respective tables.

Currently, you can only get a good sense of it when making a verbose request, because of the way our default fields are working:

**Request:**

```json
{
	"limit": 1,
	"verbose": true
}
```

**Response:**

```json
{
  "total_metadata": {
    "count": 22
  },
  "page_metadata": {
    "page_number": 1,
    "num_pages": 22,
    "count": 1
  },
  "results": [
    {
      "award": 47924,
      "type": "A",
      "type_description": "BPA Call",
      "modification_number": "0",
      "federal_action_obligation": "3200.00",
      "action_date": "2016-09-06",
      "description": "IGF::OT::IGF TOWER 2 LEAK REPAIR FOR HVAC UNIT AT HERNDON, VA",
      "update_date": "2016-12-20T22:50:57.657000Z",
      "contract_data": {
        "naics": "238220",
        "cost_or_pricing_data": "J",
        "naics_description": "238220: PLUMBING, HEATING, AND AIR-CONDITIONING CONTRACTORS",
        "product_or_service_code": "J099"
      },
      "assistance_data": {},
      "awarding_agency": null,
      "recipient": null,
      "funding_agency": null
    }
  ]
}
```

PR #120, opened 2017-01-25
Location pop recipient tags

afrasier:
* Add `recipient_flag` and `place_of_performance_flag` to location objects
* Add filtering by those tags to the hierarchical geographic autocomplete 
* Add management command `update_location_usage_flags` to update these flags. This can be called after ETL to ensure the flags are properly set. It uses set updates so it is pretty swift.
* Updated load contracts to set these flags properly by default
* Updated docs


usaspending_api/references/management/commands/update_location_usage_flags.py
```diff
@@ -0,0 +1,43 @@
...
+    def handle(self, *args, **options):
+        annotated_set = Location.objects.annotate(legalentity_count=Count('legalentity'))
+        annotated_set = annotated_set.annotate(award_count=Count('award'))
+        annotated_set = annotated_set.annotate(procurement_count=Count('procurement'))
+        annotated_set = annotated_set.annotate(assistance_count=Count('financialassistanceaward'))
+
+        # Locations have a recipient flag of true if the number of legal entities using
+        # them is greater than or equal to 1
+        # We could combine all of the following procedures into one line, but that would
+        # greatly reduce readability
```
bsweger: Good call to prioritize readability!

using_the_api.md
```diff
@@ -606,6 +607,11 @@ This is a special type of autocomplete query which allows users to search for ge
     * `domestic` - Matches only entries with the United States as the `location_country_code`
     * `foreign` - Matches only entries where the `location_country_code` is _not_ the United States
     * `all` - Matches any location entry. This is the default behavior
+  * `usage` - _Optional_ - The usage of the search. Options available are:
+    * `recipient` - Matches only entries where the location is used as a recipient location
+    * `place_of_performance` - Matches only entries where the location is used as a place of performance
+    * `both` - Matches only entries that serve as both recipient location and place of performance
```
bsweger: Ah, this answers the `both/all` question I had in `api_request_utils`. So my new question: why do we need a both option if we have an all option?

usaspending_api/references/management/commands/update_location_usage_flags.py
bsweger: Thinking about it more, I believe we could simplify things further by doing something like:

```python
# get a queryset of location records with ids that are present in the legal entity location_id field
le_locations = Location.objects.filter(location_id__in=LegalEntity.objects.values('location'))
# update the recipient flag on the locations in that queryset
le_locations.update(recipient_flag = True)

# repeat for place of performance
```

usaspending_api/common/api_request_utils.py
```diff
@@ -429,6 +430,14 @@ def build_response(self):
         elif scope == "domestic":
             scope_q = Q(**{"location_country_code": "USA"})
 
+        usage_q = Q()
+        if usage == "recipient":
+            usage_q = Q(recipient_flag=True)
+        elif usage == "place_of_performance":
+            usage_q = Q(place_of_performance_flag=True)
+        elif usage == "both":
```
afrasier: "all" is when you don't care about if it is pop or recipient, both is if you want the case where both are true. Slightly different use case. "all" has no filters, so it doesn't need to have a case here.

usaspending_api/references/management/commands/update_location_usage_flags.py
```diff
@@ -0,0 +1,43 @@
...
+import django
+
+
+class Command(BaseCommand):
+    help = "Updates Location object usage flags based upon their current usage."
+    logger = logging.getLogger('console')
+
+    def handle(self, *args, **options):
+        annotated_set = Location.objects.annotate(legalentity_count=Count('legalentity'))
+        annotated_set = annotated_set.annotate(award_count=Count('award'))
```
afrasier: True, but it is better to be explicit here in case some edge case creeps up.

using_the_api.md
afrasier: 'both' won't return a location that is neither recipient nor pop, 'all' will.

usaspending_api/references/management/commands/update_location_usage_flags.py
bsweger: Yeah, if we move toward the more simplified version of the update, it would be no biggie to handle the award table too.

PR #124, opened 2017-01-26
Initial pytest conversion of Django TestCases

catherinedevlin:
This is a simpleminded direct conversion to using py.test; will be followed by more work to take advantage of pytest capabilities

### DIFF
--- a/usaspending_api/references/test_cfda_loader.py
+++ b/usaspending_api/references/test_cfda_loader.py
@@ -1,4 +1,3 @@
-from django.test import TestCase, Client
 from usaspending_api.references.models import CFDAProgram
 from django.core.management import call_command
 from django.conf import settings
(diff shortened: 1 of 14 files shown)


usaspending_api/references/test_cfda_loader.py
```diff
@@ -1,31 +1,37 @@
...
+    """
+    call_command('loadcfda')
+
+
+@pytest.mark.django_db
+def test_program_number():
+    """
+    Make sure an instance of a program number is properly created
+    """
+    call_command('loadcfda')
```
bsweger: I know you didn't write these tests initially, but I wonder if it's really necessary to run the `loadcfda` command three times? I suspect it might work to combine these checks into a single test and only run the loader once or use a fixture if it's important to do these checks as individual tests. What do you think?

usaspending_api/references/test_cfda_loader.py
```diff
@@ -1,31 +1,37 @@
...
+    """
+    Make sure a account identication is properly mapped to program_number
+    """
+    call_command('loadcfda')
+    account_identification = CFDAProgram.objects.get(
+        program_number='98.009', account_identification='12-2278-0-1-151.')
+    #        assert(subtier.department == department)
+
+
+# TESTING TODO: put these call_commands in a fixture so they get cleaned up?
```
bsweger: Ah, I see you've already been thinking about this! If combining or fixturizing these could help reduce the run time of our test suite, I'm all for doing it as part of this PR if there's time.

usaspending_api/submissions/test_load_submission.py
```diff
@@ -1,47 +1,52 @@
...
 from usaspending_api.submissions.models import SubmissionAttributes
 
 import pytest
 
 
-class LoadSubmissionTest(TestCase):
+@pytest.fixture(scope="module")
+def endpoint_data():
+    call_command('flush', '--noinput')
+    call_command('loaddata', 'endpoint_fixture_db')
```
bsweger: This is outside the current scope of my pytest knowledge, but I wonder if there's a way to define a fixture more globally, so that all of our test files can access it. It seems there are a few tests that have to run through this endpoint loading process, and it would be nice if we only had to do it once. Presumably there's a way to rollback any db modifications that are made during an individual test, so that the initial snapshot remains the same from test to test?

This is all purely speculative 😄

PR #125, opened 2017-01-26
New homepage for the API

kaitlin:
**Note**: This should be merged after https://github.com/fedspendingtransparency/usaspending-api/pull/123  is merged, as I merged that branch into this branch already. 

This PR includes a homepage for the API. It essentially just includes a basic template at the top level domain, and pulls in the `using_the_api.md` file to generate the content. It splits out the table of contents from that file to create a sticky side menu, so I moved the TOC to the top of the markdown file for easier parsing. 

Note that you'll need to `pip install -r requirements.txt` to install the github flavored markdown python extension.

Also, I had to convert the table of contents menu and header anchor tags to be html to really leverage the styling, but everything else remains markdown. The markdown parser is a little more finicky than Github's -- most notably, you need multiline blocks of code to be un-indented and have newlines before and after to trigger the proper code block styling.


kaitlin:
Ok it looks like merging from dev into this branch fixed the code climate issues. They were on a migration that I hadn't actually edited. 

The yellow background should be fixed.. I pushed a commit last night to fix the styling and make it a dark background. I also added commit to fix one of the anchor tags.

PR #126, opened 2017-01-27
Geocomplete Tweaks

afrasier:
- [x] backend review
- [x] kudos from @mab6bf 
* Make geocomplete hierarchy an ordered dict so that results return with the following preference:
Country > State > County > City > Zip > District
* Sort the responses alphabetically
* Accept a limit, default to 10


mab6bf:
[run kudos]

usaspending_api/common/api_request_utils.py
```diff
@@ -464,12 +466,16 @@ def build_response(self):
                     "matched_ids": Location.objects.filter(Q(**{"location_congressional_code": row[0], "location_state_code": row[1], "location_state_name": row[2]})).values_list("location_id", flat=True)
                 }
                 response_object.append(response_row)
+                if len(response_object) >= limit:
```
bsweger: To make sure we don't actually pull back more records than we need from the db, would it be possible to limit results when creating the queryset instead?

I'm guessing you might have wanted everything back so you could sort it. Any way to do sorting in the queryset as well? 

We're already seeing db performance degradation as we load more data, so I want to make sure we avoid transporting extra data/looping through results if we're able to perform the equivalent logic at the queryset level.

afrasier: My bad, I didn't realize we could use `distinct()` on a values list. We now sort and limit in the query, but this line is still needed since we _may_ not fill our response list in a single query. (I.E. a search for "WASHINGTON" would give us Washington state, but we should cascade down and fill the remaining 8 slots with cities named "Washington")

bsweger: Thanks, @afrasier. As I look at this, I keep getting the feeling that we could go even deeper re: what we can accomplish in the query vs having to manually spin through its results. 

For example, do you think it would be possible to annotate the queries on lines 457 and 472 to return `place`, `place_type`, and `parent` and then do an additional join to pull in the `matched_ids`? And just serialize/return a response within the confines of DRF instead of crafting a bespoke response?

This is a high-level spitball, and maybe it wouldn't work. But would be a good exercise to think through as we work on our endpoints that don't inherit from the ViewSets.

afrasier: After toying with it for a while, I don't think it will work. This is a case where we're not returning an actual model from the DB, but a custom response so we kind of have to do it this way. Unless we want to make some new "hierarchical geography" model and a new endpoint for that.

I also think it is probably faster to do `SELECT ids WHERE blah` vs. trying to aggregate the id's together when executing a distinct, which is what we'd have to do; unless I've missed some nifty django function.

bsweger: Thanks for poking around at this. Fair point that this isn't a model-based endpoint, and I def. don't think we need to make a hierarchical geography model. What I would like to do--and we don't need to do this now--is to migrate the functions in `api_request_utils.py` to a place consistent with DRF (common.mixins, etc).

bsweger:
@afrasier Thanks for the additional work. My only remaining question is concerning unit tests. Now that we're fully converted to pytest, it's a good time to start breaking our code into testable chunks and exercising the edge cases.

afrasier:
Yes, I agree. I was rolling them into my other location based story; but it's past code-freeze so they won't be in until after Wednesday.

PR #128, opened 2017-01-30
Support single item verbose

bsweger:
Apologies for this large PR--the original intent was to finish up the single award story by making sure that single-item requests return all data (not just the default field set).

During development, I also began streamlining our pagination code by moving towards the DRF out-of-the-box pagination. For example, we now no longer need a custom ViewSet just to return request metadata in our custom format.

This PR also adds create/update date fields to the SubTier and TopTier models, to be consistent with our other models.

The PR also makes improves the performance of the test suite a bit by reducing the number of test fixtures loaded (a good next step would be to test the cfda, tas, etc. loaders with test files instead of full files).

Lastly, there are some documentation updates that add info re: pagination tweaks and the new {PK} format endpoints, as well as adding some additional clarity (for example, more robust endpoint descriptions).


using_the_api.md
```diff
@@ -34,65 +35,75 @@ In general, status codes returned are as follows:
 
 ### Data Endpoints <a name="data-endpoints"></a>
 
-Data endpoints are split by payload into POST and GET methods. In general, the format of a request and response will remain the same with the endpoint only changing the data provided in the response.
+Data endpoints are split by payload into POST and GET methods. In general, the format of a request and response will remain the same between endpoints.
```
afrasier: "among" not "between" :)

afrasier:
By pagination directions not working in post requests, you mean it requires modifying the url to traverse pages regardless of the method used to get the data?

bsweger:
@afrasier To the question re: pagination and post requests... Django Rest Framework has great pagination support, but it requires that pagination directives be query parameters rather than items in the body of post request. 

You can see that in the pagination code, which is looking for pagination directives in `request.query_params`: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/pagination.py#L206

I was thinking about this, and ultimately, I don't think it's a bad thing to have one place where we're expecting pagination parameters. Rather than thinking about it in terms of `GET` vs `POST` requests, I started thinking in terms of common, non-field-related parameters that are applicable regardless of request type.

In addition to getting some neat stuff from the DRF pagination, I like having less code to maintain and having things live in a place consistent with the DRF docs/examples to ensure that future maintainers have an easier lift.

For now, I updated the pagination code to return metadata in the same way we have it now: https://github.com/fedspendingtransparency/usaspending-api/blob/support-single-item-verbose/usaspending_api/common/pagination.py#L29.

If the front-end team signs off, I'd like to move to a flatter metadata format that includes next/previous page links for API consumers who want to use them. It looks like this: 

**Request**
```http://localhost:8000/api/v1/awards?limit=15&page=2```

**Response**
```json
{
  "total_count": 39,
  "page_count": 15,
  "page_num": 2,
  "page_total": 3,
  "next": "http://localhost:8000/api/v1/awards/?limit=15&page=3",
  "previous": "http://localhost:8000/api/v1/awards/?limit=15",
  "results": [
    {some results}
]
}
```

afrasier:
I don't see a problem with the flatter metadata - the only reason I had it separated as it was was because we were doing that one aggregation on a per-page basis; but since we now have the aggregation endpoints we can probably get rid of it.  We could just write some mixin to duplicate limit/page from the POST body into the request params in that case, so we maintain current functionality of all parameters being acceptable via POST.

usaspending_api/common/serializers.py
```diff
@@ -19,56 +19,64 @@ def __init__(self, *args, **kwargs):
...
+                }  # The child tag should be removed when child field limiting is implemented
                 self.fields[field] = children[field]["class"](**child_args)
         except AttributeError:
             # We don't have any nested serializers
             pass
 
         request = self.context.get('request')
-        if request and not self.context.get("child", False):  # The child check should be removed when child field limiting is implemented
-            # workaround to use this serializer with both GET and POST
+        if request:
```
afrasier: Removing the child check here - does the endpoint still behave properly when a filtered field is that of a child? i.e.:
```
"filters": [
{
      "field": "procurement__type",
      "operation": "is_null",
      "value": false
}
]
```

Without this check before, it would return the summary award X times, where X is the number of procurements. Your other changes may account for this but I want to make sure it doesn't creep back in.

bsweger: @afrasier Thanks for flagging! I removed the child check from this line to ensure that we display all fields on all child objects when the request is for a single item.

The child check is now being done on [line 44](https://github.com/fedspendingtransparency/usaspending-api/blob/support-single-item-verbose/usaspending_api/common/serializers.py#L44) and [line 53](url). The intent was to make sure that we don't mess with the serializer's field lists if we're in a nested serializer.

When testing, I got the same results in dev as in the branch. Can you clarify that what I'm describing is actually the issue you're raising? Would like to make sure this PR doesn't re-introduce an old issue.

afrasier: If the branch matches dev this should be fine!:)

PR #129, opened 2017-01-31
Reference City County Data

afrasier:
Not to be dealt with till after sprint review since this is past code freeze; but I’m dropping it in here so people can have a gander.

* Added a command `load_reference_data` that performs all reference
data loading in once command
* Migration to fix field length mismatch between location and reference
data
* Override location save method to update data based upon reference
country, city, and county codes
* Added test cases for new functionality, as well as some old
functionality: geocomplete scope, geocomplete limits, geocomplete
congressional code search, geocomplete usage search


usaspending_api/references/tests/test_location.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.mark.django_db
+def test_location_reference_fill():
+    country_code = mommy.make('references.RefCountryCode', country_code="USA", country_name="United States", _fill_optional=True)
+    city_county_code = mommy.make(
+        'references.RefCityCountyCode', city_code="A", county_code="B", _fill_optional=True)
+    location = mommy.make(
+        'references.Location', location_country_code=country_code, location_city_code="A", location_county_code="B")
+    assert location.location_city_name == city_county_code.city_name
```
bsweger: This is great--I personally aspire to write more tests against our model methods, so thanks for providing a good example.

usaspending_api/references/tests/test_location.py
```diff
@@ -0,0 +1,91 @@
...
+    assert location.location_county_name == city_county_code.county_name
+    assert location.location_state_code == city_county_code.state_code
+    assert location.location_country_name == country_code.country_name
+
+
+@pytest.mark.django_db
+def test_geocomplete_scope():
```
bsweger: And again, good stuff on the unit testing--exactly the direction we need to be going.

This is a minor point, but I've been thinking about these types of tests that are more specific to our common code than to a particular app or endpoint. If the code we're testing is something in `common/api_request_utils.py`, do you think the test should live there?

usaspending_api/references/tests/test_location.py
```diff
@@ -0,0 +1,91 @@
...
+    foreign_response = GeoCompleteHandler({"value": "a", "scope": "foreign"}).build_response()
+    assert len(no_scope_response) == 2
+    assert domestic_response[0]["place"] == "United States"
+    assert len(domestic_response) == 1
+    assert foreign_response[0]["place"] == "Canada"
+    assert len(foreign_response) == 1
+
+
+@pytest.mark.django_db
+def test_geocomplete_limit():
```
bsweger: Same question here about the test location. Again, it's a minor point, and not something worth holding up a merge. The tests are great!

usaspending_api/references/tests/test_location.py
afrasier: Yeah, I just put it in the Location tests because it felt "logical" to me, as it dealt with location objects. We could really put it either place. Feel free to move it later if you want :) But in the interest of time I'm just going to leave it here for now.

usaspending_api/references/models.py
```diff
@@ -181,6 +182,39 @@ def get_default_fields():
...
+                "city_code": self.location_city_code,
+                "county_code": self.location_county_code,
+                "state_code": self.location_state_code,
+                "city_name": self.location_city_name,
+                "county_name": self.location_county_name
+            }
+            # Clear out any blank or None values in our filter, so we can find the best match
+            q_kwargs = dict((k, v) for k, v in q_kwargs.items() if v)
+            matched_reference = RefCityCountyCode.objects.filter(Q(**q_kwargs))
+            # We only load the data if our matched reference count is one; otherwise,
```
afrasier: Good point. Logging now, don't know why I wasn't

afrasier:
Added the .md file for tracking those changes to models on save, and doc'd the Location changes.

We don't need to have a data migration for this, because (1) we're going to need to reload the data once the data loaders are fixed to pick up all columns (they're currently missing a few) (2) we can just run the command `update_location_usage_flags` which will update all locations and save them, thereby calling our save override if we need to update them before our re-load

PR #135, opened 2017-02-02
Add pkid to agency

afrasier:
* Update agency serializer to display id for the unique combination of top/sub/office agencies


usaspending_api/references/serializers.py
```diff
@@ -7,32 +7,42 @@ class ToptierAgencySerializer(LimitableSerializer):
 
     class Meta:
         model = ToptierAgency
-        fields = ('cgac_code', 'fpds_code', 'name')
+        fields = '__all__'
```
bsweger: Thanks for fixing this!

usaspending_api/references/models.py
```diff
@@ -63,6 +63,15 @@ class Agency(models.Model):
     subtier_agency = models.ForeignKey('SubtierAgency', models.DO_NOTHING, null=True)
     office_agency = models.ForeignKey('OfficeAgency', models.DO_NOTHING, null=True)
 
+    @staticmethod
+    def get_default_fields():
+        return [
+            "id",
```
bsweger: Do you think we need to return `id` in the toptier, subtier, and office default field list too? Or am I straying into foolish consistency territory?

afrasier: I don't think it's needed, as the cgac/fpds/subtier codes should be enough to uniquely identify them at that point; and the agencies most people will be searching on is the combination of the 3

PR #138, opened 2017-02-03
Fixing n+1 problem

afrasier:
This PR fixes performance issues relating to the typical ORM N+1 issue.
Locally, this took a 3,600 query, 40s call to `/awards/` down to 5 queries, 1.2s 

* Changed a bunch of serializers to `LimitableSerializer` and moved
nested serializers into `Meta.nested_serializers`
* Added code to prefetch and select related nested serializer fields
based upon the `Meta` class
* Updated views to use the new eager loading method
* Updated autocomplete to use eager loaded when needed
* Codeclimate fixes

### DIFF
--- a/usaspending_api/common/api_request_utils.py
+++ b/usaspending_api/common/api_request_utils.py
@@ -349,6 +349,11 @@ class AutoCompleteHandler():
         except:
             raise
 
+        # If the serializer supports eager loading, set it up
+        if serializer:
+            if hasattr(serializer, "setup_eager_loading") and callable(serializer.setup_eager_loading):
+                data_set = serializer.setup_eager_loading(data_set)
+
         return_object = {}
 
         filter_matched_ids, pk_name = AutoCompleteHandler.get_filter_matched_ids(data_set.all(), body["fields"], body["value"], body.get("mode", "contains"), body.get("limit", 10))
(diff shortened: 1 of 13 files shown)


usaspending_api/common/api_request_utils.py
```diff
@@ -349,6 +349,11 @@ def handle(data_set, body, serializer=None):
         except:
             raise
 
+        # If the serializer supports eager loading, set it up
+        if serializer:
+            if hasattr(serializer, "setup_eager_loading") and callable(serializer.setup_eager_loading):
+                data_set = serializer.setup_eager_loading(data_set)
+
```
afrasier: Before it's asked, this is to perform eager loading in the case that the autocomplete is returning objects, or the autocomplete field is on a nested child. The checks for `hasattr` and `callable` are safety measures if someone uses autocomplete with a serializer that does not inherit from `LimitableSerializer`

usaspending_api/common/serializers.py
```diff
@@ -78,6 +77,38 @@ def __init__(self, *args, **kwargs):
                     # We don't have get default fields available
                     pass
 
+    '''
```
bsweger: Tiny thing--can you move this great info into the docstring of `setup_eager_loading`?

afrasier: Sorry, what do you want moved?

bsweger: Moving the comment to what is now known as line 93 (re: after `def_setup_eager_loading`) will make it a [docstring](https://www.python.org/dev/peps/pep-0257/)

PR #139, opened 2017-02-06
Include Award.date_signed__fy by a lookup registered on the dj field …

catherinedevlin:
…type.

Filtering by __fy supported on all date and datetime fields.
Includes notes in common/helpers.py on how to add other __fy fields to endpoints.
Fields are not yet indexed in db, however.

#DS-531


catherinedevlin:
- Thanks @afrasier for the idea to use a lookup/transform

- This PR doesn't create a function-based index on FY(date_signed), because I didn't see where we're creating indexes (don't see any `db_index=True` fields in our models).  We'll need that index for good performance, though.

catherinedevlin:
@bsweger Well, the data is just as available to SQL users - they can `SELECT FY(date_signed) FROM awards` or whatever.  The only benefit to them that I can think of from having a physical column is that it will show up with `\d` and various tools to inspect the table structure.  But we should be able to offset that with documentation.

I also like that using a function makes it automatically available on all date / datetime fields, and automatically in synch - otherwise, every time we add a field (and we do seem to be in significant flux), there's a risk of forgetting to add the FY fields or forgetting to write or test the save method.  And if the synch depends on the ORM, there's the risk that somebody will inject data directly by SQL someday without realizing that there's crucial logic in the ORM. 

I have a half-finished branch with the physical column approach, and this one is definitely less code.

But I acknowledge that any sort of stored functions are a code smell by some ways of thinking, because it means there's some code "hiding" outside the regular app structure.

bsweger:
@catherinedevlin aah---ok...I wasn't making the connection to how someone could get the data in SQL. Yes, agreed that it's a tradeoff. I think you're right that having the flexibility to apply as needed vs doing a schema change every time we need to put FY magic onto a new date field is a good way to go. Thanks for clarifying and setting me straight!

PR #140, opened 2017-02-06
Ds 654 contract pricing

afrasier:
This was a quick PR to support front-end on DS-654

Changes
* Added `type_of_contract_pricing` to Procurement objects, and code to automatically populate those values
* Added `potential_total_award_value` to summary awards, which is populated when it has D1 data by the latest transaction
* Added fields to relevant default lists and serializers


usaspending_api/awards/models.py
```diff
@@ -384,12 +405,27 @@ def get_default_fields():
...
+            "type_of_contract_pricing_description",
             "naics",
             "naics_description",
             "product_or_service_code"
         ]
 
+    def get_pricing_type_description(self):
+        description = [item for item in CONTRACT_PRICING_TYPES if item[0] == self.type_of_contract_pricing]
+        if len(description) == 0:
+            return "Unknown Type"
```
bsweger: Back to the earlier question of whether or not contract pricing type is relevant/expected on all contract records. If it's not, maybe we should return `None` here. If it *is* something we expect on every record, should we also use some kind of "unknown" code in `type_of_contract_pricing` for consistency?

usaspending_api/awards/models.py
```diff
@@ -304,7 +324,8 @@ class Procurement(AwardAction):
     piid = models.CharField(max_length=50, blank=True)
     parent_award_id = models.CharField(max_length=50, blank=True, null=True, verbose_name="Parent Award ID")
     cost_or_pricing_data = models.CharField(max_length=1, blank=True, null=True)
-    type_of_contract_pricing = models.CharField(max_length=2, blank=True, null=True, verbose_name="Type of Contract Pricing")
+    type_of_contract_pricing = models.CharField(max_length=2, blank=True, null=True, choices=CONTRACT_PRICING_TYPES, verbose_name="Type of Contract Pricing")
```
bsweger: Is contract pricing type always a relevant piece of data for contracts? I'm trying to figure out if this pattern should match what we're doing on the award type/desc fields in the award model...where we default to `U`'/`Unknown`. But yeah, if contract pricing isn't always relevant, it makes sense to leave these as `NULL` with no default values.

usaspending_api/awards/models.py
afrasier: Sure, I can add that. I didn't because 'U' is actually a used type code for contract pricing, so I'll have to use a different string. (It is expected to get this data every time)

afrasier:
@bsweger I've added the default values for `type_of_contract_pricing`

usaspending_api/awards/models.py
```diff
@@ -324,7 +325,7 @@ class Procurement(AwardAction):
     piid = models.CharField(max_length=50, blank=True)
     parent_award_id = models.CharField(max_length=50, blank=True, null=True, verbose_name="Parent Award ID")
     cost_or_pricing_data = models.CharField(max_length=1, blank=True, null=True)
-    type_of_contract_pricing = models.CharField(max_length=2, blank=True, null=True, choices=CONTRACT_PRICING_TYPES, verbose_name="Type of Contract Pricing")
+    type_of_contract_pricing = models.CharField(max_length=2, default="UN", blank=True, null=True, choices=CONTRACT_PRICING_TYPES, verbose_name="Type of Contract Pricing")
     type_of_contract_pricing_description = models.CharField(max_length=150, blank=True, null=True, verbose_name="Type of Contract Pricing Description")
```
bsweger: If we're defaulting the code field, does it make sense to default the description field too, to match the pattern in use on `type_description` in the `Award` model?

afrasier: Description is never set by anything other than the procurement object itself, so if the type is given a default of "UN", then the description will also update to "Unknown Type". This is guaranteed via the save method.

bsweger: Yeah, I get that. On some level, I'm bothered that we do the same thing on `Awards` but have it set up slightly differently. Yet, that does not need to hold up this merge.

PR #141, opened 2017-02-07
Ds 653 financial data

afrasier:
This PR completes DS-653

* Adds `financial_set` to award serializer, containing file C data
* Restructures the linking from `FinancialAccountsByAwards` (fab) to link directly to `TreasuryAppropriationAccount` instead of relying on a pass-through of File A data. There is a data migration to update this relationship on existing awards; it uses raw SQL due to the fact that the `F()` functionality in django does not and is not planned to support joins
* Updates the eager loading algorithm to properly use `prefetch_related` when a child serializer's parent is using `many`, even if that child is not
* Updated `load_submissions` to use the new `treasury_account` mapping for File C data 
* Updated models to have appropriate default fields


usaspending_api/awards/migrations/0059_auto_20170207_1854.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.1 on 2017-02-07 18:54
+from __future__ import unicode_literals
+
+import django.db.models.deletion
+from django.db import connection, migrations, models
+
+
+def attach_tas(apps, schema_editor):
+    with connection.cursor() as cursor:
```
catherinedevlin: I've googled up claims that the "better" way to do this is with `migrations.RunSQL("UPDATE financial_accounts_by_awards...")` - though I'm not sure what they mean by "better" aside from saving one line of code and one import.  

I guess one other benefit is that you can define the *reverse* SQL in the same place instead of nesting it in a RunPython, FWIW.  Or (as in this case) just not define the reverse SQL, not wrap the RunSQL in a RunPython, and save maybe four lines.

https://markusholtermann.eu/2014/09/django-17-database-migrations-done-right/

afrasier: That is probably better, I will move it over to that then merge

PR #142, opened 2017-02-07
Add business_types descriptions

afrasier:
Small PR to support front end on DS-647

This adds business_types enum descriptions to the LegalEntity object, as well as a data migration to populate the data


usaspending_api/references/models.py
```diff
@@ -271,7 +300,8 @@ class LegalEntity(DataSourceTrackedModel):
     vendor_doing_as_business_name = models.CharField(max_length=400, blank=True, null=True)
     vendor_phone_number = models.CharField(max_length=30, blank=True, null=True)
     vendor_fax_number = models.CharField(max_length=30, blank=True, null=True)
-    business_types = models.CharField(max_length=3, blank=True, null=True)
+    business_types = models.CharField(max_length=3, blank=True, null=True, choices=LEGAL_ENTITY_BUSINESS_TYPES, default="UN")
```
afrasier: Not sure

usaspending_api/references/models.py
```diff
@@ -377,13 +407,25 @@ class LegalEntity(DataSourceTrackedModel):
     small_business = models.CharField(max_length=1, blank=True, null=True)
     individual = models.CharField(max_length=1, blank=True, null=True)
 
+    def get_type_description(self):
+        description = [item for item in LEGAL_ENTITY_BUSINESS_TYPES if item[0] == self.business_types]
```
afrasier: That's a good idea. This code is similar in a lot of other places, so we'd want to refactor there at the same time. We should make another story for that, since the front-end needs this sooner than later :)

PR #144, opened 2017-02-07
New tests for award autocomplete

catherinedevlin:
Not strictly speaking necessary, but they were the best way for me to figure out how our autocomplete works, and I can base my tests for DS-699 on these.


catherinedevlin:
Good idea - just pushed a new commit with matched_object tests.

PR #145, opened 2017-02-08
Aggregate tests

bsweger:
Adding some more robust tests to the aggregate mixin. There are a few more I'd like to do, but before proceeding would like to get feedback on paramterizing these.

I like the pytest parameter option, but I'm wondering if I'm not sacrificing readability for brevity here. It's hard to document for future readers what it is each input is trying to test?

What do you all think? [This commit](https://github.com/fedspendingtransparency/usaspending-api/commit/8180fbcb403825adefe29c9b17cffd6e8b53f714) shows an earlier, more verbose version of the tests.


usaspending_api/common/tests/test_aggregate_mixin.py
```diff
@@ -0,0 +1,128 @@
...
+    # Query should return two field names: 'item' and 'aggregate'
+    fields = agg.first().keys()
+    assert len(fields) == 2
+    assert 'aggregate' in fields
+    assert 'item' in fields
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('model, request_data, result', [
+    (Award, {'field': 'total_obligation', 'group': 'period_of_performance_start_date'}, [
```
catherinedevlin: I see the concern about legibility here.  It would add bulk, but maybe we could clarify with comments inside the tuples, like

    (Award,                                     # model
     {'field': 'total_obligation', ... },       # request_data
     [ {'item': datetime.date ...
     ]                                          # result  
     )
 

I don't think it's inherently bad to have a big decorator on top of a function, though.

bsweger:
@catherinedevlin Thanks for weighing in here. That's a good perspective, and I'll circle back to add some more test cases. In the meantime, this first batch is ready to go if the code passes muster!

bsweger:
@catherinedevlin ready for your 👀 again...just updated with the latest dev changes. Thanks!

PR #148, opened 2017-02-09
Ds 529 tas endpoint

afrasier:
This PR completes DS-529

* Adds File A data to TAS endpoint
* Adds File B data to TAS endpoint
* Migrations to move File B to link directly to TAS rather than via File A


usaspending_api/financial_activities/models.py
```diff
@@ -1,5 +1,5 @@
 from django.db import models
-from usaspending_api.accounts.models import AppropriationAccountBalances
+from usaspending_api.accounts.models import AppropriationAccountBalances, TreasuryAppropriationAccount
```
catherinedevlin: This PR makes AppropriationAccountBalances an unused import... we don't have anything automated in our toolchain for detecting these, actually.  I should make a ticket for looking into that, we've probably got a ton of them.

catherinedevlin: Made DS-714

PR #150, opened 2017-02-10
Update txn schema

bsweger:
This PR adds three new models:
* Transaction
* TransactionAssistance (OneToOne relationship with Transaction)
* TransactionContract (OneToOne relationship with Transaction)

Eventually, these will replace the AwardAction abstract class and it physical artifacts: Procurement and Financial Assistance Award. Those still exist until we do a data migration and update the data loaders.

The PR also adds some tests to ensure that updates to transaction data flow upward to their parent award as expected.

@catherinedevlin Hopefully this is enough to get started on data migrations while I tackle the loads. I tried not to change field names (it was tempting!) so that it would be more straightforward to update the data. Basically, we'll need to copy Procurement records to Transaction/TransactionContract and FinancialAssistanceAward records to Transaction/TransactionAssistance.


bsweger:
@catherinedevlin I see I need to address some failing tests post-dev-merge. Will have to take a look in the AM, but hopefully what's here is enough for you to consider approaches for the data migration.

catherinedevlin:
@bsweger Do you think I should add the data changes directly to 0064_auto_20170210_0416.py?   Like a `migrations.RunSQL` at the end of the `operations` list?  That way the data always comes with in the same operation.

bsweger:
@catherinedevlin good question. I've been burned in the past by putting the schema changes and data migrations into a single file. Postgres was complaining about uncompleted transactions, so I found it easier to split it up. That said, you're better positioned to sooth Postgres. 

Either way is probably fine, since we haven't actually removed the old tables yet or updated the DRF views. No matter what, we'll have to follow up with a final migration to delete the old stuff.

usaspending_api/awards/tests/test_models.py
```diff
@@ -0,0 +1,138 @@
...
+    award = mommy.make('awards.Award')
+    txn = mommy.make('awards.Transaction', award=award)
+    mommy.make(
+        'awards.TransactionContract',
+        transaction=txn,
+        potential_total_value_of_award=1000
+    )
+    assert award.potential_total_value_of_award == 1000
+
+
```
catherinedevlin: Marking this with `@pytest.mark.skip` might be better than an empty test - it will provide a reminder that we've got a desired but unwritten test

usaspending_api/awards/migrations/0064_auto_20170210_0416.py
```diff
@@ -0,0 +1,199 @@
...
+                ('submitted_type', models.CharField(blank=True, max_length=1, null=True, verbose_name='Submitted Type')),
+                ('certified_date', models.DateField(blank=True, null=True)),
+                ('create_date', models.DateTimeField(auto_now_add=True, null=True)),
+                ('update_date', models.DateTimeField(auto_now=True, null=True)),
+                ('period_of_performance_start_date', models.DateField(blank=True, null=True)),
+                ('period_of_performance_current_end_date', models.DateField(blank=True, null=True)),
+                ('submission', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='submissions.SubmissionAttributes')),
+            ],
+            options={
+                'db_table': 'transaction_asssistance',
```
bsweger: Whoa-great catch. That would be an awkwardly-named table for sure. Will fix!

usaspending_api/awards/tests/test_models.py
bsweger: Good idea--thanks!

bsweger:
@catherinedevlin ok, ready for another look!

catherinedevlin:
In doing the data migration, I am noticing a bunch of columns in `financial_assistance_award` that don't appear (at least by those names) in `transaction` or `transaction_assistance`.  Here's the query:

     select column_name, table_name 
     from information_Schema.columns 
     where table_name in  
       ('awards_procurement', 'transaction', 
        'financial_assistance_award', 'transaction_contract', 
        'transaction_assistance') 
      order by column_name, table_name;

`business_funds_indicator`, `cfda_number`, `cfda_title`,  `correction_late_delete_indicator`, `drv_federal_funding_amount`, `face_value_loan_guarantee`, `fain`, `record_type`, `sai_number`, for example... 

Is it OK to lose these columns?  Or do they exist under different names in the transaction + (transaction_contract or transaction_assistance) family?

PR #153, opened 2017-02-11
SQL-based migration to move existing data after procurement->transact…

catherinedevlin:
…ion change

**NOTE** hold off on merging until we're squared away with the front-end team.

Task DS-692

This approach turned out to be weirder than I expected, and I'm open to critiques\*.  I thought a sql-based one would be better than an ORM-based loop in case large amounts of data need migration.

\* - I mean, obviously, we're all open to critiques, that's why we're open-source programmers.

**Update 2/17**: added code to point the API at the new txn tables (see comments below)


bsweger:
@catherinedevlin As we discussed earlier in the week, am updating this branch with the code that points the API at the new transaction tables. Thus, after we merge this:
1. data from the old `AwardAction` series of tables gets migrated to the new `Transaction` series of tables (thank you for that!)
2. the API will consume the new tables

Will follow-up with a cleanup PR that removes the old stuff.

usaspending_api/awards/migrations/0065_auto_20170211_1623.py
```diff
@@ -0,0 +1,454 @@
...
+    ]
+    # migrations.noop b/c otherwise creating the
+    # temporary table throws an IrreversibleError
+
+
+'''
+These column lists were generated with queries on the
+information schema:
+
+-- 1. insertion to transaction_assistance
```
bsweger: This is neat. Thanks for leaving this here for future reference!

bsweger:
Front-end has given the ok with varying levels of enthusiasm, so gonna merge.

PR #154, opened 2017-02-13
Update usaspending loads for new transaction models

bsweger:
My apologies for this large PR. There are detailed messages on most of the individual commits, but the TL;DR of the PR:

1. Create a new ETL to hold all data loading code.
2. In that ETL app, copy over the usaspending contracts and assistance load code and modify it to work with the new `Transaction`, `TransactionContract` and `TransactionAssistance` models

A few notes:
* Because each incoming from USAspending now inserts into 2 tables instead of one, working within the construct of the threaded loader, which updates a single model at a time, became difficult, the new load commands don't use threading
* About the above, I believe the way forward is to focus on set-based data updates, rather than row-by-row processing that results in many db calls for each record. That's outside the scope of this PR, but I did use Django's bulk loading for the txn reecords and experimented with caching agency data in the contract loader
* Related, the .save() overrides on the txn models are gone, and replaced with a set-based update. This assumes that awards and transactions will never be updated outside the context of a submission or usaspending data load, which I think is a safe assumption. Overriding `save()` is very elegant, but but forces us to rely on individual row updates rather than set-based updates (for example, bulk inserts bypass `save()`.)
* There is some raw SQL in here. I know that's controversial,  but I wrestled the ORM for hours and finally decided it was worth it to update award values in a single pass rather than going through row by row

Since AwardAction, etc and Transaction, etc are both still active until we finish merging all related PRs, there's a bunch of duplicate code in the etl and awards apps. Once we're fully switched over, I'll submit a cleanup PR.


usaspending_api/etl/award_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Note: Obviously the raw SQL is not ideal. That said, the complex update
+    of award fields based on the earliest, latest, and aggregate values
+    of the child transactions was problematic to do in a set-based way
+    via the ORM. These updates do need to be set-based, as looping through
+    and updating individual award records would be an ETL bottleneck.
+
+    TODO: modify this to work with specific awards, so we aren't updating
+    every award, all the time
+    """
```
afrasier: Awards do have a FK to submission attributes, which is created for every broker/USASpending load - could use that to limit the update to only newly submitted items

usaspending_api/etl/award_helpers.py
```diff
@@ -0,0 +1,80 @@
+from django.db import connection
+
+
+def update_awards():
+    """
+    Update award data based on the information in their child transactions.
+
+    Note: Obviously the raw SQL is not ideal. That said, the complex update
```
afrasier: If we're certain this can't be done cleanly via the ORM; it would be useful to have a docstring explaining exactly what the SQL is doing so that it is more maintainable in the future for people less familiar with SQL.

bsweger: Good point--will do that.

usaspending_api/etl/award_helpers.py
bsweger: Oooooh--good idea. That is way simpler than what I was thinking of. Thanks for the suggestion!

usaspending_api/etl/award_helpers.py
```diff
@@ -0,0 +1,116 @@
...
+    # award. the joined data from earliest, latest, and summarized
+    # transactions are used to update awards fields as appropriate
+    sql_update = 'WITH {}, {}, {}'.format(sql_txn_latest, sql_txn_earliest, sql_txn_totals)
+    sql_update += (
+        'UPDATE awards a '
+        'SET awarding_agency_id = l.awarding_agency_id, '
+        'certified_date = l.certified_date, '
+        'data_source = l.data_source, '
+        'date_signed = e.action_date, '
+        'description = l.description, '
```
afrasier: Geoff has informed me that description should come from the earliest transaction; but he said everything else looked fine :)

bsweger: Great idea to put Geoff's eyes on this! Will make that change--thanks!

PR #161, opened 2017-02-21
DS-724: FY aggregation

catherinedevlin:
Wraps up aggregated results to let them use the __fy transform


usaspending_api/common/mixins.py
```diff
@@ -52,13 +52,31 @@ def aggregate(self, request, *args, **kwargs):
...
 
         return aggregate
 
+    _sql_function_transformations = {'fy': IntegerField}
+
+    def _wrapped_f_expression(self, col_name):
+        """F-expression of col, wrapped if needed with SQL function call
+
+        Assumes that there's an SQL function defined for each
+        registered lookup."""
```
afrasier: How does this play with fk traversal? (Or is this not supported in aggregate endpoints); i.e. "recipient__location__state_code"

catherinedevlin: It... does not play well, it turns out.  I took a crack at making it work, but it's not turning out to be easy.  So for now I'm adding a skipped test for it, and we can write a ticket.

afrasier: (Mainly because using_the_api.md says that fk traversal works here)

catherinedevlin: OK, reenabling that test.  Turns out it was just a poorly written test making it appear not to work...

PR #162, opened 2017-02-22
DS-605: Summary awards are specific to awarding agency

catherinedevlin:
As per https://federal-spending-transparency.atlassian.net/browse/DS-605, this makes the awards created or found for loaded contracts and assistance specific to awarding subtier agency.

Ready for review.


catherinedevlin:
OH HECK.  On contracts, I'm being tripped up by NULLS - multiple records with one piid, but null parent and awarding agency.  Curse you, three-value logic!  Working to fix...

catherinedevlin:
Turned out that the multiple records with one piid, but null parent and awarding agency are incorporated into our common/fixtures/endpoint_fixture_db.json itself.

namkyux:
No objections on the code logic from the changes made.

usaspending_api/etl/tests/test_usaspending_assistance_load.py
```diff
@@ -19,6 +22,72 @@ def test_usaspending_assistnace_load():
...
+    def fainchange(row):
+        return _replacer(row, 'federal_award_id', fain_prefixer)
+
+    def new_fain(row):
+        row['federal_award_id'] = '{}-fain2'.format(row['federal_award_id'])
+        return row
+
+    with mutated_csv(filepath, new_fain) as mutant_file:
+        call_command('load_usaspending_assistance', mutant_file.name)
+        print('ok')
```
bsweger: Is this a leftover debug?

usaspending_api/etl/tests/test_usaspending_assistance_load.py
```diff
@@ -19,6 +22,72 @@ def test_usaspending_assistnace_load():
...
+    assert Award.objects.count() == awards_before_loads + (2 * base_award_count
```
bsweger: This is great--thanks for adding such robust tests to this process. Confession: I had to look at the mutator function and this code a few times before grokking what's going on. Do you think it makes sense to augment the comments you already have by adding a note here that says you're now processing the .csv with different FAINSs?

usaspending_api/etl/tests/test_usaspending_assistance_load.py
```diff
@@ -19,6 +22,72 @@ def test_usaspending_assistnace_load():
...
+    # Find an agency with no awards
+    agencies_in_use = Award.objects.values('awarding_agency__id')
+    new_agency = Agency.objects.exclude(id__in=agencies_in_use).first()
+    # Assign loaded records to that agency
+    for award in Award.objects.all():
+        award.awarding_agency = new_agency
+        award.save()
+    # re-import original file - should have new records
+    call_command('load_usaspending_assistance', filepath)
+    assert Award.objects.count() == awards_before_loads + (3 * base_award_count
```
bsweger: Again, take this with a grain of salt coming from a reviewer who skipped breakfast, but I had to pause and think why the post-awarding agency update award count would be 3 times the original number instead of 2 times (of course, it's because of the previous test). From a code reader perspective, it might be helpful to have more narrative in the comments (_e.g._, "changing FAINS on the input file and re-running the load should create a new award for each record"...."changing awarding agency in the db and re-running the load should cause an additional set of new awards to be created).

Again, not needed for merging, and the most important thing is having these robust tests around the tricky logic for creating awards. I really like the way you've done this!

usaspending_api/etl/tests/test_usaspending_assistance_load.py
catherinedevlin: ah, it totally is, thanks

usaspending_api/etl/tests/test_usaspending_assistance_load.py
catherinedevlin: yes, and Destie pointed this out, too.  Going to add some comments now.

bsweger:
@catherinedevlin Thanks for updating the test comments!

PR #164, opened 2017-02-23
Implement DRF tracking+reporting

afrasier:
* Added drf-tracking app
* Added LoggingMixin to all endpoints _except_ AutoComplete endpoint as
these don’t play nice together. To implement these, we will need to
rework autocomplete into a mixin
* Added `generate_analytics_json_file` which takes a date range
(default of “today”) to dump json-formatted data from the tracking
table
* Added tests for the tracking and reporting
* Updated common views to handle exceptions and return nice responses
with status codes instead of diaf

### DIFF
--- a/usaspending_api/common/mixins.py
+++ b/usaspending_api/common/mixins.py
@@ -136,7 +136,11 @@ class FilterQuerysetMixin(object):
         if len(request.data):
             fg = FilterGenerator()
             filters = fg.create_from_request_body(request.data)
-            return queryset.filter(filters).distinct()
+            try:
+                filtered_qs = queryset.filter(filters).distinct()
+            except:
+                raise
+            return filtered_qs
         else:
             filter_map = kwargs.get('filter_map', {})
             fg = FilterGenerator(filter_map=filter_map)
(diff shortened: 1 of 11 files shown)


usaspending_api/common/mixins.py
```diff
@@ -136,7 +136,11 @@ def filter_records(self, request, *args, **kwargs):
         if len(request.data):
             fg = FilterGenerator()
             filters = fg.create_from_request_body(request.data)
-            return queryset.filter(filters).distinct()
+            try:
+                filtered_qs = queryset.filter(filters).distinct()
+            except:
```
catherinedevlin: AFAIK, a bare "except" with a bare "raise" is effectively the same as not having a try block at all, so this seems a little confusing.

afrasier: I think this was a debug thing, my bad I'll remove it

usaspending_api/common/views.py
```diff
@@ -22,33 +25,42 @@ def list(self, request, *args, **kwargs):
...
+            response_object = {
+                "total_metadata": metadata,
+                "page_metadata": page_metadata
+            }
+            response_object.update({'results': serialized_data})
+            status_code = status.HTTP_200_OK
+        except InvalidParameterException as e:
+            response_object = {"message": str(e)}
+            status_code = status.HTTP_400_BAD_REQUEST
+        except Exception as e:
```
afrasier: Interesting, does that also include the error message?

PR #170, opened 2017-02-28
Documentation Updates

afrasier:
This is a list of all data changes I could find, a brief summary of the change, description, and file where this modification takes place.

I think I got them all, but if I have overlooked anything people should feel free (and obliged) to update this file.


data_changes.md
```diff
@@ -0,0 +1,58 @@
...
+is found, the location object is updated with that matching data
+- _FILE_: `references/models.py`
+
+
+- _IMPORT TYPE_: USA Spending, Contract; USA Spending, Financial
+- _CHANGE_: **Extraction of Single Character Codes**
+- _DESCRIPTION_: Data loaded from legacy USA Spending data typically comes in the format of
+"<CODE>: <DESCRIPTION>" where code is a single character. In these cases, the single code is
+extracted and stored in the database, with descriptions being provided via enumerated dictionaries.
+- _FILE_:
```
bsweger: Starting here, the font gets really small. _shakes fist at Markdown_

data_changes.md
```diff
@@ -0,0 +1,58 @@
...
+- _IMPORT TYPE_: Broker, USASpending
+- _CHANGE_: **Creation of summary awards**
+- _DESCRIPTION_: Parent awards that do not currently exist in the database will be
+created with empty data, to provide references for transactions being loaded.
+- _FILE_: `submissions/management/commands/load_submission.py`, et. al.
+
+
+- _IMPORT TYPE_: Broker
+- _CHANGE_: **State code converted to foreign province**
+- _DESCRIPTION_: Cases where a location must be created, but the data specified has
```
bsweger: What do you think about something that de-emphasizes the "created" part of the narrative (which might be confusing to readers not in a database mentality)? Maybe something like "contract and assistance data that has a state code and a non 'USA' country code will be considered foreign?

Not sure that's any more helpful? This is hard.

data_changes.md
```diff
@@ -0,0 +1,58 @@
...
+a state code, but is not also specified with country code of 'USA' will be converted
+to a foreign province
+- _FILE_: `submissions/management/commands/load_submission.py`
+
+
+- _IMPORT TYPE_: All (Location Data)
+- _CHANGE_: **Location Data is updated if a valid match is found**
+- _DESCRIPTION_: Cases where some location, but not all, data is specified, if a
```
bsweger: I'm not sure what you mean here? I suspect that folks reading won't know what a "location" is, at least as we think of it. Is that we keep a running list of locations, and those with partial information will be updated if we get a subsequent record that matches and has more info?

data_changes.md
```diff
@@ -0,0 +1,58 @@
+## Changes to Imported Data ##
+
+Generally, data will be imported as-is from historical data or newly given data
+with a few exceptions:
+
+- _IMPORT TYPE_: Broker, USASpending
+- _CHANGE_: **Creation of summary awards**
```
bsweger: I'm torn about whether or not we should include info re: summary award. It's a bit different than "here's some incoming data that we changed." If we do put info about award creation into this version of the document, I think we need to elaborate a bit more about how we create these awards and watch for trigger words like "summary" and "parent awards" that will confuse agency folks who are already familiar with the domain.

Happy to work on wording if that's helpful.

data_changes.md
```diff
@@ -0,0 +1,58 @@
+## Changes to Imported Data ##
+
+Generally, data will be imported as-is from historical data or newly given data
+with a few exceptions:
+
+- _IMPORT TYPE_: Broker, USASpending
```
bsweger: My .02 on the `IMPORT TYPE` is to leave it out unless someone asks for it. If there's something that applies to one type of load but not another, we can also note it inline (which you've already done in places like line 31).

data_changes.md
```diff
@@ -0,0 +1,58 @@
...
+
+- _IMPORT TYPE_: All (Location Data)
+- _CHANGE_: **Location Data is updated if a valid match is found**
+- _DESCRIPTION_: Cases where some location, but not all, data is specified, if a
+single unique reference (STATE CODE, CITY CODE, COUNTY CODE, STATE NAME, COUNTY NAME)
+is found, the location object is updated with that matching data
+- _FILE_: `references/models.py`
+
+
+- _IMPORT TYPE_: USA Spending, Contract; USA Spending, Financial
```
bsweger: If we do keep IMPORT TYPE, `USA Spending, Financial` should be `USA Spending, Assistance`

data_changes.md
afrasier: This is when location has a county code/etc. match, it will update data from the ref_city_county data

data_changes.md
afrasier: I mean, what's the target audience here? Are we assuming people reading about changes to incoming data during an ETL are having no knowledge of databases at all?

data_changes.md
afrasier: If you have any ideas, feel free. I'm not sure how to word this stuff in a way that someone unfamiliar with generally how the data is to understand it

data_changes.md
bsweger: I think the target audience might be auditors, etc. who are interested in how we might be changing the data. Let's ask about this during sprint review.

PR #173, opened 2017-02-28
Delete old AwardAction-based transaction tables

bsweger:
This is the last cleanup step for the switch to the new
transaction tables. Delete the old tables and their models.
Also add some info to the award model about where those
fields get updated.


usaspending_api/awards/models.py
```diff
@@ -163,6 +163,20 @@ def get_queryset(self):
...
+    specific actions against an award, though the award records themselves aren't
+    assigned on the incoming data. To rectify that and to make the transactional
+    data easier for people to understand, we create Award objects and map
+    transactions to them.
+
+    Much of the award record data (for example, awarding_agency, funding_agency,
+    type) is automatically populated from info in the award's child transactions.
+    These updates happen in our USAspending and data act broker load processes:
+    see ETL\award_helpers.py for details.
+    """
```
catherinedevlin: hooray for documentation comments

bsweger:
@catherinedevlin thanks for the review! I'll fix up Travis and also need to make sure the broker load process kicks off the awards updates, since those will no longer be done via model `save()` overrides once this PR is merged

afrasier:
Broker loader does use update_awards() and update_contract_awards() :)

bsweger:
@afrasier sure enough, there it is. Excellent.

PR #181, opened 2017-03-02
`.distinct()` was forcing very slow sort on all queries

catherinedevlin:
This seems to fix our killer performance problem.


afrasier:
That was optimistic.

PR #189, opened 2017-03-06
Add Latest Transaction to Summary Award

afrasier:
Adding latest transaction to summary award, pulled when requested or when hitting the `/awards/pk` endpoint

* Added latest transaction to awards model 
* Added setting latest transaction to update_awards method
* Updated Broker ETL to set type_descriptions and to call update_awards and update_contract_awards
* Updated Documentation


usaspending_api/submissions/management/commands/load_submission.py
```diff
@@ -12,16 +12,25 @@
...
     Agency, LegalEntity, Location, RefObjectClassCode, RefCountryCode, RefProgramActivity, CFDAProgram)
 from usaspending_api.submissions.models import SubmissionAttributes
+from usaspending_api.etl.award_helpers import update_awards, update_contract_awards
 
 # This dictionary will hold a map of tas_id -> treasury_account to ensure we don't
 # keep hitting the databroker DB for account data
 TAS_ID_TO_ACCOUNT = {}
 
+# Store some additional support data needed for the type description matching
+award_type_dict = {a[0]: a[1] for a in AWARD_TYPES}
```
catherinedevlin: In my PR from earlier today, I calc the dictionary form of  `AWARD_TYPES` and  `CONTRACT_PRICING_TYPES` in awards/models.py itself, then import those here to save a little code; I'll push a commit shortly to this to re-do that in this PR.

catherinedevlin: ah, wait: I didn't understand why git didn't incorporate my changes - now I see, it's just very similar code in a dfferent file.  I'll still submit that as a refactor, but as  small separate PR after this is approved.

usaspending_api/awards/models.py
```diff
@@ -191,6 +191,7 @@ class Award(DataSourceTrackedModel):
     create_date = models.DateTimeField(auto_now_add=True, blank=True, null=True)
     update_date = models.DateTimeField(auto_now=True, null=True)
     latest_submission = models.ForeignKey(SubmissionAttributes, null=True)
+    latest_transaction = models.ForeignKey("awards.Transaction", related_name="latest_transaction", null=True)
```
catherinedevlin: The `related_name` is the fieldname from the `Transaction` side, so it's confusing to have that named `latest_transaction` when it actually points back to this award:

`In [4]: Transaction.objects.first().latest_transaction.first()
Out[4]: <Award: Definitive Contract piid: SBAHQ16C0024 fain: None uri: None>`

Not sure what would be a clearer name - `latest_transaction_of`, `latest_transaction_of_award`, or even `latest_transaction_of_which_award`, maybe?  Alternately, maybe it's not necessary to have a `related_name` here, since  there's already a link from the Transaction up to the parent Award via the Transaction's `.award`:

`
In [5]: Transaction.objects.first().award
Out[5]: <Award: Definitive Contract piid: SBAHQ16C0024 fain: None uri: None>`

... the only utility I can figure is if we want to be able to determine from a Transaction whether it is the last transaction of its parent award.  Do we?

afrasier: We have to have a related name here because there would be a collision with transaction set related names (you'll see this if you try to makemigrations without a related name) I went for 'latest_transaction' here, because I thought it was clear, however I see what you mean now. We could probably do with naming it "latest_for_award", how does that sound?

PR #191, opened 2017-03-07
DS-703 case insensitivity for 'in' and 'equals' operations

namkyux:


namkyux:
Good fix on the iexact for nonstrings!

PR #192, opened 2017-03-07
Aggregate potential total value of award

afrasier:
* `potential_total_value_of_award` is now aggregated across `TransactionContract` data rather than just taking the latest
* Updated test_award_helpers to verify the summation


usaspending_api/etl/award_helpers.py
```diff
@@ -97,21 +97,33 @@ def update_contract_awards(award_tuple=None):
         sql_txn_contract_latest += 'WHERE award_id IN %s '
```
bsweger: Since potential_total_value_of_award was the only field we were getting from the latest contract txn record, I think we can flat out replace lines 90-98 with the new `sql_txn_totals` common table expression you've added.

usaspending_api/etl/award_helpers.py
```diff
@@ -97,21 +97,33 @@ def update_contract_awards(award_tuple=None):
...
+        sql_txn_totals += 'WHERE tx.award_id IN %s '
+    sql_txn_totals += 'GROUP BY tx.award_id) '
+
     # construct a sql query that uses the latest txn contract common table
     # expression above and joins it to the corresopnding
     # award. that joined data is used to update awards fields as appropriate
     # (currently, there's only one trasnaction_contract field that trickles
     # up and updates an award record: potential_total_value_of_award)
-    sql_update = 'WITH {}'.format(sql_txn_contract_latest)
+    sql_update = 'WITH {}, {}'.format(sql_txn_contract_latest, sql_txn_totals)
```
bsweger: See above question...do we still need to use `sql_txn_contract_latest`?

usaspending_api/etl/award_helpers.py
afrasier: As it currently stands, no, we don't need it; so I suppose we can remove it. Let me know if you want me to get rid of it.

PR #195, opened 2017-03-08
Cache locations in memory during contracts load

catherinedevlin:
On the contract file I was benchmarking on (SBA 17 Contracts.csv), this sped the overall load up by 38%.

### DIFF
--- a/usaspending_api/etl/helpers.py
+++ b/usaspending_api/etl/helpers.py
@@ -1,8 +1,16 @@
 from datetime import datetime
+import warnings
 
 from django.db.models import Q
+from django.core.cache import caches, CacheKeyWarning
 
 from usaspending_api.references.models import Agency, Location, RefCountryCode
+
+warnings.simplefilter("ignore", CacheKeyWarning)
+
+def clear_caches():
+    for cache in ['locations', 'recipients', 'summary-awards']:
+        caches[cache].clear()
 
 
 def cleanse_values(row):
@@ -48,7 +56,9 @@ def fetch_country_code(vendor_country_code):
 
     return country_code
 
+location_cache = caches['locations']
 
+@profile
 def get_or_create_location(row, mapper):
     location_dict = mapper(row)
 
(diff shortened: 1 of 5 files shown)


usaspending_api/settings.py
```diff
@@ -845,3 +845,23 @@
     "original_loan_subsidy_cost": "original_loan_subsidy_cost",
     "business_funds_indicator": "business_funds_indicator"
 }
+
+# If caches added or renamed, edit clear_caches in usaspending_api/etl/helpers.py
+CACHES = {
```
bsweger: @catherinedevlin quick question...are we using the recipients and summary awards caches (want to make sure I understand how this is working...I see the explicit code for using the location cache).

if we haven't implemented the recipients/awards caching yet, is it worth putting a note somewhere to our future selves that this is a to do when we return to this story?

usaspending_api/etl/helpers.py
```diff
@@ -1,9 +1,18 @@
...
 from django.db.models import Q
+from django.core.cache import caches, CacheKeyWarning
 
 from usaspending_api.references.models import Agency, Location, RefCountryCode
 
+warnings.simplefilter("ignore", CacheKeyWarning)
+
+
+def clear_caches():
+    for cache in ['locations', 'recipients', 'summary-awards']:
```
bsweger: A very minor nit (and I know this is coming from the loader code itself)...the name `summary-award` is still hanging around from the days before we renamed the `SummaryAward` model to just `Award`. Do you think it's worth starting to make our code match that new name where we can?

usaspending_api/settings.py
destiejo: I'd echo this - I know in contracts and assistance we build a few lists (IE txn) - would we need to insert a "recipient_cache = caches['recipients']" for the two loaders? 

Wasn't sure if the default cache just automatically stored everything you've already pulled! I know we clear the caches before you run the rest of the code in the loader too.

catherinedevlin: I suppose it's clearer just to delete mention of the unused caches.  I stopped in the middle of implementing them...

usaspending_api/etl/helpers.py
catherinedevlin: good point!  Taking out that cache anyway for now, but if we ever re-do we can use a better name

PR #196, opened 2017-03-08
Update agency loader

afrasier:
* Move to `authoritative_agency_list.csv`
* Remove agencies from `reference_fixture`, instead `load_reference_data` now uses `loadagencies`, allowing us to reload and update the agencies at will
* Update places where `reference_fixture` is loaded for agencies to use `loadagencies`
* Load agencies is now idempotent and can be reloaded without causing duplicates
* Added agency abbreviations to toptier and subtier agencies


usaspending_api/references/migrations/0054_auto_20170308_2100.py
```diff
@@ -0,0 +1,25 @@
...
+
+    dependencies = [
+        ('references', '0053_auto_20170210_1814'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='subtieragency',
+            name='abbreviation',
+            field=models.CharField(blank=True, max_length=150, null=True, verbose_name='Agency Abbreviation'),
```
destiejo: This might be a dumb question - but above we have Agency Abbreviation and Subtier Abbreviation, do we need to use both here too? Or does Agency Abbreviation pull for both?

usaspending_api/references/models.py
```diff
@@ -144,13 +146,15 @@ class SubtierAgency(models.Model):
     create_date = models.DateTimeField(auto_now_add=True, blank=True, null=True)
     update_date = models.DateTimeField(auto_now=True, null=True)
     subtier_code = models.CharField(max_length=4, blank=True, null=True, verbose_name="Sub-Tier Agency Code")
+    abbreviation = models.CharField(max_length=150, blank=True, null=True, verbose_name="Agency Abbreviation")
```
destiejo: Same question as above - should this be Subtier Abbreviation?

afrasier: No, there's no need to title it like that, because to access it you still need `awarding_agency__subtier_agency__abbreviation` or `toptier_agency__abbreviation`, just like `name` so there is no confusion there. Generally, there's no need to have `MODELNAME_FIELDNAME`, it's just on some fields due to legacy code

afrasier: Oh, if you're talking about the verbose name, they're both referencing agencies so "Agency Abbreviation" should be fine.

destiejo: Ok! Makes sense - updating to approve.

usaspending_api/references/management/commands/loadagencies.py
```diff
@@ -16,33 +16,44 @@ def handle(self, *args, **options):
 
         try:
             with open(os.path.join(django.conf.settings.BASE_DIR,
-                      'usaspending_api/references/management/commands/agency_list.csv')) \
-                      as csvfile:
+                      'usaspending_api', 'data', 'authoritative_agency_list.csv'), encoding="Latin-1") \
+                    as csvfile:
 
                 reader = csv.DictReader(csvfile)
                 for row in reader:
```
bsweger: If a human will be periodically providing updates to the agency .csv file, we might want to think about explicitly lower or upper casing the incoming column headers so the loader works if someone changes the case on a future version of the file. People sometimes don't realize that changing the headings impacts the load.

PR #198, opened 2017-03-09
Potential Total Value Data Migration

afrasier:
I forgot to commit this guy, and so he got left out of the PR! Oh no!


afrasier:
I've made a change as to how the award tuple is evaluated. Instead of checking against `None` we just do a check of `if award_tuple` which will still return false if `award_tuple` is `None`, but also if `award_tuple` is empty. Previously, if an empty tuple was supplied, there would be an exception. This was preventing this migration from working on an empty (i.e. test) database

bsweger:
@afrasier good catch--thanks for updating!

PR #199, opened 2017-03-09
Fixing Full Text Search

afrasier:
While helping @ejgullo on her full text search story, I discovered that we were missing the code that performed vector annotation. It appears that during refactoring to the new mixins, the annotation of full-text search vectors was left out. This code rectifies this, cleans up the filter mixin, adds validation for full text search operation, and introduces a test case to catch this problem.

* Added extra validations for full text search queries to the filter
validator
* Added search vector annotation to the filtering mixin
* Removed deprecated code from the filtering mixin
* Added test case for full text search on multiple fields

### DIFF
--- a/usaspending_api/common/api_request_utils.py
+++ b/usaspending_api/common/api_request_utils.py
@@ -253,6 +253,13 @@ class FilterGenerator():
                                 raise InvalidParameterException("Invalid field, operation 'range_intersect' requires an array of length 2 for field")
                             if (not isinstance(filt['value'], list) or len(filt['value']) != 2) and 'value_format' not in filt:
                                 raise InvalidParameterException("Invalid value, operation 'range_intersect' requires an array value of length 2, or a single value with value_format set to a ranged format (such as fy)")
+                        if filt['operation'] == 'search':
+                            if not isinstance(filt['field'], list) and not self.is_string_field(filt['field']):
+                                raise InvalidParameterException("Invalid field: '" + filt['field'] + "', operation 'search' requires a text-field for searching")
+                            elif isinstance(filt['field'], list):
+                                for search_field in filt['field']:
+                                    if not self.is_string_field(search_field):
+                                        raise InvalidParameterException("Invalid field: '" + search_field + "', operation 'search' requires a text-field for searching")
                     else:
                         raise InvalidParameterException("Malformed filter - missing field, operation, or value")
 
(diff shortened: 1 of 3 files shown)


usaspending_api/common/api_request_utils.py
```diff
@@ -253,6 +253,13 @@ def validate_post_request(self, request):
                                 raise InvalidParameterException("Invalid field, operation 'range_intersect' requires an array of length 2 for field")
                             if (not isinstance(filt['value'], list) or len(filt['value']) != 2) and 'value_format' not in filt:
                                 raise InvalidParameterException("Invalid value, operation 'range_intersect' requires an array value of length 2, or a single value with value_format set to a ranged format (such as fy)")
+                        if filt['operation'] == 'search':
```
catherinedevlin: Took me a minute to puzzle these next few lines out; the upshot is that we're raising an error if any field in filt['field'] is non-string, whether filt['field'] is a list or a single value... is that right?  Maybe (non-urgently) a comment to that effect, or even pull it out to a function with an informative name like `raiseIfNonFieldPresent` or something...

usaspending_api/common/mixins.py
```diff
@@ -155,21 +155,24 @@ def filter_records(self, request, *args, **kwargs):
...
-            # deprecated: we plan to start storing fiscal years in the database
-            if request.query_params.get('fy'):
-                fy = FiscalYear(request.query_params.get('fy'))
-                fy_arguments = fy.get_filter_object('date_signed', as_dict=True)
-                filters = {**filters, **fy_arguments}
-            return queryset.filter(**filters).distinct()
+            filters = Q(**fg.create_from_query_params(request.query_params))
+
+        # Handle FTS vectors
+        if len(fg.search_vectors) > 0:
```
catherinedevlin: Now I'm really curious if 169-172 could be replaced by `vector_sum = sum(fg.search_vectors)`.  If search vectors support `+=` I bet they support `sum`.

usaspending_api/common/api_request_utils.py
afrasier: Yes, `search` has the option of specifying a single field or an array, so we check to see if all the fields are text.

usaspending_api/common/mixins.py
afrasier: I'm sure we'll get a chance to try that out when we (inevitably) need to add more functionality to the full text search (i.e. supporting "OR" and "AND" which I think I saw in the design docs...)

PR #200, opened 2017-03-09
WIP: Add caching decorator from drf-extensions

bsweger:
Putting up a WIP caching PR to get some initial thoughts. This does not satisfy the environment-based caching that we've been talking about, but it was so darn easy to implement that I thought it
was worth getting some feedback on.

Paging @afrasier @catherinedevlin @namkyux @destiejo for thoughts.

Limited local testing looks promising. That said, we may need think through some of the points that @kevinli-work raised via Slack (_e.g._, resetting the cache, etc.)

**/awards/**
1st request: 7 queries in 4062 ms
2nd request: 3 queries in 14.08 s

**/awards?limit=500**
1st request: 7 queries in 4284 ms
2nd request: 3 queries in 10 ms

**/transactions**
1st request: 7 queries in 863 ms
2nd request: 3 queries in 13.09 ms


afrasier:
Are there some configurations on this package so we can calibrate the expiration time/etc? If this works, it's easy and simple. My only concern is not being able to configure its behavior because it's all tucked away under the hood somewhere.

A few other questions: 
* Is this caching just the response data (i.e. the queryset) or the entire response (including html templates if you view the page directly http://blahblah/api/v1/awards
* Is it possible to enable/disable this via settings.py or some other means to implement environment based caching
* Does this use a specific cache set up in the settings.py (Is this memcache only? Can we ever switch it to a dbcache? Thinking on scalability of the cloud architecture) 

I'm sure I have other questions as we go along, but it looks promising.

catherinedevlin:
That is just so darn beautiful.  The more code we don't have to write, the less we can mess up.

http://chibisov.github.io/drf-extensions/docs/#cache-response

Looks like we can specify a named cached (from the settings.py), and that's a good place for whatever settings we want to impose - expiration time, which backend (local memory, memcached, file, etc)

Great find, Becky.

afrasier:
That is just _FANTASTIC_

If we can point them to a named cache in the settings file we can have the sandbox deployment script change the cache backend to a dummy cache.

https://docs.djangoproject.com/en/1.10/topics/cache/#dummy-caching-for-development

I think this is a great solution!

bsweger:
@afrasier @catherinedevlin Thanks for your comments! I took this out of WIP because we can disable the caching via the settings file if it gives us any trouble.

bsweger:
Putting this back in WIP so we can move away from local memory cache before merging.

bsweger:
Closing this one: caching has been rolled into #211.

PR #203, opened 2017-03-13
Updated API Documentation

afrasier:
There's a lot here, buckle up!

* Added `generate_model_markdown` management command. This generates a table in markdown format based upon a specified model, getting descriptions from the model's fields' `help_text` parameter. You can call this command via `python manage.py generate_model_markdown awards.Award` to see it in action. I didn't have time to add `help_text` to every field on every model, and I suspect I will need some help coming up with friendly descriptions of these fields. Additionally, this models functionality can be leveraged by a template tag, allowing us to make some fully automated docs that render nicely, so that we don not have to keep up with them. (That's future plans though, and I'm sure there's a library somewhere to help, but I couldn't find any I could get working in time)

* Added `help_text` to `Award`
* Moved all documentation into `/api_docs/`
* Modified `docs.py` template tag generator to accept arbitrary Markdown via new `MarkdownView`
* Extracted all documentation urls into `/api_docs/urls.py`
* Created a new landing page so it doesn't dump people straight into `using_the_api.md`
* Added an introductory tutorial, `intro_tutorial.md` accessible via `/docs/intro-tutorial`
* Added documentation index, `documentation_index.md` accessible via `/docs/`
* Added data dictionary, `data_dictionary.md`, accessible via `/docs/data-dictionary/`. This has jargon definitions from the content guidelines' linked spreadsheet; a table of every endpoint, the methods they use, and generally what they provide; a list of every field in every object with descriptions where available (generated from `generate_model_markdown`)
* Added request recipes, `request_recipes.md` accessible via `/docs/recipes`. I only had a time to add a handful, but this is easily expandable
* Added a test class called `test_documentation_examples.py` which tests _every_ example post and get request listed in the _new_ tutorials and documentation. Note that this doesn't check for the accuracy of the responses, just that the responses return 200. This way, if a change causes a break in this test and no others, it is likely the documentation needs to be updated.


afrasier:
Yes, it needs some CSS work on the table stylings. I was hoping we could page one of the front end/design folks since the CSS in there looks like it was generated by SASS and I don't want to break it.

usaspending_api/api_docs/tests/test_documentation_examples.py
```diff
@@ -0,0 +1,89 @@
+import pytest
+import json
+from datetime import date
+
+from model_mommy import mommy
+from rest_framework import status
+import json
+
+'''
+This set of tests exists to test all example requests used in the API documentation
```
bsweger: Terrific idea to write tests for the documentation examples!

usaspending_api/api_docs/markdown/data_dictionary.md
```diff
@@ -0,0 +1,750 @@
...
+    </ul>
+  </li>
+</ul>
+[//]: # (Begin Content)
+
+## Lexicon <a name="lexicon"></a>
+
+In this section you will find definitions for common terms used in the API. For detailed explanations of specific fields, please see that endpoint's entry in the Endpoint Details section.
+
+| Term | Explanation |
```
afrasier: Once we have a better place for it to live, I'm fine with either. I'm not sue if it will live here or not, but as it isn't on the main page now I've included it here for usability's sake.

afrasier:
I didn't have the time this sprint to explore automated docs; but it's something we should definitely look into. We're already using the 'self-describing' API, we just need to update the docstrings on each endpoint to get that a bit more spruced up. 

I think, generally, documenting what each field in help_text will be the first step in any documentation since we can have them pull from that for field explanations. The main benefit of the markdown generation command was making it easy to pull relation/help_text descriptions into markdown, since that's how our current documentation is displayed. Thus, as far as model field descriptions go, we could automate that pull using a custom template tag, which would be good to use in combination with one of those autodoc options; with the autodoc describing each endpoint and their responses, and the markdown docs describing precisely what each field is.

Anyway, thanks for taking a look. I know it's a lot, but it's not complete and is just a start. We should definitely have another documentation story next sprint to look at those automated endpoint doc's and getting help_text on all the fields (we may need some SME help on that one!)

usaspending_api/api_docs/markdown/data_dictionary.md
```diff
@@ -0,0 +1,750 @@
...
+
+## Endpoint Details <a name="endpoint-details"></a>
+
+#### Award Endpoints <a name="award"></a>
+This data is represented internally as the model: `Award`
+
+
+| Field | Type | Description |
+| ----- | ----- | ----- |
+| financial_set | Relation | Reverse look-up for relation from FinancialAccountsByAwards::award |
```
kevinli-work: What does type `Relation` mean? API users working with JSON responses don't/shouldn't have to care about the API's database schema. Instead, I would specify if these are string representations of other IDs (and what object and fields those IDs are...and what the API endpoint should be for that object) or if they are nested objects and link to that object's documentation. The description of these fields should also identify what the logical (not database or code) relationship is to the nested or referenced object (the Financial Account of an Award is an award's XXXXXXXX).

usaspending_api/api_docs/markdown/data_dictionary.md
```diff
@@ -0,0 +1,750 @@
...
+| date_signed | Date | The date the award was signed |
+| recipient | Relation | The recipient of the award |
+| description | String | A description of the award |
+| period_of_performance_start_date | Date | The start date for the period of performance |
+| period_of_performance_current_end_date | Date | The current, not original, period of performance end date |
+| place_of_performance | Relation | The principal place of business, where the majority of the work is performed. For example, in a manufacturing contract, this would be the main plant where items are produced. |
+| potential_total_value_of_award | Float | The sum of the potential_value_of_award from associated transactions |
+| last_modified_date | Date | The date this award was last modified |
+| certified_date | Date | The date this record was certified |
+| create_date | Datetime | The date this record was created in the API |
```
kevinli-work: How is `Datetime` different from `Date`? Is the output format (being JSON it's either numerical or a string) for these formats documented?

usaspending_api/api_docs/markdown/data_dictionary.md
```diff
@@ -0,0 +1,750 @@
...
+| update_date | Datetime | The last time this record was updated in the API |
+| latest_submission | Relation | The submission attribute object that created this award |
+| latest_transaction | Relation | The latest transaction by action_date associated with this award |
+
+#### Transaction Endpoints <a name="transaction"></a>
+
+Transaction data is returned as a combination of the three following models:
+This data is represented internally as the model: `Transaction`
```
kevinli-work: What do these "internal representation" lines mean? What would an external API user be doing with this information?

usaspending_api/api_docs/markdown/data_dictionary.md
```diff
@@ -0,0 +1,750 @@
...
+| data_source | String | The source of this entry, either Data Broker (DBR) or USASpending (USA) |
+| award | Relation | The award which this transaction is contained in |
+| usaspending_unique_transaction_id | String | If this record is legacy USASpending data, this is the unique transaction identifier from that system |
+| submission | Relation | The submission which created this record |
+| type | String | The type for this transaction. For example, A, B, C, D |
+| type_description | String | The plain text description of the transaction type |
+| period_of_performance_start_date | Date | The period of performance start date |
+| period_of_performance_current_end_date | Date | The current end date of the period of performance |
+| action_date | Date | The date this transaction was actioned |
+| action_type | String | The type of transaction. For example, A, B, C, D |
```
kevinli-work: I'm skimming through, but for fields with code responses, I think we should either document or link to an external document (DAIMS?) of all the possible response codes and what they mean.

usaspending_api/api_docs/markdown/data_dictionary.md
```diff
@@ -0,0 +1,750 @@
...
+| period_of_performance_potential_end_date | Date | The potential end date of the period of performance |
+| ordering_period_end_date | String | The end date for the ordering period |
+| current_total_value_award | Float | The current value of the award |
+| potential_total_value_of_award | Float | The potential total value of the award |
+| referenced_idv_agency_identifier | String | The agency identifier of the agency on the IDV |
+| idv_type | String | The IDV type code |
+| multiple_or_single_award_idv | String | Specifies whether the IDV is a single more multiple award vehicle |
+| type_of_idc | String | Code representing the type of IDC |
+| a76_fair_act_action | String | A-76 FAIR act action |
+| dod_claimant_program_code | String |  |
```
kevinli-work: I assume the remaining missing descriptions will eventually be added in?

usaspending_api/api_docs/markdown/landing_page.md
```diff
@@ -0,0 +1,28 @@
...
+
+The USAspending API allows the public to access data published under the DATA Act. While the API is under development, we are gradually increasing the amount of available data and the capabilities of the API. Data in the current API comes from a selection of submissions the DATA Act Data Broker and the current USASpending.gov.
+
+## Getting Started <a name="getting-started"></a>
+
+**New to APIs?** Check out our [introductory tutorial](/docs/intro-tutorial) on the API to learn the ropes
+
+**Familiar with APIs?** Roll up your sleeves with our more technical documentation [Using the API](/docs/using-the-api), or jump right into the data with some [request recipes](/docs/recipes/)
+
+**New to government financial data?** Check out our [data dictionary](/docs/data-dictioanry) to learn to navigate it
```
kevinli-work: `data-dictioanry`

usaspending_api/api_docs/markdown/request_recipes.md
```diff
@@ -0,0 +1,75 @@
...
+      "field": "type",
+      "operation": "in",
+      "value": ["A", "B", "C", "D"]
+    }
+  ]
+}
+```
+
+
+###### Get all Awards which have associate TransactionConract data
```
kevinli-work: `TransactionConract

usaspending_api/api_docs/markdown/request_recipes.md
```diff
@@ -0,0 +1,75 @@
+<ul class="nav nav-stacked" id="sidebar">
+  <li><a href="#award-recipes">Award Recipes</a></li>
+</ul>
+[//]: # (Begin Content)
+
+# Award Recipes <a name="award-recipes"></a>
+
+These example requests are for the `/api/v1/awards/` endpoint.
+
```
kevinli-work: I think these requests need detailed descriptions of what you're querying for. For example, instead of `all rewards for a specific agency`, something like:
```
Find all awards for that were awarded by agency XXXX, which has a CGAC code of 0971 and is the toptier agency.
```

usaspending_api/api_docs/markdown/data_dictionary.md
afrasier: Yes

usaspending_api/api_docs/markdown/data_dictionary.md
afrasier: I think that's generally the goal, there just wasn't enough time to populate `help_text` on all fields with that models.

usaspending_api/api_docs/markdown/request_recipes.md
afrasier: Added, thanks

PR #204, opened 2017-03-13
Defer saving awards during File C load until a bulk_create

catherinedevlin:
This cuts the time on `python manage.py load_submission 977` from 3 minutes to 2 minutes by deferring insertion of Award objects until it can insert them in bulk.  In the process, it uses a cache to avoid duplicate creation, though the cache on its own doesn't improve speed.

Instances of several other classes are also created as a submission load, and a similar strategy can be applied to them for more speed improvement.  There will be diminishing returns, though: profiling revealed that saving Awards was the worst bottleneck, and speeding up other saves will have less effect.


usaspending_api/awards/tests/test_awards.py
```diff
@@ -134,3 +134,19 @@ def test_award_date_signed_fy(client):
     for result in resp.data['results']:
         assert 'date_signed__fy' in result
         assert int(result['date_signed__fy']) > 2012
+
+
+@pytest.mark.django_db
+def test_manual_hash_eq_fain():
+    """test that records with equal FAIN hash as equal"""
+    m1 = mommy.make('awards.award', fain='ABC')
```
bsweger: Thanks for accompanying these changes with tests! Do you think that using `_fill_optional` parameter on `mommy.make` would be a simple way to get a more robust test (_i.e._, it would test that random noise in the non-hashed fields would not affect the way we need the hash to work). Feel free to tell me I'm over-thinking this!

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -202,10 +212,11 @@ def handle(self, *args, **options):
...
                     fain=row.get('fain'),
                     uri=row.get('uri'),
-                    parent_award_id=row.get('parent_award_id'))
+                    parent_award_id=row.get('parent_award_id'),
+                    use_cache=True)
                 award.latest_submission = submission_attributes
-                award.save()
-            except:
+                award_queue[award.manual_hash()] = award
+            except:   # TODO: silently swallowing a bare exception is bad mojo
```
bsweger: Indeed it is! This is an artifact of an earlier time when we needed something...anything...in our database for testing. I would argue that now is the time to remove big try/catch blocks on our loaders, since we're still loading agency submissions on demand. Let's find the problems now and fix them, before we start running stuff overnight.

usaspending_api/awards/models.py
```diff
@@ -215,7 +235,7 @@ def __str__(self):
         return '%s piid: %s fain: %s uri: %s' % (self.get_type_display(), self.piid, self.fain, self.uri)
 
     @staticmethod
-    def get_or_create_summary_award(piid=None, fain=None, uri=None, awarding_agency=None, parent_award_id=None):
+    def get_or_create_summary_award(piid=None, fain=None, uri=None, awarding_agency=None, parent_award_id=None, use_cache=False):
```
bsweger: I like how you set this up...smart to have the cache default to off.

usaspending_api/awards/models.py
```diff
@@ -225,23 +245,39 @@ def get_or_create_summary_award(piid=None, fain=None, uri=None, awarding_agency=
...
                     if parent_award_id:
                         # If we have a parent award id, recursively get/create the award for it
-                        parent_award = Award.get_or_create_summary_award(**{i[1]: parent_award_id, 'awarding_agency': awarding_agency})
+                        parent_award = Award.get_or_create_summary_award(use_cache=use_cache, **{i[1]: parent_award_id, 'awarding_agency': awarding_agency})
                     # Now create the award record for this award transaction
                     summary_award = Award(**{i[1]: i[0], "parent_award": parent_award, "awarding_agency": awarding_agency})
-                    summary_award.save()
+                    if use_cache:
+                        awards_cache.set(q_kwargs_fixed, summary_award)
+                        # when we cache a summary_award whose parent_award is not yet save()ed,
```
bsweger: I don't know the answer to this question, but is it something we need to resolve before merging?

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -220,7 +231,12 @@ def handle(self, *args, **options):
                 'program_activity_code': get_or_create_program_activity(row['program_activity_code'])
             }
 
-            load_data_into_model(award_financial_data, row, value_map=value_map, save=True)
+            afd = load_data_into_model(award_financial_data, row, value_map=value_map, save=False)
+            afd_queue.append(afd)
+
+        Award.objects.bulk_create(award_queue.values())
+        FinancialAccountsByAwards.objects.bulk_create(afd_queue)
+        awards_cache.clear()  # but now that cache is operating for other records...
```
bsweger: I'm not sure what the comment here means. Thinking of our future selves, would we interpret this as a todo that needs addressed? Or something informational?

usaspending_api/awards/models.py
afrasier: If you grab it from the cache using the key, you'll be fine as long as you eventually save it. You won't be able to grab it via something like `.get` `.get_or_create` or `.filter` until it is saved. (Until that point, it is just a transient instance of the model, rather than an instance stored in the DB)

usaspending_api/etl/management/commands/load_submission.py
afrasier: I think this is supposed to catch the exception being thrown above (if treasury_account is None) - can probably refactor that into an `ObjectDoesNotExist` exception. (Or just check if it is None in a finally block and in that case log the exception and move on)

usaspending_api/awards/models.py
bsweger: @afrasier Thanks for weighing in. Are you saying that the answer here is no action required before safely merging?

afrasier: I believe it should be fine, but I'd run some tests locally to verify to be safe.

usaspending_api/etl/management/commands/load_submission.py
catherinedevlin: I made a little separate ticket to fix this: DS-946.

usaspending_api/awards/tests/test_awards.py
catherinedevlin: That's a good idea.  I'm including that in my next commit.

usaspending_api/etl/management/commands/load_submission.py
catherinedevlin: Ah, that's actually an obsolete note to self.  Thanks.

usaspending_api/awards/models.py
catherinedevlin: I just realized that my question-to-self doesn't make sense; since the only way to indicate to `get_or_create_summary_award` that it has a parent is to include the `parent_award_id`, I'm worrying about a situation that can't happen.  So I'm deleting this comment.

usaspending_api/etl/management/commands/load_submission.py
bsweger: Thanks for putting in the story, @catherinedevlin!

PR #205, opened 2017-03-14
Adding ERD to docs

afrasier:
We should refactor the templates in the future to be more django-ific.

This is a PR into my documentation branch that will add the ERD diagrams @catherinedevlin added to the documentation pages, and style it according the the templates we use across the docs. 

Using a separate PR so we have the option of not merging this in, but I think we should as the diagram does provide some value (and we can polish it all up during next sprint!)

### DIFF
--- a/usaspending_api/templates/plate.html
+++ b/usaspending_api/templates/plate.html
@@ -1,0 +1,124 @@
+{% load docs %}
+{% load static %}
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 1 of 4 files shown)


usaspending_api/templates/plate.html
```diff
@@ -0,0 +1,124 @@
+{% load docs %}
```
catherinedevlin: Thanks for incorporating this so nicely into the overall docs!

Eventually we should see if we can figure out how to make up the hover-over info on fields more legible, but that's very non-urgent.

PR #209, opened 2017-03-20
WIP: Create and add a materialized view and model for financial accou…

catherinedevlin:
WORK IN PROGRESS: not for merging now (or maybe ever), raising it for initial visibility.

Sketching out how we might work a view with cube-based precalculated aggregates into out models.

…nt summaries


catherinedevlin:
It doesn't look like we're going anywhere with this anytime soon, so I'm closing the PR (but I'll keep the branch around)

PR #211, opened 2017-03-20
Db performance

afrasier:
**Required prior to merge**

- [x] Backend approval ( @bsweger @catherinedevlin )
- [x] DevOps deployment prepared ( @willkjackson )
- [x] Tested on Sandbox (Myself and @willkjackson )

This marks the first set of database performance fixes. Easier to follow commit by commit.
Summary:

* Tweaked how the award manager checked for null awards, twice
* Indexed commonly searched award fields. This had nominal effect, as foreign keys are already indexed by default in Django, and other fields are less commonly filtered on. We should profile a set of queries over the alpha to see what users are doing to judge further indexes. 
* Disabled table logging and related tests 
* Disabled subquery when there are no filters 
* Added caching

Note on caching: There is now a 'CACHE_ENVIRONMENT' variable in `settings.py` - it has 3 options 'disabled', 'local', 'elasticache'; the default of which is 'local' and will cache locally in memcached. 

If you wish to disable caching, simply change that variable to 'disabled'. For production, this will be switched to 'elasticache', and the relevant connection strings will be updated.

There are a lot of migrations because the help_text changes from the documentation never got migrated properly. 

Further improvements (i.e. moving to CursorPagination) will be done in another branch; this set of changes is good to go once the checklist above is complete.


usaspending_api/common/mixins.py
```diff
@@ -173,9 +173,11 @@ def filter_records(self, request, *args, **kwargs):
                 vector_sum += vector
             queryset = queryset.annotate(search=vector_sum)
 
+        subwhere = filters
         # Create structure the query so we don't need to use distinct
         # This happens by reforming the request as 'WHERE pk_id IN (SELECT pk_id FROM queryset WHERE filters)'
-        subwhere = Q(**{queryset.model._meta.pk.name + "__in": queryset.filter(filters).values_list(queryset.model._meta.pk.name, flat=True)})
+        if len(filters) > 0:
```
afrasier: Previously, we would construct this subquery even if there were no filters, which dramatically slowed down flat 'list' calls to an endpoint with no filters.

usaspending_api/settings.py
```diff
@@ -227,6 +227,63 @@
...
+
+# Cache environment - 'local', 'disabled', or 'elasticache'
+CACHE_ENVIRONMENT = 'local'
+
+# Set up the appropriate elasticache for our environment
+CACHE_ENVIRONMENTS = {
+    # Elasticache settings are changed during deployment, or can be set manually
+    'elasticache': {
+        'BACKEND': 'django_elasticache.memcached.ElastiCache',
+        'LOCATION': 'ELASTICACHE-CONNECTION-STRING',
```
afrasier: These strings are overriden by the deployment script. This allows us to _not_ store AWS strings in our repo.

usaspending_api/settings.py
```diff
@@ -227,6 +227,63 @@
...
+        'OPTIONS': {
+            'IGNORE_CLUSTER_ERRORS': True,
+        },
+    },
+    'local': {
+        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
+        'LOCATION': 'locations-loc-mem-cache',
+    },
+    'disabled': {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
```
afrasier: This fakes the caching so that no code needs to change when development wants to disable caching. (And allows for testing to see why certain queries are slow, etc.)

usaspending_api/awards/models.py
```diff
@@ -161,23 +161,23 @@ class Award(DataSourceTrackedModel):
...
+    period_of_performance_start_date = models.DateField(null=True, db_index=True, verbose_name="Start Date", help_text="The start date for the period of performance")
+    period_of_performance_current_end_date = models.DateField(null=True, db_index=True, verbose_name="End Date", help_text="The current, not original, period of performance end date")
     place_of_performance = models.ForeignKey(Location, null=True, help_text="The principal place of business, where the majority of the work is performed. For example, in a manufacturing contract, this would be the main plant where items are produced.")
-    potential_total_value_of_award = models.DecimalField(max_digits=20, decimal_places=2, blank=True, null=True, verbose_name="Potential Total Value of Award", help_text="The sum of the potential_value_of_award from associated transactions")
+    potential_total_value_of_award = models.DecimalField(max_digits=20, db_index=True, decimal_places=2, blank=True, null=True, verbose_name="Potential Total Value of Award", help_text="The sum of the potential_value_of_award from associated transactions")
```
bsweger: Yes! Thanks for indexing all the things.

usaspending_api/common/mixins.py
bsweger: Good catch--thanks for the context!

usaspending_api/common/mixins.py
```diff
@@ -254,6 +256,70 @@ class SuperLoggingMixin(LoggingMixin):
 
     events_logger = logging.getLogger("events")
 
+    """Mixin to log requests - customized to disable DB logging, remove this method to re-enable"""
```
bsweger: I'm confused about what we're disabling here. I'd vote to remove whatever code we have for logging to the db if we're not going to do that anymore. It will live on in GitHub if we really need it, but in the meantime would prefer to keep things clean.

Were we logging extra info to the db that we weren't writing to the file-based logs?

usaspending_api/common/tests/test_drf_tracking.py
```diff
@@ -7,6 +7,7 @@
 import json
 
 
+@pytest.mark.skip(reason="Table logging is now disabled")
```
bsweger: Ditto the comment above--would prefer not to keep this stuff around if we're not using it.

usaspending_api/settings.py
```diff
@@ -227,6 +227,63 @@
...
+    # Elasticache settings are changed during deployment, or can be set manually
+    'elasticache': {
+        'BACKEND': 'django_elasticache.memcached.ElastiCache',
+        'LOCATION': 'ELASTICACHE-CONNECTION-STRING',
+        'TIMEOUT': 'TIMEOUT-IN-SECONDS',
+        'OPTIONS': {
+            'IGNORE_CLUSTER_ERRORS': True,
+        },
+    },
+    'local': {
```
bsweger: Do `BACKEND` and `LOCATION` here replace the settings in `CACHES` (line 231). Or is this in addition to? If the former, should we delete `CACHES`? If the latter, what do you think about adding some additional comments here that explain how these work together? If I understand correctly, we're caching some stuff (e.g., API responses) in memcache and other things (data load info) in local memory. But I'm not sure how all of these knobs and dials work together.

usaspending_api/settings.py
bsweger: So this applies just to things we're caching via memcache? Or does this apply to the caching we do in local memory as well?

afrasier: Just things with the decorator. (Or anything specifically looking at `usaspending-cache`)

usaspending_api/awards/models.py
```diff
@@ -137,11 +137,11 @@ def get_queryset(self):
         and the award will no longer match these criteria
         '''
         q_kwargs = {
-            "type": "U",
-            "total_obligation__isnull": True,
+            "latest_transaction__isnull": True,
```
afrasier: It's on line 186 in models.py under Awards. "Empty" summaries never have one, so this check can replace the others.

README.md
```diff
@@ -28,6 +28,9 @@ More complete install documentation is available on the PostgreSQL [wiki](https:
 
 **Note:** If you're setting up the project on Mac OSX, we recommend using [homebrew](http://brew.sh) to install PostgreSQL.
 
+### Install Memcached
```
afrasier: You'll need memcached installed to `pip install -r requirements.txt` (pylibmc needs it)

usaspending_api/settings.py
catherinedevlin: Likewise, I can't tell the difference between the function of CACHES and CACHE_ENVIRONMENTS.

usaspending_api/common/mixins.py
afrasier: So this is basically duplicating some library code, but instead of sending it to the DB it's solely keeping it around for logging to `events.log`; Basically we're not really _disabling_ anything per se, we're writing some functionality to log this data to a file and omit logging it to the database.

usaspending_api/common/tests/test_drf_tracking.py
afrasier: We may want DB logging in the future, hence why I opted to leave it in and just skip it. I can remove it if we really want to.

usaspending_api/settings.py
catherinedevlin: Oh, wait.  I see.  (?)  CACHE_ENVIRONMENTS is used to add an entry to CACHES, based on the constant CACHE_ENVIRONMENT, which is hardcoded into settings.py up in line 243, but can also (?) be overridden by the deployment script.  (And if you did want to disable the usaspending-cache for a particular run, would you just do that by temporarily changing line 243, or is there a more elegant way?)

It all could probably use clarification.

usaspending_api/settings.py
catherinedevlin: I guess it applies to the cache named "usaspending-cache", which is used for the django-rest-framework caching; but the low-level in-memory caching uses the cache named "location", whose caching backend is not affected by this CACHE_ENVIRONMENTS.

README.md
bsweger: Ah--thanks for the info. Do we need pylibmc for local development? I'd prefer not to make local devs have to install memcached. Heck, I'd prefer not to install it. We did something similar to @catherinedevlin's suggestion on the broker (2 separate requirements.txt files).

afrasier: Sure, I can move the new requirements to a new file (something like requirements-caching.txt) I'll go ahead and do that

usaspending_api/common/mixins.py
bsweger: Ah, ok...let me make sure I understand....we're overriding a third party library that was logging to the db so that we can grab the info it conveniently collects for us and log it to a file instead?

usaspending_api/settings.py
bsweger: Agreed. We have to hold up the PR, but some additional clarification would be terrific.

usaspending_api/common/mixins.py
```diff
@@ -254,6 +256,70 @@ class SuperLoggingMixin(LoggingMixin):
...
+    def initial(self, request, *args, **kwargs):
+        """Set current time on request"""
+
+        # check if request method is being logged
+        if self.logging_methods != '__all__' and request.method not in self.logging_methods:
+            super(LoggingMixin, self).initial(request, *args, **kwargs)
+            return None
+
+        # get IP
```
bsweger: I hadn't noticed this before, but is this logging a user's fill IP address? Are there any privacy issues we want to consider before doing that?

usaspending_api/awards/models.py
bsweger: 👍 thanks...missed it!

bsweger:
@afrasier Thanks for the splitting up the requirements files--that was my biggest concern. Before we merge, I have one other question about IP addresses (which I realize wasn't in the scope of this PR, but it's enough of a flag for me to ask...see question inline).

usaspending_api/common/mixins.py
afrasier: This data is only available via events.log, and only logs the IP of the address sending the request. I'm not sure what privacy concern there would be, as when someone makes a request to a website you're always going to be giving them your IP.

bsweger: Sure, but my question is more about--do we, in our application logs, want to be the keepers of said IP addresses? Unless we need them for debugging, I'd prefer to take them out. Sorry for not catching this earlier!

bsweger: Arman and I chatted about this, and he made a good case for how an IP address fits into our overall understanding of how the API is being used. We arrived at leaving this but adding a story that calls for disposing of the logs at regular intervals.

PR #216, opened 2017-03-23
Add federal account model and data migrations

bsweger:
New federal account model and data migration. Doing this in two parts to minimize blocking
some of the downstream stories that need this new table.

This is part one:
* add FederalAccount model
* add nullable foreign keys to FederalAccount on the TreasuryAppropriationAccount and AppropriationAccountBalances models
* add new data to test fixture .json

A follow-up PR will:
* update the submission loader to populate the FederalAccount FKs
* update the initial data load to populate FederalAccount FKs
* depending on how the above plays out, maybe make the FKs non-nullable

### DIFF
--- a/usaspending_api/accounts/migrations/0018_auto_20170322_2024.py
+++ b/usaspending_api/accounts/migrations/0018_auto_20170322_2024.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accounts', '0017_auto_20170320_1455'),
+    ]
+
+    tas_fk_sql = '''
+        update treasury_appropriation_account t
+        set federal_account_id = fa.id
+        from federal_account fa
+        where t.agency_id = fa.agency_identifier
+        and t.main_account_code = fa.main_account_code
+        '''
+
+    ab_fk_sql = '''
+        update appropriation_account_balances ab
+        set federal_account_id = t.federal_account_id
+        from treasury_appropriation_account t
+        where ab.treasury_account_identifier = t.treasury_account_identifier
+        '''
+
+    def insert_federal_accounts(apps, schema_editor):
+        TreasuryAppropriationAccount = apps.get_model(
+            "accounts", "TreasuryAppropriationAccount")
+        FederalAccount = apps.get_model("accounts", "FederalAccount")
+        # Get a count of distinct combinations of agency id/main acct code
+        # in the TAS table.
+        expected_federal_accounts = TreasuryAppropriationAccount.objects.distinct(
+            'main_account_code', 'agency_id').count()
+
+        # Get TAS agency id (AID), main acct code (MAC), and the acct title for the most
+        # recent period end date for each agency/acct code combo. These
+        # will be the federal account records.
+        federal_accounts = TreasuryAppropriationAccount.objects.values_list(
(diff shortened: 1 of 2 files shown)


usaspending_api/accounts/models.py
```diff
@@ -105,9 +127,17 @@ def __str__(self):
         return "%s" % (self.tas_rendering_label)
 
 
-# Table #4 - Appropriation Account Balances
 class AppropriationAccountBalances(DataSourceTrackedModel):
+    """
```
catherinedevlin: Thank you for this comment!

usaspending_api/accounts/models.py
```diff
@@ -3,11 +3,33 @@
...
+    """Represents a single Treasury Account Symbol (TAS)."""
     treasury_account_identifier = models.AutoField(primary_key=True)
+    federal_account = models.ForeignKey('FederalAccount', models.DO_NOTHING, null=True)
     tas_rendering_label = models.CharField(max_length=50, blank=True, null=True)
     allocation_transfer_agency_id = models.CharField(max_length=3, blank=True, null=True)
+    # todo: update the agency details to match FederalAccounts. Is there a way that we can
+    # retain the text-based agency TAS components (since those are attributes of TAS
+    # while still having a convenient FK that links to our agency tables using Django's
+    # default fk naming standard? Something like agency_identifier for the 3 digit TAS
+    # component and agency_id for the FK?)
```
catherinedevlin: I'm pretty sure the answer is "yes, why not?"  It's a small offense against the principles of normalization, but since this isn't an OLTP system, sacrificing that for performance is fine.

usaspending_api/accounts/models.py
```diff
@@ -105,9 +127,17 @@ def __str__(self):
...
+    """
+    Represents Treasury Account Symbol (TAS) balances for each DATA Act
+    broker submission. Each submission provides a snapshot of the most
+    recent numbers for that fiscal year. In other words, the lastest
+    submission for a fiscal year reflects the balances for the entire
+    fiscal year.
+    """
     appropriation_account_balances_id = models.AutoField(primary_key=True)
+    # a convenience FK back to federal accounts for ease of traversal
+    federal_account = models.ForeignKey('FederalAccount', models.DO_NOTHING, null=True)
```
afrasier: Is this necessary? It's already accessible via `treasury_account_identifier__federal_account`?

bsweger: It's definitely not necessary. My thinking was that having the FK here too would reduce the traversals necessary when aggregating account balances by federal account (rather than by tas).

I'm fine removing this if we want a more pure data model.

afrasier: Personally I'd prefer it a bit more pure; but if you think it has good merit I'll leave it up to you. (This is not a major complaint or problem so I'll go ahead and approve :) )

usaspending_api/accounts/migrations/0018_auto_20170322_2024.py
```diff
@@ -0,0 +1,111 @@
...
+
+    ab_fk_sql = '''
+        update appropriation_account_balances ab
+        set federal_account_id = t.federal_account_id
+        from treasury_appropriation_account t
+        where ab.treasury_account_identifier = t.treasury_account_identifier
+        '''
+
+    def insert_federal_accounts(apps, schema_editor):
+        TreasuryAppropriationAccount = apps.get_model(
```
catherinedevlin: Interesting, I'd never seen `.get_model`.  Lets you fill the model in the same migration it's created it, I guess.  Nice!

PR #219, opened 2017-03-24
DS-949, Update account balances endpoint to return only one set

catherinedevlin:
... of  of numbers per fiscal year

This still needs tests, but if folks are desperate to get their hands on it, here's the preliminary version.


usaspending_api/accounts/models.py
```diff
@@ -165,7 +165,20 @@ class AppropriationAccountBalances(DataSourceTrackedModel):
...
     update_date = models.DateTimeField(auto_now=True, null=True)
+    final_of_fy = models.BooleanField(blank=False, null=False, default=False, db_index=True)
 
     class Meta:
         managed = True
         db_table = 'appropriation_account_balances'
+
+    @classmethod
+    def populate_final_of_fy(cls):
+        sql = """   UPDATE appropriation_account_balances
```
afrasier: Is this not accomplishable via regular ORM operations? We're starting to get some serious SQL creep.

bsweger: This one might be do-able via ORM. The particular thing I've found that the ORM doesn't do is `UPDATE` via a table join. Agreed that we should have a shared understanding of when we can/should drop to raw SQL, but would disagree that we have SQL creep.

Aside from migrations--where I think it's ok to be less stringent about SQL vs ORM--the only other place I see raw SQL in the code base is the award helpers (and the currently PR'd federal accounts helpers). Both of those fall into the "update via table join pattern" that isn't supported.

bsweger: @catherinedevlin Your choice of `select distinct on` is an elegant solution here! If it's helpful, here's similar `distinct on` code I via ORM (ignore the annotation and coalesce stuff..that was to handle and edge case with null, which I don't think is applicable here).

@afrasier full disclosure: I don't _think_ `distinct on` is ANSI SQL, so regardless of whether we do this via ORM or raw SQL, this wouldn't necessarily be portable to another RDBMS. I think that's fine--it's a handy feature of using Postgres--but wanted you to be aware.

afrasier: We're already pretty tied to Postgres as it is, so a little more piling on is probably fine.

catherinedevlin:
@bsweger , 

>  If we can keep things set-based but still go through the ORM

It turns out my SQL was wrong; the correct version updates from a join, so that's going to be harder.  I'll still look into it.

Doesn't look promising so far:

https://docs.djangoproject.com/en/dev/topics/db/queries/#updating-multiple-objects-at-once

> However, unlike F() objects in filter and exclude clauses, you can’t introduce joins when you use F() objects in an update – you can only reference fields local to the model being updated. If you attempt to introduce a join with an F() object, a FieldError will be raised:

Although, maybe if we use this package: https://djangopackages.org/packages/p/django-bulk-update/ 

> need a data migration 

Oh, I didn't think of that!  Yes, it's easy to add to the migration; I'm doing it now.

> Do think we should use a manager

Sure!  I'll add that.

catherinedevlin:
Still trying to figure out a way to do this within the Django ORM - this approach looks possibly promising:

```
objs = AppropriationAccountBalances.objects. \
    annotate(rank=RawSQL("""RANK() OVER 
                           (PARTITION BY FY(reporting_period_start) 
                            ORDER BY reporting_period_start DESC)""", [])   )
```
but we're looking for the reporting_period_start of the linked SubmissionAttributes object, not the AppropriationAccountBalances object itself, so... more work to do on this.

catherinedevlin:
One other possibility would be to put the calculation in the `.get_queryset` method by using a RANK OVER... PARTITION BY... ORDER BY statement.  This would still be a chunk of very raw SQL, but it on the other hand it would be more thoroughly embedded in the Django workings and not possible for it to get out-of-synch with the data, the way a separate SQL update could.  Back on the first hand, though, it would be some performance load upon getting the data.

afrasier:
I think for now, this satisfies the idea of an "MVP", so we can go ahead and merge this in for this sprint. We can make another story in JIRA to look back and give that a shot if you think it is worth it.

PR #221, opened 2017-03-27
DS-924: Seek or create program_activity_code by budget year and agency

catherinedevlin:


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -514,12 +514,16 @@ def get_or_create_object_class(row_object_class, row_direct_reimbursable, logger
     return obj_class
 
 
-def get_or_create_program_activity(program_activity_code):
+def get_or_create_program_activity(row, submission_attributes):
     # We do it this way rather than .get_or_create because we do not want to
     # duplicate existing pk's with null values
-    prg_activity = RefProgramActivity.objects.filter(ref_program_activity_id=program_activity_code).first()
-    if prg_activity is None and program_activity_code is not None:
-        prg_activity = RefProgramActivity.objects.create(ref_program_activity_id=program_activity_code)
```
catherinedevlin: I'm actually really puzzled here.

It *looks like* row['program_activity_code'] ought to be used to populate `RefProgramActivity.program_activity_code`, both because of the matching name, and because the model is defined with

`        unique_together = (('program_activity_code', 'budget_year', 'responsible_agency_id', 'allocation_transfer_agency_id', 'main_account_code'),)`

But, all this time, we've been using it to populate `ref_program_activity_id`.  

So what is `ref_program_activity_id` for, and why did we define it as an `IntegerField` rather than an `AutoField` (like most of our surrogate primary keys are)?

afrasier: That's one of the legacy models. It was reflected out of an old postgres database, and it looks like no one went back and tuned it up. So, there's no good reason why it's not an Autofield. (Also that name is just terrible...)

catherinedevlin: OK, so... 

Lots of the values in the `submissions` table have 4-character `program_activity_code`s that will fit into `RefProgramActivity.program_activity_code`, but lots of others have 12-character codes.  Should we expand `RefProgramActivity.program_activity_code` to 12 characters (or, perhaps better, simple `TEXT`), slice a 4-character portion of `program_activity_code` (which portion?), or... ?

catherinedevlin: Per conversation with Becky, the submissions with long codes currently in the broker DB are residue from earlier work.  It's OK to expect 4-character codes from now on.

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -514,12 +515,16 @@ def get_or_create_object_class(row_object_class, row_direct_reimbursable, logger
...
 
-def get_or_create_program_activity(program_activity_code):
+def get_or_create_program_activity(row, submission_attributes):
     # We do it this way rather than .get_or_create because we do not want to
     # duplicate existing pk's with null values
-    prg_activity = RefProgramActivity.objects.filter(program_activity_code=program_activity_code).first()
-    if prg_activity is None and program_activity_code is not None:
-        prg_activity = RefProgramActivity.objects.create(program_activity_code=program_activity_code)
+    filters = {'program_activity_code': row['program_activity_code'],
+               'budget_year': fy(submission_attributes.reporting_period_start),
```
bsweger: Good news... `reporting_fiscal_year` is now a part of `SubmissionAttributes`, so we don't need to derive it ourselves (that's a recent change: https://github.com/fedspendingtransparency/usaspending-api/pull/233)

If I'm reading this correctly, I think this means we can ax the fy helper function and just pull fy from the submission data.

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -514,12 +515,16 @@ def get_or_create_object_class(row_object_class, row_direct_reimbursable, logger
...
+               'responsible_agency_id': row['agency_identifier'],
```
bsweger: This is really a question for the broker team or @kaitlin, but when using an agency to lookup program activities, should we be using the `cgac_code` from the submission or `agency_identifier` from the File B/C data? 

They're probably one in the same, but worth confirming.

usaspending_api/etl/tests/test_load_submission.py
```diff
@@ -64,3 +64,18 @@ def test_load_submission_command(endpoint_data, partially_flushed):
     assert sub.reporting_fiscal_quarter == 1
     assert sub.reporting_period_start == date(2015, 10, 1)
     assert sub.reporting_period_end == date(2015, 12, 31)
+
+
+@pytest.mark.django_db
+def test_load_submission_command_program_activity_uniqueness(endpoint_data, partially_flushed):
```
bsweger: Thanks for adding this. Good to see us adding some more unit tests on the submission load process, given that we're about to kick it into high gear!

usaspending_api/references/migrations/0059_auto_20170330_1426.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.1 on 2017-03-30 14:26
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def copy_ref_program_activity_id_to_id(apps, schema_editor):
+    RefProgramActivity = apps.get_model('references.RefProgramActivity')
+    RefProgramActivity.objects.update(id=models.F('ref_program_activity_id'))
```
bsweger: Oh, nice--I always forget about the good old `F` object

usaspending_api/references/migrations/0059_auto_20170330_2107.py
```diff
@@ -12,9 +12,4 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.AlterField(
```
bsweger: Will changing an existing migration affect folks who have already applied it?

usaspending_api/references/migrations/0060_auto_20170330_1522.py
```diff
@@ -0,0 +1,24 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('references', '0059_auto_20170330_1426'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='refprogramactivity',
+            name='ref_program_activity_id',
```
bsweger: Happy to say goodbye to `ref_program_activity_id`. I know this type of work is tedious--thank for you the housekeeping.

usaspending_api/references/serializers.py
```diff
@@ -75,7 +75,6 @@ class Meta:
 
 class RefProgramActivityBriefSerializer(serializers.ModelSerializer):
     prefetchable = False
-    id = serializers.IntegerField(source='ref_program_activity_id')
```
bsweger: Do we want to serialize the new `id` field to replace the serialization of `ref_program_activity_id`, or do we just not need it?

usaspending_api/etl/management/commands/load_submission.py
kaitlin: they're not always the same, but we should be using the AID since the program activities map to the AID of the TAS, not necessarily the CGAC code of the agency submitting.

bsweger: Heh, you probably already asked this, but I confirmed that yes, we should be pulling agency_identifier from each row of the incoming data (looks like you're already doing this, so carry on!)

usaspending_api/references/serializers.py
catherinedevlin: It's actually serialized by including the name in `fields` (line 84).  The explicit line defining the serialized field is only needed when the serialized field name differs from the source name.  In fact, DRF throws an error when you define the field unnecessarily that way (I tried).

usaspending_api/references/migrations/0059_auto_20170330_2107.py
catherinedevlin: Ohhhh.... rats.  I was hoping to slip some edits to recent migrations through while the edited migrations only existed on my own machine... but I don't think this qualifies for that anymore.

I'd better figure out how to do this as separate migrations instead.

usaspending_api/etl/management/commands/load_submission.py
catherinedevlin: Actually, I hadn't asked it - I made an (apparently lucky) guess.  Thanks for getting the confirmation!

usaspending_api/references/serializers.py
bsweger: Thanks for the context!

usaspending_api/etl/management/commands/load_submission.py
catherinedevlin: Oh, thank you for that!

bsweger: Not something we need to hold the merge for, but you should now be able to replace `fy(submission_attributes.reporting_period_start)` with `submission_attributes.reporting_fiscal_year` (_i.e._, no `fy` function needed).

catherinedevlin: Ah... I dropped that fix at some point in juggling git branches.  OK, I'm re-committing it; I'm not letting this get away without it.

PR #222, opened 2017-03-27
rm submission command

afrasier:
* Add `rm_submission` command for devops
* Add tests

### DIFF
--- a/usaspending_api/submissions/management/commands/rm_submission.py
+++ b/usaspending_api/submissions/management/commands/rm_submission.py
@@ -2,6 +2,7 @@ from datetime import datetime
 import logging
 import os
 from django.core.management.base import BaseCommand
+from django.core.exceptions import ObjectDoesNotExist
 
 from usaspending_api.submissions.models import SubmissionAttributes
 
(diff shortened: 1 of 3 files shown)


usaspending_api/submissions/management/commands/rm_submission.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    help = "Removes a single submission from the configured data broker database"
+    logger = logging.getLogger('console')
+
+    def add_arguments(self, parser):
+        parser.add_argument('submission_id', nargs=1, help='the broker submission id to delete', type=int)
+
+    def handle(self, *args, **options):
+        # This will throw an exception and exit the command if the id doesn't exist
+        submission = SubmissionAttributes.objects.get(broker_submission_id=options["submission_id"][0])
```
catherinedevlin: On a nonexistent submission number, this throws an error stack ending in 

`usaspending_api.submissions.models.DoesNotExist: SubmissionAttributes matching query does not exist.`

which is not too bad as cryptic errors go, especially for a specialty task like deleting a submission, we just may want to add a better error message at some point (not necessarily part of this PR).

usaspending_api/submissions/tests/test_rm_submission_command.py
```diff
@@ -0,0 +1,82 @@
...
+    verify_zero_count(SUBMISSION_MODELS, 123)
+    verify_zero_count(LATEST_SUBMISSION_MODELS, 123, "latest_submission")
+
+    # Make sure we still have our other submission
+    verify_zero_count(SUBMISSION_MODELS, 456, eq_zero=False)
+    verify_zero_count(LATEST_SUBMISSION_MODELS, 456, "latest_submission", eq_zero=False)
+
+    call_command("rm_submission", 456)
+
+    verify_zero_count(SUBMISSION_MODELS, 456)
```
catherinedevlin: Nice that you're verifying the deletion of dependent records, too.

PR #233, opened 2017-03-31
Add fiscal year, quarter, and period to submission attributes

bsweger:
This PR adds and populates a few new fields to the `SubmissionAttributes` model. The most important addition is `reporting_fiscal_quarter`, which will support API requests that involve the fiscal quarter.

New columns:
1. `reporting_fiscal_year`: pulled from the broker's submission table
2. `reporting_fiscal_period`: pulled from the broker (note: fiscal period 1 = October)
3. `reporting_fiscal_quarter`: calculated based on `reporting_fiscal_period`
4. `quarter_format_flag`: grabbed this from the broker while I was in there (once we're live, we'll only see quarterly submissions, but during this testing period, we do have some monthly submissions loaded)

The PR also removes some old columns form `SubmissionAttributes` that were not being used/no longer match the broker's submission table.


usaspending_api/etl/helpers.py
```diff
@@ -107,3 +107,15 @@ def parse_numeric_value(string):
...
+def get_fiscal_quarter(fiscal_reporting_period):
+    """Return the fiscal quarter."""
+    if fiscal_reporting_period in [1, 2, 3]:
+        return 1
+    elif fiscal_reporting_period in [4, 5, 6]:
+        return 2
+    elif fiscal_reporting_period in [7, 8, 9]:
+        return 3
+    elif fiscal_reporting_period in [10, 11, 12]:
+        return 4
```
afrasier: In this context, does 1=October, 2=November, 3=December?  (And the counting doesn't start at 0?)

I've sure you've checked, I just want to make sure!

bsweger: Yes, thanks for double checking that! Yes, there periods are already stored in the broker as federal fiscal periods, so 1 = October. I added a note to the docstring to clarify.

PR #237, opened 2017-03-31
Dev stable ->

afrasier:
Merge data loading hotfixes from dev stable into dev

* Add repair_sequences command, which generates relevant SQL for all apps we use which will repair sequences for those apps. (This is not run automatically so that the user can verify the SQL)
* Change the primary key for program activities to autofield
* Remove tas_rendering_label from appropriation account balances
* Validate direct reimbursable exists before using it
* New TAS list (make sure you reload it via manage.py loadtas usaspending_api/data/tas_list.csv)
* Fix tuple check in award update
* Fix summary award duplication


afrasier:
Closing again, will resolve conflict in another branch

PR #238, opened 2017-03-31
Dev-stable -> dev (with conflict resolution)

afrasier:
Merge data loading hotfixes from dev stable into dev

* Add repair_sequences command, which generates relevant SQL for all apps we use which will repair sequences for those apps. (This is not run automatically so that the user can verify the SQL)
* Change the primary key for program activities to autofield
* Remove tas_rendering_label from appropriation account balances
* Validate direct reimbursable exists before using it
* New TAS list (make sure you reload it via manage.py loadtas usaspending_api/data/tas_list.csv)
* Fix tuple check in award update
* Fix summary award duplication


usaspending_api/awards/models.py
```diff
@@ -257,25 +257,25 @@ def get_or_create_summary_award(piid=None, fain=None, uri=None, awarding_agency=
                     q_kwargs_fixed.sort()
                     summary_award = awards_cache.get(q_kwargs_fixed)
                     if summary_award:
-                        return summary_award
+                        return False, summary_award
```
catherinedevlin: When there are multiple return values, it would be really good to describe them in the docstring, I think - they can be so confusing otherwise.

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -208,14 +207,15 @@ def handle(self, *args, **options):
...
-                    use_cache=True)
+                created, award = Award.get_or_create_summary_award(
+                        piid=row.get('piid'),
+                        fain=row.get('fain'),
+                        uri=row.get('uri'),
+                        parent_award_id=row.get('parent_award_id'),
+                        use_cache=True)
                 award.latest_submission = submission_attributes
-                award_queue[award.manual_hash()] = award
+                if created:
```
catherinedevlin: thanks again for catching my bug here

PR #239, opened 2017-04-02
Associate sequence with primary key of ref_program_activity

catherinedevlin:
This corrects an error I've gotten in `load_submission` since we converted `RefProgramActivity`'s primary key to an `AutoField`.

`django.db.utils.IntegrityError: duplicate key value violates unique constraint "ref_program_activity_pkey"`

I think this may be a bug in Django's migration code; I don't see any good reason for that association on a newly AutoField not to be created by default.


afrasier:
I've also written a command called "repair_sequences" that will ensure sequences are at the latest ID entry, but hopefully this migration will make it moot.

PR #240, opened 2017-04-03
Cache parents of award objects where appropriate for bulk creation

catherinedevlin:
Fixes the duplicate load bug that Kaitlin found on Friday.


catherinedevlin:
It took me hours of crawling through pdb to realize what was happening, so... no apologies warranted about not eyeballing this subtlety!

PR #246, opened 2017-04-04
Add multi-group aggregation

afrasier:
* Add multi-group aggregation
* Item is now "deprecated", but not removed
* Update serializers to respect `req` if it exists for field limiting
* Update AggregateSerializer to add any extra group items
* Update tests
* Update views to pass `req` in context to serializers


usaspending_api/common/mixins.py
```diff
@@ -114,22 +123,29 @@ def validate_request(self, params, queryset):
...
+        # make sure group fields is a list
+        if not isinstance(group_fields, list):
+            group_fields = [group_fields]
+
         # if a groupby date part is specified, make sure the groupby field is
         # a date and the groupby value is year, quarter, or month
         if date_part is not None:
+            # only allow date parts when grouping by a single field (for now)
+            if len(group_fields) > 1:
+                raise InvalidParameterException("Date parts are only valid when grouping by a single field.")
```
catherinedevlin: Good catch & nice clear exception message

usaspending_api/common/serializers.py
```diff
@@ -113,5 +119,28 @@ def setup_eager_loading(cls, queryset, prefix=""):
 
 class AggregateSerializer(serializers.Serializer):
 
+    def __init__(self, *args, **kwargs):
+        super(AggregateSerializer, self).__init__(*args, **kwargs)
+
+        include_fields = None
+
+        req = self.context.get('req')
```
catherinedevlin: Not sure, but - this harvesting of parameters from `req` and/or `request` might be a good candidate for pulling out into a helper function, both because (1) it's repeated in serializers.py, and (2) that would be a good chance to give it a clarifying name; as it stands I had to squint at it for a minute wondering what's going on.

usaspending_api/common/tests/test_aggregate_mixin.py
```diff
@@ -188,18 +196,91 @@ def test_aggregate_fy(monkeypatch, aggregate_models, model, request_data,
...
+    'field': 'total_obligation',
+    'group': ['period_of_performance_start_date__fy', 'type']
+}, _expected_fy_type_aggregated),
+    (Award, {
+        'field': 'total_obligation',
+        'group': ['uri', 'type']
+    }, _expected_type_pop_day_aggregated)])
+def test_aggregate_fy_and_type(monkeypatch, aggregate_models, model, request_data,
+                               expected):
+    request = Mock()
```
catherinedevlin: Hey, is this the first appearance in our project of Mocks?  Cool...

afrasier: Someone else wrote it first, I just stole it :)

PR #247, opened 2017-04-04
Add quarterly totals for tas/program activity/object class

bsweger:
This is another PR to support the spending over time graph on the accounts page ([DS-930](https://federal-spending-transparency.atlassian.net/browse/DS-930)).

**Updates**
* Add a quarterly table for "File B" data
* Add methods for retrieving and loading quarterly data
* Run those methods as part of the submission loader
* Do a one-time data migration to update existing submissions

After discussion the various options for getting and storing/not storing these quarterly numbers, decided on the following approach:
1. Put the quarterly numbers in their own physical table that's modeled after the annual version
2. The quarterly number load for each submission is fired from the submission loader, with a one-time data migration to handle existing submissions
3. Since doing this in a set-based way was problematic via ORM but bypassing the ORM entirely for data inserts could also be problematic due to `update_date`, `DataSourceTrackedModel`, and other things that Django manages for us, I split the difference, hoping we could get the advantages of both worlds:

    - Use `Manager.raw()` to retrieve the quarterly data, which lets us interact with a `RawQueryset`
    - Bulk load the `RawQueryset` items into the new quarterly table

If this process passes muster, I'll do the same for quarterly numbers on the File A data.


usaspending_api/financial_activities/models.py
```diff
@@ -56,3 +56,165 @@ class FinancialAccountsByProgramActivityObjectClass(DataSourceTrackedModel):
...
+        # for subsequent bulk insert
+        # TODO: maybe we don't want this entire list in memory?
+        for rec in qtr_records:
+
+            # remove any fields in the qtr_record RawQuerySet object so that we
+            # can create its TasProgramActivityObjectClassQuarterly counterpart
+            # more easily. it's a bit hacky, but avoids the need to explicitly
+            # set dozens of attributes when creating TasProgramActivityObjectClassQuarterly
+            rec_dict = rec.__dict__
+            for key in list(rec_dict.keys()):
```
catherinedevlin: I think you could boild these three lines down to 

`rec_dict = {key: rec_dict[key] for key in rec_dict if key in field_list}`

if you love comprehensions as gratuitously as I do.

usaspending_api/financial_activities/tests/test_financial_activities.py
```diff
@@ -0,0 +1,140 @@
...
+        program_activity=pa1,
+        gross_outlays_delivered_orders_paid_total_cpe=20.60,
+        obligations_incurred_by_program_object_class_cpe=5,
+        obligations_undelivered_orders_unpaid_total_fyb=99.99,
+        ussgl490800_authority_outlayed_not_yet_disbursed_fyb=30
+    )
+
+
+@pytest.mark.django_db
+def test_get_quarterly_numbers(tas_oc_pa_data):
```
catherinedevlin: This is a really nice set of tests!

usaspending_api/financial_activities/models.py
```diff
@@ -56,3 +56,165 @@ class FinancialAccountsByProgramActivityObjectClass(DataSourceTrackedModel):
...
+        """
+        Return a RawQuerySet of quarterly financial numbers by TAS,
+        object class, and program activity.
+        """
+        if current_submission_id is None:
+            return FinancialAccountsByProgramActivityObjectClass.objects.raw(
+                cls.QUARTERLY_SQL)
+        else:
+            sql = cls.QUARTERLY_SQL + ' WHERE current.submission_id = %s'
+            return FinancialAccountsByProgramActivityObjectClass.objects.raw(
```
catherinedevlin: I am fuzzy about what happens when a submission's `previous_submission_id` is `NULL` - maybe a clarifying comment?

usaspending_api/financial_activities/models.py
```diff
@@ -56,3 +56,165 @@ class FinancialAccountsByProgramActivityObjectClass(DataSourceTrackedModel):
...
+                sql, [current_submission_id])
+
+
+class TasProgramActivityObjectClassQuarterly(DataSourceTrackedModel):
```
bsweger: Yes, great idea! At the moment, the fields between these two models aren't exactly the same, since the old one has a bunch of unused columns (almost removed those as part of this PR, but decided not to pull this particular thread right now).

I added a general "accounts financial models" cleanup story: https://federal-spending-transparency.atlassian.net/browse/DS-1026

Thanks for the prompt--we should absolutely do this.

usaspending_api/financial_activities/models.py
bsweger: I do love them, and I will steal this one--thank you!

usaspending_api/financial_activities/models.py
bsweger: Good call. Overall, not my strongest docstring. Will update!

PR #248, opened 2017-04-05
File C field fixes

afrasier:
This fixes issues @ejgullo was having with the `/accounts/awards/` endpoint not serializing data properly.

* Add ProgramActivity and ObjectClass serializers
* Add said serializers to File C serializer


usaspending_api/common/tests/test_endpoints.py
```diff
@@ -102,12 +102,13 @@ def evaluateEquivalence(item1, item2):
...
                 continue
-            if 'next' in key:  # Ignore this due to checksum comparison
-                continue
-            if 'previous' in key:  # Ignore this due to checksum comparison
+            if 'page_metadata' in key:  # Ignore this due to checksum comparison
                 continue
             equality = equality and evaluateEquivalence(item1[key], item2[key])
     else:
         # Converting to string represenations to clear up some issues with Decimal casting
+        # print(str(item1))  # Leaving these here for future debugging
```
afrasier: As I said in the comment, these lines are super useful when testing why a check failed; so until we properly parameterize this I'm leaving them in here commented.

PR #261, opened 2017-04-07
Load 2017 program activity data

bsweger:
This commit removes program activity reference data from our fixtures and
loads it from a .csv file instead. Unlike some of the other loaders, it
doesn't make use of the threaded loader because that doesn't support keying
on multiple columns. 

The data is pretty straightforward, however, and wrapping the
load into a single transaction results in reasonable performance. It also helps 
that this information should only need to be updated once a year.


usaspending_api/references/management/commands/loadprogramactivity.py
```diff
@@ -0,0 +1,50 @@
...
+        if an existing row was updated
+    """
+
+    obj, created = RefProgramActivity.objects.get_or_create(
+        program_activity_code=row['pa_code'].strip().zfill(4),
+        budget_year=row['year'],
+        responsible_agency_id=row['agency_id'].strip().zfill(3),
+        main_account_code=row['account'].strip().zfill(4),
+        defaults={'program_activity_name': row['pa_name'].strip()}
+    )
```
afrasier: Will this update previous program activities that were created but are missing other data?

bsweger: It will not. Since we're removing and reloading all submissions when we go live, those submissions will pick up the correct records (once we update the loader to look for `account` in addition to `agency`, `year`, and `program activity code`. 

That will leave us with some clutter of old records. This is vexing, and in theory we could write a migration that updates our existing dev data to point to the correct PA info. Part of me really wants to write one, but I'm also not sure it's worth the time. If this all works correctly, the net effect is we have some incomplete records that no one will ever point to.

afrasier: Fair enough!

PR #262, opened 2017-04-09
Description fields

afrasier:
Shortlist:

* Adds `update_description_fields` management command
  * This updates every row, set based; however, if it becomes a slow point in the ETL it is written in a way to support future limiting by submission id
* Adds many `_description` fields per DAIMS enumerations
  * There are some fields present in the DAIMS that are not present on models: `FeeforUseofService`, `reason_for_modification`, `reason_not_competed`, `undefinitized_action`, `who_can_use_idv`
  * Adding these fields falls outside of the scope of this story, and will be left for future sprints
* Update `LimitableSerializer` so that if any coded field is requested, its description will always come back with it. (Thus, we do not have to update every set of default fields)
* Refactors current description update functionality for `Award`, `Transaction`, and `LegalEntity`
* Adds data migration to fix old 'U', 'UN' types
  * Prior to this, we had a default award_type of 'U' with description 'Unknown'; however, 'U' is actually a valid type for financial data with the now-updated DAIMS data. Thus, this data migration takes any previous 'U' and sets it to null, such that the description will remain as 'Unknown'. Due to our lack of knowledge of the 'U' type prior to this, we cannot be sure if the type is meant to be 'U' or null. Recommend to re-load any pertinent submissions if possible, or if data integrity issues arise.
* Adds data migration to run description update function
* Adds test for description update function. This test checks a subset of description fields which should automatically update.

**How it Works**

DAIMS data is stored in `usaspending_api/data/daims_maps.py`. Each key is formatted thusly:
* If the field is unique among objects, `<field_name>_map`; e.g. `business_types_map` or `sea_transportation_map`
* If the field is not unique among objects (e.g. many models use the field `type`) and the map should not be shared among them, prepend the model name followed by a period, `<model_name>.<field_name>_map`; e.g. `Award.type_map` or `Transaction.type_map` (This is just an example, as the map for `type` can and is shared between Transaction and Award models)

The command `update_description_fields` calls the helper method `update_model_description_fields` from `usaspending_api/etl/helpers.py`. This method has a list of updatable apps (apps that we want to automatically update) which allows us to exclude library, Django, or uninteresting apps (like corsheaders, or admin). This method then iterates over every model Django has registered, validates it belongs to an updatable app, then iterates over every field in the model. If the filed ends in `_description`, and there is a field present that matches the prior parts, it will attempt to update the description by using the relevant map. 

For example:
* Begins checking every field in TransactionContract
* Finds field named `idv_type_description`
* Sees `_description`, so checks for a field of `idv_type` on the same Model
* `idv_type` is present, so it now looks for a map called `Transaction.idv_type_map` or `idv_type_map` in `daims.maps.py`
* It finds the map, and thus programmatically constructs an array of `When()` expressions
* Finally, it takes the `When()` expressions, creates a `Case()` using them, and performs a set-based bulk update on the model


usaspending_api/etl/helpers.py
```diff
@@ -141,3 +144,116 @@ def get_previous_submission(cgac_code, fiscal_year, fiscal_period):
         .order_by('-reporting_fiscal_period') \
         .first()
     return previous_submission
+
+
+def update_model_description_fields():
```
catherinedevlin: Impressive!  Hopefully it will run in a tolerable amount of time, but I like being able to handle all those fields with a single general chunk of code.

usaspending_api/etl/management/commands/load_usaspending_contracts.py
```diff
@@ -84,7 +84,6 @@ def handle(self, *args, **options):
                 "gfe_gfp": h.up2colon(row['gfe_gfp']),
                 "cost_or_pricing_data": h.up2colon(row['costorpricingdata']),
                 "type_of_contract_pricing": h.up2colon(row['typeofcontractpricing']),
-                "type_of_contract_pricing_description": CONTRACT_PRICING_TYPES_D.get(h.up2colon(row['typeofcontractpricing'])),
```
catherinedevlin: I like how pulling all this stuff out into update_model_description_fields makes this function shorter and less complex.

afrasier:
@catherinedevlin New commit:
* Add ability to specify cases in DAIMS maps
* Refactor helpers.py to support cases in DAIMS maps
* Add test case for cases in DAIMS maps

Please leave a comment once you've looked over that. Just a few structural changes and extra tests to support this functionality.

afrasier:
No one winces as hard as I.

PR #263, opened 2017-04-10
Canonicalize locations

catherinedevlin:
Satisfies DS-762 by transforming location fields into a canonicalized form (capitalized, and with excess whitespace removed).  A migration transforms existing records to match.


usaspending_api/references/helpers.py
```diff
@@ -0,0 +1,39 @@
+def canonicalize_string(val):
+    """
+    Return version of string in UPPERCASE and without redundant whitespace.
+    """
+
+    try:
+        return " ".join(val.upper().split())
```
afrasier: This is a neat trick!

PR #265, opened 2017-04-10
Read replicas

willkjackson:
This PR adds two environment variables (with capability for more, up to 5 RRs on AWS), DB_SOURCE and DB_R1. If either of these are set, Django will use the replicas.py router to randomize reads between the two databases.

This *does not* account for replication lag in any way. The replication lag can be up to 5 minutes to clone any changes from the source to the replicas. Our timeouts for Elasticache are currently much higher than this, so it really won't affect anything.

Continuing to use DATABASE_URL for the connection string is fine; this won't affect any existing setups.

Plan to test this for acceptance is to (manually) disable caching temporarily and show multiple db connections (currently on sandbox).


usaspending_api/settings.py
```diff
@@ -107,11 +107,19 @@
 
 DATABASES = {'default': dj_database_url.config(conn_max_age=10)}
 
+# read replica env vars... if not set, default DATABASE_URL will get used
+# if only one set, this will error out (single DB should use DATABASE_URL)
+if os.environ.get('DB_SOURCE') or os.environ.get('DB_R1'):
```
bsweger: @willkjackson I know this PR has already been merged, but adding a note that we may want to explicitly exclude these new databases from the test suite. I had previously run into an issue where pytest was setting up extra test databases everywhere because I had the BROKER environment variable set.

Pinging @afrasier and @catherinedevlin for their thoughts. The place we ended up excluding the broker db from the test suite is: https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/conftest.py#L9

PR #266, opened 2017-04-10
WIP: Contract modification

bsweger:
This PR adds two additional fields to the transaction contract model:

1. contract modification reason (code)
2. contract modification reason (desc)

In addition, it updates the usaspending loader to load the code from `reasonformodification`.

Still to do:
Track down where this data exists on the D1 records coming from the broker.


bsweger:
Closing this--as it turns out, the field we need to load is already on the model...we just didn't have it mapped to the USAspending data.

PR #270, opened 2017-04-11
Update DAIMS map

afrasier:
Turns out the DAIMS was wrong on a few things. This fixes them. No code changes, just dict updates.


usaspending_api/data/daims_maps.py
```diff
@@ -222,7 +222,7 @@
         "B": "Indefinite Delivery Contract (IDC)",
         "C": "Federal Supply Schedule (FSS)",
         "D": "Basic Ordering Agreement (BOA)",
-        "E": "Blanket Purcchase Agreeement (BPA)",
+        "E": "Blanket Purchase Agreeement (BPA)",
```
bsweger: Heh, there's still a typo here, don't you `agreeeeee`? 😂

afrasier: This was how it was in the DAIMS, so don't blame me lol

bsweger: Shouldn't we correct the typo, tho?

PR #274, opened 2017-04-13
Executive compensation models

afrasier:
* Add `LegalEntityOfficers` model
* Add 1-1 relationship from `LegalEntityOfficers` to `LegalEntity`
* Ensure a `LegalEntity` always has a set of officers, even if all the
officers are null
* Data migration to initialize `LegalEntityOfficers` on existing
`LegalEntity` objects

ETL to come in future PR when Broker has updated their table structure.


usaspending_api/references/models.py
```diff
@@ -400,6 +408,43 @@ class Meta:
...
+    officer_1_amount = models.DecimalField(max_digits=20, decimal_places=2, blank=True, null=True)
+    officer_2_name = models.TextField(null=True, blank=True)
+    officer_2_amount = models.DecimalField(max_digits=20, decimal_places=2, blank=True, null=True)
+    officer_3_name = models.TextField(null=True, blank=True)
+    officer_3_amount = models.DecimalField(max_digits=20, decimal_places=2, blank=True, null=True)
+    officer_4_name = models.TextField(null=True, blank=True)
+    officer_4_amount = models.DecimalField(max_digits=20, decimal_places=2, blank=True, null=True)
+    officer_5_name = models.TextField(null=True, blank=True)
+    officer_5_amount = models.DecimalField(max_digits=20, decimal_places=2, blank=True, null=True)
+
```
bsweger: Should we have a `create_date` field here too?

usaspending_api/references/migrations/0067_create_officers.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.1 on 2017-04-13 17:41
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def update_legal_entities(apps, schema_editor):
```
bsweger: Thanks for thinking to do a data migration. I'm not sure why we need to ensure that every LegalEntity has a set of officers? If this is ensure that we have something to serialize for the front-end while we're loading in the real data, is there another way we can mock something out for them? 

Or is this to ensure that the `LegalEntity.officer` query pattern will always work? That's something I didn't think of re: the `OneToOne` pattern.

usaspending_api/references/models.py
```diff
@@ -383,6 +383,14 @@ class LegalEntity(DataSourceTrackedModel):
     small_business_description = models.TextField(blank=True, null=True)
     individual = models.CharField(max_length=1, blank=True, null=True)
 
+    def save(self, *args, **kwargs):
+        super(LegalEntity, self).save(*args, **kwargs)
+
+        try:
```
bsweger: Is there a way that Django's `get_or_create` pattern would work here? Overall, that seems like a cleaner way than try/catching db errors?

usaspending_api/references/models.py
afrasier: In all future cases, the Officer object is created at the same time as the LegalEntity object; but we need update_date for the ETL process

usaspending_api/references/migrations/0067_create_officers.py
afrasier: Yeah, this is to ensure that there is _always_ an Officer for each legal entity item. It has two bonuses, both of which you point out: the `LegalEntity.officers` pattern always works, and the front end has an easier life of just checking if an officer is null vs. the entire nested object being null.

usaspending_api/references/models.py
afrasier: Yeah, I could do it that way. This is one of those cases where you write it in a more complex way, then start paring things down as you debug, and then you get to this point, hah. I'll update it first thing.

usaspending_api/references/migrations/0067_create_officers.py
afrasier: Although _one_ reason to use this pattern is I think it might be faster to do a reverse lookup from this item, rather than search through all LegalEntityOfficers for the specific duns. (i.e. a get_or_create)

PR #276, opened 2017-04-13
Submission loader award mapping updates

bsweger:
This PR represents part of the work necessary to fix up the award mapping in the submission loader. Note to reviewer--more detailed info in the commit messages. Might be easiest to review changeset by changeset.

**Included updates:**
* Add awarding agency as a criteria when mapping D1 and D2 records to awards
* Process the C file after the D1 and D2 files to ensure that award records are created with the most detailed possible data (i.e., awarding agency data includes subtier)
* When processing C files, try to match them to D1/D2 files so we can grab the subtier agency info for award matching
* Update award matching to match on toptier agency alone is we don't have awarding subtier info
* Remove `--delete` parameter on the`load_submission` command. Deleting an existing submission is now the default behavior.
* Remove `latest_submission` from the `Award` model

**Still to do:**
* More complete unit tests re: `Award.get_or_create_summary_award` (to test partial agency matching)


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -103,342 +96,43 @@ def handle(self, *args, **options):
...
-            field_map = {}
-
-            load_data_into_model(appropriation_balances, row, field_map=field_map, value_map=value_map, save=True, reverse=reverse)
-
-        AppropriationAccountBalances.populate_final_of_fy()
-
-        # Insert File A quarterly numbers for this submission
-        AppropriationAccountBalancesQuarterly.insert_quarterly_numbers(
-            submission_attributes.submission_id)
+        load_file_a(submission_attributes, appropriation_data, db_cursor)
```
afrasier: 👍 👍 👍 👍 finally refactored, thank you for taking the time!!

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -755,6 +445,383 @@ def dictfetchall(cursor):
...
+            field_map=contract_field_map,
+            value_map=contract_value_map,
+            save=True)
+
+    # Update awards for new linkages
+    update_awards(tuple(AWARD_UPDATE_ID_LIST))
+    update_contract_awards(tuple(AWARD_CONTRACT_UPDATE_ID_LIST))
+
+    # Update the descriptions TODO: If this is slow, add ID limiting as above
+    update_model_description_fields()
```
afrasier: Should this be in load_d1? Should we run this after all records have been created (i.e. after all files have been loaded)

usaspending_api/submissions/management/commands/submission_stats.py
```diff
@@ -0,0 +1,65 @@
...
+    TransactionContract)
+from usaspending_api.financial_activities.models import (
+    FinancialAccountsByProgramActivityObjectClass, TasProgramActivityObjectClassQuarterly
+)
+from usaspending_api.submissions.models import SubmissionAttributes
+
+logger = logging.getLogger('console')
+
+
+class Command(BaseCommand):
```
afrasier: This is probably a pretty useful command for @willkjackson

usaspending_api/etl/award_helpers.py
```diff
@@ -119,3 +120,59 @@ def update_contract_awards(award_tuple=None):
...
+        uri: uri from File C (assistance awards only)
+
+    Returns:
+        A Transaction model instance
+    """
+    # if both fain and uri are supplied as paramaters, look up by fain first
+    incoming_fain = fain
+    incoming_uri = uri
+    if incoming_fain is not None and incoming_uri is not None:
+        uri = None
```
catherinedevlin: If the intent here is "do not filter on URI if `incoming_fain` is present", I don't think this does it.  Filtering for `assistance_data__uri=None` is different from not filtering on it.  In development:

```
Transaction.objects.filter(assistance_data__fain__isnull=False).count()

58098

Transaction.objects.filter(assistance_data__fain__isnull=False). \
    filter(assistance_data__uri=None).count()

2510
```

usaspending_api/submissions/management/commands/submission_stats.py
```diff
@@ -0,0 +1,65 @@
...
+        AppropriationAccountBalances,
+        AppropriationAccountBalancesQuarterly,
+        FinancialAccountsByProgramActivityObjectClass,
+        TasProgramActivityObjectClassQuarterly,
+        Transaction,
+        TransactionAssistance,
+        TransactionContract,
+        FinancialAccountsByAwards]
+
+    stats = [(m._meta.model_name, m.objects.filter(
```
catherinedevlin: Wow, what a nice concise implementation!

usaspending_api/etl/award_helpers.py
afrasier: Right, I think we just don't want `assistance_data__uri` in the filter _at all_, rather than checking for None. Unless we actually do want to only find awards where the FAIN matches but URI is none.

catherinedevlin: OK, I'll make a little PR to the PR to that end...

catherinedevlin: OK, there's now a small PR to this PR: https://github.com/fedspendingtransparency/usaspending-api/pull/280 - @afrasier , can you review it?

usaspending_api/etl/management/commands/load_submission.py
bsweger: Good call--I think this got swept up in moving the code around...will push a change that runs this after all files are processed.

bsweger:
Closing this in favor #286 (which includes the same submission loader changes, but w/o the `dev` merges.

usaspending_api/awards/models.py
```diff
@@ -217,38 +219,66 @@ def get_or_create_summary_award(piid=None, fain=None, uri=None, awarding_agency=
...
+                    .filter(awarding_agency=awarding_agency) \
+                    .first()
+                if (summary_award is None and
+                        awarding_agency is not None and
+                        awarding_agency.toptier_agency.name != awarding_agency.subtier_agency.name):
+                    # No award match found when searching by award id info +
+                    # awarding subtier agency. Relax the awarding agency
+                    # critera to just the toptier agency instead of the subtier
+                    # agency and try the search again.
+                    awarding_agency_toptier = Agency.objects.filter(
```
bsweger: Yes, great idea! I have to revisit this for a late-breaking agency edge-case, so will take you up on this suggestion.

PR #278, opened 2017-04-14
Subaward models

afrasier:
* Model for File F
* Aggregate fields on `Award` for subawards (`total_subaward_amount` and `subaward_count`)
* Add serializer for subawards to award; it is not default so must be specifically requested or it will only appear on a detail request (i.e. `/awards/1234`)

ETL + Aggregate methods will come in a future PR, this is "stubby" for the front end.


usaspending_api/awards/models.py
```diff
@@ -494,29 +498,27 @@ class Meta:
 
 
 class SubAward(DataSourceTrackedModel):
```
bsweger: Very minor point, but what do you think about un-CamelCasing this for future ease of typing?

usaspending_api/awards/models.py
```diff
@@ -494,29 +498,27 @@ class Meta:
...
+    submission = models.ForeignKey(SubmissionAttributes, models.CASCADE)
+    cfda = models.ForeignKey(CFDAProgram, models.DO_NOTHING, null=True)
+    funding_agency = models.ForeignKey(Agency, models.DO_NOTHING, null=True)
+    place_of_performance = models.ForeignKey(Location, models.DO_NOTHING, null=True)
+
+    subaward_number = models.IntegerField(db_index=True)
+    amount = models.DecimalField(max_digits=20, decimal_places=2)
+    description = models.TextField(null=True, blank=True)
+
+    recovery_model_question1 = models.CharField(max_length=1, null=True, blank=True)
```
bsweger: Given that subaward data has a reputation for being...problematic and that it's not really our job to validate the incoming data (and not to mention the timeline), what do you think about making questions 1 & 2 straight up text fields?

usaspending_api/awards/models.py
```diff
@@ -494,29 +498,27 @@ class Meta:
...
+    recovery_model_question2 = models.CharField(max_length=1, null=True, blank=True)
+
+    action_date = models.DateField(blank=True, null=True)
+    award_report_month = models.IntegerField()
```
bsweger: Did we ever find out if this is a fiscal period or a calendar month?

usaspending_api/awards/models.py
```diff
@@ -494,29 +498,27 @@ class Meta:
...
+    award_report_year = models.IntegerField()
```
bsweger: Same question here...is this a fiscal year or calendar year?

usaspending_api/awards/models.py
```diff
@@ -494,29 +498,27 @@ class Meta:
...
+
+    naics = models.CharField(max_length=6, blank=True, null=True, verbose_name="NAICS", help_text="Specified which industry the work for this transaction falls into. A 6-digit code")
```
bsweger: Again, I'm inclined to be less aggressive on the field lengths here, since we don't really know what's coming our way on the data loads.

usaspending_api/awards/models.py
afrasier: I have absolutely no qualms, that's just how it was before.

usaspending_api/awards/models.py
afrasier: I was thinking of doing that, this is just how they were before. TextFields it is!

usaspending_api/awards/models.py
afrasier: I never could tell from looking at a few rows, I'll ask Kristine

usaspending_api/awards/models.py
afrasier: Agreed, I'll probably just TextField it. There's not a real big reason not to.

afrasier:
Per Kevin's request, added:
* /api/v1/subawards/
* /api/v1/subawards/{pk}
* /api/v1/subawards/total
* /api/v1/subawards/autocomplete

usaspending_api/awards/models.py
```diff
@@ -493,30 +497,28 @@ class Meta:
...
-    subaward_number = models.CharField(max_length=32, blank=True, null=True)
-    reporting_period_start = models.DateField(blank=True, null=True)
-    reporting_period_end = models.DateField(blank=True, null=True)
-    last_modified_date = models.DateField(blank=True, null=True)
-    certified_date = models.DateField(blank=True, null=True)
-    create_date = models.DateTimeField(auto_now_add=True, blank=True, null=True)
-    update_date = models.DateTimeField(auto_now=True, null=True)
+class Subaward(DataSourceTrackedModel):
+    # Foreign keys
+    award = models.ForeignKey(Award, models.CASCADE, related_name="subawards")
```
catherinedevlin: I appreciate gathering the columns into whitespace-separated logical clumps.

PR #279, opened 2017-04-16
Uppercase ref city county

catherinedevlin:
Earlier work on DS-762 (making a single canonical location name, in uppercase) didn't finish the problem, because during the save, mixed-case names were pulled from RefCityCounty.  Now fixed.


usaspending_api/references/helpers.py
```diff
@@ -10,20 +10,32 @@ def canonicalize_string(val):
...
+
+
+def fields_by_partial_names(model, substrings):
+    """
+    For any model, all field names containing any substring from `substrings`.
+    """
+    for f in model._meta.fields:
+        for substring in substrings:
+            if substring in f.name:
+                yield f.name
```
afrasier: Nice use of `yield` here

PR #280, opened 2017-04-17
Fix to the logic of filtering by FAIN and/or URI

catherinedevlin:
The original filters specifically for records where URI is null;
this one simply does not filter by URI when it is not supplied.
I believe that's the intended effect.

### DIFF
--- a/usaspending_api/etl/award_helpers.py
+++ b/usaspending_api/etl/award_helpers.py
@@ -158,21 +158,13 @@ def get_award_financial_transaction(
     txn = Transaction.objects.filter(
         awarding_agency__toptier_agency__cgac_code=toptier_agency_cgac,
         contract_data__piid=piid,
-        contract_data__parent_award_id=parent_award_id,
-        assistance_data__fain=fain,
-        assistance_data__uri=uri) \
-        .order_by('-action_date').first()
+        contract_data__parent_award_id=parent_award_id
+        ).order_by('-action_date')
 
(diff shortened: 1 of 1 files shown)


afrasier:
I'll let becky approve this because I'm not sure if it's the intended effect or not.

kaitlin:
I believe you're correct. Search on FAIN first (regardless of the URI value). If nothing matches, _then_ search on URI. But if two records have the same FAIN and different URIs, they should match.

bsweger:
Excellent, thanks for the clarification, @kaitlin. I feel like I'm still missing something about these changes, @catherinedevlin. It looks like:

1. We're filtering by piid/parent id
2. Then filtering the results of #1 by fain/uri

Would we want to do two separate and independent searches of the `Transaction` data depending on the award type?

kaitlin:
yeah that doesn't make sense to me ... we should either filter on PIID or FAIN but not both.

bsweger:
Ok, I opened story [DS-111](https://federal-spending-transparency.atlassian.net/browse/DS-1111) to circle back to this. I think we can merge `misc-submission` and start re-running the test loads without this fix.

@catherinedevlin It's a great catch--thank you! Sounds like we need to think through the solution a bit. 

Loading submisisons w/o this fix means, essentially, that we might do some odd matching to D records when trying to find a sub-tier agency for award matching. Because we do have a backup plan to match/create awards using top-tier only, that should be a reasonable backup.

PR #281, opened 2017-04-18
updated submission loader to capture PA name

kaitlin:
After noticing that a bunch of program activities were being created but remaining nameless, I added an update to at least capture the name as we first see it in a submission. We won't necessarily track all variations that exist for a particular PA across submissions and if an agency needs to update the title of a PA, we can ask agency submitters to update their PA lists with OMB and provide the "official" name. 

Also include a quicky unit test.

### DIFF
--- a/usaspending_api/etl/management/commands/load_submission.py
+++ b/usaspending_api/etl/management/commands/load_submission.py
@@ -205,14 +205,11 @@ def get_or_create_program_activity(row, submission_attributes):
                'main_account_code': row['main_account_code'], }
     prg_activity = RefProgramActivity.objects.filter(**filters).first()
     if prg_activity is None and row['program_activity_code'] is not None:
-        prg_activity = RefProgramActivity.objects.create(**filters)
+        # If the PA has a blank name, create it with the value in the row.
+        # PA loader should overwrite the names for the unique PAs from the official
+        # domain values list if the title needs updating, but for now grab it from the submission
+        prg_activity = RefProgramActivity.objects.create(**filters, program_activity_name=row['program_activity_name'])
         logger.warning('Created missing program activity record for {}'.format(str(filters)))
-    # If the PA has a blank name, update it with the value in the row.
-    # PA loader should overwrite the names for the unique PAs from the official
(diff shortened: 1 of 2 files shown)


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -207,6 +207,13 @@ def get_or_create_program_activity(row, submission_attributes):
     if prg_activity is None and row['program_activity_code'] is not None:
         prg_activity = RefProgramActivity.objects.create(**filters)
         logger.warning('Created missing program activity record for {}'.format(str(filters)))
+    # If the PA has a blank name, update it with the value in the row.
+    # PA loader should overwrite the names for the unique PAs from the official
+    # domain values list if the title needs updating, but for now grab it from the submission
+    if prg_activity and not prg_activity.program_activity_name:
```
bsweger: Since the list of PA data we originally load from the .csv (https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/data/program_activity.csv) doesn't have any blank names, do we need this extra `if` statement? In other words, is it safe just to add the name when creating a new PA record up in line 208?

If you see existing records with an empty PA name in our data, those likely predate our updated PA load from the last sprint. We didn't clean those up as part of that work, choosing not to spend the time on a data migration that would fix up soon-to-be-deleted submissions. In retrospect, leaving those old, partial records there probably created some unnecessary confusion.

kaitlin: I don't _think_ so because the names may vary slightly across submissions and we don't want to create duplicate program activities for slightly different names. I definitely got blank PA names when loading everything fresh from the `misc-submission` branch so I don't think it was addressed before this.

bsweger: Yeah, definitely we were not putting in program activity name when creating PA records on the fly, so not surprising that you saw blank names. What I'm trying to figure out is why we can't just pop in the name when creating a new PA record up in lines 207-8. Does that make sense?

kaitlin:
Ah I had in my mind that the call to `create` was a call to `get_or_create`. I've consolidated the updates.

PR #283, opened 2017-04-18
No dupe transactions

catherinedevlin:
Addresses most of DR-1048, except for tests and change logging (change logging is in #282)


usaspending_api/awards/models.py
```diff
@@ -310,6 +328,24 @@ def get_default_fields(path=None):
...
 
+    @classmethod
+    def get_or_create(cls, **kwargs):
+        """Gets and updates, or creates, a Transaction
+
+        Transactions must be unique on Award, Awarding Agency, and Mod Number
+        """
+        transaction = cls.objects.filter(
+            award=kwargs.get('award'),
+            awarding_agency=kwargs.get('awarding_agency'),
```
bsweger: _In theory_, the awarding agency is part of the award. So, it might not be necessary to include `awarding_agency` in this lookup.

Pinging @kaitlin because awards and awarding agencies have been vexing lately, and would be grateful for her .02.

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -315,16 +315,17 @@ def handle(self, *args, **options):
                 "action_date": format_date(row['action_date']),
             }
 
-            transaction_instance = load_data_into_model(
-                Transaction(), row,
+            transaction_dict = load_data_into_model(
```
bsweger: Thanks for renaming this to match the type of object coming back...it's the little things!

usaspending_api/awards/models.py
```diff
@@ -310,6 +328,24 @@ def get_default_fields(path=None):
             "assistance_data"  # must match related_name in TransactionAssistance
         ]
 
+    @classmethod
+    def get_or_create(cls, **kwargs):
```
bsweger: Outing myself as on the noob-ier side of Django here, but is it correct to assume that this overrides the model's `get_or_create` function? Would someone still be able to use the regular `get_or_create` with the `defaults` parameter and whatnot?

usaspending_api/awards/models.py
```diff
@@ -310,6 +328,24 @@ def get_default_fields(path=None):
...
+        """Gets and updates, or creates, a Transaction
+
+        Transactions must be unique on Award, Awarding Agency, and Mod Number
+        """
+        transaction = cls.objects.filter(
+            award=kwargs.get('award'),
+            awarding_agency=kwargs.get('awarding_agency'),
+            modification_number=kwargs.get('modification_number')
+        ).first()
```
bsweger: To your question on the JIRA ticketk about a unique index. That sure would be preferable. Again, _in theory_, if we start with a fresh db, that should be fine, once the loaders are all fixed up to updated existing txns instead of inserting new ones.

The only issue I foresee is the existing historical data we have loaded. If none of that would violate then unique constraint, I think we should add it. Otherwise, we may have to put a story for that in the backlog.

If we're unable to add the unique constraint (which I agree is the preferable solution), what do you think about adding a descending sort here on `update_date` to ensure that we're getting the latest existing record when there are more than one? Then we'll at least know that if there are somehow multiple txn records, we're always updating the same one.

usaspending_api/awards/models.py
```diff
@@ -258,7 +259,24 @@ class Meta:
         db_table = 'awards'
 
 
-class Transaction(DataSourceTrackedModel):
+class TransactionAgeComparisonMixin:
+
+    def newer_than(self, dct):
+        """Compares age of this instance to a dict
+
+        Compares (last_modified_date or certified_date) of each
```
bsweger: For code readers who aren't super well-versed in Python, it might be worth expanded this description to specifically state that we check last_modified date first and then certified date.

catherinedevlin: Good point, will fix.

usaspending_api/awards/models.py
catherinedevlin: OK, I'll sort the records, and write a ticket for a potential unique constraint.

usaspending_api/awards/models.py
catherinedevlin: Good question, but it doesn't override, because the usual `get_or_create` is defined on `Transaction.objects`, not on `Transaction`.  Note the difference:
```
In [4]: Transaction.objects.get_or_create
Out[4]: <bound method BaseManager._get_queryset_methods.<locals>.create_method.<locals>.manager_method of <django.db.models.manager.Manager object at 0x1044a84e0>>

In [5]: Transaction.get_or_create
Out[5]: <bound method Transaction.get_or_create of <class 'usaspending_api.awards.models.Transaction'>>
```
I chose the name because that's a customary name for getting or creating, but maybe I should change the name to avoid confusion?

bsweger: I don't have overly-strong feelings. We already have a `get_or_create_summary_award` method on `Award`, so precedent for doing that, even though I appreciate your initial approach to being less verbose.

Given that there's an argument for either name, might was well keep what you have.

bsweger:
Just committed the last of the unit tests, so taking this out of WIP status. In addition to the tests:
* removed awarding subtier agency as part of the txn lookup, since we've already checked that information when linking the txn record to an award
* made a tweak to pull `certified_date` from the submission table, since the `certified_date` field on the `Transaction` model isn't populated (we have a model cleanup story in the backlog to fix that)

There was a bit of thrashing around as we refined the requirements, so wouldn't recommend reviewing changeset by changeset.

usaspending_api/awards/models.py
catherinedevlin: Precedent sounds like a good reason - I'm making that change.

bsweger:
Thanks for the updates, @catherinedevlin. I'd love for @afrasier (or @kaitlin if she wants to review the test cases) to take a look since you and I are both in too deep now...

PR #284, opened 2017-04-18
Browsable API performance -> dev-stable

afrasier:
@willkjackson 

This PR merges in the changes from the file-f-etl branch which update the Browsable API renderer to make it not hammer the DB to build an HTML form

This just customizes the BrowsableAPIRenderer and changes the settings file to use this renderer.


afrasier:
This may be broken, fixing...

PR #285, opened 2017-04-18
Subaward ETL Process

afrasier:
* Adds `subaward_etl.py`, which contains a method to perform File F etl
* Adds a management command `load_subawards`
  * Loading a single submission's subawards: `load_subawards -s <submission id>`
  * Loading the subawards for every submission in the datastore: `load_subawards --all`
* Add a call to the subaward loader at the end of `load_submissions` so any new or re-loaded will have their subawards also loaded
* Added tests for the subaward ETL, and added test data to relevant JSON files
* Refactored `dictfetchall` and `PhonyCursor` out of `load_submission.py` into `broker_etl_helpers.py` since it is also needed by the File F etl (and the incoming File E etl)
* Added `update_award_subawards` to `award_helpers`, which sums the total subaward amounts and attaches the counts to the prime awards. 
* Updated subaward serializer to serialize certain nested objects where needed

* Added custom renderer that removes the HTML form and raw context at the bottom of the Browsable API, which were previously hammering the DB when using the browsable API. It is much faster and easier to use now.


usaspending_api/etl/subaward_etl.py
```diff
@@ -0,0 +1,246 @@
+import logging
```
catherinedevlin: Thank you for splitting this out into a new file; load_submission.py is so big already.

usaspending_api/common/renderers.py
```diff
@@ -0,0 +1,61 @@
...
+
+class BrowsableAPIRendererWithoutForms(BrowsableAPIRenderer):
+    """Renders the browsable api, but excludes the HTML form."""
+
+    def get_rendered_html_form(self, data, view, method, request):
+        """Don't render the HTML form"""
+        return None
+
+    # Lifted from the DRF, but modified to remove content + HTML form
+    def get_raw_data_form(self, data, view, method, request):
```
afrasier: This is called by a templating function in DRF :)

usaspending_api/etl/award_helpers.py
```diff
@@ -119,3 +119,35 @@ def update_contract_awards(award_tuple=None):
         rows = cursor.rowcount
 
     return rows
+
+
+def update_award_subawards(award_tuple=None):
```
afrasier: I wanted to keep all the SQL helpers together so that hopefully we can refactor them all together in the future.

usaspending_api/etl/award_helpers.py
```diff
@@ -119,3 +119,35 @@ def update_contract_awards(award_tuple=None):
...
+    """
+    Updates awards' subaward counts and totals
+    """
+    # Sum and count subaward_amounts
+    sql_sub_totals = (
+        'subaward_totals AS ('
+        'SELECT award_id, SUM(amount) AS total_subaward_amount, COUNT(*) AS subaward_count '
+        'FROM awards_subaward ')
+    if award_tuple:
+        sql_sub_totals += 'WHERE award_id IN %s '
```
afrasier: Not for these commands, no. I was just using the code we currently have from prior helpers. We should look to refactor/tidy them up as time permits because I agree it is a bit fast and loose.

PR #286, opened 2017-04-19
Dev stable misc submission

bsweger:
This replaces PR #276. It includes the same updates to the submission loader, but excludes the `dev` merges so we can hotfix to `dev-stable`/`staging`.


bsweger:
@kaitlin @willkjackson @afrasier @catherinedevlin Added some logic to handle the unexpected case when an incoming D file doesn't have the awarding agency toptier cgac code. This scenario was causing the following behavior, which we don't want:

1. D file is processed and creates an award record w/o the awarding agency info
2. C file is processed, tries to find an award record by matching on awarding agency/award id, can't find anything
3. Thus, C file routine creates and maps to a new award record, this one w/ the awarding agency info
4. Now we have two awards records...one maps to C, and one maps to D

Committed some changes to inspect toptier agency info on the incoming D files. If the upstream process has not filled in that info, look it up using the subtier code.

I re-ran submission 2450 with these changes, and only saw a single award record for FAIN `DEFE0012829` (prior to this, we got two).

usaspending_api/references/models.py
```diff
@@ -72,6 +72,54 @@ def get_default_fields(path=None):
             "office_agency"
         ]
 
+    @staticmethod
+    def get_by_toptier(toptier_cgac_code):
```
catherinedevlin: These specifically-named methods are very nicely clarifying.

PR #293, opened 2017-04-20
Fix Bug, Speed up

afrasier:
Speeds up File F ETL by only pulling subawards we have the ability to
match and load.

Fixes a bug with FAIN lookup.

Updates tests


usaspending_api/etl/subaward_etl.py
```diff
@@ -18,6 +18,7 @@
                                          AND fsrs_procurement.idv_reference_number = award_procurement.parent_award_id
        LEFT OUTER JOIN fsrs_subcontract  ON fsrs_subcontract.parent_id = fsrs_procurement.id
 WHERE  award_procurement.submission_id = %s
+       AND award_procurement.piid IN %s
```
bsweger: I haven't been following along--do we need to care about parent award id for the purposes of this lookup?

usaspending_api/etl/subaward_etl.py
```diff
@@ -26,6 +27,7 @@
        LEFT OUTER JOIN award_financial_assistance ON fsrs_grant.fain = award_financial_assistance.fain
        LEFT OUTER JOIN fsrs_subgrant ON fsrs_grant.id = fsrs_subgrant.parent_id
 WHERE  award_financial_assistance.submission_id = %s
+       AND award_financial_assistance.fain IN %s
```
bsweger: Similar question to above--do we care about URI here?

usaspending_api/etl/subaward_etl.py
afrasier: No, we check for a matching parent award ID at the bottom; this is just a small check that ensures we only pull Records we have the _possibilty_ of matching.

afrasier:
I do have tests for submission/piid/fain/etc. matching, making sure they attach properly. They were part of the previous PR.

usaspending_api/etl/subaward_etl.py
```diff
@@ -68,6 +65,7 @@ def load_subawards(submission_attributes, db_cursor):
         # a matching parent award id, piid, and submission attributes
         award = Award.objects.filter(transaction__submission=submission_attributes,
                                      transaction__contract_data__piid=row['piid'],
+                                     transaction__contract_data__isnull=False,
```
bsweger: If the filters are AND'd together, would `transaction__contract_data__isnull=False` always be true if `transaction__contract_data__piid=row['piid']`?

afrasier: Nope! OneToOne relationships do this weird thing where even though contract_data is null, it still lets you check contract_data__piid, and returns null. Just making sure to be explicit here.

afrasier: i.e. `Transaction.objects.filter(contract_data__isnull=True, contract_data__piid__isnull=True)` is perfectly valid (which is kind of weird, but... django)

usaspending_api/etl/subaward_etl.py
```diff
@@ -68,6 +65,7 @@ def load_subawards(submission_attributes, db_cursor):
         # a matching parent award id, piid, and submission attributes
         award = Award.objects.filter(transaction__submission=submission_attributes,
```
bsweger: I know this code is already in dev, but something I found helpful when doing various submission loader fixes is make a discrete method on the model (in this case, `Award`) for these types of matches. That made it a lot easier to unit test the edge cases with good old model mommy.

usaspending_api/etl/subaward_etl.py
bsweger: We discussed offline. All is well.

usaspending_api/etl/tests/test_subaward_etl.py
```diff
@@ -27,7 +27,9 @@ def test_subaward_etl_fixture():
 
     # Transaction with subaward, by FAIN
     txn2 = mommy.make(Transaction, award=prime_award_2, submission=submission)
-    mommy.make(TransactionAssistance, transaction=txn2, fain=test_fain, uri=test_uri)
+    txn3 = mommy.make(Transaction, award=prime_award_2, submission=submission)
+    mommy.make(TransactionAssistance, transaction=txn2, fain=test_fain, uri=None)
+    mommy.make(TransactionAssistance, transaction=txn3, fain=None, uri=test_uri)
```
bsweger: Thanks for adding this!

PR #295, opened 2017-04-21
Add certified date to SubmissionAttributes

bsweger:
This changeset adds a certified date to our submission table. Because
the broker's submission table doesn't have a certified date field, we
grab their latest submission update date when when loading broker data.


afrasier:
Code looks good; though, should we use a different name if it's not _actually_ the certified date? We already have confusing field names so we should avoid proliferating them. Or is the latest update date guaranteed to be the same date as the certification from broker?

bsweger:
@afrasier Yeah, it's a bit awkward b/c the broker doesn't actually have a field called `certified_date`. I confirmed with @nmonga91 that `updated_at` on a submission that's ready for publication is essentially the certified date.

For sure, there should be a follow-up PR to clean up the remainder of the `drv_` fields and repeated submission attributes that are proliferating. I had to skip that b/c there isn't time right now to update the various corresponding fixtures and .json files. I added those tasks to our "model clean-up" ticket: https://federal-spending-transparency.atlassian.net/browse/DS-1026

Whaddaya think?

bsweger:
@afrasier.... @kaitlin is putting in a broker story to make an explicit `certified_date` on the submission table. She confirmed that we can use `updated_at` in the meantime

PR #297, opened 2017-04-21
Subaward etl databugfixes

afrasier:
This fixes some issue with Subaward ETL and it fighting with data constraints

* Widen the Zip Last 4 to take longer char strings
* Allow `load_subawards` to take a list of multiple submissions
* Change query SQL to eliminate duplicated rows (saves processing time)
* Include `awarding_agency` as a linkage for subaward->award
* Update tests for these new cases/constraints


usaspending_api/etl/management/commands/load_subawards.py
```diff
@@ -60,17 +62,29 @@ def handle(self, *args, **options):
...
         if options["update_all"]:
             submissions_to_update = SubmissionAttributes.objects.exclude(broker_submission_id__isnull=True)
         else:
-            submission = SubmissionAttributes.objects.filter(broker_submission_id=options['submission_id']).first()
-            if not submission:
-                logger.critical("Submission {} not found in datastore".format(options['submission_id']))
-                return
-            submissions_to_update = [submission]
+            for submission_id in options['submission_id']:
+                sub = SubmissionAttributes.objects.filter(broker_submission_id=submission_id).first()
```
bsweger: Now that the submission loader automatically deletes any submission with the same id, do you think it's time to put a unique index on `broker_submission_id`? I guess that's out of scope, but all of these `first()` queries make me slightly uneasy.

Sometimes when I'm using that pattern, I add a `sort_by(-update_date)` to the query so at least know we're always operating on the same record in the event that something went awry. Do you think that's warranted here?

usaspending_api/etl/management/commands/load_subawards.py
```diff
@@ -60,17 +62,29 @@ def handle(self, *args, **options):
...
+                    logger.critical("Submissions not found in datastore".format(options['submission_id']))
+                else:
+                    submissions_to_update.append(sub)
 
+        failed_submissions = []
+        success_submissions = []
         for submission in submissions_to_update:
-            logger.info("Loading subaward data for submission {}".format(submission.broker_submission_id))
-            load_subawards(submission, db_cursor)
+            try:
```
bsweger: I'm torn here...I see the need for a soft landing as we're loading all of this data, but am also wary of wrapping everything in a `try/catch` with a generic exception. Are there more specific exceptions we can anticipate and handle?

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -139,7 +139,10 @@ def handle(self, *args, **options):
         # Once all the files have been processed, run any global
         # cleanup/post-load tasks.
         # 1. Load subawards
-        load_subawards(submission_attributes, db_cursor)
+        try:
```
bsweger: I think we need a larger conversation about how the loader should behave on a hard fail. Nothing else is operating in a try/catch right now (mostly because that was causing a lot of uncaught data issues).

If we need to handle fails, does it make sense to percolate loader errors up to a higher entity that can catch them? I'd rather not wrap one part of the process in such a generic `try/catch`.

usaspending_api/etl/subaward_etl.py
```diff
@@ -12,7 +12,7 @@
 
 # These queries are directly from the broker
 D1_FILE_F_QUERY = """
-SELECT *
+SELECT DISTINCT ON (subcontract_num, fsrs_procurement.contract_number, fsrs_procurement.idv_reference_number) *
```
bsweger: Nice-- `DISTINCT ON` is on of my favorite Postgres features. Another subaward n00b question...does this mean that our source data has duplicate records and we are trying to filter those out? Are we certain we can throw away the extras?

usaspending_api/etl/subaward_etl.py
```diff
@@ -22,7 +22,7 @@
 """
 
 D2_FILE_F_QUERY = """
-SELECT *
+SELECT DISTINCT ON (subaward_num, award_financial_assistance.fain, award_financial_assistance.uri) *
```
bsweger: Same question as above re: `DISTINCT ON`

usaspending_api/etl/subaward_etl.py
```diff
@@ -60,10 +64,18 @@ def load_subawards(submission_attributes, db_cursor):
...
+            logger.warn("Subaward number {} cannot find matching agency with toptier code {} and subtier code {}".format(row['subcontract_num'], row['awarding_agency_code'], row['awarding_sub_tier_agency_c']))
+            continue
+
         # Find the award to attach this sub-contract to
         # We perform this lookup by finding the Award containing a transaction with
         # a matching parent award id, piid, and submission attributes
-        award = Award.objects.filter(transaction__submission=submission_attributes,
+        award = Award.objects.filter(awarding_agency=agency,
+                                     transaction__submission=submission_attributes,
                                      transaction__contract_data__piid=row['piid'],
```
bsweger: I know we discussed this at length in the last go-round, but could you just search the award itself for `piid` and `parent_award_id`, thus avoiding the need get into `transaction__contract_data`?

usaspending_api/etl/subaward_etl.py
```diff
@@ -60,10 +64,18 @@ def load_subawards(submission_attributes, db_cursor):
...
                                      transaction__contract_data__isnull=False,
                                      transaction__contract_data__parent_award_id=row['parent_award_id']).distinct().first()
```
bsweger: Apologies for not noticing this before, but what do you think about adding a `sort_by` here as a hedge against some kind of data malfunction?

usaspending_api/etl/subaward_etl.py
```diff
@@ -135,17 +147,25 @@ def load_subawards(submission_attributes, db_cursor):
...
+        if not agency:
+            logger.warn("Subaward number {} cannot find matching agency with toptier code {} and subtier code {}".format(row['subaward_num'], row['awarding_agency_code'], row['awarding_sub_tier_agency_c']))
+            continue
+
         # Find the award to attach this sub-award to
         # We perform this lookup by finding the Award containing a transaction with
         # a matching fain and submission. If this fails, try submission and uri
         if row['fain'] and len(row['fain']) > 0:
-            award = Award.objects.filter(transaction__submission=submission_attributes,
+            award = Award.objects.filter(awarding_agency=agency,
```
bsweger: Mulling this over some more, I realize we use this same award lookup logic a few times in the code base. Perhaps not a thread we want to pull right now, but perhaps it would make sense to split `Award.get_or_create_summary_award` into two: `get_award` that does the lookup and can be used as needed and `create_award`.

usaspending_api/etl/subaward_etl.py
```diff
@@ -135,17 +147,25 @@ def load_subawards(submission_attributes, db_cursor):
...
+                                         transaction__submission=submission_attributes,
                                          transaction__assistance_data__isnull=False,
                                          transaction__assistance_data__fain=row['fain']).distinct().first()
 
         # Couldn't find a match on FAIN, try URI if it exists
         if not award and row['uri'] and len(row['uri']) > 0:
-            award = Award.objects.filter(transaction__submission=submission_attributes,
+            award = Award.objects.filter(awarding_agency=agency,
```
bsweger: Same question as above...would it be useful/more readable to query the `uri` and `fain` on the award itself instead of digging into `transaction__assistance_data`?

usaspending_api/references/models.py
```diff
@@ -230,7 +230,7 @@ class Location(DataSourceTrackedModel):
     zip_4a = models.CharField(max_length=10, blank=True, null=True)
     congressional_code = models.CharField(max_length=2, blank=True, null=True, verbose_name="Congressional District Code")
     performance_code = models.CharField(max_length=9, blank=True, null=True, verbose_name="Primary Place Of Performance Location Code")
-    zip_last4 = models.CharField(max_length=4, blank=True, null=True)
+    zip_last4 = models.CharField(max_length=10, blank=True, null=True)
```
bsweger: Booyeah, no zip code parsing!

usaspending_api/etl/management/commands/load_subawards.py
afrasier: I mainly use .first() to get None back, rather than throwing an error (And thus avoid using try/except as flow control)

A unique key on broker_submission_id is definitely warranted.

usaspending_api/etl/management/commands/load_subawards.py
afrasier: Not particularly, any exception in the loading will be needed to investigate manually, so catching all exceptions, logging it, and moving to the next submission is probably the best course of action when loading a large set of submissions' subawards.

usaspending_api/etl/management/commands/load_submission.py
afrasier: The reason I wrapped this portion in it's own block is because it's a totally separate process from the rest of the data load - if it errors out, there's no reason not to continue on and update the award transactions/etc. using the other helpers. 

We can also just remove it from load_submissions all together, and then just call load_subawards separately.

usaspending_api/etl/subaward_etl.py
afrasier: Yes, I chatted with the broker folks - duplicates are largely due to how the join is performed, so this just removes the handful of cases where it duplicates rows.

usaspending_api/etl/subaward_etl.py
afrasier: The identifying data from broker is from the transactions, as well, awards aren't linked to submissions themselves, so this preserves the most information during the lookup. That is, since row['piid'] is coming from the broker transaction data, I look up by transaction to ensure it's always 1:1.

usaspending_api/etl/subaward_etl.py
afrasier: Sure, update date?

usaspending_api/etl/subaward_etl.py
afrasier: Yeah, that's probably a good idea.

usaspending_api/etl/management/commands/load_subawards.py
bsweger: Gotcha--that makes sense. I might go rogue and sneak in a PR for that unique index.

usaspending_api/etl/subaward_etl.py
bsweger: Ok, thanks for the context!

usaspending_api/etl/management/commands/load_submission.py
bsweger: Yeah, it feels strange to treat the subaward load process as "it's okay if this fails" when every other part of a submission load is critical enough that we can't consider the submission done if it fails. That said, I think this is a question for the larger group...will ping the channel.

usaspending_api/etl/subaward_etl.py
bsweger: Maybe `id` in this case? Though whatever you think makes the most sense is fine. You know my motto--if the off-chance that we're wrong, let's be wrong consistently.

bsweger:
@afrasier Thanks for the additional info. I think this is good to go once the award sort is in. I think we can leave the subaward exception handling as is until we answer the question of how we want to treat this process in relation to the rest of the submission load.

PR #298, opened 2017-04-21
DS-1139 - no agency when toptier and subtier are both None

catherinedevlin:
Go ahead and call .get_by_subtier with the `None` subtier code, just return `None` instead of the first agency that happens to have a null subtier code


usaspending_api/references/tests/test_agency.py
```diff
@@ -83,6 +83,9 @@ def test_get_by_subtier():
     assert Agency.get_by_subtier('abc') == agency1
     # if there's no match, we should get none
     assert Agency.get_by_subtier('nope') is None
+    # if called with an empty argument, we should get None
+    assert Agency.get_by_subtier('') is None
```
bsweger: Nice--thanks for checking both `''` and `None`.

PR #299, opened 2017-04-24
Executive Compensation ETL

afrasier:
This PR adds an ETL command for Executive Compensation data, pulled from the Broker.

* Adds command `load_executive_compensation`
  * `load_executive_compensation -d DUNS DUNS DUNS . . .` will load only information for the specified DUNSes
  * `load_executive_compensation -a` loads executive compensation data for every DUNS currently in the datastore
* Adds tests for `load_executive_compensation`

This command is not called anywhere, and must be explicitly called. (i.e. this does not hook in to our regular ETL process anywhere)

**! NOTE !**: The broker currently only has the executive compensation table in dev and staging; thus, we won't be able to run this pointing at the production broker until their next deploy. This has been tested against their dev and staging databases and it all works.


usaspending_api/etl/management/commands/load_executive_compensation.py
```diff
@@ -0,0 +1,62 @@
...
+        parser.add_argument(
+            '-d',
+            '--duns',
+            dest="duns",
+            nargs='+',
+            type=int,
+            help="DUNS to load subawards for"
+        )
+
+        parser.add_argument(
```
catherinedevlin: Does this option actually end up being implemented?  It looks like it's what happens if you omit --duns, but it looks like --all doesn't actually do anything.  And I think calling with `--all --duns 123` would do the same thing, which is confusing.

usaspending_api/etl/tests/test_executive_compensation_etl.py
```diff
@@ -0,0 +1,31 @@
...
+import json
+
+from django.core.management import call_command
+from usaspending_api.references.models import LegalEntityOfficers
+
+from model_mommy import mommy
+
+
+@pytest.fixture
+def test_exec_etl_fixture():
```
catherinedevlin: I always get confused about the role of etl_test_data.json, and it's especially weird in a case like that depends partially on data generated here, but partially on data from the "virtual broker database" represented by etl_test_data.json.  I feel like a comment to that effect would be helpful (even if that comment appeared in multiple test files.)

usaspending_api/etl/executive_compensation_etl.py
```diff
@@ -0,0 +1,56 @@
...
+from usaspending_api.references.models import LegalEntity, LegalEntityOfficers
+from usaspending_api.etl.broker_etl_helpers import dictfetchall, PhonyCursor
+
+logger = logging.getLogger("console")
+
+# Broker SQL query
+FILE_E_QUERY = """
+SELECT *
+FROM executive_compensation e
+INNER JOIN (
```
afrasier: I believe the group by in the join performs a similar function as a DISTINCT ON but I can switch it anyway to be more explicit.

usaspending_api/etl/management/commands/load_executive_compensation.py
afrasier: Ah yes, this is just a leftover from a bit of copy and pasting, I will fix that and clarify.

usaspending_api/etl/tests/test_executive_compensation_etl.py
afrasier: Will do.

usaspending_api/etl/management/commands/load_executive_compensation.py
```diff
@@ -0,0 +1,67 @@
...
+                return
+        else:
+            db_cursor = PhonyCursor()
+
+        # Require users to explicitly call -a to use the "All" case so that
+        # it is not accidentally triggered by omitting an option
+        if options["duns"] is None and options["update_all"]:
+            load_executive_compensation(db_cursor, None)
+        elif options["duns"] is not None:
+            load_executive_compensation(db_cursor, options["duns"])
```
catherinedevlin: Hmm... now will it silently do nothing if neither `--all` nor `--duns` is passed?

usaspending_api/etl/management/commands/load_executive_compensation.py
```diff
@@ -0,0 +1,69 @@
...
+        else:
+            logger.info("Please specify either a list of DUNS (using the --duns) flag, or set to update all mode using the (--all) flag.")
```
catherinedevlin: Great!

PR #300, opened 2017-04-24
Fix D file loads when funding agency subtier code is not provided

bsweger:
This is a hotfix for the scenario where we don't have funding subtier agency codes on incoming D records. Last week we merged a fix to ensure that lookups in that scenario don't return the wrong Agency object. This fixes a downstream error caused when said funding agency object is `null`.

Additionally, updated the ETL test data to reflect a few missing agency data scenarios.


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -660,14 +660,20 @@ def load_file_d1(submission_attributes, procurement_data, db_cursor):
         pop_location, created = get_or_create_location(
             place_of_performance_field_map, row, place_of_performance_value_map)
 
-        # If awarding/funding toptier agency code (aka CGAC) is not supplied on the D1 record,
-        # use the sub tier code to look it up
-        if row['awarding_agency_code'] is None:
+        # If awarding toptier agency code (aka CGAC) is not supplied on the D1 record,
+        # use the sub tier code to look it up. This code assumes that all incoming
+        # records will supply an awarding subtier agency code
+        if row['awarding_agency_code'] is None or len(row['awarding_agency_code'].strip()) < 1:
```
catherinedevlin: Hmm, for a future improvement, it would be nice to strip all fields at once upon load from the file instead of having to remember to do so at various points during processing...

bsweger: Yes, totally agree that if we need to `.strip()`, we should strip all the things in a single place. I believe the broker actually does `strip()` everything before inserting, so this was me being overly-paranoid.

Following your example from earlier today, I'm gonna start an "idea" ticket for submission loader 2.0.

PR #301, opened 2017-04-24
Tests for rm_submission of dependent TasProgramActivityObjectClassQua…

catherinedevlin:
…rterly

Tiny PR toward verifying fix of DS-1110.


afrasier:
Did you mean to add that model to the SUBMISSION_MODELS which it checks to validate are 0 counts after deletion?

catherinedevlin:
Oh, of course!  Thanks for catching that, @afrasier/

bsweger:
Good call, @catherinedevlin! Do you think we should also add `AppropriationAccountBalancesQuarterly` to the list of things to check?

catherinedevlin:
A good suggestion, @bsweger  - incorporated.

PR #303, opened 2017-04-24
Fixes a problem where recipient name could be null

afrasier:


usaspending_api/etl/subaward_etl.py
```diff
@@ -180,9 +180,16 @@ def load_subawards(submission_attributes, db_cursor):
         recipient = LegalEntity.objects.filter(recipient_unique_id=row['duns']).first()
 
         if not recipient and row['duns']:
+            recipient_name = row['awardee_name']
+            if recipient_name is None:
```
bsweger: Do we have to worry about an incoming recipient name of `""` in addition to checking for `None`?

usaspending_api/etl/subaward_etl.py
```diff
@@ -180,9 +180,16 @@ def load_subawards(submission_attributes, db_cursor):
...
+                recipient_name = row['awardee_or_recipient_legal']
+            if recipient_name is None:
+                logger.warn("Subaward number {} has no recipient name; skipping...".format(row['subaward_num']))
```
bsweger: Should we be skipping the row entirely or just loading it with a blank recipient name? `recipient_name` on the `LegalEntity` model is kinda strange...it allows `blank=True` but doesn't allow NULLS. 

You've probably already discussed with @kaitlin  the ramifications of skipping a subaward record, but wanted to double check, since we're crossing a line into territory that would affect dollar amounts.

kaitlin: I think if both name fields are null we should probably skip, unless there's a DUNS. Otherwise how useful is the record?

bsweger: @kaitlin Thanks for weighing in...if I'm reading correctly, I think there's a DUNS here (see line 182). @afrasier am I interpreting right?

usaspending_api/etl/subaward_etl.py
afrasier: I'm not sure, might be something worth asking @kaitlin

kaitlin: I think we should cover the empty string case, JIC. The data is a bit unpredictable.

usaspending_api/etl/subaward_etl.py
afrasier: Yes, that is correct. Currently, it will pick up a record where a DUNS matches, or, if there is a supplied duns and no match, it will create the recipient. Otherwise it leaves the recipient NULL on the subaward.

usaspending_api/etl/subaward_etl.py
afrasier: So as long as there is a DUNS accept empty strings?

kaitlin: yeah, because perhaps the recipient record we're linking to has the name. Or we could add something later that goes through and adds names to all recipients with a DUNS but with an empty name.

afrasier: Ok, can do.

usaspending_api/etl/subaward_etl.py
bsweger: So if there's a DUNS but we can't find a recipient, it looks like we're skipping the record. Is that what we want?

kaitlin: IMO we should create a recipient record with the DUNS, even if there's no name.

afrasier: Got it.

usaspending_api/etl/subaward_etl.py
```diff
@@ -180,9 +180,15 @@ def load_subawards(submission_attributes, db_cursor):
         recipient = LegalEntity.objects.filter(recipient_unique_id=row['duns']).first()
 
         if not recipient and row['duns']:
+            recipient_name = row['awardee_name']
+            if recipient_name is None:
+                recipient_name = row['awardee_or_recipient_legal']
+            if recipient_name is None:
```
bsweger: Do we still need a check for `recipient_name` == ``?

afrasier: No? If it is "" we allow it to be stored as the blank string as long as the duns exist, or am I misunderstanding here?

bsweger: Heh, no, I'm misreading. Carry on!

PR #304, opened 2017-04-24
Recipient duns fixes

afrasier:
This fixes the issue brought up in DS-1143 where null DUNS's were causing some serious linkage issues.

* Adds `get_or_create_by_duns(duns)` to the `LegalEntity` Objects
  * This method returns `LegalEntity, Boolean` similar to other django get_or_create patterns.
  * If the specified DUNS is null, it will always create a new LegalEntity
  * If the specified DUNS exists, it will return that LegalEntity
  * If the specified DUNS does not exist, it will create a LegalEntity with that DUNS
* Update all usages of `.get()`, `.filter().first()` and other accessor methods for `LegalEntity` throughout the codebase where DUNS were used to `get_or_create_by_duns`
* Update recipient autocomplete to exclude any null DUNS, per acceptance criteria
* Update file E ETL to not check for null DUNS to prevent erroneous executive compensation linkages
* Added tests to verify the behavior of `get_or_create_by_duns`


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -647,14 +647,12 @@ def load_file_d1(submission_attributes, procurement_data, db_cursor):
         legal_entity_location, created = get_or_create_location(legal_entity_location_field_map, row, legal_entity_location_value_map)
 
         # Create the legal entity if it doesn't exist
-        try:
-            legal_entity = LegalEntity.objects.get(recipient_unique_id=row['awardee_or_recipient_uniqu'])
-        except ObjectDoesNotExist:
+        legal_entity, created = LegalEntity.get_or_create_by_duns(duns=row['awardee_or_recipient_uniqu'])
+        if created:
```
bsweger: Maybe not something we need to address right now, but do you think we'll ever have to update attributes on an existing legal entity based on the incoming data? I suspect not, since our incoming records aren't the canonical DUNS source, but seems worth asking.

usaspending_api/etl/management/commands/load_usaspending_assistance.py
```diff
@@ -168,16 +168,17 @@ def get_or_create_recipient(self, row):
...
         recipient_dict.update(self.recipient_flags_by_type(recipient_type))
 
-        le = LegalEntity.objects.filter(
-            recipient_unique_id=row['duns_no']).first()
-        if not le:
-            le = LegalEntity.objects.create(**recipient_dict)
+        le, created = LegalEntity.get_or_create_by_duns(duns=row['duns_no'])
+        if created:
+            # Update from our recipient dictionary
+            for attr, value in recipient_dict.items():
```
bsweger: I'm not familiar with the different types of recipient info we get from a historic data submission vs a broker submission. Am I interpreting correctly that here we're adding additional attributes to a new LegalEntity (e.g., name, recipient flags by type) that we don't add when creating a new LegalEntity via broke data?

usaspending_api/etl/subaward_etl.py
```diff
@@ -88,14 +88,13 @@ def load_subawards(submission_attributes, db_cursor):
...
-        recipient = LegalEntity.objects.filter(recipient_unique_id=row['duns']).first()
+        recipient, created = LegalEntity.get_or_create_by_duns(duns=row['duns'])
 
-        if not recipient and row['duns']:
-            recipient, created = LegalEntity.objects.get_or_create(recipient_unique_id=row['duns'],
-                                                                   parent_recipient_unique_id=row['parent_duns'],
-                                                                   recipient_name=row["company_name"],
-                                                                   location=get_or_create_location(row, location_d1_recipient_mapper)
-                                                                   )
+        if created:
```
bsweger: Similar question here. Is this a third way of filling in new LegalEntity data depending on its source? It seems feasible that it would be handled different for a subaward, but I don't know enough to be sure.

usaspending_api/references/tests/test_legal_entity.py
```diff
@@ -0,0 +1,17 @@
...
+
+from usaspending_api.references.models import LegalEntity
+
+
+@pytest.mark.django_db
+def test_get_or_create_by_duns():
+    # Null duns should always return True for created
+    assert LegalEntity.get_or_create_by_duns(duns=None)[1]  # Assert created is True
+    assert LegalEntity.get_or_create_by_duns(duns=None)[1]  # Assert created is True
+    assert LegalEntity.get_or_create_by_duns(duns=None)[1]  # Assert created is True
```
bsweger: Great test--what do you think about adding a line for the `duns=""` case that you handle in `get_or_create_by_duns`?

usaspending_api/references/tests/test_recipients_autocomplete.py
```diff
@@ -47,3 +51,12 @@ def test_bad_recipients_autocomplete_request(client):
         content_type='application/json',
         data=json.dumps({}))
     assert resp.status_code == status.HTTP_400_BAD_REQUEST
+
+
+@pytest.mark.django_db
+def test_recipient_autocomplete_null_duns_exclusion(client, recipients_data):
```
bsweger: Nice--thanks for adding this test!

usaspending_api/etl/management/commands/load_submission.py
afrasier: I think at this point no. As much as I hate to suggest it we may just want to connect directly to SAMS and pull this information for ourselves rather than relying on it filtering in via submissions to the broker... :\

usaspending_api/etl/subaward_etl.py
afrasier: We have less information coming in for the subawards, hence why it is so sparse. We can try and pull more information by joining other tables in the future if we need to.

usaspending_api/etl/management/commands/load_usaspending_assistance.py
afrasier: This performs the same function as the other locations. It just iterates over the dictionary and updates the instance with the dictionary's values. It needed to be done here because we were can no longer call `.create(**dict)`

usaspending_api/references/tests/test_legal_entity.py
afrasier: Ok, can do.

usaspending_api/etl/management/commands/load_submission.py
bsweger: Agreed that this should stand for now. Thanks!

usaspending_api/etl/management/commands/load_usaspending_assistance.py
bsweger: Gotcha...I initially flagged because the only I saw being explicitly set on new `LegalEntity` objects created by the submission loader is the location (_i.e._ not the name): https://github.com/fedspendingtransparency/usaspending-api/blob/2ad3d7f49baab760a44df67481148a1e0f1312d0/usaspending_api/etl/management/commands/load_submission.py#L652.

Will the name be handled on the submission side because the incoming field name matches the model name?

afrasier: Correct!

PR #305, opened 2017-04-25
increased length of number_of_actions on transaction contract model

kaitlin:
After running some test submissions I got an error on the `number_of_actions` field. This PR increases the allowable length from 6 to 10 to accommodate exceedingly large values that are showing up in the data.


bsweger:
Checked in with Kaitlin...am going to push an update that makes this a text field and gives the same treatment to some of the other character fields on the D file models.

bsweger:
Closing this in favor of #311.

PR #306, opened 2017-04-25
[WIP] Add initial support for docker compose.

toolness:
I wanted to familiarize myself a bit with this project and noticed that it didn't have a Docker Compose setup, so I thought contributing one would be a nice way to get started!

This adds `docker_django_management.py`, which is an entrypoint we use in CALC (and a few other projects) to make Docker play nicely with Django for development purposes. It's already been code reviewed by other 18F'ers so you don't need to review it if you don't want to (we vendor it so devs don't need to do virtualenv stuff on their Docker host).

This also adds a `docker-compose.yml` that leverages your existing `Dockerfile`.

## Quick start

Until I add proper docs to the README, here's the basics.

The `docker-compose.yml` sets up the database for you, so you don't need to configure anything.

For the most part you can run everything you'd normally run to set things up, but prefix it all with `docker-compose run app`.  So e.g.:

* to run database migrations, run `docker-compose run app python manage.py migrate`.
* to run tests, run `docker-compose run app py.test`.

The one major exception to this rule is starting the development server: you want to run `docker-compose up` for that, and then visit port 8000 on your Docker host (which probably means `http://localhost:8000`).

Unlike the instructions in your existing `Dockerfile`, the Docker Compose setup mounts your working directory directly into the container, so you can make changes to files and see their effects without needing to rebuild the container: the development server will immediately notice when files change and restart itself, just as with local development.

## Notes

* The tests appear to "hang" at certain places but eventually resume--I'm not sure if this is something that happens outside of Docker. All the tests take about 170 seconds to run on my system, and they complete successfully (3 tests are skipped), but much of the time is taken up by this "hanging".

* Running `manage.py load_reference_data` works, but logs a _lot_ of logging errors--from what I can tell, it seems like multiple processes are trying to write to the log file at the same time or something. Does this happen on non-Docker setups too?

## To do

- [ ] Add Docker instructions to the README.  This will basically be a riff off of [CALC's Docker guide](https://calc.gsa.gov/docs/docker.html).

- [ ] Consider moving the requirements into a virtualenv stored in a Docker volume; this will make it a lot easier to iterate on dependencies, so that they don't all need to be re-fetched and re-installed whenever you make a minor change to `requirements.txt`--see https://github.com/18F/calc/issues/1230 for more context. This could also be done later in a separate PR.

- [ ] Consider using the official [memcached Docker image](https://hub.docker.com/_/memcached/) for better dev/prod parity. Can be done in a later PR, or ignored if this dev/prod distinction isn't likely to be an issue.

Anyhow, let me know if this kind of thing seems useful to you. No worries if it's not, as it was a good way for me to familiarize myself with some of the code regardless.


manage.py
```diff
@@ -3,9 +3,10 @@
 import sys
 
 if __name__ == "__main__":
+    os.environ.setdefault("DDM_CONTAINER_NAME", "app")
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "usaspending_api.settings")
     try:
-        from django.core.management import execute_from_command_line
+        from docker_django_management import execute_from_command_line
```
toolness: This sets up a convenient shortcut that allows `manage.py` to be used directly from the Docker Host without having to prefix it with `docker-compose run app`.  However, the shortcut relies on `import django` raising an `ImportError` on the Docker host, which is usually but not always the case, which is why we still document everything using the long-form notation.

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+      - PYTHONUNBUFFERED=yup
+  db:
+    image: postgres:9.5.4
+    volumes:
+      - pgdata:/var/lib/postgresql/data/
+    environment:
+      - POSTGRES_DB=usaspending
+      - POSTGRES_USER=usaspender
+volumes:
+  pgdata:
```
toolness: The official postgres container already uses an unnamed [data volume](https://docs.docker.com/engine/tutorials/dockervolumes/) for its storage, but here we give it a name so that it's easy to do things with. For instance, `docker volume rm usaspendingapi_pgdata` will delete it, which can be useful if you ever want to blow away your development db.

(All Docker Compose containers and volumes are prefixed with a slugified form of your project's root directory name, so if you have this project checked out to a directory called `usaspending-api`, all of Docker Compose's stuff will be prefixed with `usaspendingapi_` when examining it through the `docker` CLI tool.)

catherinedevlin:
This is so nice, Atul!  Thank you!

One thing to add would be directions for how to set environment variables within the containers without putting them in docker-compose.yml.  I can see you're setting some of the crucial ones in docker-compose.yml, but there are others like DATA_BROKER_DATABASE_URL that the submission load relies on, and that we wouldn't want to check into the repo anyway (it includes credentials to the actual data broker DB).

toolness:
Thanks Catherine!  That's a great point about the environment variables.  For CALC, we use Python's [`dotenv` package](https://github.com/theskumar/python-dotenv#readme) to read environment variables from a `.env` file during development, as it makes things easy for both docker and non-docker development setups. Do you think I should add support for that to this PR?

I think the alternative is to have the developer make a `docker-compose.override.yml` file that explicitly passes environment variables to the container. This requires a bit more familiarity with the [compose file format](https://docs.docker.com/compose/compose-file/) but doesn't involve adding any more dependencies to the project.

Let me know which one you prefer and I'll add it to the PR!

catherinedevlin:
Using an .env sounds good - I know that Arman's already using one of his own, and I'm using one adapted from him; I don't know if it's an official project standard, but it's in use.

docker-compose.yml
willkjackson: Atul,
Thanks for this, I've added some comments and will continue tinkering with containerized psql.

PR #314, opened 2017-04-25
updating django-memcache dependency

willkjackson:


afrasier:
We'll need to make sure to keep an eye on the main repo for when it is properly fixed, and switch back to it.

PR #317, opened 2017-04-26
Update transaction lookup to separate fain and uri

bsweger:
This simplifies get_award_financial_transaction to try the
lookup by fain and then by uri (and not together as we were
previously doing). @catherinedevlin had suggested this approach
as a better way in a prior code review, and she was right!


bsweger:
Thanks for the review--will hold off on merging until post sprint review!

PR #320, opened 2017-04-27
Add data formatting note about changing the signs for display

bsweger:
Adding Kristine's notes about changing the signs for display.


bsweger:
FYI--updated the base branch to `dev-stable`. We have a few hotfixes being promoted to stable, and this updated documentation seems worth including for the beta release.

PR #323, opened 2017-05-01
Hotfix File B object class duplicates

bsweger:
This hotfix adds a check to see if a submission's file B is in the edge case where agencies use duplicate 4 digit object classes with different D/R flags to report deobligations.

See the detailed commit messages and code comments for more information. We need to hotfix this and promote to staging to ensure that the following certified submissions can process:

* 3570
* 3914

The `load_submission` test has been updated to check this edge case. That said, we're checking against our own canned test data, so be aware of that. This needs careful eyes, and once it's merged to `dev-stable` would recommend doing local loads of the submissions above.


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -495,6 +494,129 @@ def load_file_a(submission_attributes, appropriation_data, db_cursor):
...
+
+    Args:
+        submission_attrbitues: submission object currently being loaded
+        db_cursor: db connection info
+    """
+    submission_id = submission_attributes.broker_submission_id
+
+    # does this file B have the dupe object class edge case?
+    check_dupe_oc = (
+        'SELECT count(*) '
```
catherinedevlin: A possible performance improvement would be to replace this count where we don't actually care about the number with an exists: https://blog.jooq.org/2016/09/14/avoid-using-count-in-sql-when-you-could-use-exists/

The first step would be to determine if this is a useful improvement or not - if the query is not much of a choke point and/or will hopefully be cleansed with fire soon, it's not worth worrying about, and in any case shouldn't hold up this commit; it could be noted for future investigation.

bsweger: Ah, great point. So good to have your eyes on this SQL! I've been running the `count(*)` version against the broker db frequently/obsessively, making sure that no new submissions fall into this object class edge case. It runs pretty quickly against the entire body of certified submissions, so I don't expect it to be much of a chokepoint when running for an individual submission.

The schema team agrees that agencies should not submit data this way, so let's go with the "cleansed with fire soon" line of thinking.

bsweger:
Thanks, @catherinedevlin! `I like how well this code encapsulates the ugliness` goes down as one of my favorite programming feedback of all time.

PR #326, opened 2017-05-02
Guide in API

catherinedevlin:
This exposes the Guide of definitions in our API.  It loads them from a copy of the USAspendingGuide.xlsx that can be downloaded at https://staging.spendingdata.us/graphics/USAspendingGuide.xlsx.


usaspending_api/references/tests/test_guide.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+
+from django.core.management import call_command
+
+@pytest.fixture()
```
afrasier: session='module' will make the fixture only get created once, then reused. :)

usaspending_api/references/views.py
```diff
@@ -94,3 +94,19 @@ def get_queryset(self):
         queryset = self.serializer_class.setup_eager_loading(queryset)
         filtered_queryset = self.filter_records(self.request, queryset=queryset)
         return filtered_queryset
+
+
+class GuideViewSet(FilterQuerysetMixin, DetailViewSet):
```
afrasier: I think front end will also need an Autocomplete endpoint for these, right?

bsweger: The guide search function (which will need the autocomplete) is a story in the next sprint.

afrasier: Ah, right! carry on

usaspending_api/references/tests/test_guide.py
catherinedevlin: That's what I tried, but my `guide_data` fixture needs to use `db`, and the `db` fixture doesn't work with a module-scoped fixture:

`ScopeMismatch: You tried to access the 'function' scoped fixture 'db' with a 'module' scoped request object, involved factories`

... there might be some way around that (I'm trying to figure it out from http://pytest-django.readthedocs.io/en/latest/database.html - do you know one?)

afrasier: Hmm, I can't think of one off the top of my head. Ah well :\

usaspending_api/references/migrations/0003_add_definition.py
```diff
@@ -0,0 +1,43 @@
...
+                ('title', models.TextField()),
+                ('url', models.TextField()),
+                ('definition', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='references', to='references.Definition')),
+            ],
+            options={
+                'managed': True,
+                'db_table': 'definition_resource',
+            },
+        ),
+        migrations.AlterUniqueTogether(
```
bsweger: This is ponderous! Want to make sure that the recent migration smooshing didn't accidentally blow away an index...is this unique index something you removed specifically?

usaspending_api/references/models.py
```diff
@@ -620,3 +621,27 @@ def get_default_fields(path=None):
...
+    def save(self, *arg, **kwarg):
+        self.slug = slugify(self.term)
+        return super(Definition, self).save(*arg, **kwarg)
+
+
+class DefinitionResource(models.Model):
+    id = models.AutoField(primary_key=True)
+    title = models.TextField()
+    url = models.TextField(null=False, blank=False)
+    definition = models.ForeignKey(Definition, related_name='references')
```
bsweger: Will setting the related_name to `references` cause any issues or confusion because the application name is also `references`? It's probably fine, but just wanted to make sure!

usaspending_api/references/tests/test_guide.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Sorry to squash these together, but I don't want to load the guide data
+    # multiple times
+
+    resp = client.get('/api/v1/references/guide/')
+    assert resp.status_code == 200
+    assert len(resp.data["results"]) > 70
+
+    resp = client.get(
+        'http://localhost:8000/api/v1/references/guide/agency-identifier/')
```
bsweger: Is the localhost reference here an artifact of local testing that should come out before merging?

usaspending_api/references/tests/test_guide.py
```diff
@@ -0,0 +1,49 @@
...
+    assert resp.status_code == 200
+    assert resp.data['term'] == 'Agency Identifier'
+
+    resp = client.get('/api/v1/references/guide/frumious-bandersnatch/')
```
bsweger: :joy:

usaspending_api/references/tests/test_guide.py
```diff
@@ -0,0 +1,49 @@
...
+    for itm in resp.data['results']:
+        assert itm['data_act_term'] == 'Budget Authority Appropriated'
+
+    resp = client.get('/api/v1/references/guide/?plain__contains=Congress')
+    assert resp.status_code == 200
+    assert len(resp.data['results']) > 0
+    for itm in resp.data['results']:
+        assert 'congress' in itm['plain'].lower()
+
+    # case-insensitivity fail?
```
bsweger: Huh, good question. Assuming we're good to go without this test case, since the lookup part of the glossary is out of scope for the current sprint. Is that right?

usaspending_api/references/urls.py
```diff
@@ -1,11 +1,16 @@
-from django.conf.urls import url
+from django.conf.urls import url, include
 from usaspending_api.references import views
+from rest_framework.routers import DefaultRouter
+
+guide_router = DefaultRouter()
```
bsweger: Bless you for getting the router working! Nice!

usaspending_api/references/management/commands/load_guide.py
```diff
@@ -0,0 +1,72 @@
...
+            help='the path to the Excel spreadsheet to load',
+            default=default)
+        parser.add_argument(
+            '-a',
+            '--append',
+            help='Append to existing guide',
+            action='store_true',
+            default=False)
+
+    def handle(self, *args, **options):
```
bsweger: I struggle with this question, so am curious to hear your thoughts--do you think we need to unit test infrequently-run loader code like this?

usaspending_api/references/migrations/0003_add_definition.py
catherinedevlin: Whoah!  I did not notice that!  No, and I don't even know how that happened.

Let me check on that...

usaspending_api/references/tests/test_guide.py
catherinedevlin: Right,  I just noticed that our API docs for __contains says it's case-insensitive, but I don't see that working anywhere (I should confirm that and make a ticket)

usaspending_api/references/tests/test_guide.py
catherinedevlin: Oh, it so is!  Thank you!

usaspending_api/references/models.py
catherinedevlin: I... don't think it will, but 'resources' would almost certainly be a better name anyway, wouldn't it?  I might have meant to type that in the first place...

usaspending_api/references/management/commands/load_guide.py
catherinedevlin: Well... "Yes, test it" is never the wrong answer, right?

My earlier excuse was that I was rushing to get something for Kevin to develop against, but that doesn't hold up anymore.  I should add at least some testing.

usaspending_api/references/tests/test_guide.py
catherinedevlin: @bsweger Captured this in a ticket: https://federal-spending-transparency.atlassian.net/browse/DS-1212

bsweger:
@catherinedevlin One thing I forgot to note in the PR...we should follow this up with an update the API docs and make sure this new endpoint is listed.

PR #328, opened 2017-05-03
Reset database migrations

bsweger:
To reset Django migrations:
1. merge this PR
2. run the new `reset_migrations` management command

Have tested locally, but am keeping this in WIP status to coordinate with @catherinedevlin's PR (which contains a migration) and to test the build on sandbox.

This changeset removes the existing migration .py files for our Django apps.
It adds a new series of 0001_initial migrations that reflect the current state
of the database. Lastly, it adds a management command that removes the existing
migration record from the django_migrations table and runs fake migrations for
each app. Once the management command (reset_migrations) is run, we'll be back
at ground zero for migrations.


bsweger:
Took this out of WIP status, because I think it needs to merge before other PRs with migrations.

PR #329, opened 2017-05-03
Agency autocomplete sorting

russellbodine:
This pull request is in response to story DS-1182

Added toptier_flag to agency model.

Added ability for autocomplete api calls to sort.

Added tests for the new functionality.

Updated the loader to account for the new toptier_flag. 

**After the merge, we need to run loadagencies to set the new toptier_flag**


russellbodine:
Sample postman request!

{"fields":["subtier_agency__name", "toptier_agency__name"],"value":"Department","mode":"contains","matched_objects":true,"limit":100, "order": ["-toptier_flag"]}

PR #338, opened 2017-05-09
New notebook with queries useful for DBA checkups

catherinedevlin:
This doesn't actually touch any existing code, it's just a notebook of queries to check for common problems with DDL configuration, etc.


nmonga91:
Moving this to a different location outside of the API repo

PR #339, opened 2017-05-10
RequestCatalog Deployment Changes

afrasier:
* RequestCatalog only reads from the source database
* RequestCatalog doesn't update it's access time every time


willkjackson:
Added comments and check for migrate, looks good to test in staging environment.

PR #344, opened 2017-05-11
Dev stable --> Dev

bsweger:
Bring the request catalog hotfixes back down to dev.


bsweger:
Will fix the code climate issue in a subsequent PR to ensure we don't get things out of sync.

Issue #342, opened 2017-05-10
"fields" option does not appear to support nested field names using double underscores


pschmied:
Hello, I'm looking at an example on https://api.usaspending.gov/docs/using-the-api that seems to suggest that double underscores should be usable with the "fields" option in a POST request:
```
{
    fields": ["toptier_agency__name", "subtier_agency__name"],
    "value": "DEFENSE",
    "mode": "contains",
    "limit": 100,
    "matched_objects": true
  "filters": []
}
```

Unfortunately this approach does not seem to work in my testing. Using this bit of json (`examplequery.json`)
```
{
    "limit": 1,
    "fields": ["recipient__location"],
    "filters": [
        {
            "field": "place_of_performance__state_code",
            "operation": "equals",
            "value": "CT"
        }
    ]
}
```

...and this curl invocation:
```
curl -H "Content-Type: application/json" -X POST --data-binary "@examplequery.json" https://api.usaspending.gov/api/v1/awards/ |jq  .|more
```

The API endpoint returns none of the recipient fields:

```
{
  "page_metadata": {
    "page": 1,
    "has_next_page": true,
    "has_previous_page": false,
    "next": "https://api.usaspending.gov/api/v1/awards/?limit=1&page=2&req=b0275bf10b3",
    "current": "https://api.usaspending.gov/api/v1/awards/?limit=1&page=1&req=b0275bf10b3",
    "previous": null
  },
  "req": "b0275bf10b3",
  "results": [
    {
      "type_description": "Definitive Contract",
      "description": "121NAG17S0010 INTL AID GRAIN"
    }
  ]
}
```

If, however, I do not attempt to specify any of the nested field names (i.e. ` "fields": ["recipient"]`, I can see that `recipient__location` is valid:

```
{
  "page_metadata": {
    "page": 1,
    "has_next_page": true,
    "has_previous_page": false,
    "next": "https://api.usaspending.gov/api/v1/awards/?limit=1&page=2&req=f2458c12573",
    "current": "https://api.usaspending.gov/api/v1/awards/?limit=1&page=1&req=f2458c12573",
    "previous": null
  },
  "req": "f2458c12573",
  "results": [
    {
      "type_description": "Definitive Contract",
      "description": "121NAG17S0010 INTL AID GRAIN",
      "recipient": {
        "legal_entity_id": 505323,
        "parent_recipient_unique_id": null,
[...]

PR #345, opened 2017-05-11
Nested field limiting

afrasier:
This resolves issue #342 which also affected @ebdabbs front-end work

* Move default fields from models to serializers
* Limitable serializer meta objects now support `default_fields` and
`default_exclude`, which take an array of strings and provide the
default exclude and include fields. (The `fields` variable should
remain '__all__' to allow access to all variables by API users)
* Nested serializers can now include `default_fields` and
`default_exclude` in their kwargs
* Limitable serializer now properly passes down include/exclude fields
to nested objects using Django double-underscore traversal
* If a child object's field is requested, but the child itself is not
requested, it will be automatically added (otherwise it would not
serialize)
* The verbose flag now _only_ applies to the top-level request object.
That is, if you request verbose on `/awards/` only the award will be
verbose, the nested objects will still use their defaults. This
super-full-detail view is still available by `retrieve`ing an object
using the `/awards/pk` endpoint (or any other retrieve route)
* Added tests to support nested object include and exclude fields

### DIFF
--- a/usaspending_api/accounts/models.py
+++ b/usaspending_api/accounts/models.py
@@ -89,49 +89,6 @@ class TreasuryAppropriationAccount(DataSourceTrackedModel):
         concatenated_tas = ATA + AID + TYPECODE + POAPHRASE + ACCTPHRASE
         return concatenated_tas
 
-    @staticmethod
-    def get_default_fields(path=None):
-        if "awards" in path:
-            return [
-                "treasury_account_identifier",
-                "tas_rendering_label",
-                "account_title",
-                "reporting_agency_id",
-                "reporting_agency_name",
-                "federal_account"
-            ]
-
-        return [
-            "treasury_account_identifier",
-            "tas_rendering_label",
-            "federal_account",
-            "allocation_transfer_agency_id",
-            "agency_id",
-            "beginning_period_of_availability",
-            "ending_period_of_availability",
-            "availability_type_code",
-            "main_account_code",
-            "sub_account_code",
-            "account_title",
-            "reporting_agency_id",
-            "reporting_agency_name",
-            "budget_bureau_code",
-            "budget_bureau_name",
-            "fr_entity_code",
-            "fr_entity_description",
-            "budget_function_code",
-            "budget_function_title",
-            "budget_subfunction_code",
-            "budget_subfunction_title",
-            "account_balances",
-            "program_balances",
-            "program_activities",
-            "object_classes",
-            "totals_program_activity",
-            "totals_object_class",
-            "totals",
-        ]
-
     @property
     def program_activities(self):
         return [
--- a/usaspending_api/common/serializers.py
+++ b/usaspending_api/common/serializers.py
@@ -56,7 +56,7 @@ class LimitableSerializer(serializers.ModelSerializer):
                 include_fields = include_fields + self.identify_missing_children(self.Meta.model, include_fields)
 
         # Create and initialize the child serializers
-        try:
+        if hasattr(self.Meta, "nested_serializers"):
             # Initialize the child serializers
             children = self.Meta.nested_serializers
             for field in children.keys():
@@ -84,11 +84,8 @@ class LimitableSerializer(serializers.ModelSerializer):
                     "context": {**self.context, 'verbose': False},  # Do not verbos-ify child objects
                     "fields": child_include_fields,
                     "exclude": child_exclude_fields
-                }  # The child tag should be removed when child field limiting is implemented
+                }
                 self.fields[field] = children[field]["class"](**child_args)
-        except AttributeError:
-            # We don't have any nested serializers
-            pass
 
         # Now we alter our own field sets
         # We must exclude before include to avoid conflicts from user error
--- a/usaspending_api/common/serializers.py
+++ b/usaspending_api/common/serializers.py
@@ -7,83 +7,130 @@ class LimitableSerializer(serializers.ModelSerializer):
 
     """Extends the model serializer to support field limiting."""
     def __init__(self, *args, **kwargs):
-        # next two lines are deprecated and will be removed
-        # once all views inherit from a generic class or are
-        # refactored into viewsets (i.e., once the serializer
-        # consistently has direct access to the request)
-        include_fields = kwargs.pop('fields', None)
-        exclude_fields = kwargs.pop('exclude', None)
+        # Grab any kwargs include and exclude fields, these are typically
+        # passed in by a parent serializer to the child serializer
+        kwargs_has_include = 'fields' in kwargs
+        kwargs_has_exclude = 'exclude' in kwargs
+        kwargs_include_fields = kwargs.pop('fields', [])
+        kwargs_exclude_fields = kwargs.pop('exclude', [])
+
+        # Intialize now that kwargs have been cleared
         super(LimitableSerializer, self).__init__(*args, **kwargs)
+
+        # Get params from our request or req object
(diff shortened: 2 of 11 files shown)


usaspending_api/common/serializers.py
```diff
@@ -7,83 +7,127 @@ class LimitableSerializer(serializers.ModelSerializer):
...
+        # Now we alter our own field sets
+        # We must exclude before include to avoid conflicts from user error
+        for field_name in exclude_fields:
+            self.fields.pop(field_name)
+
+        if len(include_fields) > 0:
+            allowed = set(include_fields)
+            existing = set(self.fields.keys())
+            for field_name in existing - allowed:
+                # If we have a coded field, always include its description
```
bsweger: I can't remember if we've always had this feature of automatically including the description of coded fields (I think it's been in since the story to add all of those `_description` fields?). 

If it's existing logic we should leave it in, but maybe a consideration for API 2.0 is am thinking more about the trade-offs between a convenience vs coupling the serializer to specific model names and deviating from our standard pattern of explicitly requesting fields when they're needed.

I know it's more nuanced here because these fields aren't always (ever?) in the DAIMS, so harder to surface them.

usaspending_api/common/serializers.py
```diff
@@ -7,83 +7,127 @@ class LimitableSerializer(serializers.ModelSerializer):
...
+                if field_name.split("_")[-1] == "description" and "_".join(field_name.split("_")[:-1]) not in allowed:
+                    continue
+                self.fields.pop(field_name)
+
+    # Returns a list of child names for fields in the field list
```
bsweger: Thanks, as always, for taking the time to document. Now that there's a function for this logic, what do you think about converting these remarks to the function's docstring?

usaspending_api/common/serializers.py
```diff
@@ -7,83 +7,127 @@ class LimitableSerializer(serializers.ModelSerializer):
...
+        children = []
+        model_fields = [f.name for f in model._meta.get_fields()]
+        for field in fields:
+            split = field.split("__")
+            if len(split) > 0 and split[0] in model_fields and split[0] not in fields:
+                children.append(split[0])
+
+        return children
+
+    # Takes a child's name and a list of fields, and returns a set of fields
```
bsweger: Same question as above re: docstring

usaspending_api/common/serializers.py
```diff
@@ -7,83 +7,127 @@ class LimitableSerializer(serializers.ModelSerializer):
...
+    # that belong to that child
+    def get_child_fields(self, child, fields):
```
bsweger: Great to see this logic as well as the code in `identify_missing_children` in their own functions. This sets the stage nicely for small unit tests that can focus on these very discrete, but important pieces of logic.

Three months ago I might have said that testing the larger unit of work is sufficient, but I've recently had a lot of luck (and caught some things) by getting more specific. Reasonable people can disagree where the line is, though. What do you think?

usaspending_api/common/tests/test_limitable_serializer.py
```diff
@@ -0,0 +1,52 @@
...
+from datetime import date
+
+from model_mommy import mommy
+from rest_framework import status
+
+from usaspending_api.awards.models import Award
+from usaspending_api.common.api_request_utils import FilterGenerator
+
+
+@pytest.fixture
```
bsweger: Thanks for rounding out the test cases in `common`--much needed, and great to see!

usaspending_api/accounts/models.py
```diff
@@ -89,49 +89,6 @@ def generate_tas_rendering_label(ATA, AID, TYPECODE, BPOA, EPOA, MAC, SUB):
         concatenated_tas = ATA + AID + TYPECODE + POAPHRASE + ACCTPHRASE
         return concatenated_tas
 
-    @staticmethod
```
bsweger: So great to remove all of this from the models--thank you for tackling this bit of tech debt as part of the story.

usaspending_api/common/serializers.py
```diff
@@ -7,83 +7,127 @@ class LimitableSerializer(serializers.ModelSerializer):
...
     def __init__(self, *args, **kwargs):
-        # next two lines are deprecated and will be removed
-        # once all views inherit from a generic class or are
-        # refactored into viewsets (i.e., once the serializer
-        # consistently has direct access to the request)
-        include_fields = kwargs.pop('fields', None)
-        exclude_fields = kwargs.pop('exclude', None)
+        # Grab any kwargs include and exclude fields, these are typically
+        # passed in by a parent serializer to the child serializer
+        kwargs_has_include = 'fields' in kwargs
```
bsweger: Just to make sure I understand what's happening here... by checking `**kwargs` for include and exclude fields, the views/viewsets to use the `get_serializer_context()` method to override the field defaults as necessary?

If that's true, do we need to do that in whatever viewset (tas/file a something or another?) returns the non-default list of fields? Or am I misunderstanding (entirely possible)?

usaspending_api/common/serializers.py
```diff
@@ -7,83 +7,127 @@ class LimitableSerializer(serializers.ModelSerializer):
...
-                try:
-                    include_fields = self.Meta.model.get_default_fields(path=request._request.path_info)
-                    allowed = set(include_fields)
-                    existing = set(self.fields.keys())
-                    for field_name in existing - allowed:
-                        self.fields.pop(field_name)
-                except AttributeError:
-                    # We don't have get default fields available
-                    pass
+        # Now we alter our own field sets
```
bsweger: It's great to see the default field logic de-coupled from the models. Since the trade-off for doing so (and a worthwhile one, imo) is to add some complexity to the serializer, does it make sense to encapsulate the logic above hat ultimately produces the final `include_fields` and `exclude_fields` list into a function that can be unit-tested?

Where I'm going with this thought is giving us a low barrier to entry for testing various permutations of the `**kwargs` and `default_fields` options to ensure that the resulting include/exclude field lists are what we expect.

usaspending_api/common/serializers.py
afrasier: This allows us to set the fields by passing them in via context if we want to. If they are not present in the context, we check the request object. Typically, the process flows:

Top level serializer checks kwargs, finds nothing
Top level serializer checks request, finds fields + excludes
Top level serializer passes down relevant child fields via their context
Then the process repeats all the way down the nesting chain.

usaspending_api/common/serializers.py
afrasier: Sure, although as  you mentioned above I'd like to tackle that later and get this in so it's not blocking front end work.

usaspending_api/common/serializers.py
afrasier: It's existing logic, yeah. And I agree with your points.

usaspending_api/common/serializers.py
afrasier: If you don't mind I can switch them over when I loop back to do the extra unit tests.

usaspending_api/common/serializers.py
afrasier: I think its important to test the smaller bits of logic as long is it makes sense to test them, as well as testing them incorporated into the larger scope. I can write some unit tests for these (And fix up the doc strings) when I loop back around.

afrasier:
As mentioned, I'll be sure to write some extra unit tests + fix up those doc strings when I loop back around, but I'd love to get this functionality in sooner than later so we don't block front-end any longer than we have to :)

PR #346, opened 2017-05-11
ATA / AID toptier agency linkages

afrasier:
In support of DS-1248 @mab6bf 

* Add `awarding_toptier_agency` and `funding_toptier_agency` linkages
based on ATA and AID as cgac
* Add command `update_tas_agencies` that updates all agencies with
proper linkages.
* Added call to `update_tas_agencies` to `load_reference_data` after
agencies and TAS have been loaded

When this is deployed, we'll need to ad-hoc call `update_tas_agencies` command on dev / stg / prod, etc.


afrasier:
Yeah I was trying to avoid doing a data migration because of past issues, but if you hate the idea of another management command then I can just make it a data migration.

bsweger:
@afrasier yeah, it's always a tricky decision, but I think in this case it makes sense to go the incorporate into loader/one-time migration route so our future selves won't have to do this extra step when the TAS file gets updated

afrasier:
Ok, migration has been added and the extra command removed!

usaspending_api/references/management/commands/load_reference_data.py
```diff
@@ -23,12 +23,10 @@ def handle(self, *args, **options):
         self.logger.info("Loading agency list")
         call_command('loadagencies')
 
+        # TAS's should only be loaded after agencies to ensure they can properly link to agencies
```
bsweger: good note, thanks!

PR #348, opened 2017-05-11
Remove State Code -> Foreign Province mapping

afrasier:
Removes State Code -> Foreign Province mapping from the submission
loader, and the relevant text from `data_reformatting.md`


data_reformatting.md
```diff
@@ -11,14 +11,6 @@ Generally, data will be imported as-is. In some cases, however, we reformat the
 
 Making sure our location-related fields are as robust and accurate as possible powers the geographic filters used by our website and API.
 
-### State Code and Foreign Province
```
bsweger: Thanks for remembering to update this!

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -400,16 +400,10 @@ def get_or_create_location(location_map, row, location_value_map={}):
     location_country = RefCountryCode.objects.filter(
         country_code=row[location_map.get('location_country_code')]).first()
 
-    # temporary fix until broker is patched: remove later
     state_code = row.get(location_map.get('state_code'))
     if state_code is not None:
-        # Fix for procurement data foreign provinces stored as state_code
-        if location_country and location_country.country_code != "USA":
-            location_value_map.update({'foreign_province': state_code})
```
bsweger: Just want to confirm: will we still load the `foreign_province` field when there's actually a value there in the source data? (I think so, but worth checking).

afrasier: Yes! :)

PR #353, opened 2017-05-12
Recipient tag groupings

afrasier:
This PR is in support of @mab6bf front end story regarding recipient filtering

* Adds a set of "category" buckets for recipient types, per Geoff and Ross
* Adds a method to execute the logic of updating these category flags
* Adds a data migration to set these category flags on currently existing data
* Adds `/references/recipients/` and `/references/recipients/<pk>` endpoint for better access/filtering of recipients
* Adds the "high level" categories to the default serialized field of LegalEntities


usaspending_api/references/models.py
```diff
@@ -294,6 +294,75 @@ class LegalEntity(DataSourceTrackedModel):
...
+    category_small_business = models.BooleanField(default=False)
+    category_other_than_small_business = models.BooleanField(default=False)
+
+    category_minority_owned_business = models.BooleanField(default=False)
+    category_alaskan_native_owned_business = models.BooleanField(default=False)
+    category_american_indian_owned_business = models.BooleanField(default=False)
+    category_asian_pacific_american_owned_business = models.BooleanField(default=False)
+    category_black_american_owned_business = models.BooleanField(default=False)
+    category_hispanic_american_owned_business = models.BooleanField(default=False)
+    category_native_american_owned_business = models.BooleanField(default=False)
```
catherinedevlin: OK, this is probably just carrying data broker weirdness through, but... `category_native_american_owned_business` and `category_american_indian_owned_business` are separate?  Just verifying...

usaspending_api/references/models.py
```diff
@@ -294,6 +294,75 @@ class LegalEntity(DataSourceTrackedModel):
...
+    category_native_hawaiian_owned_business = models.BooleanField(default=False)
+    category_subcontinent_asian_indian_american_owned_business = models.BooleanField(default=False)
+    category_tribally_owned_business = models.BooleanField(default=False)
+    category_other_minority_owned_business = models.BooleanField(default=False)
+
+    category_woman_owned_business = models.BooleanField(default=False)
```
russellbodine: Woman vs Women owned business.  Are they really different?

usaspending_api/references/models.py
```diff
@@ -401,10 +470,303 @@ class LegalEntity(DataSourceTrackedModel):
...
+
+        # Non-profit category
+        le.category_foundation = (
+            le.foundation == "1"
+        )
+        le.category_community_development_corporations = (
+            le.community_developed_corporation_owned_firm == "1"
+        )
+        le.category_nonprofit = (
+            le.business_types == "M" or  # Nonprofit with 501(c)(3) IRS Status (Other than Institution of Higher Education)
```
catherinedevlin: Thanks for these inline comments making the business types less crpytic

usaspending_api/references/models.py
```diff
@@ -401,10 +470,303 @@ class LegalEntity(DataSourceTrackedModel):
...
+            le.for_profit_organization == "1" or
+
+            le.category_small_business or
+            le.category_other_than_small_business
+        )
+        # End Business Category
+
+        # Minority Owned Business Category
+        le.category_alaskan_native_owned_business = (
+            le.alaskan_native_owned_corporation_or_firm == "1"
```
catherinedevlin: Very broadly, on all these fields, have we verified that we don't have any "uncleansed" versions of the same data?  For instance, if we're looking for "1", might there be variants like "1 ", "1.0", or maybe "Y", "T", etc., hiding out in our data?  And, for the ones holding letters, that there aren't lowercase variants?

usaspending_api/references/models.py
afrasier: Yes, I'm not sure what the legal distinction is there, but they are separate categories and flags.

usaspending_api/references/models.py
afrasier: These are all (so far) coming from the Broker, so they are all going to be the same (1 or 0 or null (since the source field in the broker is boolean)) and the business flags are a similar case. When we load historical items we'll need to ensure they're properly formatted, true.

afrasier:
Closing this for now in favor of trying an array field

usaspending_api/references/models.py
catherinedevlin: OK, thanks for the reassurance!

PR #354, opened 2017-05-15
Surrogate CFDA keys

afrasier:
This PR will move us to CFDA program models with appropriate surrogate keys.

How it works:
1. Create new CFDA model that has the default surrogate key (id)
2. Rename current CFDA fields to cfda_old
3. Create new cfda fields called cfda_new (This is to prevent relationship naming collisions)
4. Migration that (1) Creates the new CFDA model entries from the old and (2) Sets appropriate foreign key linkages from cfda_new to the new CFDA model based upon the old cfda link
5. Delete cfda_old links, rename cfda_new to cfda
6. Delete old CFDA model

This should be transparent to the front end.

Additional things
* Updated usages of CFDAProgram to use the new model
* Updated old loaders to use the fact we now have a surrogate key
* Updated test data to use the new model


usaspending_api/awards/models.py
```diff
@@ -455,7 +455,7 @@ class TransactionAssistance(DataSourceTrackedModel):
     uri = models.TextField(blank=True, null=True)
     cfda_number = models.TextField(blank=True, null=True, verbose_name="CFDA Number")
     cfda_title = models.TextField(blank=True, null=True, verbose_name="CFDA Title")
-    cfda = models.ForeignKey(CFDAProgram, models.DO_NOTHING, null=True)
+    cfda = models.ForeignKey(CFDA, models.DO_NOTHING, related_name="related_assistance", null=True)
```
bsweger: I like these explicitly-named relationships! Do you think it makes sense to take this once step further and be specific about the particular assistance attribute we're referencing here (e.g., `related_cfda`)?

usaspending_api/references/models.py
```diff
@@ -482,8 +482,8 @@ class Meta:
         unique_together = (('program_activity_code', 'budget_year', 'responsible_agency_id', 'allocation_transfer_agency_id', 'main_account_code'),)
 
 
-class CFDAProgram(DataSourceTrackedModel):
-    program_number = models.TextField(primary_key=True, max_length=7)
+class CFDA(DataSourceTrackedModel):
```
bsweger: I'm irrationally happy that you've taken this opportunity to give this model a better name. I know it's awkward when the name is actually an acronym, but do you think there's any value in calling it Cfda for ease of typing?

usaspending_api/references/models.py
```diff
@@ -521,16 +521,13 @@ class CFDAProgram(DataSourceTrackedModel):
...
     omb_agency_code = models.TextField(blank=True, null=True)
     omb_bureau_code = models.TextField(blank=True, null=True)
-    # published_date = models.DateTimeField(blank=True, null=True)
-    # archived_date = models.DateTimeField(blank=True, null=True)
     published_date = models.TextField(blank=True, null=True)
     archived_date = models.TextField(blank=True, null=True)
     create_date = models.DateTimeField(auto_now_add=True, blank=True, null=True)
     update_date = models.DateTimeField(auto_now=True, null=True)
 
     class Meta:
```
bsweger: Since we're recreating the table, what do you think about adding a unique index on `program_number`? At least on my local db, that field is already unique, but could be a nice line of defense for future updates this info.

usaspending_api/awards/models.py
```diff
@@ -500,7 +500,7 @@ class Subaward(DataSourceTrackedModel):
     award = models.ForeignKey(Award, models.CASCADE, related_name="subawards")
     recipient = models.ForeignKey(LegalEntity, models.DO_NOTHING)
     submission = models.ForeignKey(SubmissionAttributes, models.CASCADE)
-    cfda = models.ForeignKey(CFDAProgram, models.DO_NOTHING, null=True)
+    cfda = models.ForeignKey(CFDA, models.DO_NOTHING, related_name="related_subawards", null=True)
```
bsweger: What do you think about `related_cfda` or `cfda_subawards` or whatnot here, to make it clear that we're talking about CFDA?

usaspending_api/references/models.py
```diff
@@ -482,8 +482,8 @@ class Meta:
         unique_together = (('program_activity_code', 'budget_year', 'responsible_agency_id', 'allocation_transfer_agency_id', 'main_account_code'),)
 
 
-class CFDAProgram(DataSourceTrackedModel):
-    program_number = models.TextField(primary_key=True, max_length=7)
+class CFDA(DataSourceTrackedModel):
+    program_number = models.TextField(null=False)
```
bsweger: Good call to keep this as a required field even though it's no longer the PK.

usaspending_api/references/serializers.py
```diff
@@ -72,7 +72,7 @@ class Meta:
 class CfdaSerializer(LimitableSerializer):
 
     class Meta:
-        model = CFDAProgram
+        model = CFDA
         fields = '__all__'
         default_fields = [
```
bsweger: Now that we have a surrogate key, should we serialize `id` as well?

usaspending_api/awards/models.py
```diff
@@ -455,7 +455,7 @@ class TransactionAssistance(DataSourceTrackedModel):
     uri = models.TextField(blank=True, null=True)
     cfda_number = models.TextField(blank=True, null=True, verbose_name="CFDA Number")
```
bsweger: Can we remove `cfda_number` and `cfda_title` fields now that cfda is properly foreign-keyed? That would also be a forcing function to make sure that API users are grabbing this info from the one, canonical place. Not sure if this would require further updates to the data loaders...didn't check on that.

afrasier: It won't require any extra updates, but front end may need to make some changes if we remove those fields. I'm totally game to remove them though. @mab6bf any thoughts on that?

usaspending_api/awards/models.py
afrasier: So change it to something like "related_assistance_cfda"? We could, it might be overkill, because that reverse accessor is already going to be on the CFDA object already. (i.e. CFDA.objects.first().related_assistance)

usaspending_api/references/models.py
afrasier: I'm OK either way. Hopefully it doesn't cause any migration funkyness to change it. But just let me know if you want to pull the trigger on that and I can change it.

usaspending_api/references/models.py
afrasier: Good idea! Will do.

usaspending_api/references/serializers.py
afrasier: Seems like a good idea, good catch

usaspending_api/awards/models.py
bsweger: To close the loop publicly, front end is using these old fields, and we hadn't planned for their time when estimating the story, so we'll keep these redundant fields for now. Hopefully when the front-end has to use some of the new data we just surfaced, we'll be able to do this bit of cleanup.

usaspending_api/references/models.py
```diff
@@ -482,8 +482,8 @@ class Meta:
         unique_together = (('program_activity_code', 'budget_year', 'responsible_agency_id', 'allocation_transfer_agency_id', 'main_account_code'),)
 
 
-class CFDA(DataSourceTrackedModel):
-    program_number = models.TextField(null=False)
+class Cfda(DataSourceTrackedModel):
+    program_number = models.TextField(null=False, db_index=True)
```
bsweger: Thanks for swinging back for the index. I _think_ this will put an index on the field (which is good), but not necessary a unique index. IIRC, using `unique=True` will automatically create a unique index.

afrasier: You're right, I meant to put a unique on this (which comes with an index for free) instead of vice versa, let me fix that.

afrasier: Should be all set now, thanks!

PR #355, opened 2017-05-16
WIP: Materialized

catherinedevlin:
WORK IN PROGRESS.  Replacing the various operations to find an awarding agency with lookups in materialized views.


catherinedevlin:
Withdrawing this, at least for the time being, b/c the Pandas approach seems more promising.

PR #356, opened 2017-05-16
guide-autocomplete

russellbodine:
Make an autocomplete endpoint for the guide

Added 

- url connection
- autocomplete view
- test case

Arman fixed a routing issue.


usaspending_api/references/urls.py
```diff
@@ -10,12 +10,13 @@
 
 
 urlpatterns = [
-    url(r'', include(guide_router.urls)),
     url(r'^locations/$', views.LocationEndpoint.as_view(mode_list)),
     url(r'^locations/geocomplete', views.LocationGeoCompleteEndpoint.as_view()),
     url(r'^agency/$', views.AgencyEndpoint.as_view(mode_list)),
     url(r'^agency/autocomplete', views.AgencyAutocomplete.as_view()),
     url(r'^recipients/autocomplete', views.RecipientAutocomplete.as_view()),
+    url(r'^guide/autocomplete', views.GuideAutocomplete.as_view()),
```
bsweger: Pinging @catherinedevlin, since she implemented the guide router when creating the original endpoint last sprint. I think taking this step makes a lot of sense, am not quite clear if/how we should continue to follow that pattern when adding, say, an `autocomplete`.

catherinedevlin: Well, I tried making that work, but it's a little complicated because he's subclassing `AutocompleteView`, which is a `View` and not a `ViewSet`, which works differently with routers...

Since this would be the first autocomplete endpoint using a router, I think it's OK to leave that refactoring to some future date where we move everything to routers rather than trying to work it into this ticket.

PR #357, opened 2017-05-16
Recipient categories

afrasier:
This implements the recipient business type categories using an ArrayField. @mab6bf tagging for visibility

* Adds `business_categories` field to LegalEntity
* Adds logic to fill this array list with appropriate business categories
* Updates filter generator to support proper contains for ArrayField, i.e. you can do the following request:
```
{
    "field": "business_categories",
    "operation": "contains",
    "value": ["nonprofit", "us_owned_business"]
}
```
* Adds test for new `references/recipients/` and `references/recipients/<pk>` endpoints
* Adds missing `recipient_unique_id` to default legal entity serializer
* Migration to set business categories for currently existing LegalEntities


usaspending_api/references/serializers.py
```diff
@@ -132,9 +132,11 @@ class Meta:
         default_fields = [
             "legal_entity_id",
             "parent_recipient_unique_id",
+            "recipient_unique_id",
```
bsweger: Thanks for catching this. Oof, looks like another piece of confusion re: our legacy data models (_i.e._, `legal_entity_id` as the surrogate key). Out of scope for this work, but just noting since I thought your CFDA work was the end of our FK-renaming days. I added this note to our db model cleanup story.

usaspending_api/references/urls.py
```diff
@@ -15,7 +15,9 @@
     url(r'^locations/geocomplete', views.LocationGeoCompleteEndpoint.as_view()),
     url(r'^agency/$', views.AgencyEndpoint.as_view(mode_list)),
     url(r'^agency/autocomplete', views.AgencyAutocomplete.as_view()),
-    url(r'^recipients/autocomplete', views.RecipientAutocomplete.as_view()),
+    url(r'^recipients/$', views.RecipientViewSet.as_view(mode_list), name='recipient-list'),
```
bsweger: While we're in here making changes, what do you think about moving away from the `references/` prefix in our endpoints? Technically doing this now would be jumping the gun, since we haven't laid down our new API guiding principles yet, but I've long thought the `references` prefix doesn't have any meaning, since objects like recipients, cfda, and the like are full-fledged members of the fedspending data ecosystem.

Not something we need to change right now, but just planting the seed...

usaspending_api/references/views.py
```diff
@@ -83,6 +83,24 @@ def get_queryset(self):
...
+                       DetailViewSet):
+    """
+    Returns information about award recipients and vendors
+    """
+
+    serializer_class = LegalEntitySerializer
+
+    def get_queryset(self):
+        """Return the view's queryset."""
+        queryset = LegalEntity.objects.all().exclude(recipient_unique_id__isnull=True)
```
bsweger: Cool, I'm glad we're adding a full-fledged recipients endpoint!

usaspending_api/common/api_request_utils.py
```diff
@@ -212,6 +212,11 @@ def create_q_from_filter(self, filt):
                 else:
                     # Otherwise, use built in django in
                     operation = "__in"
+            if operation is '__icontains' and isinstance(value, list):
```
bsweger: It's a bit unclear what this means from a user perspective...does that mean that the search for business types will be case sensitive?  Would putting in a test case for this type of array field searching might be a good way to make the expected behavior clearer while making sure we have coverage for this new feature?

usaspending_api/references/models.py
```diff
@@ -401,10 +474,362 @@ class LegalEntity(DataSourceTrackedModel):
     individual = models.TextField(blank=True, null=True)
 
     def save(self, *args, **kwargs):
+        LegalEntity.update_business_type_categories(self)
         super(LegalEntity, self).save(*args, **kwargs)
 
         LegalEntityOfficers.objects.get_or_create(legal_entity=self)
 
+    @staticmethod
+    def update_business_type_categories(le):
```
bsweger: Good call on putting this logic into the model instead of one of our separate helper files.

usaspending_api/references/models.py
```diff
@@ -401,10 +474,362 @@ class LegalEntity(DataSourceTrackedModel):
...
+        # Create a new list of categories
+        categories = []
```
bsweger: Just to make sure I understand correctly--we're adding this biz type array to support searching and displaying those, but we're also keeping the long list of individual business designations on the model in case folks need to reference them individually?

usaspending_api/references/migrations/0007_auto_20170516_1702.py
```diff
@@ -0,0 +1,24 @@
...
+
+from django.db import migrations
+from usaspending_api.references.models import LegalEntity
+
+
+def forwards_func(apps, schema_editor):
+    legalentities = apps.get_model("references", "LegalEntity")
+    for le in legalentities.objects.all():
+        LegalEntity.update_business_type_categories(le)
+        le.save()
```
bsweger: Does `le.save()` already trigger `update_business_type_categories(le)?

usaspending_api/references/urls.py
afrasier: I'd agree.

usaspending_api/common/api_request_utils.py
afrasier: The Django implementation of ArrayField does not override "icontains", it only works using the "contains" method; and due to that is always case sensitive. I can add some extra tests.

usaspending_api/references/models.py
afrasier: Yes, these categories "bin" a few of them together.

usaspending_api/references/migrations/0007_auto_20170516_1702.py
afrasier: I wasn't sure if Django was smart enough to not save it because there weren't any changes :) Do you have any insight on that?

afrasier:
@bsweger Tests have been added, if you'd care to re-review

usaspending_api/references/migrations/0007_auto_20170516_1702.py
bsweger: Well, no particular insight. Though it occurs to me that we've been getting away from `save` methods on the models since we never update our data except through the load jobs. So this convo brings me to a different question...do we need the `save()` method, especially if we're going towards bulk inserts and it won't be triggered anyway?

bsweger:
Thanks for the tests, looks good! I don't think the `save()` line of questions needs to hold up a merge. If you can update the docs to include the new field, we should be all set!

afrasier:
Alrighty, docs updated!

PR #358, opened 2017-05-17
Fix erroneous mapping

afrasier:
When we updated the parent unique id name it looks like we overlooked
changing the default mappings to match.


afrasier:
What response? From the broker?

bsweger:
@afrasier From the API. If I understand correctly, we weren't loading values for `parent_recipient_unique_id` because we had it mapped wrong to the broker data. Do we have any unit tests or something in our big old endpoint series of tests that check this field to make sure we've grabbed it correctly now?

Am approving this so people can get the data...we can follow up with tests.

afrasier:
I suppose we'd need to greatly expand the load_submission test suite to cover that.

PR #359, opened 2017-05-17
Aggregation Null Opt-In

afrasier:
Per suggestion from @mab6bf and @kevinli-work this PR introduces functionality to support "opt-in" for null values in aggregates.

Unless `show_nulls` is set to `true` on an aggregate endpoint request, instances where _all_ group fields are null will be filtered out. For instance:

```
{
    "field": "federal_action_obligation",
    "group": ["assistance_data__cfda__program_number", "contract_data__naics"]
}
```
Will only show results where either of the groups are not null; hiding those where they are both null. In the case of a single group:
```
{
    "field": "federal_action_obligation",
    "group": "assistance_data__cfda__program_number"
}
```
Will only show results where the CFDA program number is not null.
```
{
    "field": "federal_action_obligation",
    "group": "assistance_data__cfda__program_number",
    "show_nulls": true
}
```
This will replicate current functionality, and show the aggregate where that group is null.

* Adds functionality to support filtering null groups on aggregate requests
* Adds tests for new functionality
* Adds `show_nulls` option to aggregate endpoint requests
* Updates documentation for the new `show_nulls` option


bsweger:
Per our Slack convo, we're going to hold this one until there's a corresponding story and more robust discussion about how the NULLs are impacting front-end and the best way to fix this.

PR #363, opened 2017-05-18
Removed financing accounts and stopped adding them to db

russellbodine:
This is done by adding a function in loadtas.py called
skip_and_remove_financing_tas and calling that function before every
row.  

loadtas.py MUST BE CALLED AFTER PR is accepted to remove all current financing accounts!!!


russellbodine:
woops, i messed up and created a nother pr instead of this one...

PR #364, opened 2017-05-18
Use pandas

catherinedevlin:
Moved several operations from within the file C loop to Pandas operation on the overall data set before the loop begins.  

There's some tidying up I ought to do on function names and docstrings, but right now I need to scramble for my plane... I can come back around to it when I'm available, or others can go ahead and make changes as appropriate.


afrasier:
Poking around a bit:

Loading 2637 locally, on dev
```
real	4m27.583s
user	1m28.559s
sys	0m4.834s
```

With this branch:
```
real	2m16.321s
user	1m9.542s
sys	0m3.398s
```

Very exciting!
I had to edit where you pass in an environment variable as it seems pandas wants a db connection instead, I've committed this change.

PR #369, opened 2017-05-24
Remove federal accounts added to loadtas workflow

russellbodine:
added cleanup function to remove old federal accounts


usaspending_api/references/reference_helpers.py
```diff
@@ -5,6 +5,10 @@
 from usaspending_api.accounts.models import FederalAccount, TreasuryAppropriationAccount
 
 
+def remove_federal_accounts(tas_tuple=None):
```
catherinedevlin: Could we get a docstring on this?  Just `remove_federal_accounts` doesn't answer many questions (remove which ones?  Remove them why?)

... granted, a single line of code is not hard to read, but still.

afrasier: Maybe renaming it to `remove_empty_federal_accounts` would clarify things, as well :)

usaspending_api/references/reference_helpers.py
```diff
@@ -5,6 +5,10 @@
...
+    FederalAccount.objects.filter(treasuryappropriationaccount__isnull=True).distinct().delete()
```
russellbodine: this only runs once anyway. (we are removing the bad federal accounts that are currently in db)

usaspending_api/references/reference_helpers.py
russellbodine: renamed and added docstring

catherinedevlin: Renaming was a great suggestion; thanks @afrasier  !

PR #372, opened 2017-05-31
Filter hash endpoint

russellbodine:
Endpoints made in references.

Explanation of these endpoints can be found here:

https://gist.github.com/russellbodine/ad03c78cf15325ff1bdb29fb6fab1ee5

PLS MESSAGE KEVIN AFTER PR COMPLETES


usaspending_api/references/models.py
```diff
@@ -185,6 +186,18 @@ class Meta:
         db_table = 'office_agency'
 
 
+class FilterHash(models.Model):
+    filter_hash_id = models.AutoField(primary_key=True)
```
catherinedevlin: Since you don't seem to be using `filter_hash_id` anywhere, making this change should be as simple as deleting line 190 (and regenerating the migration).  Django will automatically set up the `id`.

usaspending_api/references/views.py
```diff
@@ -2,11 +2,70 @@
...
+        hash = m.hexdigest().encode('utf8')
+        # check for hash in db, store if not in db
+        # global fh
+        # fh = FilterHash.objects.all()
+        try:
+            fh = FilterHash.objects.get(hash=hash)
+        except FilterHash.DoesNotExist:
+            # store in DB
+            try:
+                json_dict = json.loads(json_dict)
```
catherinedevlin: You can get this with a single step: it's in `request.data` - then you can drop some of the confusing intermediary steps.

usaspending_api/references/views.py
```diff
@@ -2,11 +2,70 @@
...
+    def post(self, request, format=None):
+        """return the hash for a json"""
+        # get hash
+        body_unicode = request.body.decode('utf-8')
+        body = json.loads(body_unicode)
+        hash = body["hash"]
+        # check for hash in db, if not in db
+        try:
+            fh = FilterHash.objects.get(hash=hash)
+            json_dict = json.loads(fh.filter)
```
catherinedevlin: Same comment here about using `request.data`

usaspending_api/references/models.py
```diff
@@ -185,6 +186,18 @@ class Meta:
...
+    create_date = models.DateTimeField(auto_now_add=True, blank=True, null=True)
```
russellbodine: I noticed that all the models seem to have null=true for create_data. I believe this is set because null=True is required bj Djago if the field is not unique.

usaspending_api/references/views.py
russellbodine: Fixed

usaspending_api/references/views.py
russellbodine: running into some typing errors.  gonna skip this one and load from db.

russellbodine: nvm I fixed it.

russellbodine:
I did a fair bit of unit testing, but im all for making another story for documentation and a pytest.  This does need to go out the door ASAP tho.

PR #374, opened 2017-06-02
DS 1320 award spending endpoints

nmonga91:
Adding 2 new v2 endpoints: `/award_spending/award_type` and `/award_spending/recipient`.

Associated gists:

- Award Type: https://gist.github.com/nmonga91/fbc8c26739b66c3136ac4184fed6e665
- Recipient: https://gist.github.com/nmonga91/e85d076f057df2657f00191c10952911

Requirements:
- [x] API Docs updated


usaspending_api/awards/migrations/0007_auto_20170601_1715.py
```diff
@@ -0,0 +1,25 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='historicaltransaction',
+            name='fiscal_year',
+            field=models.IntegerField(blank=True, help_text='Fiscal Year calculated based on Action Date', null=True),
+        ),
+        migrations.AddField(
+            model_name='transaction',
+            name='fiscal_year',
+            field=models.IntegerField(blank=True, help_text='Fiscal Year calculated based on Action Date', null=True),
```
catherinedevlin: Oh, I'd been forgetting about `help_text`!  Thank you for using that.

usaspending_api/awards/views.py
```diff
@@ -6,15 +6,81 @@
...
+
+        try:
+            queryset = Transaction.objects.all()
+            # Filter based on fiscal year and agency id
+            queryset = queryset.filter(fiscal_year=fiscal_year, awarding_agency=awarding_agency_id)
+            # alias awards.category to be award_type
+            queryset = queryset.annotate(award_type=F('award__category'))
+            # sum obligations for each category type
+            queryset = queryset.values('award_type').annotate(obligated_amount=Sum('federal_action_obligation'))
+        except Error:
```
catherinedevlin: I'm not sure the reasoning of this catch - it just catches a specific exception and re-raises it as a vague exception, destroying information that could be useful for debugging; unless I'm missing something, that doesn't seem good.  Could we catch it with `except Error as e` and add `str(e)` to the exception message?

usaspending_api/api_docs/markdown/endpoints.md
```diff
@@ -42,6 +42,8 @@ To reduce unnecessary data transfer, most endpoints return a default set of info
 | [/api/v1/accounts/awards/](/api/v1/accounts/awards/) | GET, POST | Returns a list of financial account data by treasury account symbol, program activity, object class, and award |
 | /api/v1/accounts/awards/:id | GET, POST | Returns a single treasury account symbol/program activity/object class/award record with all fields |
 | [/api/v1/accounts/awards/total/](/api/v1/accounts/awards/total/) | POST | Supports aggregation on treasury account symbol/program activity/object class/award records |
+| [/api/v1/award_spending/award_type/](/api/v1/award_spending/award_type/) | GET | Returns a list of total obligations by award_type |
```
catherinedevlin: I'm confused - this doc lists them as in `v1` but below they're going in as `v2`?

usaspending_api/awards/views.py
nmonga91: The original thought behind this was to only catch django ORM error types. I can remove this try/except all together if it's acceptable for the exception to just be handled at the top most level.

usaspending_api/api_docs/markdown/endpoints.md
nmonga91: Yes, this should be `v2`. Thanks for catching this! Will update

PR #376, opened 2017-06-05
Financial spending object class skunkworks

russellbodine:


usaspending_api/accounts/serializers.py
```diff
@@ -115,3 +115,10 @@ class Meta:
                 "kwargs": {"read_only": True}
             }
         }
+
+
+class ObjectClassFinancialSpendingSerializer(serializers.Serializer):
+
+    major_object_class_code = serializers.CharField()
+    major_object_class_name = serializers.CharField()
+    obligated_amount = serializers.DecimalField(21, 2)
```
bsweger: @nmonga91 Wanted to pull you in here...I started a pattern of serializing amounts as `DecimalField` instead of characters. It doesn't seem to affect the actual output, which is still rendered in quotes. What are your feelings on this?

usaspending_api/accounts/views/federal_spending.py
```diff
@@ -23,7 +24,7 @@ def get_queryset(self):
 
         # required query parameters were not provided
         if not (fiscal_year and agency_id):
-            raise ParseError('Missing one or more required query parameters: fiscal_year, agency_id')
+            raise InvalidParameterException('Missing one or more required query parameters: fiscal_year, agency_id')
```
bsweger: Good call to change the error type here--it didn't make sense to raise a 500 when the incoming parameters were wrong.

usaspending_api/accounts/views/financial_spending.py
```diff
@@ -24,7 +23,7 @@ def get_queryset(self):
 
         # required query parameters were not provided
         if not (fiscal_year and agency_id):
-            raise InvalidParameterException('Missing one or more required query parameters: fiscal_year, agency_id')
+            raise ParseError('Missing one or more required query parameters: fiscal_year, agency_id')
```
bsweger: Heh, saw you removed `ParseError` in a previous commit, and now it's back. Is there a reason not to use `InvalidParameterException` here? I do think we should be returning a 400 if the request is invalid.

usaspending_api/references/tests/test_financial_spending.py
```diff
@@ -0,0 +1,39 @@
...
+    # agency w/ tt agency
+    ttagency1 = mommy.make('references.ToptierAgency', cgac_code="cgac_code", name="tta_name")
+    mommy.make('references.Agency', id=1, toptier_agency=ttagency1)
+    # TAS w/ same ttagency
+    tas1 = mommy.make('accounts.TreasuryAppropriationAccount', awarding_toptier_agency=ttagency1, funding_toptier_agency=ttagency1)
+
+    # Financial Account with Object class and submission
+    object_class_1 = mommy.make('references.ObjectClass', major_object_class="mocCode",
+                                major_object_class_name="mocName", object_class="ocCode", object_class_name="ocName")
+    submission_1 = mommy.make('submissions.SubmissionAttributes', reporting_fiscal_year=2017, cgac_code="cgac_code")
```
bsweger: What do you think about not explicitly setting`cgac_code` on the submission? 

Thinking from the perspective of a new dev who is learning the code base by reading the tests....setting `cgac_code="cgac_code"` implies that it's significant to the API lookup, when in reality the only toptier agency info that matters is the `funding_toptier_agency` that you set in `tas1`.

usaspending_api/references/tests/test_financial_spending.py
```diff
@@ -0,0 +1,39 @@
...
+@pytest.mark.django_db
+def test_award_type_endpoint(client, financial_spending_data):
+    """Test the award_type endpoint."""
+
+    resp = client.get('/api/v2/financial_spending/object_class/?fiscal_year=2017&funding_agency_id=1')
+    print(resp.data)
+    assert resp.status_code == status.HTTP_200_OK
+    assert len(resp.data['results']) > 0
+
+    resp = client.get('/api/v2/financial_spending/object_class/')
```
bsweger: Good idea to check that missing params cause an error. I had to read this a few times to understand the intent, so it might be a good idea to preface this with a comment about what you're testing for.

usaspending_api/references/tests/test_financial_spending.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@pytest.mark.django_db
+def test_award_type_endpoint(client, financial_spending_data):
+    """Test the award_type endpoint."""
+
+    resp = client.get('/api/v2/financial_spending/object_class/?fiscal_year=2017&funding_agency_id=1')
+    print(resp.data)
+    assert resp.status_code == status.HTTP_200_OK
+    assert len(resp.data['results']) > 0
```
bsweger: I know that setting up the test models here is tedious (having done it for the `financial_balances/agencies/` endpoint). Since you've gotten this far, I think it would be worthwhile to keep going and:

1. explicitly test for the number of result you're expecting (instead of `> 0`)
2. add a `FinancialAccountsByProgramActivityObjectClass` object with a different tas and/or object class to ensure that those numbers don't show up in the response
3. add at least one more `FinancialAccountsByProgramActivityObjectClass` with `tas1` and `object_class_1` and check the response to make sure the numbers summed correctly

usaspending_api/references/tests/test_financial_spending.py
russellbodine: comment added!

usaspending_api/accounts/views/financial_spending.py
russellbodine: the exception handler doesnt know how to handle ParseError, so it throws a 500 after.  It knows about InvalidParameterException though, so it can through a 400 after it gets raised. Thanks @nmonga91 for figuring this out.

usaspending_api/references/tests/test_financial_spending.py
russellbodine: this was left over from the last code update.  I'll take it out!

usaspending_api/accounts/views/financial_spending.py
bsweger: Oooh, good sleuthing. Should we fix that? Asking because I'm using `InvalidParameterException` for the `financial_balances/agencies` endpoint and want to be consistent. But it doesn't seem right to return a 500 for bad params.

usaspending_api/references/tests/test_financial_spending.py
russellbodine: Added all 3 points!

usaspending_api/accounts/serializers.py
russellbodine: Discussed in group chat.

usaspending_api/accounts/views/federal_spending.py
russellbodine: 👍

usaspending_api/accounts/serializers.py
bsweger: @russellbodine I think we decided in the chat to change this to `DecimalField(None, 2)` to get rid of unneeded specificity.

usaspending_api/accounts/serializers.py
```diff
@@ -122,3 +122,11 @@ class ObjectClassFinancialSpendingSerializer(serializers.Serializer):
     major_object_class_code = serializers.CharField()
     major_object_class_name = serializers.CharField()
     obligated_amount = serializers.DecimalField(21, 2)
+
+
+class AgenciesFinancialBalancesSerializer(serializers.Serializer):
+
+    fiscal_year = serializers.IntegerField()
```
bsweger: Note--these are mine, so you can ignore them...I have a hotfix in to set these to `None`

usaspending_api/references/tests/test_financial_spending.py
```diff
@@ -33,7 +49,8 @@ def test_award_type_endpoint(client, financial_spending_data):
     resp = client.get('/api/v2/financial_spending/object_class/?fiscal_year=2017&funding_agency_id=1')
     print(resp.data)
     assert resp.status_code == status.HTTP_200_OK
-    assert len(resp.data['results']) > 0
+    assert len(resp.data['results']) == 2
```
bsweger: Thanks for expanding these tests! I think we still need a check for the actual dollar amounts being returned in the response, particularly for a value that's been summed. You may need to one more object class/tas record to support this (_i.e._, make sure that at least one object class/tas combo has more than one record in `FinancialAccountsByProgramActivityObjectClass` to ensure that things are added correctly.

PR #377, opened 2017-06-05
Add /v2/financial_balances/agencies endpoint

bsweger:
This PR is part of the work for [DS-1316](https://federal-spending-transparency.atlassian.net/browse/DS-1316). The acceptance criteria for the `financial_balances/agencies/` endpoint have changed as the sprint progressed--our latest thinking for how this endpoint should behave is here: https://gist.github.com/bsweger/91d85b7f4ef1fb9c4f0a54c541ef5da6

### DIFF
--- a/usaspending_api/accounts/serializers.py
+++ b/usaspending_api/accounts/serializers.py
@@ -115,3 +115,11 @@ class TasCategorySerializer(LimitableSerializer):
                 "kwargs": {"read_only": True}
             }
         }
+
+
+class AgenciesFinancialBalancesSerializer(serializers.Serializer):
+
+    fiscal_year = serializers.IntegerField()
+    budget_authority_amount = serializers.DecimalField(21, 2)
+    obligated_amount = serializers.DecimalField(21, 2)
+    outlay_amount = serializers.DecimalField(21, 2)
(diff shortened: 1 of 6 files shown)


usaspending_api/accounts/serializers.py
```diff
@@ -115,3 +115,11 @@ class Meta:
                 "kwargs": {"read_only": True}
             }
         }
+
+
+class AgenciesFinancialBalancesSerializer(serializers.Serializer):
+
+    fiscal_year = serializers.IntegerField()
+    budget_authority_amount = serializers.DecimalField(21, 2)
```
catherinedevlin: I've got no experience with specifying serializer columns this way, so bear with me if I'm off-base, but: is it necessary to specify the precision (especially `max_digits`) for a field in the serializer?  If it's possible, it seems more flexible to leave `max_digits` unspecified.  Or is there a benefit to specifying it I don't realize?

(It's not that I'm trying to build the API to accommodate hyperinflation...)

bsweger: Yes--we think alike on this matter! However, when I removed those precisions, I got this error:
`TypeError: __init__() missing 2 required positional arguments: 'max_digits' and 'decimal_places`

As much as I agree that we don't need to be this specific given the upstream validations, with this precision, we're at $`99,999,999,999,999,999,999.99`. Hopefully we are safe for some time to come!

PR #380, opened 2017-06-06
Remove fiscal_year from AgenciesFinancialBalancesSerializer

bsweger:
Because fiscal_year is required in the request parameters,
remove it from the response. This change makes the
/financial_balances/agencies/ endpoint behavior consistent
with the others that we're building out for the agency page.

### DIFF
--- a/usaspending_api/accounts/serializers.py
+++ b/usaspending_api/accounts/serializers.py
@@ -126,7 +126,6 @@ class ObjectClassFinancialSpendingSerializer(serializers.Serializer):
 
 class AgenciesFinancialBalancesSerializer(serializers.Serializer):
 
-    fiscal_year = serializers.IntegerField()
     budget_authority_amount = serializers.DecimalField(None, 2)
     obligated_amount = serializers.DecimalField(None, 2)
     outlay_amount = serializers.DecimalField(None, 2)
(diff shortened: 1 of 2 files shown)


usaspending_api/accounts/serializers.py
```diff
@@ -126,7 +126,6 @@ class ObjectClassFinancialSpendingSerializer(serializers.Serializer):
 
 class AgenciesFinancialBalancesSerializer(serializers.Serializer):
 
-    fiscal_year = serializers.IntegerField()
```
nmonga91: Since we're removing it from the serializer, do we want to alter the queryset to no longer include `.values('fiscal_year')`?

bsweger: Yeah, good point...will remove.

bsweger: @nmonga91 Took another look at this--the `annotate` statement Sums the values does need the fiscal year to be specified, because that's the group by field. We could technically don't need to annotate `submission__reporting_fiscal_year` into `fiscal_year` for this to work, but I kind of like it that way for readability. What do you think?

nmonga91: I'm all for leaving it as is for readability. So if it's removed from the serializer, but not the queryset, it won't be part of the response at all?

nmonga91: ^ confirmed it won't be in the response

PR #385, opened 2017-06-12
Financial spending object class endpoint added, url, serializer, and Test

russellbodine:


usaspending_api/accounts/views/financial_spending.py
```diff
@@ -44,5 +45,46 @@ def get_queryset(self):
...
+            raise InvalidParameterException('Missing one or more required query parameters: fiscal_year, funding_agency_id')
 
+        # using final_objects below ensures that we're only pulling the latest
+        # set of financial information for each fiscal year
+        queryset = FinancialAccountsByProgramActivityObjectClass.final_objects.all()
+        # get the incoming agency's toptier agency, because that's what we'll
+        # need to filter on
+        # (used filter() instead of get() b/c we likely don't want to raise an
+        # error on a bad agency id)
+        toptier_agency = Agency.objects.filter(id=funding_agency_id).first().toptier_agency
```
nmonga91: What happens when the `funcing_agency_id` isn't provided? Is there an error calling `.first()` on an empty result set? and then possibly calling `.toptier_agency` on None?

usaspending_api/accounts/views/financial_spending.py
```diff
@@ -44,5 +45,46 @@ def get_queryset(self):
...
+        queryset = queryset.filter(
+            submission__reporting_fiscal_year=fiscal_year,
+            treasury_account__funding_toptier_agency=toptier_agency,
```
nmonga91: May need to reconsider how we handle this filter for `funding_toptier_agency` if the user passes a non-existent ID. We could possibly short-circuit the response and just return an empty query set immediately.

usaspending_api/accounts/views/financial_spending.py
russellbodine: line 24 checks to make sure the value is provided (if not provided an invalad parameter exception is thrown).  if the provided id doesnt exist, an empty results array is returned.  This is the behavior in other endpoints.

usaspending_api/accounts/views/financial_spending.py
russellbodine: if the provided id doesnt exist, an empty results array is returned. This is the behavior in other endpoints.

usaspending_api/accounts/views/financial_spending.py
```diff
@@ -44,5 +45,51 @@ def get_queryset(self):
...
+        # using final_objects below ensures that we're only pulling the latest
+        # set of financial information for each fiscal year
+        queryset = FinancialAccountsByProgramActivityObjectClass.final_objects.all()
+        # get the incoming agency's toptier agency, because that's what we'll
+        # need to filter on
+        # (used filter() instead of get() b/c we likely don't want to raise an
+        # error on a bad agency id)
+        agency = Agency.objects.filter(id=funding_agency_id).first()
+        if agency is None:
+            queryset = queryset.filter(
```
nmonga91: here can we immediately return the empty queryset without having to continue with the annotations below? Also another way to return an empty queryset is `Model.objects.none()` (found it and thought it could be useful)

usaspending_api/references/tests/test_financial_spending.py
```diff
@@ -49,13 +49,25 @@ def financial_spending_data(db):
...
     # make sure resp.data['results'] contains a 3000 value
-    assert resp.data['results'][1]['obligated_amount'] == "3000.00"
+    assert (resp.data['results'][1]['obligated_amount'] == "3000.00" or resp.data['results'][0]['obligated_amount'] == "3000.00")
+
+    # check for bad request due to missing params
+    resp = client.get('/api/v2/financial_spending/object_class/')
+    assert resp.status_code == status.HTTP_400_BAD_REQUEST
+
+    resp = client.get('/api/v2/financial_spending/object_class/?fiscal_year=2017&funding_agency_id=1&major_object_class_code=10')
+    print(resp.data)
```
nmonga91: Should probably remove this print statement

usaspending_api/references/tests/test_financial_spending.py
```diff
@@ -49,13 +49,25 @@ def financial_spending_data(db):
 def test_award_type_endpoint(client, financial_spending_data):
     """Test the award_type endpoint."""
 
-    resp = client.get('/api/v2/financial_spending/object_class/?fiscal_year=2017&funding_agency_id=1')
+    resp = client.get('/api/v2/financial_spending/major_object_class/?fiscal_year=2017&funding_agency_id=1')
     print(resp.data)
```
nmonga91: Should probably remove this print statement

usaspending_api/references/tests/test_financial_spending.py
russellbodine: Im removing them!  fyi, they dont print anything unless the code fails or the explicit flag is raised.

usaspending_api/references/tests/test_financial_spending.py
russellbodine: done

PR #386, opened 2017-06-13
Agency endpoint

russellbodine:


usaspending_api/references/serializers_v2.py
```diff
@@ -0,0 +1,12 @@
...
+from usaspending_api.references.models import (
+    Agency, Cfda, LegalEntity, Location, ObjectClass, OfficeAgency,
+    RefProgramActivity, SubtierAgency, ToptierAgency, LegalEntityOfficers, Definition)
+from usaspending_api.common.serializers import LimitableSerializer
+
+
+class AgencyV2Serializer(serializers.Serializer):
+
+    agency_name = serializers.CharField()
+    # active_fiscal_year = serializers.IntegerField()
```
nmonga91: We made the obligations a `DecimalField`, should we stick with this instead of the line below it?

usaspending_api/references/views.py
```diff
@@ -109,7 +109,6 @@ class AgencyEndpoint(SuperLoggingMixin,
                      FilterQuerysetMixin,
                      DetailViewSet):
     """Return an agency"""
-    serializer_class = AgencySerializer
```
nmonga91: Should this be removed? would it break the existing endpoint?

usaspending_api/references/urls.py
```diff
@@ -15,6 +15,7 @@
     url(r'^locations/$', views.LocationEndpoint.as_view(mode_list)),
     url(r'^locations/geocomplete', views.LocationGeoCompleteEndpoint.as_view()),
     url(r'^agency/$', views.AgencyEndpoint.as_view(mode_list)),
+    url(r'^agency/(?P<pk>[0-9]+)/$', views.AgencyEndpoint.as_view(mode_detail), name='agency-detail'),
```
nmonga91: Is this only a new v2 endpoint? If so, wondering about the addition here

usaspending_api/references/urls_v2.py
```diff
@@ -0,0 +1,16 @@
...
+
+guide_router = DefaultRouter()
+guide_router.register('guide', views.GuideViewSet)
+
+mode_list = {'get': 'list', 'post': 'list'}
+mode_detail = {'get': 'retrieve', 'post': 'retrieve'}
+
+
+urlpatterns = [
+    # url(r'^agency/$', views.AgencyEndpoint.as_view(mode_list)),
```
nmonga91: If the commented out lines aren't being used, we should probably remove them

usaspending_api/urls.py
```diff
@@ -36,6 +36,7 @@
     url(r'^api/v2/financial_balances/', include('usaspending_api.accounts.urls_financial_balances')),
     url(r'^api/v1/tas/', include('usaspending_api.accounts.urls_tas')),
     url(r'^api/v1/references/', include('usaspending_api.references.urls')),
+    url(r'^api/v2/references/', include('usaspending_api.references.urls_v2')),
```
nmonga91: I altered this in my branch a bit, but let's group all the v2 endpoints together below all the v1 endpoints so it's consolidated as much as possilbe

usaspending_api/references/serializers_v2.py
russellbodine: woops, forgot to change it!

nmonga91:
It may be better to use a `views_v2` directory in `references/` (same for serializers, urls) so that we can organize them a bit better. I have created this in my branch as an example: https://github.com/fedspendingtransparency/usaspending-api/tree/ds-1298-budget-function-autocomplete/usaspending_api/references

Just something to consider as we reorganize the backend

usaspending_api/references/views.py
russellbodine: huh dont know how this happened.  fixed

usaspending_api/urls.py
russellbodine: fixed!

usaspending_api/references/urls_v2.py
russellbodine: yep! done

usaspending_api/references/urls.py
russellbodine: woops!

russellbodine:
I think it will be better to do these migrations as its own story later this week.  That way we can make sure there are no weird merge conflicts.

usaspending_api/references/serializers_v2.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import serializers
+from usaspending_api.references.models import (
+    Agency, Cfda, LegalEntity, Location, ObjectClass, OfficeAgency,
+    RefProgramActivity, SubtierAgency, ToptierAgency, LegalEntityOfficers, Definition)
+from usaspending_api.common.serializers import LimitableSerializer
+
+
+class AgencyV2Serializer(serializers.Serializer):
```
nmonga91: I didn't see this being used anywhere, do we still need this file?

usaspending_api/references/views_v2.py
```diff
@@ -0,0 +1,38 @@
...
+        queryset = queryset.filter(cgac_code=toptier_agency.cgac_code)
+
+        # get the most up to date fy
+        queryset = queryset.order_by('-reporting_fiscal_year')
+        queryset = queryset.annotate(
+            fiscal_year=F('reporting_fiscal_year')
+        )
+        submission = queryset.first()
+        if submission is None:
+            return Response({'results': {}})
```
nmonga91: Here, above line 31, I would create a response dict: `response = {}` and return `Response({'results': response})`. And then later where you're setting up the agency name and fiscal year, you can just add that to the response dict: `response['agency_name'] = toptier_agency.name` and etc. and do a similar return as above.

This method allows easier expansion to the response contents as opposed to adding to the return statement on line 38

usaspending_api/references/serializers_v2.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework import serializers
+
+
+class AgencyV2Serializer(serializers.Serializer):
```
nmonga91: I didn't see this being used anywhere, do we still need this file?

usaspending_api/references/serializers_v2.py
russellbodine: nope! Removed

usaspending_api/references/views_v2.py
russellbodine: done!

usaspending_api/references/serializers_v2.py
russellbodine: nope removed!

PR #390, opened 2017-06-16
Update to recipient and award_type v2 endpoints

nmonga91:
adding check to ensure awarding_agency_id actually corresponds to a toptier agency. if not, an appropriate error message is returned. Also removed unused imports.

### DIFF
--- a/usaspending_api/awards/views.py
+++ b/usaspending_api/awards/views.py
@@ -1,18 +1,13 @@
 from collections import namedtuple
-import json
-
-from rest_framework import status
-from rest_framework.views import APIView
-from rest_framework.response import Response
 
 from usaspending_api.awards.models import Award, Transaction, Subaward
 from usaspending_api.awards.serializers import AwardSerializer, AwardTypeAwardSpendingSerializer, \
     RecipientAwardSpendingSerializer, SubawardSerializer, TransactionSerializer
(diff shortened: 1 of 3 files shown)


usaspending_api/awards/views.py
```diff
@@ -67,6 +62,9 @@ def get_queryset(self):
         if not (fiscal_year and awarding_agency_id):
             raise InvalidParameterException('Missing one or more required query parameters: fiscal_year, awarding_agency_id')
 
+        if not check_valid_toptier_agency(awarding_agency_id):
```
russellbodine: If the agency id does not exist in the db, calling check_valid_toptier_agency will throw a Null pointer error because it calls .first()

usaspending_api/awards/views.py
```diff
@@ -121,6 +119,9 @@ def get_queryset(self):
...
```
russellbodine: same error as above.  need to check if awarding_agency_id actually corresponds to an agency b4 calling .first() in check_valid_toptier_agency

usaspending_api/awards/views.py
nmonga91: `first()` returns None so this shouldn't be an issue

usaspending_api/awards/views.py
nmonga91: `first()` returns None so this shouldn't be an issue

PR #391, opened 2017-06-16
Fill empty state info

catherinedevlin:
Locations with state abbrev or state name, but not both, are the source of DS-1336.

This cleans them up during loading.  We can also run `.save()` on the existing location records to clean up the already-loaded ones.


usaspending_api/references/models.py
```diff
@@ -236,8 +237,38 @@ class Location(DataSourceTrackedModel):
...
+
+        if (self.state_code in code_to_state) or (self.state_name in state_to_code):
+            return True
+
+    def fill_missing_state_data(self):
+        """Fills in blank US state names or codes from its counterpart"""
+
+        if self.state_code and self.state_name:
+            return
+        if self.state_name_mismatch():
```
russellbodine: State mismatch cant be true since it already returned if state_code and state_name both exist

usaspending_api/references/models.py
```diff
@@ -236,8 +237,38 @@ class Location(DataSourceTrackedModel):
...
         super(Location, self).save(*args, **kwargs)
 
+    def state_name_mismatch(self):
+        "True if state_name and state_code not in agreement"
+
+        if (not self.state_name) or (not self.state_code):
+            return False
+        return (self.state_name != code_to_state.get(self.state_code))
+
+    def is_us_state(self):
```
russellbodine: this function "is_us_state" doesnt seem to be used.

catherinedevlin: Oh, good point.  That was required by the code to fill in missing countries, which I took out b/c I couldn't get a clear answer on whether we needed it.  I'll take it out of here.

usaspending_api/references/models.py
```diff
@@ -236,8 +237,38 @@ class Location(DataSourceTrackedModel):
...
+        "Is this unambiguously a US state?"
+
+        if self.state_name_mismatch():
+            return False
+
+        if (self.state_code in code_to_state) or (self.state_name in state_to_code):
+            return True
```
catherinedevlin: Yup, boolean logic in Python short-circuits - as soon as one fails, evaluation stops.

usaspending_api/references/abbreviations.py
```diff
@@ -0,0 +1,63 @@
+code_to_state = {
```
nmonga91: I came across this library: https://github.com/unitedstates/python-us

Not sure if that would be a better solution compared to having our own abbreviations file, but something look at.

catherinedevlin: Hmm... this is a tough call!

I am usually all "third-party modules FTW", and this looks like a cool one (I'll keep it in mind for the future).  On the other hand, it makes our dependencies a little more complex (esp. since it has some dependencies of its own), and I don't think the state abbreviations are going to require ongoing maintenance.

I think I'll throw this out for discussion in #datastore!

usaspending_api/references/models.py
```diff
@@ -236,8 +237,29 @@ class Location(DataSourceTrackedModel):
...
+        if (not self.state_name) or (not self.state_code):
+            return False
+        return (self.state_name != code_to_state.get(self.state_code))
+
+    def fill_missing_state_data(self):
+        """Fills in blank US state names or codes from its counterpart"""
+
+        if self.state_code and self.state_name:
+            return
+        if self.state_name_mismatch():
```
russellbodine: wont state_name_mismatch never return true, because it requires that both values are not None, and the boolean-if above states:

if self.state_code and self.state_name:
     return

catherinedevlin: Erg, you're right.  Again, this was only needed for the code filling in country, which I dropped.

PR #392, opened 2017-06-19
AgencyID MAC Name

russellbodine:


usaspending_api/accounts/models.py
```diff
@@ -19,6 +19,8 @@ class FederalAccount(models.Model):
     agency_identifier = models.TextField(db_index=True)
     main_account_code = models.TextField(db_index=True)
     account_title = account_title = models.TextField()
+    # agency_identifier + '-' + main_account_code + ' - ' + account_title
```
nmonga91: Some additional text here like "The federal_account_code field is made up of: ..." will make it clear it's describing the field below it

usaspending_api/references/reference_helpers.py
```diff
@@ -23,6 +23,14 @@ def update_federal_accounts(tas_tuple=None):
         Number of rows updated
     """
 
+    # add federal_account_code
+    queryset = FederalAccount.objects.all()
```
nmonga91: How often will this be run?

russellbodine: any time loadtas is called, very rarely

usaspending_api/accounts/models.py
russellbodine: added text

usaspending_api/references/reference_helpers.py
russellbodine: also it skips my code

PR #394, opened 2017-06-19
Load historical contracts and financial assistance

catherinedevlin:
Makes usaspending_api/etl/management/commands/load_usaspending_assistance.py and load_usaspending_contracts.py obsolete by pulling data from tables in the broker rather than from CSV dumps.  

Note that, at the moment, the target tables only exist and are (partly) populated in specific broker database instances: `published_award_financial_assistance`, for financial assistance, is in the broker sandbox, and `detached_award_procurement`, for contracts, is in broker dev.  To load each one, $DATA_BROKER_DATABASE_URL must be pointed to the corresponding instance.

... which makes using the new `load_historical` command to actually use, for now.


usaspending_api/etl/management/commands/load_historical.py
```diff
@@ -9,8 +11,17 @@
...
 
-def to_date(date_str):
-    return datetime.strptime(date_str, '%Y-%m-%d').date()
+def before_colon(txt):
+
+    return txt.split(':', 1)[0].strip()
+
+
+def preprocess_historical_d2_row(row):
+    "Change values `code: full description` to simply `code` for certain row keys"
```
nmonga91: Docstrings should be between triple quotes `"""text"""`

usaspending_api/etl/management/commands/load_historical.py
```diff
@@ -22,27 +33,43 @@ class Command(load_base.Command):
...
-        # File D1
-        sql = 'SELECT * FROM detached_award_procurement WHERE true'
+        if options['contracts'] and options['financial_assistance']:
+            raise CommandError('Default is to load both contracts and financial_assistance')
+
+        if not options['financial_assistance']:
+            procurement_data = self.broker_data(db_cursor, 'detached_award_procurement', options)
+            load_base.load_file_d1(submission_attributes, procurement_data, db_cursor)
+
+        if not options['contracts']:
```
nmonga91: So this will only load the D2 (financial assistance) data if 'contracts' _isn't_ one of the options?

catherinedevlin: That's the idea - if you don't specify `--contracts` or `--financial_assistance`, you get both.

Alternately, it could load nothing if you don't specify either.  I could go either way...

usaspending_api/etl/management/commands/load_historical.py
catherinedevlin: OK, that's fixed with the latest commit.

PR #396, opened 2017-06-21
New resave_instances utility to run .save() on existing rows

catherinedevlin:
Changes like the state name/abbreviation synching added to Location.save() for DS-1336 require running .save() on all existing records.  This `resave_instances` utility does that.


usaspending_api/common/management/commands/resave_instances.py
```diff
@@ -0,0 +1,51 @@
...
+from django.core.management import call_command
+from django.db import connection
+
+import django.db.models.base
+from django.apps import apps
+
+
+def models_with_custom_save():
+
+    result = [m for m in apps.get_models() if m.save != django.db.models.base.Model.save]
```
nmonga91: Just for my knowledge, this is where Python is able to see that the `.save` was overridden?

nmonga91:
Somewhere (either in the docstring, or the message that's printed) can we put an example of it's usage? That way a user wouldn't need to experience an error to know that an additional parameter is needed.

usaspending_api/common/management/commands/resave_instances.py
catherinedevlin: that's right!

nmonga91:
[run tests]

PR #401, opened 2017-06-23
Do not store DUNS as an integer, it's an identifier

catherinedevlin:
A DUNS with leading zeroes, when stored as an integer, changes
(like 079150498 to 79150498, and results in not finding and filling the
officer pay info.  There's no reason for it to be handled as a number
anyway, we don't do any math on it.  It's just an identifier.

This is the cause of the bug in https://federal-spending-transparency.atlassian.net/browse/DS-1378


usaspending_api/etl/management/commands/load_executive_compensation.py
```diff
@@ -24,7 +24,6 @@ def add_arguments(self, parser):
             '--duns',
             dest="duns",
             nargs='+',
-            type=int,
```
nmonga91: If type is simply removed, does it default to a string?

catherinedevlin: That's right!

PR #404, opened 2017-06-27
Load historical budget authorization numbers, then expose via endpoint

catherinedevlin:
Load historical budget data, and expose it through a /api/v2/budget_authority/agency/ endpoint with optional FREC argument.

Associated gist: https://gist.github.com/catherinedevlin/a65347b8526f44ce6d2ece5607f80902

- [ ] Gist approved by frontend ( @kevinli-work )


usaspending_api/accounts/serializers.py
```diff
@@ -1,7 +1,9 @@
 from rest_framework import serializers
 
 from usaspending_api.accounts.models import (
-    AppropriationAccountBalances, FederalAccount, TreasuryAppropriationAccount
+    AppropriationAccountBalances, FederalAccount, TreasuryAppropriationAccount)
+from usaspending_api.financial_activities.models import (
+    FinancialAccountsByProgramActivityObjectClass
```
russellbodine: why do we need this searializer

requirements/requirements.txt
```diff
@@ -37,3 +37,4 @@ django-simple-history==1.8.2
 openpyxl==2.4.7
 pandas==0.18.1
 python-dateutil==2.6.0
+xlrd==1.0.0
```
russellbodine: what is xlrd

catherinedevlin: excel reader - for reading the files, like the FREC mappings, supplied as .xlsx

usaspending_api/accounts/serializers.py
catherinedevlin: it gets used down in `TasCategorySerializer` - but hey, I see this import happens twice.  I'll fix that.

usaspending_api/etl/management/commands/load_budget_authority.py
```diff
@@ -0,0 +1,165 @@
...
+        headers = [cell.value for cell in sheet.row(3)]
+
+        FrecMap.objects.all().delete()
+
+        instances = []
+        for i in range(4, sheet.nrows):
+            row = dict(zip(headers, (cell.value for cell in sheet.row(i))))
+            instance = FrecMap(
+                agency_identifier=row['AID'],
+                main_account_code=row['MAIN'],
```
nmonga91: `main_account_code` is referred to by `MAIN` as opposed to `MAC` in `broker_rules_fr_entity.xlsx`? Just want to confirm it's mapping.

usaspending_api/etl/management/commands/load_budget_authority.py
```diff
@@ -0,0 +1,165 @@
...
+                fr_entity_code=row['FR Entity Type'])
+            instances.append(instance)
+        FrecMap.objects.bulk_create(instances)
+
+    def load_quarterly_spreadsheets(self, quarter, results):
+        """Special procedure for getting quarterly update .xls files
+
+        These are downloaded from
+        MAX Information and Reports (Executive, Legislative, and Judicial Users)
+        https://max.omb.gov/maxportal/document/SF133/Budget/FY%202017%20-%20SF%20133%20Reports%20on%20Budget%20Execution%20and%20Budgetary%20Resources.html
```
nmonga91: So for this we would have to download them manually an run the loader? This would be run at the end of every quarter?

usaspending_api/urls.py
```diff
@@ -38,6 +38,7 @@
     url(r'^api/v2/financial_balances/', include('usaspending_api.accounts.urls_financial_balances')),
     url(r'^api/v2/financial_spending/', include('usaspending_api.accounts.urls_financial_spending')),
     url(r'^api/v2/references/', include('usaspending_api.references.urls_v2')),
+    url(r'^api/v2/budget_authority/', include('usaspending_api.accounts.urls_budget_authority')),
```
nmonga91: can we put this in alphabetical order within the v2 section?

usaspending_api/etl/management/commands/load_budget_authority.py
catherinedevlin: Right, as written.  I'd be happy to make something more sophisticated but I'm not sure what that would be.

usaspending_api/urls.py
catherinedevlin: Sure, good suggestion.

usaspending_api/etl/management/commands/load_budget_authority.py
nmonga91: That works, we should document that as part of our processes that need to run periodically.

catherinedevlin: Do you know where we have such documentation?

usaspending_api/etl/management/commands/load_budget_authority.py
catherinedevlin: That's correct - broker_rules_fr_entity.xlsx (second tab, DT_END) doesn't actually have a MAC column, and Kristine specified MAIN as the correct source for main account codes in a Slack message.

usaspending_api/etl/management/commands/load_budget_authority.py
nmonga91: I don't think we have anything formal for it. We may need to add it to a README somewhere, or maybe on the wiki. But that shouldn't hold up the PR. It can happen independently.

nmonga91:
@catherinedevlin, there was an update posted on the gist by @kevinli-work with regards to sorting the results

usaspending_api/etl/management/commands/load_budget_authority.py
catherinedevlin: Started something at https://github.com/fedspendingtransparency/usaspending-api/wiki/Data-maintenance - probably that will change as I talk with Will about how he wants to manage these tasks.

catherinedevlin: Oops, never mind, there was a better place for it after all: https://github.com/fedspendingtransparency/usaspending-api/blob/dev/loading_data.md

PR #405, opened 2017-06-27
Clean up old Location records when redoing a submission

catherinedevlin:
An old bug in the submission loader left bogus Location instances
in the database.  This forces non-US Locations to specifically
seek or create locations without states, and removes Location
intances attached only to the old, deleted version of the submission.

### DIFF
--- a/usaspending_api/submissions/management/commands/rm_submission.py
+++ b/usaspending_api/submissions/management/commands/rm_submission.py
@@ -1,6 +1,8 @@
+from collections import defaultdict
 from datetime import datetime
 import logging
 import os
+
 from django.core.management.base import BaseCommand
 from django.core.exceptions import ObjectDoesNotExist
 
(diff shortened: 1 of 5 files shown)


usaspending_api/submissions/management/commands/rm_submission.py
```diff
@@ -18,6 +20,25 @@ class Command(BaseCommand):
...
+
+        Could be expanded to other models.
+        """
+
+        potential_childless = []
+        for transaction in submission.transaction_set.all():
+            potential_childless.append(transaction.award.place_of_performance)
+            potential_childless.append(transaction.award.recipient.location)
+            for subaward in transaction.award.subawards.all():
+                # potential_childless.append(subaward)
```
nmonga91: Do we want to leave this line in even though it's commented out?

usaspending_api/submissions/management/commands/rm_submission.py
```diff
@@ -18,6 +20,25 @@ class Command(BaseCommand):
...
+                potential_childless.append(subaward.place_of_performance)
+                potential_childless.append(subaward.recipient.place_of_performance)
+            # potential_childless.append(transaction.award)
```
nmonga91: Same question as above

catherinedevlin: Ended up cleaning those out.

PR #406, opened 2017-06-28
Top federal accounts by obligation

phroiland:
Federal obligations test success. Renamed federal_accounts_v2 to federal_obligations. Updated urls accordingly.

Ad hoc, Edited a couple files for lines being too far extended (PEP 8). 

https://gist.github.com/phroiland/79d905bc0dc24374fb7da91ba9825c9c


kevinli-work:
Left a comment on the gist

usaspending_api/accounts/urls_federal_obligations.py
```diff
@@ -0,0 +1,14 @@
...
+
+# bind ViewSets to URLs
+federal_list = views.FederalAccountByObligationViewSet.as_view(
+    {'get': 'list', 'post': 'list'})
+federal_detail = views.FederalAccountByObligationViewSet.as_view(
+    {'get': 'retrieve', 'post': 'retrieve'})
+
+urlpatterns = [
+    url(r'^$', federal_list),
+    url(r'(?P<pk>[0-9]+)/$', federal_detail)
```
nmonga91: We have this route here to allow the PK to be passed in (`/api/v2/federal_obligations/:id`) but I didn't notice it being used in the `FederalAccountByObligationViewSet`. Is it still necessary to publish this route? I also didn't didn't see it in the associated gist.

usaspending_api/accounts/views/federal_obligations.py
```diff
@@ -0,0 +1,47 @@
...
+            raise InvalidParameterException(
+                'Missing required query parameters: fiscal_year & funding_agency_id'
+            )
+        # Use filter() instead of get() on Agency.objects
+        # get() will likely raise an error on a bad agency id
+        # Get the top_tier_agency_id from the Agency set
+        top_tier_agency_id = Agency.objects.filter(id=funding_agency_id).first().toptier_agency_id
+        # Using final_objects below ensures that we're only pulling the latest
+        # set of financial information for each fiscal year
+        queryset = FinancialAccountsByProgramActivityObjectClass.final_objects.filter(
```
nmonga91: If we have to use `final_objects` does this mean we'll have stale data for the fiscal year in question in the database?

usaspending_api/urls.py
```diff
@@ -23,7 +23,9 @@
 
 urlpatterns = [
     url(r'^status/', views.StatusView.as_view()),
+    url(r'^status2/', views.StatusView.as_view()),
```
nmonga91: Why is this duplicated under a different endpoint here?

usaspending_api/urls.py
```diff
@@ -23,7 +23,9 @@
...
     url(r'^api/v1/federal_accounts/', include('usaspending_api.accounts.urls_federal_account')),
+    url(r'^api/v2/federal_obligations/', include('usaspending_api.accounts.urls_federal_obligations')),
```
nmonga91: If we're already fixing other things in this file, can you group all the v2 endpoints under the v1 endpoints? ideally we'd want them grouped and alphabetized within their groups.

usaspending_api/accounts/urls_federal_obligations.py
phroiland: I was referencing the previous federal_account url, but just added an id field, so maybe?

usaspending_api/accounts/views/federal_obligations.py
phroiland: I'm not sure. I thought the use of final_objects was standard. How would I notice stale data?

usaspending_api/urls.py
phroiland: I think that was just a test Russ and I did, I'll remove it.

usaspending_api/urls.py
phroiland: Agreed. Was just keeping the v1 with it's corresponding v2 for the time being. I'll fix it.

usaspending_api/accounts/views/federal_obligations.py
nmonga91: I was thinking that we could query the table in question, manually, for the fiscal year and agency specified and if we have duplicates, that would clarify that there may be stale data. So  we get back a row with everything identical, except the updated_at timestamp (i think that's what `final_objects` is checking?), it would take the latest. 

I did see it used in other places, but wasn't clear as to why it was needed.

usaspending_api/accounts/urls_federal_obligations.py
nmonga91: The primary key Kevin was requesting is part should be part of the response, so we may not actually need this specific endpoint then.

usaspending_api/accounts/views/federal_obligations.py
phroiland: So when I initially wrote the code, I was printing out and sorting by account title to look at the numbers before I accepted the sum total (which seems pretty high for some accounts). There could be some stale data, and this might need a deeper analytical dive.

usaspending_api/accounts/urls_federal_obligations.py
phroiland: and it's gone.

nmonga91:
Everything looks good. Just waiting on changes to be pushed to the branch so that the PR also gets updated.

PR #408, opened 2017-06-29
add none check

russellbodine:


usaspending_api/references/views_v2.py
```diff
@@ -62,6 +62,10 @@ def get(self, request, pk, format=None):
         total_budget_authority_queryset = total_budget_authority_queryset.filter(fiscal_year=active_fiscal_year)
 
         total_budget_authority_submission = total_budget_authority_queryset.first()
+        total_budget_authority_amount = ""
```
nmonga91: should we default to 0 or just blank? this should be coordinated with frontend so we know what to display in these kinds of situations

russellbodine: im gonna change to 0

russellbodine:
https://gist.github.com/russellbodine/09dc80f4aacf2ff988a8d01faa8baf2a

PR #433, opened 2017-07-12
PSC description

phroiland:
Built industry_codes scripts to represent screenshots in DS-1195.
There is no description field for PSC in the database, so NAICS was used instead.
PSC and NAICS have a many-to-many relationship.
PSC Codes describe “WHAT” was bought for each contract action reported.
NAICS Codes describe “HOW” purchased products and services will be used.

https://gist.github.com/phroiland/5791b961b4c624b2a2b0687adc1ed129

### DIFF
--- a/.idea/dictionaries/johnfroiland.xml
+++ b/.idea/dictionaries/johnfroiland.xml
@@ -1,0 +1,7 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="johnfroiland">
+    <words>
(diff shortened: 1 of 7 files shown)


nmonga91:
there should be an actual field for PSC description. It's in the atom feed but not in the D1 files broker is generating. Until it's part of the submission we're loading from the broker, we won't be able to consume/display that description. Closing the PR for now.

PR #441, opened 2017-07-21
DUNS Recipient

phroiland:
Updated DUNS Recipient autocomplete to seperate out parents from other recipients.

- [ ] https://gist.github.com/phroiland/5533691e65dc9eba3b61acca314c4276


nmonga91:
There are some code climate issues, may have been issues with the merge of dev into the working branch.

usaspending_api/awards/serializers_v2/serializers.py
```diff
@@ -6,11 +6,11 @@ class AwardTypeAwardSpendingSerializer(serializers.Serializer):
     obligated_amount = serializers.DecimalField(None, 2)
```
nmonga91: Can you restructure the directories in the `awards` django app to match what `references` is doing with v1 and v2?

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -136,23 +136,48 @@ def post(self, request):
...
+            'parent_recipient': [
+                parents.values('legal_entity_id',
+                               'recipient_name',
+                               'parent_recipient_unique_id')[:limit]],
+            'recipient': [
+                recipients.values('legal_entity_id',
+                                  'recipient_name',
+                                  'recipient_unique_id')[:limit]]}
+
+        response['counts'] = {'parent_recipient': len(parents),
```
nmonga91: I would double check with frontend, but if `counts` isn't actually being used, we shouldn't include it in the response.

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -136,23 +136,48 @@ def post(self, request):
...
+
+        # Search and filter for Recipients, excluding Parent Recipients
+        recipients = queryset.annotate(
+            similarity=Greatest(
+                TrigramSimilarity('recipient_name', search_text),
+                TrigramSimilarity('recipient_unique_id', search_text))
+        ).distinct().order_by('-similarity').exclude(
+            parent_recipient_unique_id__in=F('recipient_unique_id'))[:limit]
+
+        response['results'] = {
```
nmonga91: Is there a reason the response is broken into `parent_recipient` and `recipient`? For the autocomplete, with this type of response, the frontend wouldn't be able to properly assess the order of relevancy between a parent recipient and recipient. I would check in with front-end to see what kind of response they would be able to utilize do render results based on the relevancy.

phroiland: Design wanted to look more like budget function and have the parent(s) essentially bubble up to the top.

usaspending_api/references/v2/views/autocomplete.py
phroiland: It's probably not, was just a check for what I was trying to do.

usaspending_api/references/v2/views/autocomplete.py
nmonga91: Budget function is being re-done to properly display based on relevance, so the current implementation may not be the best example. I would run this by @kevinli-work to ensure it can be handled as intended because with this solution, a recipient that is a better match will show up below a parent which has a lower similarity ranking. Does that make sense?

phroiland: If someone is searching by text, then they would receive the parent along with the recipient of the same name. If the search text is for a recipient name that is different from the parent, then maybe there might be an issue. If someone is searching by DUNS, they well get that result. If the recipient DUNS is not also the Parent DUNS there will be no results for the Parent.

An alternative I considered is with _icontains. It's much faster, but would probably need to be applied using a try/else block.

nmonga91: unfortunately `icontains` doesn't help with sorting my relevancy. There is also a case with trigram where random text still generates results. That would also cause both lists to be populated, potentially. Has frontend confirmed they're able to work with this response to display them with the proper relevance order?

PR #445, opened 2017-07-24
new tests, fixed code to have float wrappers

russellbodine:
without the float(decimal) in the response, the response contained:

“key”: Decimal(“0.00”)

now its

“key”: 0.0


usaspending_api/references/v2/views/toptier_agencies.py
```diff
@@ -69,26 +69,25 @@ def get(self, request, format=None):
...
             # craft response
             response['results'].append({'agency_id': agency.id,
                                         'agency_name': toptier_agency.name,
                                         'active_fy': str(active_fiscal_year),
                                         'active_fq': str(active_fiscal_quarter),
-                                        'outlay_amount': str(aggregate_dict['outlay_amount']),
-                                        'obligated_amount': str(aggregate_dict['obligated_amount']),
-                                        'budget_authority_amount': str(aggregate_dict['budget_authority_amount']),
-                                        'current_total_budget_authority_amount': total_budget_authority_amount,
+                                        'outlay_amount': float(aggregate_dict['outlay_amount']),
```
nmonga91: These values from the `aggregate_dict` seem to be already coming back as `float`, I'm not sure if the casting here is necessary.

russellbodine: they are returning as 
```'budget_authority_amount': Decimal("0.00")```

PR #446, opened 2017-07-24
Agency Recipients

phroiland:
Updated API endpoint to allow filtering based on award category. Default is null.

- [x] https://gist.github.com/phroiland/320c341249c9221b2f758394f68263ee


usaspending_api/awards/v2/views/award_spending.py
```diff
@@ -61,39 +56,41 @@ def get_queryset(self):
...
         queryset = queryset.filter(
             fiscal_year=fiscal_year,
-            awarding_agency__toptier_agency=top_tier_agency_id
-        )
-        # alias recipient column names with top and sub tier agencies
-        queryset = queryset.annotate(
-            recipient_name=F('recipient__recipient_name'),
-            recipient_id=F('recipient__legal_entity_id')
+            awarding_agency__toptier_agency=top_tier_agency_id,
+            award__category=award_category
```
catherinedevlin: If `award_category` is not supplied, won't this filter out everything with a non-null `award__category`?  Is that the intent?

phroiland: It will return null category if the user doesn't specify one.

phroiland: Working on the test.

catherinedevlin: Okay, as long as that's the known intent.

phroiland: Unless you know of a way to populate all categories if there is no award_category specified by the user?

PR #447, opened 2017-07-25
recipient_icontains

phroiland:
Alternative autocomplete search for recipient using SearchVector and icontains.

Results using Trigram
- [ ] https://gist.github.com/phroiland/db21b19772cf017718ce7629af4a2988

Results using icontains
- [x] https://gist.github.com/phroiland/bef81a8feec5a88a234d96f1c7d0d687


usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -204,23 +204,44 @@ def post(self, request):
...
-
-        return Response({'results': list(queryset.values(*column_names)[:limit])})
+        # Filter based on search text
+        response = {}
+        # Search and filter for Parent Recipients
+        try:
+            parents = queryset.annotate(
+                search=SearchVector('recipient_name', 'parent_recipient_unique_id')
+            ).filter(
+                search__icontains=search_text,
```
nmonga91: Are there any performance implications using `icontains` that we need to be aware of?

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -204,23 +204,44 @@ def post(self, request):
...
-        if exact_match_queryset.count() > 0:
-            queryset = exact_match_queryset
-
-        column_names = ['legal_entity_id', 'recipient_name', 'recipient_unique_id']
-
-        return Response({'results': list(queryset.values(*column_names)[:limit])})
+        # Filter based on search text
+        response = {}
+        # Search and filter for Parent Recipients
+        try:
```
nmonga91: What situations would cause an exception to be raised here? If there aren't any parents DUNS found, wouldn't `parents` just be an empty queryset?

usaspending_api/references/v2/views/autocomplete.py
phroiland: It's faster.

phroiland: I posted the gists with sample results

usaspending_api/references/v2/views/autocomplete.py
phroiland: Since there really aren't any exceptions to be raised, it's passed. So yes, it would just be empty. I just updated the gist to show what it would look like for a search on a recipient DUNS.

nmonga91: Ok, in that case, is the try/except block still necessary? Can that be replaced with a check like `if parents is None`?

phroiland: Just saw this. Let me try it out...

phroiland: I really don't think we need either. It will either return results for the user entered search_text or it won't. Search can be done with text, recipient_unique_id, or parent_recipient_unique_id. 

So a user can search on a parent id and get both parent and the corresponding recipient results--if they exist, but will not receive a parent if searching on a specific recipient id only.

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,8 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
```
nmonga91: Probably don't want this file in here

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -204,23 +204,63 @@ def post(self, request):
...
+        ).filter(
+            search__icontains=search_text).exclude(
+            parent_recipient_unique_id__in=F('recipient_unique_id'))[:limit]
+
+        # Counts to check for more than 10 total results
+        p_limit = parents.count()
+        r_limit = recipients.count()
+        counts = p_limit + r_limit
+        # If both counts are equal to the limit, distribute results evenly.
+        if p_limit and r_limit == limit:
```
nmonga91: Should be `if p_limit == limit and r_limit == limit:`. As is, if `p_limit` is < 10 but `r_limit` is 10, it'll still evaluate to true.

nmonga91: Also, i think we want to have the limit checking here be against the `limit` request parameter

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -204,23 +204,63 @@ def post(self, request):
...
+            parents = parents[:5]
+            recipients = recipients[:5]
+        # If counts are 10 or less, pass.
+        elif counts < 11:
+            pass
+        # If the total counts are greater than 10, set new limits.
+        elif counts > 10:
+            # If both limits are greater than 5, set limits to 5.
+            if p_limit and r_limit > 5:
```
nmonga91: This could be combined with the top-level check to be something like: `if p_limit >= limit and r_limit >= limit:`, since they both result in capping each type to half of `limit`

PR #453, opened 2017-07-28
recipient_autocomplete_trigram

phroiland:
Inadvertently created and submitted a pull request on a branch that already exists.

Reverted back to TriGramSimilarity. Able to get complete results with few characters and multiple string entries. Excess brackets removed. Tests complete.

Added exact_match calculation to improve results. Limit filter applied to results. parent_recipient_unique_id, and recipient_unique_id included in both result fields for TriGramSimilarity search

@kevinli-work 
- [x] https://gist.github.com/phroiland/bef81a8feec5a88a234d96f1c7d0d687

### DIFF
--- a/usaspending_api/references/v2/views/autocomplete.py
+++ b/usaspending_api/references/v2/views/autocomplete.py
@@ -219,28 +219,38 @@ class RecipientAutocompleteViewSet(BaseAutocompleteViewSet):
         parents = queryset.annotate(
             similarity=Greatest(
                 TrigramSimilarity('recipient_name', search_text),
-                TrigramSimilarity('parent_recipient_unique_id', search_text))
+                TrigramSimilarity('parent_recipient_unique_id', search_text),
+                TrigramSimilarity('recipient_unique_id', search_text))
         ).distinct().order_by('-similarity').filter(
-            parent_recipient_unique_id__in=F('recipient_unique_id'))[:limit]
+            parent_recipient_unique_id__in=F('recipient_unique_id'))
 
         # Search and filter for Recipients, excluding Parent Recipients
         recipients = queryset.annotate(
             similarity=Greatest(
                 TrigramSimilarity('recipient_name', search_text),
+                TrigramSimilarity('recipient_unique_id', search_text),
                 TrigramSimilarity('parent_recipient_unique_id', search_text))
         ).distinct().order_by('-similarity').exclude(
-            parent_recipient_unique_id__in=F('recipient_unique_id'))[:limit]
+            parent_recipient_unique_id__in=F('recipient_unique_id'))
+
+        parents_exact_match_queryset = parents.filter(similarity=1.0)
+        if parents_exact_match_queryset.count() > 0:
+            parents = parents_exact_match_queryset
+
+        recipients_exact_match_queryset = recipients.filter(similarity=1.0)
+        if recipients_exact_match_queryset.count() > 0:
+            recipients = recipients_exact_match_queryset
 
         # Format response
         response['results'] = {
             'parent_recipient':
                 parents.values('legal_entity_id',
                                'recipient_name',
-                               'parent_recipient_unique_id'),
+                               'parent_recipient_unique_id')[:limit],
             'recipient':
                 recipients.values('legal_entity_id',
                                   'recipient_name',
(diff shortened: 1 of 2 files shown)


usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -207,36 +207,40 @@ def post(self, request):
...
             parent_recipient_unique_id__in=F('recipient_unique_id'))[:limit]
+
         # Search and filter for Recipients, excluding Parent Recipients
         recipients = queryset.annotate(
-            search=SearchVector('recipient_name', 'recipient_unique_id')
-        ).filter(
-            search__icontains=search_text).exclude(
+            similarity=Greatest(
+                TrigramSimilarity('recipient_name', search_text),
+                TrigramSimilarity('parent_recipient_unique_id', search_text))
```
nmonga91: This should be doing the similarity on `recipient_unique_id` and not `parent_recipient_unique_id`, right?

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -207,36 +207,40 @@ def post(self, request):
...
-        ).filter(
-            search__icontains=search_text,
+            similarity=Greatest(
+                TrigramSimilarity('recipient_name', search_text),
+                TrigramSimilarity('parent_recipient_unique_id', search_text))
+        ).distinct().order_by('-similarity').filter(
             parent_recipient_unique_id__in=F('recipient_unique_id'))[:limit]
+
         # Search and filter for Recipients, excluding Parent Recipients
         recipients = queryset.annotate(
```
nmonga91: For recipients, how is exact matches handled here? Was something like https://github.com/fedspendingtransparency/usaspending-api/blob/85a84d70cddb6d7e808d26874a19f083a1122724/usaspending_api/references/v2/views/autocomplete.py#L50 not necessary?

usaspending_api/references/v2/views/autocomplete.py
phroiland: I don't know what happened, but I made those changes. I'll look into it.

usaspending_api/references/v2/views/autocomplete.py
phroiland: Same answer as above. Not sure why those changes aren't there.

phroiland: I really don't know what happened, because my test file is updated to reflect those changes. Anyway. Fixed.

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -207,36 +207,50 @@ def post(self, request):
...
-            search=SearchVector('recipient_name', 'recipient_unique_id')
-        ).filter(
-            search__icontains=search_text).exclude(
-            parent_recipient_unique_id__in=F('recipient_unique_id'))[:limit]
+            similarity=Greatest(
+                TrigramSimilarity('recipient_name', search_text),
+                TrigramSimilarity('recipient_unique_id', search_text),
+                TrigramSimilarity('parent_recipient_unique_id', search_text))
```
nmonga91: If we're looking for _just_ recipients here, is the parent DUNS column necessary for the trigram lookup?

phroiland: Wasn't sure if the algorithm results are better with more related search parameters or not.

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -207,36 +207,49 @@ def post(self, request):
...
         # Search and filter for Parent Recipients
         parents = queryset.annotate(
-            search=SearchVector('recipient_name', 'parent_recipient_unique_id')
-        ).filter(
-            search__icontains=search_text,
-            parent_recipient_unique_id__in=F('recipient_unique_id'))[:limit]
+            similarity=Greatest(
+                TrigramSimilarity('recipient_name', search_text),
+                TrigramSimilarity('parent_recipient_unique_id', search_text),
+                TrigramSimilarity('recipient_unique_id', search_text))
```
nmonga91: Is it necessary to have the recipient DUNS part of the search when we're only concerned with the parent DUNS?

PR #461, opened 2017-08-04
Broker historical models

catherinedevlin:
Models to access historical-data tables on the broker database.

Use the `.using` method when querying these models, like:

`PublishedAwardFinancialAssistance.objects.using('data_broker').all()`

The views, serializers, and endpoints are made as simple as possible, just to get it out there.  I presume we will want to do some renaming of endpoints, models, the works.  But, for now:

http://localhost:8000/api/v2/broker/publishedawardfinancialassistance/

http://localhost:8000/api/v2/broker/detachedawardprocurementviewset/

Note that at the moment the `detached_award_procurement` table on the broker is empty.


usaspending_api/broker/v2/urls.py
```diff
@@ -0,0 +1,17 @@
...
+broker_router = DefaultRouter()
+broker_router.register('broker', views.DetachedAwardProcurementViewSet)
+broker_router.register('broker', views.PublishedAwardFinancialAssistanceViewSet)
+
+mode_list = {'get': 'list', 'post': 'list'}
+mode_detail = {'get': 'retrieve', 'post': 'retrieve'}
+
+
+urlpatterns = [
+    url(r'^detachedawardprocurementviewset/$', views.DetachedAwardProcurementViewSet.as_view({'get': 'list'})),
```
nmonga91: Do we need these endpoints published?

catherinedevlin: Probably not - I wanted to throw something up there for visibility, but we can drop views.py, serializer.py, and both urls.py files from this branch and just have the models, settings.py, and migrations.

PR #473, opened 2017-08-10
Removed request catalog

catherinedevlin:
I haven't yet been able to confirm that the csv downloads still work - somebody will have to try that out, either locally or in the sandbox.  (You'll need to get S3 credentials set up on your machine to try it locally).

Aside from that, however, this seems to be working.

- [ ] Tested on sandbox with successful results


catherinedevlin:
I'm not clear on why this shows so many changes, so I'm closing this to open a new PR with just the changes I actually want.  It's easier than actually figuring out what git did here.

PR #497, opened 2017-08-17
generic DUNS recipient

phroiland:
Updated loads for generic duns. Tests passed. Major change was the removal of get_or_create_by_duns. It would create new entries for null or '' DUNS regardless of the recipient.


usaspending_api/etl/subaward_etl.py
```diff
@@ -175,17 +183,21 @@ def load_subawards(submission_attributes, db_cursor):
...
 
-        # Find the recipient by looking up by duns
-        recipient, created = LegalEntity.get_or_create_by_duns(duns=row['duns'])
+        recipient_name = row['awardee_name']
+        if recipient_name is None:
+            recipient_name = row['awardee_or_recipient_legal']
+        if recipient_name is None:
+            recipient_name = row['company_name']
+        if recipient_name is None:
+            recipient_name = ""
```
russellbodine: I need to look back here

usaspending_api/etl/subaward_etl.py
```diff
@@ -175,17 +183,21 @@ def load_subawards(submission_attributes, db_cursor):
 
         award_ids_to_update.add(award.id)
 
-        # Find the recipient by looking up by duns
-        recipient, created = LegalEntity.get_or_create_by_duns(duns=row['duns'])
+        recipient_name = row['awardee_name']
+        if recipient_name is None:
+            recipient_name = row['awardee_or_recipient_legal']
+        if recipient_name is None:
+            recipient_name = row['company_name']
```
russellbodine: setting recipient_name =  company_name is not in the original code, although it kinda makes sense.  Im worried we are not legally allowed to bind the legal entity to the company name because I know we are not legally allowed to bind a legal entity to its parent legal entity.

PR #514, opened 2017-08-27
Rename GWA_TAS_NAME column to GWA_TAS NAME

catherinedevlin:
This matches the column name that load_tas looks for.

Our code assumes that the `tas_list.csv` stores account titles in a column named “GWA_TAS NAME”, which used to be true, but in our latest updates of TAS, that column’s new name is “GWA_TAS_NAME” (note the first has one underscore, the second has two).  This undoes that accidental change.

The real problem is that the threaded data loader just lets this pass silently and mysteriously.  I want to think next week about how to fix that.


willkjackson:
Since this file came from CARS, we are not going to modify it. We have already updated lookups to reflect this change. If it changes again, we'll switch it back.

PR #515, opened 2017-08-29
Feature spending by time visualization

dpb-bah:
- [x] Reviewed
- [x] Gists reviewed and approved by front-end ( @mab6bf )

Gist:
  -[This Endpoint](https://gist.github.com/russellbodine/b3d90268b03f3cc067101d8ba4290eeb)
  -[Visualizations](https://github.com/fedspendingtransparency/usaspending-website/wiki/Award-Search-Visualizations)
  -[V2 Filters](https://gist.github.com/nmonga91/ba0e172b6d3f2aaf50f0ef1bb5d708bc#agencies-awardingfunding-agency)

Acceptance Criteria:
  - Apply v2 filters on search/spending_by_time


usaspending_api/awards/v2/filters/award.py
```diff
@@ -0,0 +1,204 @@
...
+        # extent_competed_type_codes - DONE
+        elif key == "extent_competed_type_codes":
+            or_queryset = Award.objects.none()
+            for v in value:
+                or_queryset |= Award.fitler(
+                    transaction__contract_data__extent_competed=v)
+            queryset &= or_queryset
+
+        else:
+            raise InvalidParameterException('Invalid filter: ' + key + ' does not exist.')
```
nmonga91: I think we can make this check a bit more efficient. Right now it goes through and checks if it matches any of the previous keys, if not _then_ raise the exception. We can do something like `if key not in list_of_valid_keys` and raise the exception earlier. Just a thought on performance. Not a major impact but if more and more filter options are added, it can take time to error out on a faulty key.

usaspending_api/common/helpers.py
```diff
@@ -28,6 +28,27 @@ def get_params_from_req_or_request(request=None, req=None):
     return params
 
 
+def generate_fiscal_year(context):
```
nmonga91: Can we alter the parameter names for this function and the 2 below? if it's a date object coming in, it should be named appropriately.

nmonga91:
What are the plans for long-term maintenance as filters get added/removed/updated? Right now we're manually checking for a specific filter, but this can be definitely a bit more dynamic (but don't want it too abstracted).

nmonga91:
Can we also remove any unused code that has been commented out? if it's being kept in, please add a comment as to why it's being kept.

usaspending_api/awards/v2/filters/award.py
russellbodine: While it would look nicer, that would only slow down the code.
for success cases (key is in list_of_valid_keys): it would slow everything down by~ list_of_valid_keys/2 because it would iterate through the list_of_valid_keys and then the if, elif statements

for failure cases it would offer no speedup (it would simply go through the elifs and fail which takes exactly the same amount of time as iterating through list_of_valid_keys and failing)

usaspending_api/common/helpers.py
russellbodine: can do

usaspending_api/awards/v2/filters/award.py
nmonga91: Lookups in a dictionary are O(1) and would be the fastest for detecting a failing key

russellbodine: ah list_of_valid_keys is actually dict_of_valid_keys.  Thats fine then

russellbodine:
@nmonga91 For adding a new filter, it is possible to add dynamic filters, but it is a little too crazy for MVP.

We are removing unused code now!

nmonga91:
@russellbodine can you add a README for this linking to the V2 gist and how to move towards dynamic filtering so that planning can account for the enhancements in the future?

usaspending_api/awards/v2/filters/transaction.py
```diff
@@ -0,0 +1,286 @@
...
+                    'award_amounts',
+                    'award_ids',
+                    'program_numbers',
+                    'naics_codes',
+                    'psc_codes',
+                    'contract_pricing_type_codes',
+                    'set_aside_type_codes',
+                    'extent_competed_type_codes']
+
+        if key not in key_list:
```
nmonga91: This one is checking against a key list while `awards` is not. Can we make them consistent with each other?

russellbodine: awards is actually completely overhauled in spending_by_geography.

nmonga91: Isn't `spending_by_geography` also based on Transactions also?

nmonga91:
Is there a backend gist created specific to this endpoint? similar to how we've done others?

PR #520, opened 2017-08-30
WIP: Positive obligated amounts, DS-1601

catherinedevlin:
WORK IN PROGRESS - hold off merging while I ask for clarification

Should fulfill backend part of DS-1601, https://federal-spending-transparency.atlassian.net/browse/DS-1601

The v2/references/agency/${id}/
endpoint returns an obligated amount that is outflow only 
(does not include negative obligations)


catherinedevlin:
Turns out we were wrong about the requirements!

PR #523, opened 2017-09-05
Feature unknown object class ds 1602

catherinedevlin:
When the major object class is 00, the v2/references/agency endpoint reports the major object class name as "Unknown Object Type", regardless of what's in the database.

See DS-1602: https://federal-spending-transparency.atlassian.net/browse/DS-1602


nmonga91:
The endpoint this primarily affects is `/api/v2/financial_spending/major_object_class?fiscal_year=2017&funding_agency_id=163` instead of `/api/v2/agency/:id`, is that right?

nmonga91:
Tested against `/api/v2/financial_spending/major_object_class?fiscal_year=2017&funding_agency_id=163` and everything looks good.

PR #526, opened 2017-09-06
Feature spending by geography

dpb-bah:
- [x] Reviewed
- [x] Gists reviewed and approved by front-end 

Gist:
  -[This Endpoint](https://gist.github.com/russellbodine/2b3e00d8ec1223cfeb91e73e9b7cc1fd)
  -[Visualizations](https://github.com/fedspendingtransparency/usaspending-website/wiki/Award-Search-Visualizations)
  -[V2 Filters](https://gist.github.com/nmonga91/ba0e172b6d3f2aaf50f0ef1bb5d708bc#agencies-awardingfunding-agency)

Acceptance Criteria:
  - Apply v2 filters on search/spending_by_geography


usaspending_api/awards/v2/filters/award.py
```diff
@@ -1,204 +1,204 @@
...
+                    or_queryset = Award.objects.filter(recipient__business_types_description=v)
+            if or_queryset is not None:
+                queryset &= or_queryset
 
-        # place_of_performance_scope (broken till data reload)- DONE
+        # place_of_performance_scope (broken till data reload
         elif key == "place_of_performance_scope":
             if value == "domestic":
-                queryset = queryset.Award.filter(place_of_performance__country_name="UNITED STATES")
+                queryset = queryset.filter(place_of_performance__country_name="UNITED STATES")
```
nmonga91: This may need to have an `iexact`, if the format for the country name is anything besides `UNITED STATES` exactly, we may be overlooking some elements. For example, what would happen if the country name is lower cased? Is this something that is auto-populated by us during loads so we can control what it's set to?

usaspending_api/search/v2/views/search.py
```diff
@@ -344,3 +346,60 @@ def post(self, request):
...
+
+class SpendingByGeographyVisualizationViewSet(APIView):
+
+    def post(self, request):
+        """Return all budget function/subfunction titles matching the provided search text"""
+        json_request = request.data
+        scope = json_request.get('scope', None)
+        filters = json_request.get('filters', None)
+
+        if scope is None:
```
nmonga91: This should be combined into a single if statement as such: `if not scope or not filter` which can be further simplified to `if not (scope and filter)`.

usaspending_api/search/v2/views/search.py
```diff
@@ -344,3 +346,60 @@ def post(self, request):
...
+            raise InvalidParameterException('Missing one or more required request parameters: filters')
+        potential_scopes = ["recipient_location", "place_of_performance"]
+        if scope not in potential_scopes:
+            raise InvalidParameterException('scope does not have a valid value')
+
+        # build sql query filters
+        queryset = transaction_filter(filters)
+
+        # define what values are needed in the sql query
+        # queryset = queryset.values('action_date', 'federal_action_obligation')
```
nmonga91: let's remove this if it's not being used.

dpb-bah: This will be commented back in for the speedup branch.

usaspending_api/search/v2/views/search.py
nmonga91: Decided we're leaving it as is to be specific for error handling

usaspending_api/awards/v2/filters/award.py
nmonga91: Awesome

PR #533, opened 2017-09-07
Hotfix pagination metadata v2 search

dpb-bah:
- [x] Updated gists below
- [x] Reviewed
- [x] Gists reviewed by @mab6bf  

- Including `page_metadata` object to pagination requests
- Sorting `spending-by-category` results by descending `aggregated_amount`

Updated Gists:
- [Spending by Category](https://gist.github.com/dpb-bah/7f12df270210d7a830c98af71cfa580d)
- [Spending by Award Type](https://gist.github.com/dpb-bah/6cf34627ae6d221fafab20b05fca7d58)

### DIFF
--- a/usaspending_api/common/helpers.py
+++ b/usaspending_api/common/helpers.py
@@ -55,20 +55,20 @@ def generate_fiscal_month(date):
 def get_pagination(results, limit, page, benchmarks=False):
     if benchmarks:
         start_pagination = time.time()
-    next_previous_metadata = {"next": None, "previous": None, "hasNext": False, "hasPrevious": False}
+    page_metadata = {"next": None, "previous": None, "hasNext": False, "hasPrevious": False}
     if limit < 1 or page < 1:
-        return [], next_previous_metadata
-    next_previous_metadata["hasNext"] = (limit*page < len(results))
-    next_previous_metadata["hasPrevious"] = (page > 1 and limit*(page-2) < len(results))
-    if not next_previous_metadata["hasNext"]:
+        return [], page_metadata
+    page_metadata["hasNext"] = (limit*page < len(results))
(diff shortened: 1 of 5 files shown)


usaspending_api/common/helpers.py
```diff
@@ -55,13 +55,20 @@ def generate_fiscal_month(date):
 def get_pagination(results, limit, page, benchmarks=False):
```
nmonga91: Here we're bypassing the default django pagination we've previously utilized (`usaspending-api/common/pagination.py`). Is that primarily due to what the default pagination operates on? Default I believe operates on the queryset where as this operates on the serialized results (the JSON formatted resultset)?

dpb-bah: Yes, that's correct. `usaspending-api/common/pagination.py`'s code is hardcoded for querysets whereas this function is designed for any python list.

usaspending_api/search/v2/views/search.py
```diff
@@ -169,9 +169,10 @@ def post(self, request):
             for key, value in name_dict.items():
                 results.append({"agency_name": key, "agency_abbreviation": value["abbreviation"],
                                 "aggregated_amount": value["aggregated_amount"]})
-
-            results = get_pagination(results, limit, page)
-            response = {'category': category, 'scope': scope, 'limit': limit, 'page': page, 'results': results}
+            results = sorted(results, key=lambda result: result["aggregated_amount"], reverse=True)
+            results, page_metadata = get_pagination(results, limit, page)
+            response = {'category': category, 'scope': scope, 'limit': limit, 'results': results,
+                        "page_metadata": page_metadata}
```
nmonga91: We should keep the quoting consistent here as well as other occurrences

PR #534, opened 2017-09-07
Feature object class filter

shinson:
Story ID: DS-1591

Updating federal account, object class  endpoint to retrieve minor object classes rolled up under major classes filtered by federal account.

Updated object class endpoint test

Gist link: https://gist.github.com/shinson/9638d5a2c8fa40dd5949d0e29d3d6e9a

- [x] Reviewed
- [x] Gists reviewed and approved by front-end


usaspending_api/accounts/urls_federal_accounts_v2.py
```diff
@@ -6,5 +6,5 @@
 object_class_federal_accounts = views.ObjectClassFederalAccountsViewSet.as_view()
 
 urlpatterns = [
-    url(r'^object_class/(?P<pk>[0-9]+)/$', object_class_federal_accounts)
+    url(r'^available_object_classes/(?P<pk>[0-9]+)$', object_class_federal_accounts)
```
nmonga91: After looking at the gist, is it possible to alter the endpoint to be something like `/api/v2/federal_account/:federal_account_id/available_object_classes/`? This clarifies a bit with what's going on: We're getting a specific federal account by the ID and then getting the available object classes for that federal account.

usaspending_api/accounts/views/federal_accounts_v2.py
```diff
@@ -26,24 +26,29 @@ def get(self, request, pk, format=None):
...
-                oc_list.append(financial_account.object_class)
-
-        # remove non-uniques
-        oc_set = set(oc_list)
-
-        # structure response
-        major_classes = set(oc.major_object_class_name for oc in oc_set)
+        # get fin based on tas, select oc, make distict values
+        # add conditions since '__in' will not work with non-querysets
+        financial_account_queryset = \
```
nmonga91: Here, to potentially be able to use `__in`, you could try updating `tas_queryset` to look something like this: `tas_queryset = TreasuryAppropriationAccount.objects.filter(federal_account=fa).values_list('treasury_account_identifier', flat=True).distinct()` (I don't think the `.distinct()` is necessary though).

This will return a list of IDs for the matching entries in `TreasuryAppropriationAccount` and that should work with `__in` regardless of how many come back.

PR #541, opened 2017-09-12
/download/transactions and /download/awards endpoints

catherinedevlin:
- [x] Gist verified: https://gist.github.com/catherinedevlin/eed19de0d9193cfff1f17f2f85d73bd9
- [x] Deploy process has been updated to accommodate new feature ( @willkjackson )


.gitignore
```diff
@@ -7,6 +7,7 @@
 
 log/*
 csv_downloads/*
+tmp/*
```
shinson: Should add zipfiles  under /downloads to the .gitignore so when people in development are testing download it gets tracked in git as a new file

catherinedevlin: Thanks, good suggestion.  Actually, I notice that `csv_downloads` was already in .gitignore and earlier code was saving csvs to there, which seems like a better name, so I just changed this code to save there.

ejgullo:
gist verified!

PR #545, opened 2017-09-14
.get fix that was appearing on dev but not locally

russellbodine:


nmonga91:
Not necessary. Needed a frontend fix.

PR #552, opened 2017-09-22
Hotfix: Adding lookup to a constant value for total budget authority

nmonga91:


usaspending_api/references/constants.py
```diff
@@ -0,0 +1,2 @@
+TOTAL_BUDGET_AUTHORITY = 8361447130497.72
+TOTAL_OBLIGATIONS_INCURRED = 4690484214947.31
```
shinson: `TOTAL_OBLIGATIONS_INCURRED` added but not used in the application.

nmonga91: Yup, added it in case we have issues with any calculations around that. It was just a preemptive attempt to solve a problem that hasn't happened yet.

PR #554, opened 2017-09-23
Historical data load and integration

nmonga91:
Added loader for loading historical data from the Data Broker as well as code updates to integrate with new models to surface the historical data.


usaspending_api/broker/management/commands/update_transactions.py
```diff
@@ -0,0 +1,525 @@
...
+        query += ' ORDER BY published_award_financial_assistance_id LIMIT %s OFFSET %s'
+        arguments += [limit, (page-1)*limit]
+
+        logger.info("Executing query on Broker DB => " + query % (arguments[0], arguments[1],
+                                                                  arguments[2], arguments[3]))
+
+        db_cursor.execute(query, arguments)
+
+        logger.info("Running dictfetchall on db_cursor")
+        award_financial_assistance_data = dictfetchall(db_cursor)
```
catherinedevlin: Our dictfetchall method pulls the entire result set into memory up-front - another possible performance issue, just FYI...

usaspending_api/broker/management/commands/update_transactions.py
```diff
@@ -0,0 +1,525 @@
...
+# Lists to store for update_awards and update_contract_awards
+AWARD_UPDATE_ID_LIST = []
+AWARD_CONTRACT_UPDATE_ID_LIST = []
+
+subtier_agency_map = {subtier_agency['subtier_code']: subtier_agency['subtier_agency_id'] for subtier_agency in SubtierAgency.objects.values('subtier_code', 'subtier_agency_id')}
+subtier_to_agency_map = {agency['subtier_agency_id']: {'agency_id': agency['id'], 'toptier_agency_id': agency['toptier_agency_id']} for agency in Agency.objects.values('id', 'toptier_agency_id', 'subtier_agency_id')}
+toptier_agency_map = {toptier_agency['toptier_agency_id']: toptier_agency['cgac_code'] for toptier_agency in ToptierAgency.objects.values('toptier_agency_id', 'cgac_code')}
+
+class Command(BaseCommand):
+    help = "Checks what TASs are in Broker but not in Data Store"
```
catherinedevlin: This help was copy/pasted from a different command, right?

usaspending_api/broker/management/commands/update_transactions.py
```diff
@@ -0,0 +1,525 @@
...
+
+    @staticmethod
+    def update_transaction_assistance(db_cursor, fiscal_year=None, page=1, limit=500000):
```
catherinedevlin: It looks like this is mostly just copied over from our current submission loader, right?

usaspending_api/broker/management/commands/update_transactions.py
```diff
@@ -0,0 +1,525 @@
...
+                    TransactionFABS(),  # thrown away
+                    row,
+                    as_dict=True)
+
+                transaction_assistance = TransactionFABS(transaction=transaction, **financial_assistance_data)
+                # catch exception and do nothing if we see
+                # "django.db.utils.IntegrityError: duplicate key value violates unique constraint"
+                try:
+                    transaction_assistance.save()
+                except IntegrityError:
```
catherinedevlin: In this case, should the existing row be updated?

usaspending_api/broker/management/commands/update_transactions.py
```diff
@@ -0,0 +1,525 @@
...
+    def handle(self, *args, **options):
+        logger.info('Starting historical data load...')
+
+        db_cursor = connections['data_broker'].cursor()
+        fiscal_year = options.get('fiscal_year')
+        page = options.get('page')
+        limit = options.get('limit')
+
+        if fiscal_year:
+            fiscal_year = fiscal_year[0]
```
catherinedevlin: same question as above about silently discarding >1 argument

usaspending_api/etl/helpers.py
```diff
@@ -204,20 +204,34 @@ def update_model_description_fields():
...
+
         # Loop through each of the models fields to construct a case for each
         # applicable field
         for field in model_fields:
             # We're looking for field names ending in _description
             split_name = field.split("_")
 
             # If the last element in our split name isn't description, skip it
-            if len(split_name) == 1 or split_name[-1] != "description":
+            if len(split_name) == 1 or split_name[-1] != "description"[:len(split_name[-1])]:
```
catherinedevlin: It took me a bit to figure out what this was doing.  Maybe update the comment to say "If the last element in our split name isn't description or an abbreviation for description"?

Also, is there any risk that we'll accidentally pick up something that isn't meant to be a description?

```
field = 'german_language_field_de'

split_name = field.split("_")

split_name[-1] == "description"[:len(split_name[-1])]```

File: usaspending_api/etl/management/commands/load_fpds_csv.py [hunk 7]
```diff
@@ -0,0 +1,108 @@
...
+# Lists to store for update_awards and update_contract_awards
+AWARD_UPDATE_ID_LIST = []
+
+awards_cache = caches['awards']
+logger = logging.getLogger('console')
+
+
+class Command(BaseCommand):
+    """
+    This command will load a single submission from the DATA Act broker. If
```
catherinedevlin: The "single submission" language here is out of date, right?

usaspending_api/etl/management/commands/load_fpds_csv.py
```diff
@@ -0,0 +1,108 @@
...
+        fiscal_year = options.get('fiscal_year')
+
+        if fiscal_year:
+            fiscal_year = fiscal_year[0]
+            logger.info('Processing data for Fiscal Year ' + str(fiscal_year))
+        else:
+            fiscal_year = 2017
+
+        def signal_handler(signal, frame):
+            transaction.set_rollback(True)
```
catherinedevlin: Should this be `rollback` instead of `set_rollback`?  I could be wrong, not sure what's going on here.

usaspending_api/etl/management/commands/load_fpds_csv.py
```diff
@@ -0,0 +1,108 @@
...
+            "base_and_all_options_value": 2
+        }
+        column_header_mapping_ordered = OrderedDict(sorted(column_header_mapping.items(), key=lambda c: c[1]))
+
+        logger.info('Getting FPDS data from CSV...')
+
+        fpds_data = pd.read_csv(fpds_csv, usecols=column_header_mapping_ordered.values(), names=column_header_mapping_ordered.keys(), header=None)
+
+        logger.info('Processing data for Fiscal Year ' + str(fiscal_year))
+        models = {transaction_fpds.detached_award_proc_unique: transaction_fpds for transaction_fpds in
```
catherinedevlin: `models` seems like a funny name for this variable, since it's for instances, right?

usaspending_api/broker/management/commands/update_transactions.py
```diff
@@ -0,0 +1,525 @@
...
+
+        fy_begin = '10/01/' + str(fiscal_year - 1)
+        fy_end = '09/30/' + str(fiscal_year)
+
+        if fiscal_year:
+            if arguments:
+                query += " AND"
+            else:
+                query += " WHERE"
+            query += ' action_date::Date BETWEEN %s AND %s'
```
catherinedevlin: Don't know if performance is an issue here, but all this in-query conversion of dates to strings is going to hurt it.

usaspending_api/broker/management/commands/update_awards.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        logger.info('Starting updates to award data...')
+
+        all_records_flag = options.get('all')
+        fiscal_year = options.get('fiscal_year')
+
+        AWARD_UPDATE_ID_LIST = []
```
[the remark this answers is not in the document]
willkjackson: Updated.

usaspending_api/broker/management/commands/update_transactions.py
willkjackson: This works faster than using the FY(Date) call, we were getting 15+ minutes for this query to come back before making this change.

usaspending_api/broker/management/commands/update_transactions.py
willkjackson: We account for this on our instances, limiting the results will ease the pain.

usaspending_api/etl/management/commands/load_fpds_csv.py
willkjackson: set_rollback is the correct syntax here. the function will kick out, check that this is set, and roll it back.

PR #561, opened 2017-09-26
Hotfix: Dev stable -> dev

nmonga91:


usaspending_api/accounts/views/financial_balances.py
```diff
@@ -24,16 +25,30 @@ def get_queryset(self, pk=None):
         if not (fiscal_year and funding_agency_id):
             raise InvalidParameterException('Missing one or more required query parameters: fiscal_year, funding_agency_id')
 
+        toptier_agency = Agency.objects.filter(id=funding_agency_id).first().toptier_agency
```
russellbodine: if the first returns none, this errors out

russellbodine: looks like the code was already like that though.

PR #572, opened 2017-09-28
Import CommandError

catherinedevlin:
Noticed this little nuisance: when the command tried to throw a CommandError,
it had not been imported, so threw NameError instead


catherinedevlin:
On second thought, this code may have no future, let's not bother

PR #586, opened 2017-10-02
Hotfix award spending fiscal year fix

russellbodine:


usaspending_api/common/helpers.py
```diff
@@ -74,10 +74,17 @@ def generate_last_completed_fiscal_quarter(fiscal_year):
             fiscal_quarter = current_fiscal_quarter - 1  # can also do: current_fiscal_date.prev_quarter.quarter
             fiscal_date = FiscalQuarter(fiscal_year, fiscal_quarter).end
     elif requested_fiscal_year.fiscal_year < current_fiscal_date.fiscal_year:
-        # If the retrieving a previous FY, get the date for the end of the final quarter
-        fiscal_quarter = requested_fiscal_year.end.quarter
-        fiscal_date = requested_fiscal_year.end
+        # If the retrieving a previous FY, get the last quarter data UNLESS its the most recent quarter, then account
+        # for data loading.
+        if (requested_fiscal_year.fiscal_year == current_fiscal_date.fiscal_year - 1) and \
```
nmonga91: This block doesn't account for altering the `fiscal_date` based on the quarter change. It also doesn't seem necessary to have the first part of the conditional, the latter part should suffice in making sure to offset correctly. If FY 16 is requested, it's > 45 days from the current fiscal date so it should go into the else.

Also, the 2nd part of the conditional should be looking at the `requested_fiscal_year.end` instead of `.start`. Something like this:
```
        if (current_fiscal_date - requested_fiscal_year.end).days <= 45:
            fiscal_quarter = requested_fiscal_year.end.quarter - 1
            fiscal_date = getattr(requested_fiscal_year, 'q'+str(fiscal_quarter)).end
```

PR #612, opened 2017-10-16
add an if for the agency page incase the agency is DOD

russellbodine:


usaspending_api/references/v2/views/agency.py
```diff
@@ -25,7 +25,8 @@ def get(self, request, pk, format=None):
         toptier_agency = agency.toptier_agency
         # get corresponding submissions through cgac code
         queryset = SubmissionAttributes.objects.all()
-        queryset = queryset.filter(cgac_code=toptier_agency.cgac_code)
+        # queryset = queryset.filter(cgac_code=toptier_agency.cgac_code)
+        queryset = queryset.filter(cgac_code="097")
```
nmonga91: Is this a typo? Probably shouldn't be hardcoded here in filter.

usaspending_api/references/v2/views/agency.py
```diff
@@ -46,11 +47,21 @@ def get(self, request, pk, format=None):
...
         # (used filter() instead of get() b/c we likely don't want to raise an
         # error on a bad agency id)
-        queryset = queryset.filter(
-            submission__reporting_fiscal_year=active_fiscal_year,
-            submission__reporting_fiscal_quarter=active_fiscal_quarter,
-            treasury_account_identifier__funding_toptier_agency=toptier_agency
-        )
+        # DS-1655: if the AID is "097" (DOD), Include the branches of the military in the queryset
+        if toptier_agency and toptier_agency.cgac_code and toptier_agency.cgac_code == "097":
+            print("hello")
```
nmonga91: Extraneous print statement

usaspending_api/references/v2/views/agency.py
```diff
@@ -46,11 +47,21 @@ def get(self, request, pk, format=None):
         # need to filter on
         # (used filter() instead of get() b/c we likely don't want to raise an
         # error on a bad agency id)
-        queryset = queryset.filter(
-            submission__reporting_fiscal_year=active_fiscal_year,
-            submission__reporting_fiscal_quarter=active_fiscal_quarter,
-            treasury_account_identifier__funding_toptier_agency=toptier_agency
-        )
+        # DS-1655: if the AID is "097" (DOD), Include the branches of the military in the queryset
+        if toptier_agency and toptier_agency.cgac_code and toptier_agency.cgac_code == "097":
```
nmonga91: Why are these conditional statements necessary? `if toptier_agency and toptier_agency.cgac_code`

above that we have the following:
```
         agency = Agency.objects.filter(id=agency_id).first()
 
         if agency is None:
             return Response(response)
 
          toptier_agency = agency.toptier_agency
```
So if we get to the later lines of code, we assume we have a valid agency. With that, are those checks still necessary?

russellbodine:
confirm we want Army Corp of Engineers as part of the DoD agency page

usaspending_api/references/v2/views/agency.py
russellbodine: Their are empty agencies in the db (from when we didnt run the cleanup script), which can cause these ifs to fail

nmonga91: So it's possible for `toptier_agency = agency.toptier_agency` to be `None`? If that's the case, this is a larger issue since it _shouldn't_ be the case if it's being inserted into `Agency` and `ToptierAgency`

usaspending_api/references/tests/test_agency_v2.py
```diff
@@ -42,13 +42,18 @@ def test_award_type_endpoint(client, financial_spending_data):
...
-                                     'mission': 'test',
-                                     'icon_filename': 'test'}}
+    # expected resp.data:
+    # assert resp.data == {'results': {'agency_name': 'tta_name', 'active_fy': '2017', 'active_fq': '2',
+    #                                  'outlay_amount': '2.00', 'obligated_amount': '2.00',
+    #                                  'budget_authority_amount': '2.00',
+    #                                  'current_total_budget_authority_amount': '8361447130497.72',
+    #                                  'website': 'http://test.com',
+    #                                  'mission': 'test',
+    #                                  'icon_filename': 'test'}}
```
nmonga91: If this portion isn't being used, we should remove it

usaspending_api/references/v2/views/agency.py
russellbodine: woops thats an award bug not an agency bug...

usaspending_api/references/tests/test_agency_v2.py
russellbodine: gone!

PR #614, opened 2017-10-16
Hotfix: Fixing bug where awards table isn't populated

dpb-bah:
Fixes the issue where the awards weren't appearing when awards should be available


dpb-bah:
PR is not updating with latest commits, remaking.

PR #615, opened 2017-10-16
remove D1/D2 loading, award/transaction cleanups

willkjackson:


usaspending_api/submissions/management/commands/rm_submission.py
```diff
@@ -67,33 +29,19 @@ def signal_handler(signal, frame):
...
-        self.logger.info('Getting childless submissions')
-        potentially_childless = self.instances_potentially_left_childless(submission)
-        print('done')
-
-        deleted = defaultdict(int)
-        deleted_total = 0
-        deletion_tally = submission.delete()
-        deleted_total += deletion_tally[0]
-        deleted.update(deletion_tally[1])
+            raise "Broker submission id {} does not exist".format(broker_submission_id)
```
tony-sappe: It is preferable to throw a specific exception, perhaps `raise KeyError("Broker submission id {} does not exist".format(broker_submission_id))`

[Lengthly SO answer](https://stackoverflow.com/a/24065533)

usaspending_api/submissions/management/commands/rm_submission.py
```diff
@@ -67,33 +29,19 @@ def signal_handler(signal, frame):
...
 
-        self.logger.info('Finished deletions')
+        self.logger.info('Finished deletions.')
 
-        statistics = "Statistics:\n  Total objects Removed: {}".format(deleted_total)
-        for (model, count) in deleted.items():
-            statistics = statistics + "\n  {}: {}".format(model, count)
+        statistics = "Statistics:\n  Total objects removed: {}".format(deleted_stats[0])
+        for (model, count) in deleted_stats[1].items():
+            statistics += "\n  {}: {}".format(model, count)
```
tony-sappe: If deleted_stats is a large dict the string += will be costly. It is better to append to a list and then use `'\n '.join(temp_list)`

willkjackson: This is expected to be < 10 lines. This can stay.

tony-sappe: no problem

usaspending_api/etl/management/load_base.py
```diff
@@ -85,14 +84,15 @@ def handle(self, *args, **options):
             logger.info('Updating model description fields...')
             update_model_description_fields()
 
-        logger.info('Updating awards to reflect their latest associated transaction info...')
-        update_awards(tuple(award_update_id_list))
```
russellbodine: We need this method to make sure an award's latest_transaction is correct and to make sure its totals (federal action obligation) is correct. 

 This code becomes even more valuable because of additional transactions being created by the nightly load.

usaspending_api/etl/management/load_base.py
```diff
@@ -85,14 +84,15 @@ def handle(self, *args, **options):
...
-
-        logger.info('Updating contract-specific awards to reflect their latest transaction info...')
-        update_contract_awards(tuple(award_contract_update_id_list))
```
russellbodine: This code updates the transaction contract's base_and_all_options_value.  I have never seen this value used...

usaspending_api/etl/management/load_base.py
```diff
@@ -85,14 +84,15 @@ def handle(self, *args, **options):
...
-
-        logger.info('Updating award category variables...')
-        update_award_categories(tuple(award_update_id_list))
```
russellbodine: Need this to fill in the categories field.

PR #616, opened 2017-10-17
updating to add frec matching

willkjackson:


usaspending_api/references/management/commands/loadtas.py
```diff
@@ -50,6 +50,24 @@ def handle(self, *args, **options):
...
 
+        # Match funding toptiers by FREC if they didn't match by AID
+        unmapped_funding_agencies = TreasuryAppropriationAccount.objects.filter(funding_toptier_agency=None)
+        match_count = 0
+        self.logger.info('Found {} unmatched funding agencies across all TAS objects. '
+                         'Attempting to match on FREC.'.format(unmapped_funding_agencies.count()))
+        for next_tas in unmapped_funding_agencies:
+            # CGAC code is a combination of FRECs and CGACs. It will never be empty and it will always
+            # be unique in ToptierAgencies; this should be safe to do.
+            frec_match = ToptierAgency.objects.filter(cgac_code=next_tas.fr_entity_code).first()
```
nmonga91: So when we load the TAS's, we're already consuming the FR Entity Code for the TAS? Is it _always_ populated?

willkjackson: Yes, line 28.

PR #618, opened 2017-10-17
Hotfix Agency Page Recipient Section

dpb-bah:
Adding `Insurance` Awards under Other tab in the Recipient Section of the Agency page

### DIFF
--- a/usaspending_api/awards/v2/views/award_spending.py
+++ b/usaspending_api/awards/v2/views/award_spending.py
@@ -7,6 +7,10 @@ from usaspending_api.common.exceptions import InvalidParameterException
 from usaspending_api.common.helpers import check_valid_toptier_agency
 from usaspending_api.common.views import DetailViewSet
 from usaspending_api.references.models import Agency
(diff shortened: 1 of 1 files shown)


usaspending_api/awards/v2/views/award_spending.py
```diff
@@ -52,6 +56,48 @@ class RecipientAwardSpendingViewSet(DetailViewSet):
...
+    # Unfortunately we can only modify the results after the queryset has run and before
+    #  it gets put into a response. It is entirely copied from the superclass except for
+    #  the award_category lines
+    # TODO: Find a better solution to lump insurance into the other category without
+    #       modifying the database.
+    @cache_response()
+    def list(self, request, *args, **kwargs):
+        try:
+            # First, get the req object from RequestCatalog
+            created, self.req = RequestCatalog.get_or_create_from_request(request)
```
nmonga91: I would try to utilize this overwritten function without the use of the RequestCatalog. We don't want to rely on a legacy feature that doesn't work consistently.

PR #620, opened 2017-10-17
Hotfix NAICS autocomplete

dpb-bah:
- `load_references` includes `load_naics` which populates a simple NAICS table using a csv from census.gov
- NAICS autocomplete uses NAICS table instead


usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -152,21 +152,23 @@ def post(self, request):
...
+            TrigramSimilarity('description', search_text)))\
             .distinct().order_by('-similarity')
 
         naics_exact_match_queryset = queryset.filter(similarity=1.0)
         if naics_exact_match_queryset.count() > 0:
             queryset = naics_exact_match_queryset
 
-        results_set = list(queryset.values('naics', 'naics_description')[:limit]) if limit else list(
-            queryset.values('naics', 'naics_description'))
+        results_set = list(queryset.values('code', 'description')[:limit]) if limit else list(
```
nmonga91: Can we use `F` to alias the columns from `code` to `naics` and same for description? would that get ride of the need for the list comprehension below?

PR #633, opened 2017-10-20
Recipient Duns Performance Improvement

dpb-bah:


usaspending_api/search/v2/views/search.py
```diff
@@ -311,37 +311,31 @@ def post(self, request):
...
         elif category == "recipient":
             # filter the transactions by scope name
-            name_dict = {}  # {recipient_name: {legal_entity_id: "1111", aggregated_amount: "1111"}
-            # define what values are needed in the sql query
-            queryset = queryset.values("federal_action_obligation",
-                                       "recipient",
-                                       "recipient__recipient_name",
-                                       "recipient__legal_entity_id",
-                                       "recipient__parent_recipient_unique_id")
+            name_dict = OrderedDict()  # {recipient_name: {legal_entity_id: "1111", aggregated_amount: "1111"}
```
nmonga91: can you prepend "e.g.," to the inline comment so it's clear that it's an example of the contents of `name_dict`?

PR #654, opened 2017-10-27
Hotfix downloads recipient state name-D1 column 

shinson:
Adding recipient state name column back to D1 downloads
-Recipient state name under d1 is now a description field of foreign province and domestic state names
-Added obligated amount column to award downloads, assistance and contracts
- Updated budget function autocomplete test to be skipped since throwing in error in travis


usaspending_api/download/v2/download_column_historical_lookups.py
```diff
@@ -572,6 +574,7 @@ def set_if_found(dl_name, path, model, file):
             'recipient_parent_name': 'contract_data__ultimate_parent_legal_enti',
             'recipient_phone_number': 'contract_data__vendor_phone_number',
             'recipient_state_code': 'contract_data__legal_entity_state_code',
+            'recipient_state_name': 'contract_data__legal_entity_state_descrip',
```
nmonga91: Just something i noticed is that `TransactionNormalized` has `recipient` and you could go down an alternate route to get the same data here through TransactionNormalized -> Recipient -> Location. This would cause additional joins, so this current way is still probably better performance-wise

PR #665, opened 2017-11-01
Update award transaction filters geocoding

shinson:
Updated the  place of performance and recipient locations fields in the transaction and award filters to include new response format for geocoding on the website.

- [x]  Front end approved Gists
- [x] Matches updated gist: 
1. https://gist.github.com/kevinli-work/4180c23503d7f64075451a53f43f5bc4#place-of-performance-location
2. https://gist.github.com/kevinli-work/4180c23503d7f64075451a53f43f5bc4#recipient-location
- [x] Performance Test Reviewed:
https://gist.github.com/shinson/0d03280be699c8838ce0559b350c4429
- [x] Backend reviewed


usaspending_api/awards/v2/filters/award.py
```diff
@@ -149,11 +148,9 @@ def award_filter(filters):
 
         # place_of_performance
         elif key == "place_of_performance_locations":
-            or_queryset = []
-            for v in value:
-                or_queryset.append(v)
-            if len(or_queryset) != 0:
-                queryset &= Award.objects.filter(place_of_performance__location_id__in=or_queryset)
+            or_queryset = geocode_filter_locations('place_of_performance', value, 'award')
```
dpb-bah: Seeing the calls of this function in the `transaction_filter` use `TransactionNormalized`, I'm surprised `apps.get_model('awards', model)` works just as fine with `award` (not `Award`). Minuscule but for clarification can you capitalize these to make it clear we're dealing with the `Award` table?

usaspending_api/awards/v2/filters/location_filter_geocode.py
```diff
@@ -0,0 +1,55 @@
...
+
+        if or_queryset is not None:
+            or_queryset |= qs
+        else:
+            or_queryset = qs
+
+    return or_queryset
+
+
+def check_location_fields(filters, fields):
```
dpb-bah: We're passing in `filters` here but it doesn't look like we're using them.

usaspending_api/awards/v2/filters/award.py
shinson: Fixed to Award

usaspending_api/awards/v2/filters/location_filter_geocode.py
shinson: Removed filters from function and then when the function is called

PR #669, opened 2017-11-03
Idempotent fixer for POP & recip location missing states

catherinedevlin:
So far it only works against FPDS data (not FABS)


usaspending_api/broker/management/commands/fix_load.py
```diff
@@ -0,0 +1,222 @@
+"""
+Idempotently re-runs location creation / matching code from broker FPDS data
```
nmonga91: Can we name this command to be a little more specific to its functionality?

nmonga91:
We discussed parallelizing this as well, but logging that here for awareness.

PR #670, opened 2017-11-03
DS-1750 Spending by Category Refactor

tony-sappe:
Decent refactor to implement better Django ORM practices. See 
[DS-1750](https://federal-spending-transparency.atlassian.net/browse/DS-1750) comments for some comparison benchmarks.

AFAIK refactor meets previous [Gist](https://github.com/fedspendingtransparency/usaspending-api/wiki#spending-by-category) requirement. See WIP gist [here](https://hackmd.io/s/SyIK1HcRb)

- [x] Backend Dev Review
- [x] Front End Review
- [x] DevOps Review


usaspending_api/common/helpers.py
```diff
@@ -94,32 +95,62 @@ def generate_last_completed_fiscal_quarter(fiscal_year):
...
-    page_metadata["hasPrevious"] = (page > 1 and limit*(page-2) < total_return_count)
-    page_metadata["next"] = page+1 if page_metadata["hasNext"] else None
-    page_metadata["previous"] = page-1 if page_metadata["hasPrevious"] else None
+    page_metadata["hasPrevious"] = (page > 1 and limit * (page - 2) < total_return_count)
+    page_metadata["next"] = page + 1 if page_metadata["hasNext"] else None
+    page_metadata["previous"] = page - 1 if page_metadata["hasPrevious"] else None
+    return page_metadata
+
+
+def get_simple_pagination_metadata(results_plus_one, limit, page):
```
dpb-bah: I remember there were instances where this function could be useful. Is there any reason we haven't used this instead?

usaspending_api/search/v2/views/search.py
```diff
@@ -108,8 +108,13 @@ def post(self, request):
         # 'time_period': {'fy': '2017', 'quarter': '3'},
         # 	'aggregated_amount': '200000000'
         # }]
-        for key, value in sorted(group_results.items(), key=lambda k: (ast.literal_eval(k[0])['fiscal_year'],
-                                                                       int(ast.literal_eval(k[0])[nested_order])) if nested_order else (ast.literal_eval(k[0])['fiscal_year'])):
+        flake8_nightmare = sorted(
```
dpb-bah: As apt as the name is, could we change it to `sorted_group_results` or the like?

usaspending_api/search/v2/views/search.py
```diff
@@ -150,307 +155,249 @@ def post(self, request):
...
-                for trans in agency_set[:limit]:
-                    ttname = trans["awarding_agency__toptier_agency__name"]
-                    ttabv = trans["awarding_agency__toptier_agency__abbreviation"]
-                    ttob = trans["federal_action_obligation"] if trans["federal_action_obligation"] else 0
-                    if ttname in name_dict:
-                        name_dict[ttname]["aggregated_amount"] += ttob
-                    else:
-                        name_dict[ttname] = {"aggregated_amount": ttob, "abbreviation": ttabv}
+                agency_set = queryset \
+                    .filter(awarding_agency__toptier_agency__name__isnull=False) \
```
dpb-bah: Below in funding agency, we're checking to see if the `funding_agency` is not null. Is there a reason why we're not doing the same check here for `awarding_agency`? Both are allowed to be null in the models. This also applies to line 177.

usaspending_api/search/v2/views/search.py
```diff
@@ -150,307 +155,249 @@ def post(self, request):
...
+                    .values(
+                        agency_name=F('awarding_agency__toptier_agency__name'),
+                        abbreviation=F('awarding_agency__toptier_agency__abbreviation')) \
+                    .annotate(aggregated_amount=Sum('federal_action_obligation')) \
+                    .order_by('-aggregated_amount')
+
+                # Begin DB hits here
+                total = agency_set.count()
```
dpb-bah: Regarding the first comment, here's an example and really it applies to most the cases (see lines 185, 237, 252, 298 specifically). Is there a reason we're not using the simple pagination and using the `count`s?

usaspending_api/common/helpers.py
tony-sappe: I just made this function. It is useful if the count query is too expensive. Since I kept count in spending_by_category I didn't utilize this function.

usaspending_api/search/v2/views/search.py
tony-sappe: I was focusing on refactoring spending_by_category. I didn't want to change the Gist. count is listed in the gist file

usaspending_api/search/v2/views/search.py
tony-sappe: Good catch. I'll add the null check

usaspending_api/common/helpers.py
dpb-bah: So for some of the endpoints currently we aren't returning `count` contrary to the gist. This was removed as the front end doesn't use count for this endpoint. I'd suggest double checking with the frontend team to ensure that is the case, update the gist accordingly, and use the simple_page_metadata solution. If the `count`s aren't as taxing as we thought they were, we can readd them back in but we made sure to remove them to save on performance and the frontend wasn't using them.

usaspending_api/search/v2/views/search.py
tony-sappe: Making sure the reviewers are reviewing :)

tony-sappe:
WIP gist: https://hackmd.io/s/SyIK1HcRb

usaspending_api/common/helpers.py
tony-sappe: Verified front end isn't relying on them. Updated gist [here](https://hackmd.io/s/SyIK1HcRb)

PR #671, opened 2017-11-03
Adding IDV filtering to award type

dylanbah:


usaspending_api/awards/v2/filters/award.py
```diff
@@ -65,10 +65,15 @@ def award_filter(filters):
...
             or_queryset = []
+
+            idv_flag = all(i in value for i in ['A', 'B', 'C', 'D'])
+
             for v in value:
                 or_queryset.append(v)
             if len(or_queryset) != 0:
                 queryset &= Award.objects.filter(type__in=or_queryset)
+            if idv_flag:
+                queryset |= Award.objects.filter(type__isnull=True, latest_transaction__to_contract__pulled_from='IDV')
```
nmonga91: So if it's an IDV contract, the type is always null?

dylanbah: From Broker: "IDV contracts will not have an award type"

PR #674, opened 2017-11-06
Feature fsrs loader

alburde1:
Script to load FSRS subgrant and subcontract data from broker.

Technical Notes:
- Migration for award_subaward table
- Running script: `python manage.py load_fsrs`


usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -0,0 +1,301 @@
...
+        return dictfetchall(db_cursor)
+
+    @staticmethod
+    def get_award(row, award_type):
+        if award_type == 'procurement':
+            # we don't need the agency for grants
+            agency = get_valid_awarding_agency(row)
+
+            if not agency:
+                # logger.warning(
```
nmonga91: Is this warning no longer necessary? if it's not, can we just remove it as well the one below that's similar?

alburde1: Crud, I knew I'd forgotten to uncomment something. I commented them out because they were causing too many warnings locally and I wanted to test. We want them printing so we know what internal IDs had problems

usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -108,10 +109,15 @@ def gather_shared_award_data(self, data, award_type):
...
-            )
+            try:
+                recipient, created = LegalEntity.objects.get_or_create(
+                    recipient_unique_id=row['duns'],
+                    recipient_name=recipient_name
+                )
+            except MultipleObjectsReturned:
+                created = False
+                print('Legal Entity with DUNS: {} and name: {} returned two rows. '
+                      'Skipping...'.format(row['duns'], recipient_name))
```
alburde1: This isn't actually skipping anything, you need to put a `continue` after this to skip this record. If you want to get the first one you have to do something else. If you want a NULL in this case you have to set recipient to NULL

usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -108,10 +109,15 @@ def gather_shared_award_data(self, data, award_type):
...
-                recipient_name=recipient_name
-            )
+            try:
+                recipient, created = LegalEntity.objects.get_or_create(
+                    recipient_unique_id=row['duns'],
+                    recipient_name=recipient_name
+                )
+            except MultipleObjectsReturned:
+                created = False
+                print('Legal Entity with DUNS: {} and name: {} returned two rows. '
```
alburde1: This should be a logger not a print

usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -68,9 +68,9 @@ def get_award(row, award_type):
 
             # We don't have a matching award for this subcontract, log a warning and continue to the next row
             if not award:
-                logger.warning(
-                    "Internal ID {} cannot find award with piid {}, parent_award_id {}; skipping...".
-                    format(row['internal_id'], row['contract_number'], row['idv_reference_number']))
+                #logger.warning(
+                #    "Internal ID {} cannot find award with piid {}, parent_award_id {}; skipping...".
+                #    format(row['internal_id'], row['contract_number'], row['idv_reference_number']))
```
alburde1: This shouldn't be commented out forever, not sure what the plan is in the end but we need to remember to uncomment at some point.

usaspending_api/etl/helpers.py
```diff
@@ -78,12 +78,15 @@ def get_or_create_location(row, mapper):
...
     if country_code and country_code.country_code == "USA":
+        
         # Apparently zip codes are optional...
         if location_dict["location_zip"]:
             location_dict.update(
                 zip5=location_dict["location_zip"][:5],
                 zip_last4=location_dict["location_zip"][5:])
-            location_dict.pop("location_zip")
+
+        location_dict.pop("location_zip")
```
alburde1: Why was this moved out of the `if`? Do we always want to pop even if it's NULL?

PR #681, opened 2017-11-08
Update agency loans 1887

russellbodine:
By removing loans with null federal_action_obligation from the queryset, the user is able to see loans that are non null.


russellbodine:
Unfortunately, this is not right, we want to use the load $$$ field, which is not federal_obligation_amount.

PR #684, opened 2017-11-08
Fix award search download last modified lookup

nmonga91:


nmonga91:
@tony-sappe There are not any other changes or branches necessary. This should be everything.

PR #686, opened 2017-11-09
Populate `Location.zip5` from `Location.zip4` when appropriate

catherinedevlin:
... "appropriate" means that `zip5` is empty, and `zip4` appears to begin with a legitimate 5-digit zip.  It makes `Location.zip5` the appropriate field to use for ZIP code queries.  Actually incorporating those zips into endpoint outputs is saved for a future PR.

- `Location.save()` and broker loaders will fill `zip5`
- `manage.py fix_zip_codes` for a one-time fix of existing Locations

I don't know if we want to wrap a simple SQL statement like the one in `fix_zip_codes.py` as a Django management command - are we trying to keep our `manage.py` "uncluttered"?

My very cheesy time estimate on running `manage.py fix_zip_codes` against the full production dataset - linearly extrapolating from a small sample run on my laptop (thus the cheesiness) - is ten hours.  As written, it will create row-level locks as it goes; I could change it to run in batches with intermediate commits, in case we want to run other location-updating code at the same time.


usaspending_api/broker/management/commands/fix_load.py
```diff
@@ -0,0 +1,217 @@
...
+                    recip.save()
+
+    def add_arguments(self, parser):
+
+        parser.add_argument('--piid',
+                            nargs='+',
+                            help="PIID of award(s) to reload")
+
+    @db_transaction.atomic
+    def handle(self, *args, **options):
```
shinson: Is this script reloading FPDS data to fix the zipcode data at the Award/TransactionFPDS level?

usaspending_api/references/management/commands/fix_zip_codes.py
```diff
@@ -0,0 +1,23 @@
...
+    logger = logging.getLogger('console')
+
+    update_qry = """
+       UPDATE references_location
+       SET    zip5 = SUBSTRING(zip4 FROM '^(\d{5})\-?(\d{4})?$')
+       WHERE  zip5 IS NULL
+       AND    zip4 IS NOT NULL
+    """
+
+    def handle(self, *args, **options):
```
shinson: Adding log statements to show when the updates have started and finished, and any errors would be helpful.

usaspending_api/references/management/commands/fix_zip_codes.py
```diff
@@ -0,0 +1,23 @@
+import logging
+
+from django.core.management.base import BaseCommand
+from django.db import connection
+
+
+class Command(BaseCommand):
```
shinson: I believe this should be fine as a stand alone command since it doesn't have to be part of the nightly load process and just needs to be run once.

usaspending_api/references/management/commands/fix_zip_codes.py
willkjackson: Agree with @shinson, this should print out every 10k rows as we do in our other broker scripts.

catherinedevlin: Okay, latest commit breaks the update into 10k batches.

usaspending_api/broker/management/commands/fix_load.py
catherinedevlin: Oops, this was a half-finished part of a different branch that I apparently included in this PR with the popular git command, `git-fumble`.  Now removed.  Thanks!

PR #694, opened 2017-11-15
State fixer, using local tables

catherinedevlin:
Different and much faster approach for a one-time correction of missing state names and abbreviations.  

Currently running on dev, where it's on track to update the 98525403 locations in 3.5 hours.

For speed, this only updates place of performance locations, not recipient locations.


usaspending_api/broker/management/commands/fix_locations.py
```diff
@@ -0,0 +1,149 @@
+"""
+Idempotently re-runs location creation / matching code from broker FPDS data
```
nmonga91: @catherinedevlin is a recipient fix in the works or is only place of performance an issue at the moment?

catherinedevlin: I can add the recipient fixer pretty quickly.  I didn't know if we wanted to triage that out for the sake of runtime, though since the whole thing only took about 7 hours on dev anyway, there's probably no need to.  I'll add that now.

usaspending_api/broker/management/commands/fix_locations.py
```diff
@@ -0,0 +1,192 @@
...
+        """), ('Recipient for FABS (get state name from code)', 'transaction_fabs', """
+        UPDATE references_location l
+        SET    state_name = sa.name,
+               state_code = UPPER(REPLACE(tf.legal_entity_state_code, '.', ''))
+        FROM   transaction_normalized tn
+        JOIN   transaction_fabs tf ON (tf.transaction_id = tn.id)
+        JOIN   legal_entity le ON (tn.recipient_id = le.legal_entity_id)
+        JOIN   state_abbrevs sa ON (UPPER(REPLACE(tf.legal_entity_state_code, '.', '')) = sa.abbrev)
+        WHERE  le.location_id = l.location_id
+        AND    tf.legal_entity_state_code IS NOT NULL
```
nmonga91: Would this overlap any records that were recently filled in with their state codes if it's checking for `NOT NULL` here?

catherinedevlin: Lines 90-91 mean that it shouldn't touch and location records that presently have a state code.

PR #695, opened 2017-11-15
Generic location geofilter uses zip for zip5

catherinedevlin:
Can now geofilter on zip.


usaspending_api/awards/v2/filters/location_filter_geocode.py
```diff
@@ -7,7 +7,8 @@
     'country': 'location_country_code',
     'state': 'state_code',
     'county': 'county_code',
-    'district': 'congressional_code'
+    'district': 'congressional_code',
+    'zip': 'zip5',
 }
```
shinson: In addition to the `loc_dict` you need to include zip5 in the list  under the `get_fields_list` function since the zip5 is just one format (XXXXX). Down the line we will remove this function once county_codes and congresional_codes are uniform.

shinson: In addition to the `loc_dict` you need to include zip5 in the list under the get_fields_list function since the `zip5` is just one format (XXXXX). Down the line we will remove this function once `county_codes` and `congresional_codes` are uniform.

catherinedevlin: Ah, I see... it took me a while to figure out the purpose of that function, so I added a comment, too.  Thanks!

PR #709, opened 2017-11-17
Subaward loader and endpoint enhancements

nmonga91:
Related JIRA Story: https://federal-spending-transparency.atlassian.net/browse/DS-1896

* Updating FSRS loader to account for updating related award metadata (subaward counts and totals).
* Updated subaward serializer to include associated award data to retrieve the metadata in the response.

### DIFF
--- a/usaspending_api/broker/management/commands/load_fsrs.py
+++ b/usaspending_api/broker/management/commands/load_fsrs.py
@@ -161,16 +161,21 @@ class Command(BaseCommand):
                                   'sub_award.overall_description', 'sub_award.recovery_model_q1 AS q1_flag',
                                   'sub_award.recovery_model_q2 AS q2_flag',
                                   'sub_award.subcontract_date AS subaward_date'])
+
+            query = "SELECT " + ",".join(query_columns) + " FROM fsrs_" + award_type + " AS award " + \
+                    "JOIN fsrs_" + subaward_type + " AS sub_award ON sub_award.parent_id = award.id WHERE award.id > " + \
+                    str(max_id) + " AND sub_award.subcontract_num IS NOT NULL ORDER BY award.id, sub_award.id LIMIT " + \
+                    str(QUERY_LIMIT) + " OFFSET " + str(offset)
         else:
             query_columns.extend(['sub_award.cfda_numbers', 'sub_award.subaward_num', 'sub_award.subaward_amount',
                                   'sub_award.project_description AS overall_description',
                                   'sub_award.compensation_q1 AS q1_flag', 'sub_award.compensation_q2 AS q2_flag',
                                   'sub_award.subaward_date'])
 
-        query = "SELECT " + ",".join(query_columns) + " FROM fsrs_" + award_type + " AS award " + \
-                "JOIN fsrs_" + subaward_type + " AS sub_award ON sub_award.parent_id = award.id WHERE award.id > " + \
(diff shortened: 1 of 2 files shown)


usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -53,9 +55,9 @@ def get_award(row, award_type):
             agency = get_valid_awarding_agency(row)
 
             if not agency:
-                logger.warning(
-                    "Internal ID {} cannot find matching agency with subtier code {}".
-                    format(row['internal_id'], row['contracting_office_aid']))
+                # logger.warning(
```
dpb-bah: Any reason why we're commenting the warnings out?

usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -101,7 +103,7 @@ def gather_shared_award_data(self, data, award_type):
         skip_count = 0
         for row in data:
             counter += 1
-            if counter % 1000 == 0:
+            if counter % 10000 == 0:
```
dpb-bah: Nickpicky but I'd suggest making this a constant for readability.

usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -165,11 +167,9 @@ def gather_next_subawards(db_cursor, award_type, subaward_type, max_id, offset):
 
         query = "SELECT " + ",".join(query_columns) + " FROM fsrs_" + award_type + " AS award " + \
                 "JOIN fsrs_" + subaward_type + " AS sub_award ON sub_award.parent_id = award.id WHERE award.id > " + \
-                str(max_id) + "AND sub_award.subcontract_num IS NOT NULL ORDER BY award.id, sub_award.id LIMIT " + \
```
dpb-bah: I'm assuming sub_award.subcontract_num isn't required now?

usaspending_api/broker/management/commands/load_fsrs.py
nmonga91: Commented this out since it was outputting at a high volume during local testing. In order to avoid the same when running on other environments, I temporarily commented it out. I can re-add it if necessary.

usaspending_api/broker/management/commands/load_fsrs.py
nmonga91: Agreed, will make the change.

usaspending_api/broker/management/commands/load_fsrs.py
nmonga91: The query was having problems with saying `sub_award.subcontract_num IS NOT NULL`. It didn't think the column existed which was odd. Removing it fixed the query and that field looks to be populated on our end for all subawards.

nmonga91: Confirming the need for this with @alburde1

usaspending_api/broker/management/commands/load_fsrs.py
dpb-bah: If it affects performance I'd stick with keeping it out. If not, it definitely can help with debugging.

nmonga91: it was a very large volume of output. i'll leave it out for now.

PR #715, opened 2017-11-21
Updating nightly loaders to convert last modified date to a date object

nmonga91:
* Fixing date comparison for deletion files to operate on 2 dates, instead of 1 date and 1 string
* Updating "last_modified_date" to be converted from a string representation of a timestamp to a date object


dpb-bah:
Simple enough. Approved

PR #717, opened 2017-11-21
add unique together to the Legal Entity table.

russellbodine:


russellbodine:
CANT UNIQUE TOGETHER FIELDS WITH A NULL (ALWAYS PASSES)

PR #730, opened 2017-11-26
Added support for materialized views in spending by category

tony-sappe:
spending by category Funding and Awarding agency categories now use materialized views.

Requires summary_subagencies_view [here](https://github.com/fedspendingtransparency/usaspending-config/tree/R%26D_filter_matview)


usaspending_api/search/v2/views/search.py
```diff
@@ -158,10 +158,45 @@ def post(self, request):
...
+                    .annotate(aggregated_amount=Sum('federal_action_obligation')) \
+                    .order_by('-aggregated_amount')
+            elif category == "funding_agency":
+                queryset = queryset \
+                    .filter(funding_agency_name__isnull=False) \
+                    .values(
+                        agency_name=F('funding_agency_name'),
+                        agency_abbreviation=F('funding_agency_abbr')) \
+                    .annotate(aggregated_amount=Sum('federal_action_obligation')) \
+                    .order_by('-aggregated_amount')
```
dpb-bah: Very similar code here, would recommend something like:
```
agency_category = None
if category in ["awarding_agency", "funding_agency"]:
    agency_filters = {"{}_name__isnull".format(category): False}
    agency_values = {"agency_name": F('{}_name'.format(category)),
                     "agency_abbreviation": F('{}_abbr'.format(category))}
    queryset = queryset \
        .filter(**agency_filters) \
        .values(**agency_values) \
        .annotate(aggregated_amount=Sum('federal_action_obligation')) \
        .order_by('-aggregated_amount')
```
But then again, that'd effect the whole file as well so it's not crucial. Just something to keep in mind for this file.

usaspending_api/search/v2/views/search.py
```diff
@@ -158,10 +158,45 @@ def post(self, request):
...
+
+            # Begin DB hits here
+            results = list(queryset[lower_limit:upper_limit + 1])
+            page_metadata = get_simple_pagination_metadata(len(results), limit, page)
+            results = results[:limit]
+
+            response = {"category": category, "scope": scope, "limit": limit, "results": results,
```
dpb-bah: This file also splits between using single quotes or double quotes for strings. Again, not required for this PR, just making a note.

tony-sappe: Which quotes do we prefer for this project?

usaspending_api/search/v2/views/search.py
tony-sappe: Perhaps we can look at stuff like this when we can refactor. It adds unnecessary complexity when there are small differences.

usaspending_api/search/v2/views/search.py
dpb-bah: I think single quotes? So long as it's consistent.

PR #735, opened 2017-11-27
Hotfix: Bulk Download filtering only by Awarding Agency, IDV fix

dpb-bah:


usaspending_api/bulk_download/v2/views.py
```diff
@@ -274,13 +274,12 @@ def process_filters(self, filters, award_level):
...
             type_queryset_filters['{}__in'.format(value_mappings[award_level]['type'])] = award_types
-            type_queryset = table.objects.filter(**type_queryset_filters)
+            type_queryset = Q(**type_queryset_filters)
             if (filters['award_types'] == ['contracts']):
                 # IDV Flag
-                idv_queryset_filters = {'{}__isnull'.format(value_mappings[award_level]['type']): True,
-                                        '{}__pulled_from'.format(value_mappings[award_level]['contract_data']): 'IDV'}
-                type_queryset |= table.objects.filter(**idv_queryset_filters)
-            queryset &= type_queryset
+                idv_queryset_filters = {'{}__pulled_from'.format(value_mappings[award_level]['contract_data']): 'IDV'}
```
nmonga91: Here, we're not able to filter on specific types of contracts for the bulk download? I.e., we select all contracts or none every time?

dpb-bah: Correct, bulk download currently doesn't filter by specific award types (like only "A" contracts). Minimal changes would be required to update the gist and edit the code to allow this variance. Definitely doable and better for API users, just would require updating frontend as well.

PR #740, opened 2017-11-27
Fixing matview for geofilters state/county/district

shinson:
Fixing matview mappings throwing 500 errors:
- State: no geolayer_filters throwing an error when no geo layer filters
- County District geolayer_filters were not mapping correctly because geocode location filter was not set to matview = True

### DIFF
--- a/usaspending_api/search/v2/views/search.py
+++ b/usaspending_api/search/v2/views/search.py
@@ -500,7 +500,7 @@ class SpendingByGeographyVisualizationViewSet(APIView):
         else:
             # Adding null filter for state for specific partial index
             # when not using geocode_filter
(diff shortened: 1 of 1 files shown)


usaspending_api/search/v2/views/search.py
```diff
@@ -551,7 +551,7 @@ def county_district_queryset_matview(self, kwargs, fields_list, loc_lookup, stat
             self.queryset &= geocode_filter_locations(scope_field_name, [
                 {'state': fips_to_code.get(x[:2]), self.geo_layer: x[2:], 'country': 'USA'}
                 for x in self.geo_layer_filters
-            ], 'UniversalTransactionView')
+            ], 'UniversalTransactionView', True)
```
dpb-bah: Can you just elaborate on what's being added here?

PR #752, opened 2017-11-28
Recipient Filter Update

alburde1:
- Updating recipient filter to use the exact text sent instead of a list of `legal_entity` IDs

### DIFF
--- a/usaspending_api/awards/v2/filters/award.py
+++ b/usaspending_api/awards/v2/filters/award.py
@@ -24,6 +24,7 @@ def award_filter(filters):
                     'award_type_codes',
                     'agencies',
                     'legal_entities',
+                    'recipient_search_text',
                     'recipient_scope',
                     'recipient_locations',
                     'recipient_type_names',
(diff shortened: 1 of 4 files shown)


usaspending_api/awards/v2/filters/transaction.py
```diff
@@ -171,6 +171,7 @@ def transaction_filter(filters):
                     awarding_agency__subtier_agency__name__in=awarding_subtier
                 )
 
+        # TODO update this to recipient_search_text if we ever switch back from the matview
```
tony-sappe: AFAIK this is still used by bulk downloader, it isn't dead code yet

usaspending_api/awards/v2/filters/award.py
```diff
@@ -171,6 +171,7 @@ def award_filter(filters):
             if len(awarding_subtier) != 0:
                 queryset &= Award.objects.filter(awarding_agency__subtier_agency__name__in=awarding_subtier)
 
+        # TODO update this to recipient_search_text if we ever switch back from the matview
```
nmonga91: We should update this now

usaspending_api/awards/v2/filters/transaction.py
nmonga91: We should update this here now as well

usaspending_api/awards/v2/filters/matview_transaction.py
```diff
@@ -115,7 +115,7 @@ def transaction_filter(filters):
                     'time_period',
                     'award_type_codes',
                     'agencies',
-                    'legal_entities',
```
tony-sappe: In the interest of not breaking anything, could we keep the existing filter inplace and add the new search_text field? Are we 100% sure nothing will use `legal_entities` after the website PR?

nmonga91: I would think we shouldn't allow it going forward. @tony-sappe if we're not 100% sure, then I agree, we should keep it.

willkjackson: Keep the filter.

PR #754, opened 2017-11-29
add extent_competed to model (already in matview)

willkjackson:


usaspending_api/awards/models_matviews.py
```diff
@@ -139,60 +140,29 @@ class UniversalAwardView(models.Model):
...
     product_or_service_code = models.TextField()
     product_or_service_description = models.TextField()
     naics_code = models.TextField()
     naics_description = models.TextField()
-    dod_claimant_program_code = models.TextField()
-    program_system_or_equipmen = models.TextField()
-    information_technolog_desc = models.TextField()
-    sea_transportation_desc = models.TextField()
-    clinger_cohen_act_planning = models.TextField()
-    davis_bacon_act_descrip = models.TextField()
```
nmonga91: Why were a bunch of these removed?

PR #758, opened 2017-11-29
New command to tidy up duplicate locations

catherinedevlin:
Since we already had a bunch of duplicate Locations, and the recent data fixes created some more, I thought we should have a way to condense the duplicates down into a single canonical Location.

One question I have for reviewers is whether this is the right way to handle the date fields in a location - that, generally, the canonical Location should span from the first begin date to the last end date of the Locations that get squashed down into it.  I left `certified_date` out of that squashing, however, assuming that two otherwise identical Locations with different `certified_date`s should not be squashed together.  

This can be run at our leisure.  It's idempotent and can be in bits.

The first step is SLOW, and unfortunately can't be broken into parts or log intermediate progress, thanks to the nature of the query.  The other queries report intermediate progres..


catherinedevlin:
New matviews make this unneeded.

PR #768, opened 2017-11-30
Hotfix correct recipient type filter

alburde1:
- Recipient type filter for awards was passing for DUNS that had that type in any of their instances, plus it was slower.
- Cleaned up transaction filter a little because it was doing extraneous work

### DIFF
--- a/usaspending_api/awards/v2/filters/matview_transaction.py
+++ b/usaspending_api/awards/v2/filters/matview_transaction.py
@@ -267,13 +267,8 @@ def transaction_filter(filters):
 
         # recipient_type_names
         elif key == "recipient_type_names":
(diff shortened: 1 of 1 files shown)


tony-sappe:
The codebase is confusing at the moment. You should also update transaction.py and possibly award.py in the same dir as this file.

alburde1:
From what I can see the logic is already correct in both of these files:
https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/awards/v2/filters/award.py#L206

https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/awards/v2/filters/transaction.py#L227

PR #780, opened 2017-12-05
removing request catalog from all v1 mixins and viewsets

willkjackson:


usaspending_api/common/mixins.py
```diff
@@ -25,8 +22,8 @@ def aggregate(self, request, *args, **kwargs):
         # regardless of request type (e.g., GET, POST)
         # (not sure if this is a good practice, or we should be more
         # prescriptive that aggregate requests can only be of one type)
-        params = dict(self.req.request["query_params"])
```
dpb-bah: @nmonga91 @willkjackson and I looked over the code and weren't entirely sure if `self.request` or just `request` should be used for these changes. There's no explicit initialization of `self.request` anywhere in the class or its superclasses to mention. With the latest changes, the endpoints seem to work as expected for the front end but we'd like your input here as well.

usaspending_api/common/views.py
```diff
@@ -55,15 +50,11 @@ class DetailViewSet(viewsets.ReadOnlyModelViewSet):
...
-        return {'req': self.req, **context}
+        return {**context}
 
     @cache_response()
     def list(self, request, *args, **kwargs):
         try:
-            # First, get the req object from RequestCatalog
-            created, self.req = RequestCatalog.get_or_create_from_request(request)
-            # Pass this to the paginator
-            self.paginator.req = self.req
```
nmonga91: is nothing going to be passed to the paginator now? will that break pagination?

nmonga91: Still works as expected

PR #781, opened 2017-12-06
updated models to match new matviews

tony-sappe:
This PR is to create consistent column names across the materialized views.
Instead of replacing columns the new cols names were added as duplicate fields to provide temporary backward compatibility.

See [PR21 in usaspending-config](https://github.com/fedspendingtransparency/usaspending-config/pull/21)

- [x] Confirmed matview changes in @tony-sappe 's DB
- [x] Matview changes finished in dev DB


tony-sappe:
PR21 in usaspending-config Merged.

New matviews created early 12/8/17

PR #786, opened 2017-12-07
Resolving ability of FPDS nightly loader to save last load date

nmonga91:
Steps to successfully apply changes:

1. Before merging run: `manage_api migrate broker zero`
2. After merging run: 
    * `manage_api migrate broker`
    * `manage_api load_broker_static_data`


willkjackson:
Why is this not a new migration? It will get ignored...

PR #796, opened 2017-12-08
Converting Award ID filters from database ID to text filters

nmonga91:
* Updating Award ID filter to support a list of strings representing PIIDs, FAINs, URIs instead of database IDs
* Updating Spending by Award viewset to handle all award types instead of only one set at a time

Merge Requirements:
- [x] Filter documentation updates approved by Frontend (@kevinli-work): https://hackmd.io/AwEwTAzALAZgrGAtANigQ2YqBjA7HRAThBmCLAFNk1CYKK00wg==?both#award-id
- [x] Merged with Frontend updates ([Frontend PR #571](https://github.com/fedspendingtransparency/usaspending-website/pull/571))
- [x] Award & Transaction matviews are updated so new GIN indexes are created in dev (other environments will need to also be updated as code is pushed through the pipeline)

GIN Indexes that need to be created (already done in `dev` manually):
```
CREATE INDEX universal_award_matview_piid_gin_trgm_idx ON universal_award_matview USING gin (UPPER(piid) gin_trgm_ops);
CREATE INDEX universal_award_matview_fain_gin_trgm_idx ON universal_award_matview USING gin (UPPER(fain) gin_trgm_ops);
CREATE INDEX universal_award_matview_uri_gin_trgm_idx ON universal_award_matview USING gin (UPPER(uri) gin_trgm_ops);

CREATE INDEX universal_transaction_matview_piid_gin_trgm_idx ON universal_transaction_matview USING gin (UPPER(piid) gin_trgm_ops);
CREATE INDEX universal_transaction_matview_fain_gin_trgm_idx ON universal_transaction_matview USING gin (UPPER(fain) gin_trgm_ops);
CREATE INDEX universal_transaction_matview_uri_gin_trgm_idx ON universal_transaction_matview USING gin (UPPER(uri) gin_trgm_ops);
```


usaspending_api/search/v2/views/search.py
```diff
@@ -622,84 +623,67 @@ def post(self, request):
...
                 for field in fields:
-                    row[field] = award[award_contracts_mapping[field]]
-            elif set(filters["award_type_codes"]) <= set(loan_type_mapping):  # loans
+                    row[field] = award.get(award_contracts_mapping.get(field))
+            elif award['type'] in loan_type_mapping.keys():  # loans
                 for field in fields:
-                    row[field] = award[loan_award_mapping[field]]
-            elif set(filters["award_type_codes"]) <= set(non_loan_assistance_type_mapping):  # assistance data
+                    row[field] = award.get(loan_award_mapping.get(field))
+            elif award['type'] in non_loan_assistance_type_mapping.keys():  # assistance data
```
dpb-bah: Like the checks here, only comment is that I think in Python you don't need the `.keys()` calls, just checking to see if it's in the dictionary should check if the value is in the dictionary's keys. That being said, the `.keys()` functions are more readable so it really just comes down to a matter of taste.

nmonga91: You are correct, the `.keys()` is unnecessary. The best practice also dictates to use it without the `.keys()` so I will update that.

PR #800, opened 2017-12-08
adding events.log output from all endpoints

willkjackson:


willkjackson:
Do not merge until new ELK stack is up.

PR #801, opened 2017-12-11
removing req/request param puller, update serializers

willkjackson:


usaspending_api/common/tests/autocomplete.py
```diff
@@ -18,6 +17,7 @@ def check_autocomplete(route, client, fields, value, expected):
                 'value': value,
                 'matched_objects': match_objs
             }))
+        print(resp.status_code)
```
nmonga91: This should be removed

PR #809, opened 2017-12-12
adding multi submission loader (for nightly loads)

willkjackson:
This calls the existing "load_submission" command we have after diffing the broker DB and the API based on the FY and fiscal quarter.

EX: `python manage.py load_multiple_submissions 2017 3` will check against the broker (assuming DATABASE_URL and DATA_BROKER_DATABASE_URL are set), list any discrepancies (and when they last certified), and kick off any missing or recertified submissions for FY2017 Q3.


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -80,9 +81,9 @@ def signal_handler(signal, frame):
         logger.info('Finished getting submission {} from broker'.format(submission_id))
 
         if len(submission_data) == 0:
-            raise 'Could not find submission with id ' + str(submission_id)
```
willkjackson: We were raising a TypeError (not derived from Base) exception here. 😱

PR #816, opened 2017-12-15
Feature monthly downloads

dpb-bah:
- [x] [List Monthly download gist](https://gist.github.com/dpb-bah/3fc16bc5b9c4cce03743c94efb6c9883) approved
- [x] Backend reviewed

Includes:
- Monthly Download Script
- Includes Bulk Download Migrations
- List Monthly Files Endpoint


usaspending_api/bulk_download/management/commands/populate_monthly_files.py
```diff
@@ -0,0 +1,211 @@
...
+        for award_level in award_levels:
+            download_job_kwargs[award_level] = True
+        for award_type in award_types:
+            download_job_kwargs[award_type] = True
+        if agency and agency != 'all':
+            download_job_kwargs['agency'] = ToptierAgency.objects.filter(toptier_agency_id=agency).first()
+        if sub_agency:
+            download_job_kwargs['sub_agency'] = SubtierAgency.objects.filter(subtier_agency_id=sub_agency).first()
+        if start_date:
+            download_job_kwargs['start_date'] = start_date
```
tony-sappe: Does this default to all FYs?

dpb-bah: In this function, if start_date and end_date are not provided, it assumed it will cover all FYs. The way this function is called currently is that the start/end dates are set to each fiscal year (see lines 194-211)

tony-sappe: makes sense

tony-sappe:
Tested locally. Worked once psql was in path

PR #817, opened 2017-12-18
Bug DS-2008 - FAADS+ business types

catherinedevlin:
https://federal-spending-transparency.atlassian.net/browse/DS-2008

Incorporates the equivalence of the FAADS+ versions into the `update_business_type_categories` static method that bulk_load_fabs and bulk_load_fpds use. That is called by `LegalEntity.save`, so it should be applied automatically to future data loads.

To fix existing records, run `python manage.py resave_instances references.LegalEntity`. That processes about 135 rows per second (when run from a local machine), so estimate 27 hours for the full 99K legal entities in production.


catherinedevlin:
@nmonga91 `resave_instances` loops over individual records, so it won't lock up records, just add a little load to the server.

usaspending_api/awards/models_matviews.py
```diff
@@ -288,14 +264,11 @@ class Meta:
         db_table = 'summary_view'
 
 
-class SumaryNaicsCodesView(models.Model):
-
-    # Fields
+class SummaryNaicsCodesView(models.Model):
```
nmonga91: Since there are some matview updates to this file (for the models and associated use thereof), @tony-sappe can you please verify that this will not interfere with anything anywhere else?

catherinedevlin: Looks like those matview updates were an artifact of this branch being behind dev - when I merged dev back in, they disappeared from the diff (they're certainly nothing I wrote)

PR #819, opened 2017-12-19
Step 2 of col rename process and fixed minor matview issues

tony-sappe:
- Added ability to use summaryTransactionMonthView when award amounts match enum limits
- Fixed misspelled model names
- Removed the old cols from models so they can be removed from matviews' create SQL next sprint
- Fixed index SQL syntax errors

- [x] Backend Review

### DIFF
--- a/usaspending_api/awards/models_matviews.py
+++ b/usaspending_api/awards/models_matviews.py
@@ -345,8 +345,6 @@ class SummaryTransactionMonthView(models.Model):
     cfda_popular_name = models.TextField(blank=True, null=True)
     product_or_service_code = models.TextField()
     product_or_service_description = models.TextField()
-    psc_code = models.TextField()
-    psc_description = models.TextField()
     naics_code = models.TextField(blank=True, null=True)
     naics_description = models.TextField(blank=True, null=True)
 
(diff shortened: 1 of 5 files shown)


usaspending_api/awards/models_matviews.py
```diff
@@ -261,18 +245,10 @@ class Meta:
...
     action_date = models.DateField(blank=True, null=True)
     fiscal_year = models.IntegerField()
     type = models.TextField(blank=True, null=True)
     pulled_from = models.TextField()
-    awarding_agency_id = models.TextField(blank=True, null=True)
-    awarding_agency_name = models.TextField(blank=True, null=True)
-    awarding_agency_abbr = models.TextField(blank=True, null=True)
-    funding_agency_id = models.TextField(blank=True, null=True)
-    funding_agency_name = models.TextField(blank=True, null=True)
-    funding_agency_abbr = models.TextField(blank=True, null=True)
```
dpb-bah: Looks good, is there a list or some documentation you're referring to for which fields should be removed?

tony-sappe: I don't have a nice list, the cols are comments in the matview SQL located in usaspending_api/database_scripts/matviews/

dpb-bah: That's good enough, just so we have something to refer to for this PR.

PR #821, opened 2017-12-19
Cached bulk downloads

dpb-bah:
Fixes bug where if someone submits the same Bulk Download requested earlier that day, it will refer to that instead of processing a new one.

- [x] [Monthly Download PR](https://github.com/fedspendingtransparency/usaspending-api/pull/816) is merged
- [x] Backend review

### DIFF
--- a/usaspending_api/bulk_download/v2/views.py
+++ b/usaspending_api/bulk_download/v2/views.py
@@ -4,6 +4,7 @@ import json
 import logging
 import os
 import pandas as pd
+import datetime
 from collections import OrderedDict
 
 import django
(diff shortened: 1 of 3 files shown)


usaspending_api/bulk_download/v2/views.py
```diff
@@ -160,10 +173,40 @@ def post(self, request):
 
         # create download job in database to track progress. Starts at 'ready'
         # status by default.
-        download_job = BulkDownloadJob(job_status_id=JOB_STATUS_DICT['ready'],
-                                       file_name=timestamped_file_name)
+        download_job_kwargs = {'job_status_id': JOB_STATUS_DICT['ready'],
+                               'json_request': json.dumps(order_nested_object(json_request)),
+                               'file_name': timestamped_file_name}
+        award_levels = json_request['award_levels']
+        award_types = json_request['filters']['award_types']
```
tony-sappe: I saw the check for required parameters on lines 152-154. However I didn't see anything for 'filters' dict keys.

usaspending_api/bulk_download/v2/views.py
```diff
@@ -152,6 +153,18 @@ def post(self, request):
             if required_param not in json_request:
                 raise InvalidParameterException('{} parameter not provided'.format(required_param))
 
+        # TODO: Refactor with the bulk_download method in populate_monthly_files.py
+        # Check if the same request has been called today
+        current_date = datetime.date.today()
```
tony-sappe: Its minor, but it is good practice to explicitly use UTC time on servers. This is documented to be local time

dpb-bah: Definitely can do.

usaspending_api/bulk_download/v2/views.py
dpb-bah: That gets checked via 
`csv_sources = self.get_csv_sources(json_request)` on line 168 which calls 
`queryset = self.process_filters(json_request['filters'], award_level)` on line 451 which calls
```
for required_param in ['award_types', 'agency', 'date_type', 'date_range']:
            if required_param not in filters:
                raise InvalidParameterException('{} filter not provided'.format(required_param))
``` on lines 377-379

tony-sappe: thanks

PR #822, opened 2017-12-19
Add idv contracts to the contracts count

russellbodine:


usaspending_api/search/v2/views/search.py
```diff
@@ -698,21 +698,30 @@ def post(self, request):
...
+
         # DB hit here
         for award in queryset:
-            result_key = award['category'].replace(' ', '_')
-            result_key += 's' if result_key not in ['other', 'loans'] else ''
+            if award['category'] not in categories.keys():
+                result_key = 'other'
+            else:
+                result_key = categories[award['category']]
             results[result_key] = award['category_count']
```
shinson: Instead of `=` it should be `+=`, Just having an equal sign replaces the old value with the new one. For instance, the end result for FY 2016 is 
```
"results": {
		"contracts": 1689409,
		"grants": 626707,
		"direct_payments": 1467337,
		"loans": 192531,
		"other": 47707
	}
```
Other should equal 47707+ 611535, 47707 is just the None value

usaspending_api/search/v2/views/search.py
```diff
@@ -698,21 +698,30 @@ def post(self, request):
...
+            'loans': 'loans',
+            'other': 'other'
+        }
+
         # DB hit here
         for award in queryset:
-            result_key = award['category'].replace(' ', '_')
-            result_key += 's' if result_key not in ['other', 'loans'] else ''
+            if award['category'] not in categories.keys():
+                result_key = 'other'
```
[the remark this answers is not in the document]
tony-sappe: Good find. We only accounted for one matview

usaspending_api/search/v2/views/search.py
tony-sappe: will fix

PR #824, opened 2017-12-20
Hotfix: list monthly downloads endpoint api updates

dpb-bah:
- Including 'all' functionality for list monthly downloads endpoint
- Including 'type' in list monthly downloads endpoint

### DIFF
--- a/usaspending_api/bulk_download/v2/views.py
+++ b/usaspending_api/bulk_download/v2/views.py
@@ -315,13 +315,19 @@ class ListMonthylDownloadsViewset(APIView):
         downloads = (BulkDownloadJob.objects.filter(**bulk_download_filters)
                      .annotate(max_updated_date=Max('update_date'))
                      .filter(update_date=F('max_updated_date'))
-                     .values('file_name', 'end_date', updated_date=Cast('update_date', DateField()),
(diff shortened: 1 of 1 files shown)


usaspending_api/bulk_download/v2/views.py
```diff
@@ -307,6 +307,8 @@ def post(self, request):
                           'end_date': '{}-09-30'.format(fiscal_year)}
             bulk_download_filters.update(date_range)
         if agency_id:
+            if agency_id == 'all':
+                agency_id = None
```
tony-sappe: I don't fully understand the decision to set agency_id to None if it is "all" and add it to the bulk_download_filters, but if the agency_id is originally None, it isn't added to the bulk_download_filters. Would it be better to do something like this if statement?
`if agency_id and agency_id != 'all':`

dpb-bah: Yeah that's the problem of the wording here. There's a difference between "get the monthly files that represent all the agencies for a fiscal year or a fiscal type" (`all_contracts_2017`, `all_assistance_2017`),  vs "get all agency files" (`dod_contracts_2017`, `doe_contracts_2017`,`all_contracts_2017`,`all_assistance_2017`). So to handle this difference, if `agency='all'`, it will only include the "all" files whereas removing it will include the all files as well as the other agency files.

PR #827, opened 2017-12-21
Promote Dev stable to Staging

nmonga91:
Sprint 54


tony-sappe:
I examined the files I am familiar with and edited. Need more eyes on this, but the part I reviewed is good

PR #828, opened 2017-12-21
Improve subtier agency search performance

tony-sappe:
Subtier funding and awarding agency names have been added to 4 of the "summary" matviews to produce faster queries than going to the larger matviews for simpler filters.

I also removed the unnecessary agency_ids from the same matviews since they have not been used anywhere in the code for some time.

### Example performance diff (non-django cached)
**FY2018, FY2017, Subagency USACE - Over Time**
- Original: [timeout, 46s, 12s] (due to DB cache warming)
- New: 1s

### Before merge
- [x] Run new matview SQL on dev

### Edit
Included hotfix to fix county vs country SQL column mapping


tony-sappe:
Thanks @dpb-bah! You are correct about the action_date to fiscal_year conversion in code. That was found to be more performant than action_date range(s). It only occurs when the action_dates match fiscal year start/end dates.

PR #838, opened 2017-12-22
Hotfix: Fix for location country code between matview and normal filters

dpb-bah:
For baby download, the matviews are not used and thus the `loc_dict` should use the `location_country_code` instead of the `country_code`

### DIFF
--- a/usaspending_api/awards/v2/filters/location_filter_geocode.py
+++ b/usaspending_api/awards/v2/filters/location_filter_geocode.py
@@ -72,13 +72,11 @@ def get_fields_list(scope, field_value, loc_dict):
 
 def return_query_strings(use_matview):
     # Returns query strings according based on mat view or database
-    loc_dict = LOCATION_MAPPING
+    loc_dict = LOCATION_MAPPING.copy()
     q_str = '{0}__{1}__in'
 
     if use_matview:
         q_str = '{0}_{1}__in'
         loc_dict['country'] = 'country_code'
-    else:
-        loc_dict['country'] = 'location_country_code'
 
     return q_str, loc_dict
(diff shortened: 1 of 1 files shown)


usaspending_api/awards/v2/filters/location_filter_geocode.py
```diff
@@ -78,5 +78,7 @@ def return_query_strings(use_matview):
     if use_matview:
         q_str = '{0}_{1}__in'
         loc_dict['country'] = 'country_code'
+    else:
+        loc_dict['country'] = 'location_country_code'
```
nmonga91: what does `location_country_code` refer to here?

dpb-bah: It refers to what the country code column name in the table the function is working from. Matviews use `country_code` while baby download uses the transaction filter, not using the matview, so it goes through the location table which uses `location_country_code`. @shinson  was the one who discovered the fix and can explain it better. Without said fix, we found some of the baby downloads were returning 500s.

tony-sappe: Look at lines 5-6 on that file. `location_country_code` is the default value for loc_dict['country'] and doesn't need to be set.

dpb-bah:
Agreed to your point @tony-sappe and @nmonga91 . I looked into why that dictionary value was changing and realized that we're working off of the global `LOCATION_MAPPING` and not a *copy* of it like the code implicitly wants to do. So when someone uses a geocode filter with matview and the next person doesn't use the matview, they will be working with different values. I've included the fix and tested it locally with the frontend.

PR #839, opened 2017-12-26
Reducing code: combining the API search filter logic for matviews

tony-sappe:
After renaming the matview columns to match there are no differences between the search filters.

Can be tested on Dev DB


dpb-bah:
Can confirm, doing a diff between `matview_award` and `matview_transaction` in dev only yields the changes that are covered in this PR.

dpb-bah:
Tested locally with matview changes, ready to merge

PR #847, opened 2017-12-29
Bugfix: List Downloads pulls from S3 Bucket

dpb-bah:
List Downloads endpoint currently finds the downloads via the BulkDownloadJob table which stores the information needed. While useful and quick, this restricts the list download endpoint to only provide downloads if the `populate_monthly_download` script was run on that environment's database. This bypasses the download job table and uses the data stored in the filename from the S3 bucket to filter and provide the files requested, regardless of environment.


usaspending_api/bulk_download/v2/views.py
```diff
@@ -337,42 +337,66 @@ class ListMonthylDownloadsViewset(APIView):
 
     s3_handler = S3Handler(name=settings.MONTHLY_DOWNLOAD_S3_BUCKET_NAME, region=settings.BULK_DOWNLOAD_AWS_REGION)
 
+    @cache_response()
```
tony-sappe: How long does this request take when it doesn't hit the cache?

dpb-bah: Depends on the filters. If no filters are provided, it takes ~60s. If the fiscal year is provided, it takes ~5s. If the fiscal year and agency are provided, it takes ~0.5s. This is due to the major bottlenecks being getting the filenames from the s3 bucket (which speeds up faster when a prefix is provided, larger prefixes -> faster results) and getting the agency name/acronym from the results returned as it checks with the database. Those longer cases though will not show up as the front end currently *requires* all three filters so the average will be 0.5s for the endpoint without caching.

usaspending_api/bulk_download/v2/views.py
```diff
@@ -337,42 +337,71 @@ class ListMonthylDownloadsViewset(APIView):
...
+            name_data = re.findall('(.*)_(.*)_(.*)_Full_(.*)\.zip', name)[0]
+            agency_name = None
+            agency_abbr = None
+            agency_cgac = name_data[1]
+            if agency_cgac != 'all':
+                agency = ToptierAgency.objects.filter(cgac_code=agency_cgac).values('name', 'abbreviation')
+                if agency:
+                    agency_name = agency[0]['name']
+                    agency_abbr = agency[0]['abbreviation']
+            updated_date = '-'.join([name_data[3][:4], name_data[3][4:6], name_data[3][6:]])
```
tony-sappe: Please add comments for this. Perhaps an example of the final string

PR #849, opened 2017-12-29
Fix idv column values

russellbodine:


usaspending_api/search/v2/views/search.py
```diff
@@ -654,23 +654,22 @@ def post(self, request):
         for award in limited_queryset[:limit]:
             row = {"internal_id": award["award_id"]}
 
-            if award['type'] in contract_type_mapping:
-                for field in fields:
-                    row[field] = award.get(award_contracts_mapping.get(field))
-            elif award['type'] in loan_type_mapping:  # loans
+            if award['type'] in loan_type_mapping:  # loans
```
tony-sappe: shouldn't this remain `elif`

PR #850, opened 2018-01-03
DS-2052: sync CARS-TAS with Broker

michaeldhess:
https://federal-spending-transparency.atlassian.net/browse/DS-2052

Acceptance Criteria:
- API data loader pulls the TAS file from S3 in order to reference the same file Broker relies on

### DIFF
--- a/usaspending_api/common/threaded_data_loader.py
+++ b/usaspending_api/common/threaded_data_loader.py
@@ -11,6 +11,7 @@ import csv
 import queue
 import sys
 import json
+import smart_open
 
 
 # This class is a threaded data loader
@@ -64,7 +65,7 @@ class ThreadedDataLoader():
 
     # Loads data from a file using parameters set during creation of the loader
     # The filepath parameter should be the string location of the file for use with open()
-    def load_from_file(self, filepath, encoding='utf-8'):
+    def load_from_file(self, filepath, encoding='utf-8', bucket_name=None):
         self.logger.info('Started processing file ' + filepath)
         # Create the Queue object - this will hold all the rows in the CSV
         row_queue = JoinableQueue(500)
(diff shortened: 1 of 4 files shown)


usaspending_api/common/threaded_data_loader.py
```diff
@@ -112,6 +116,17 @@ def load_from_file(self, filepath, encoding='utf-8'):
 
         self.logger.info('Finished processing all rows')
 
+    def csv_file_to_queue(self, csv_file, row_queue):
+        reader = csv.DictReader(csv_file)
+        temp_row_queue = row_queue
```
dpb-bah: Is this meant to be `temp_row_queue = row_queue.copy()`?

usaspending_api/references/management/commands/loadtas.py
```diff
@@ -74,16 +88,12 @@ def handle(self, *args, **options):
...
-                                                                         row["BPOA"],
-                                                                         row["EPOA"],
-                                                                         row["MAIN"],
+        return TreasuryAppropriationAccount.generate_tas_rendering_label(row["ATA"], row["Agency AID"], row["A"],
+                                                                         row["BPOA"], row["EPOA"], row["MAIN"],
                                                                          row["SUB"])
 
     def skip_and_remove_financing_tas(self, row, instance):
-        if row["Financial Indicator Type2"] == "F":
+        if row["financial_indicator_type2"] == "F":
```
dpb-bah: I'm assuming the column name got lowered?

usaspending_api/references/management/commands/loadtas.py
```diff
@@ -15,9 +17,17 @@ class Command(BaseCommand):
     logger = logging.getLogger('console')
 
     def add_arguments(self, parser):
-        parser.add_argument('file', nargs=1, help='the file to load')
+        parser.add_argument('location', nargs=1, help='The location of the file to load')
 
     def handle(self, *args, **options):
+        is_remote_file = len(options['location'][0].split('.')) == 1
+        if is_remote_file:
+            s3connection = boto.s3.connect_to_region(os.environ.get('AWS_REGION'))
```
dpb-bah: I don't know if using an environmental variable or using a django setting (like `CSV_AWS_REGION` or `BULK_DOWNLOAD_AWS_REGION`) would be more useful to those running the script. I can run it by the Ops team if that hasn't been discussed.

dpb-bah:
Tested script locally to pull from S3 👍

usaspending_api/references/management/commands/loadtas.py
michaeldhess: This was the solution that the Ops team and I decided on, instead of using an env variable or hardcoding the bucket name into the script

usaspending_api/references/management/commands/loadtas.py
michaeldhess: Yes, the new files has it as `financial_indicator_type2`

usaspending_api/common/threaded_data_loader.py
michaeldhess: Correct, I can change this

dpb-bah: For reference, Mike and I found out it was more difficult making copies of JoinableQueue (tried .copy and the copy library which didn't work out). The script worked as intended without the copying so we're disregarding this point.

PR #855, opened 2018-01-04
Bugfix: Display 'All' Agency Name monthly downloads

dpb-bah:
Previously it was blank, this sets it to 'All'. Tested with local dev frontend.


usaspending_api/bulk_download/v2/views.py
```diff
@@ -394,6 +394,8 @@ def post(self, request):
                 if agency:
                     agency_name = agency[0]['name']
                     agency_abbr = agency[0]['abbreviation']
+            else:
+                agency_name = 'All'
```
tony-sappe: An empty/None `agency_abbr` is OK?

dpb-bah: Correct, I confirmed that @ebdabbs.

PR #861, opened 2018-01-09
DEV-104: maj obj class 90 is Other, others are Unknown

catherinedevlin:
Fixes https://federal-spending-transparency.atlassian.net/browse/DEV-104

This fixes the bug in the submission loader that created the ObjectClasses with major_object_class='00', major_object_class_name='Other'.

In addition, a one-time SQL fixer should fix the existing ObjectClasses:

`UPDATE object_class SET major_object_class_name = 'Unknown' WHERE major_object_class = '00';`

This will change only 6 rows and will take essentially zero time.

 It can be run before or after this PR is merged.


catherinedevlin:
Thanks, @shinson , I put the SQL in a migration.  Confirmed that it runs in negligible time against the validation database and updates the data.

PR #875, opened 2018-01-12
Adding new Award Matview generation SQL and database models

nmonga91:
JIRA Story: https://federal-spending-transparency.atlassian.net/browse/DEV-27


usaspending_api/database_scripts/broker_matviews/award_category_table.sql
```diff
@@ -0,0 +1,30 @@
+create table if not exists award_category
+    (
+        type_code text not null,
+        type_name text not null
+    );
+
+create index if not exists award_category_type_code_idx on award_category using btree (type_code);
```
tony-sappe: Typically it best practice to create the index after the table is populated with data. Not an issue with this small table

usaspending_api/database_scripts/broker_matviews/award_matview.sql
```diff
@@ -0,0 +1,1880 @@
+drop materialized view if exists award_matview;
```
tony-sappe: This is problematic for materialized views based upon this materialized view. Either you couldn't drop it, or you would need to do a cascading drop which would delete the API matviews created from this as well.

usaspending_api/database_scripts/broker_matviews/transaction_matview.sql
```diff
@@ -0,0 +1,262 @@
+drop materialized view if exists transaction_matview;
+
+create materialized view transaction_matview as
+(
+select
+    *,
+    fy(action_date) as fiscal_year
```
tony-sappe: is fy() already defined by django migrations?

usaspending_api/database_scripts/broker_matviews/broker_indexes.sql
```diff
@@ -0,0 +1,29 @@
+-- FPDS Indexes
```
tony-sappe: What is the justification for storing broker indexes in this repo?

usaspending_api/awards/migrations/0013_add_unmanaged_tables.py
```diff
@@ -0,0 +1,387 @@
...
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('type_code', models.TextField(blank=True)),
+                ('type_name', models.TextField(blank=True)),
+            ],
+            options={
+                'db_table': 'award_category',
+                'managed': False,
+            },
+        ),
+        migrations.CreateModel(
```
tony-sappe: Will there be issues with django "creating" but not managing these materialized views?

usaspending_api/awards/models_matviews.py
```diff
@@ -334,3 +334,318 @@ class SummaryTransactionMonthView(models.Model):
     class Meta:
         managed = False
         db_table = 'summary_transaction_month_view'
+
+
+class AwardMatview(models.Model):
+    generated_unique_award_id = models.TextField(blank=True, null=True)
```
tony-sappe: This col should never be null, but does it make sense to make every single col completely accurate? It might be better just to remove the kwargs from these classes since Django won't manage/enforce them. And future devs won't be confused by mismatched definitions

nmonga91: I like the suggestion of removing the kwargs on unmanaged models like this. Let's move forward with that. Will make that change.

usaspending_api/awards/migrations/0013_add_unmanaged_tables.py
nmonga91: There shouldn't be a problem with this. It's just doing some accounting by having it in the migrations. Doesn't affect anything outside of the code.

usaspending_api/database_scripts/broker_matviews/broker_indexes.sql
nmonga91: That should be removed.

usaspending_api/database_scripts/broker_matviews/transaction_matview.sql
nmonga91: This is in `common/helpers.py` as `FY_PG_FUNCTION_DEF`. It's run as part of the initial migration in the `accounts` django app

usaspending_api/database_scripts/broker_matviews/award_matview.sql
nmonga91: Agreed. This was fine for testing but we'll need a better way to manage these once other dependencies are incorporated. We can leave it in for now and handle this iteratively as we add the dependencies.

usaspending_api/database_scripts/broker_matviews/award_category_table.sql
nmonga91: Will move it down.

usaspending_api/database_scripts/broker_matviews/agency_lookup_matview.sql
```diff
@@ -0,0 +1,25 @@
+drop materialized view if exists agency_lookup;
+
+create materialized view agency_lookup as
```
tony-sappe: very minor: For the existing "API" matview SQL scripts all keywords are uppercase. Postgres doesn't care. We should keep it consistent as much as we can.

usaspending_api/database_scripts/broker_matviews/transaction_matview.sql
tony-sappe: 👍

usaspending_api/database_scripts/broker_matviews/agency_lookup_matview.sql
nmonga91: We can alter this. Lowercase was just a personal preference but Postgres docs also use uppercase so we should be consistent with that in addition to how we're doing it in other areas of our codebase.

usaspending_api/database_scripts/broker_matviews/award_matview.sql
tony-sappe: I would have agreed with you two months ago. After wrestling with my own matview creations, I think it would be better to make this change earlier rather than later. 

What I had to do was create the new materialized view with a temporary name. After it is created the existing matview is renamed and the new matview name is changed from the temporary name to the desire final name.

nmonga91:
@tony-sappe definition SQL for business category functions live in the Broker database. There will be future stories to add them to the Broker migrations in the upcoming sprint.

PR #877, opened 2018-01-16
Fixing columns for new matviews

nmonga91:
Resolving missing column names and alphabetizing columns for new matviews for ease of use


usaspending_api/awards/models_matviews.py
```diff
@@ -558,87 +518,77 @@ class Meta:
...
+    legal_entity_state_code = models.TextField()
+    legal_entity_state_name = models.TextField()
     legal_entity_zip5 = models.TextField()
-    legal_entity_congressional = models.TextField()
-
-    # place of performance
+    naics = models.TextField()
+    naics_description = models.TextField()
+    original_loan_subsidy_cost = models.DecimalField()
+    parent_award_id = models.TextField()
```
tony-sappe: Perhaps better to make this `parent_award_piid`

nmonga91: updated!

PR #881, opened 2018-01-17
DEV-37: Add logging to API endpoints

shinson:
Added logging to endpoints through a Django middleware to process API requests.
- Unified all requests into one server.log file
- Updated logging config to remove unnecessary log files and update format/config of server.log
- Removed original logging mixin from endpoints to avoid duplication

Dev Story: https://federal-spending-transparency.atlassian.net/browse/DEV-37

- [x] Backend Review
- [x] Dev Ops Review


usaspending_api/common/logging.py
```diff
@@ -0,0 +1,112 @@
...
+
+        status_code = response.status_code
+
+        self.log["status_code"] = status_code
+        self.log["response_ms"] = self.get_response_ms()
+        self.log["traceback"] = None
+
+        if 100 <= status_code < 400:
+            self.log["status"] = 'INFO'
+            self.log["timestamp"] = now().strftime('%d/%m/%y %H:%M:%S')
```
tony-sappe: Should the timestamp value be moved out of this if statement? 
404 method can overwrite it if necessary

usaspending_api/common/logging.py
```diff
@@ -0,0 +1,112 @@
...
+    def handle_404(self, request, exception):
+        """Logs 404 status"""
+        self.log["status_code"] = 404
+        self.log["response_ms"] = self.get_response_ms()
+        self.log["status"] = 'WARNING'
+        self.log["timestamp"] = now().strftime('%d/%m/%y %H:%M:%S')
+
+        self.server_logger.warning(self.get_message_string(), extra=self.log)
+
+    def handle_500(self, request):
```
tony-sappe: Does the middleware call this directly?

usaspending_api/common/logging.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+
+        if isinstance(exception, http.Http404):
+            return self.handle_404(request, exception)
+        else:
+            return self.handle_500(request)
+
+    def get_response_ms(self):
+        """Returns time elapsed from request to response/exception"""
+        duration = now() - self.start
```
tony-sappe: This might be good enough. Look at using Py3's performance counter https://docs.python.org/3/library/time.html#time.perf_counter to see if this is more accurate for calculating duration

usaspending_api/settings.py
```diff
@@ -198,49 +198,24 @@
...
-            'formatter': 'specifics'
-        },
-        'events_file': {
+        'server': {
             'level': 'INFO',
             'class': 'logging.handlers.RotatingFileHandler',
-            'filename': os.path.join(BASE_DIR, 'usaspending_api/logs/events.log'),
-            'maxBytes': 1024*1024*2,  # 2 MB
+            'filename': os.path.join(BASE_DIR, 'usaspending_api/logs/server.log'),
+            'maxBytes': 1024*1024*20,  # 20 MB
```
tony-sappe: surprised this isn't flagged as PEP8 issue

usaspending_api/common/logging.py
shinson: Timestamp is required field for all calls. 404 status calls are not included in this condition since 404 status calls are logged as a warning while 100 -399 are logged as INFO

usaspending_api/common/logging.py
shinson: This gets called through the process_exception function which gets called when there's an exception thrown in the code.

usaspending_api/common/logging.py
shinson: Time matches more accurately with the response time the browser turns. Adding it to the next commit.

usaspending_api/settings.py
shinson: This value might be adjusted later. Makes it easier to update, but I can change it if you have a suggestion.

usaspending_api/common/logging.py
tony-sappe: right, thanks. I meant to add this comment on process_exception

usaspending_api/common/logging.py
tony-sappe: Does it make more sense to log the timestamp on request instead of response?

PR #882, opened 2018-01-17
Data: Award Matview IDC Columns and Primary Key

dpb-bah:
- Adding idc columns to award matview
- Adding primary key to award matview model


usaspending_api/awards/models_matviews.py
```diff
@@ -355,7 +355,7 @@ class Meta:
 
 
 class AwardMatview(models.Model):
-    generated_unique_award_id = models.TextField()
+    generated_unique_award_id = models.TextField(primary_key=True)
```
tony-sappe: Was this for testing JOINS in the future?

dpb-bah: This was a solution to querying the matview with Django's ORM. Without it, doing `AwardMatview.objects.filter(...)` would error out as the ORM automatically includes and ID column if no `primary_key=True` is provided.

tony-sappe: Good to know. That might have been useful for me two weeks ago

PR #898, opened 2018-01-22
adding file C-D linkage to submission loader

russellbodine:


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -595,6 +595,7 @@ def load_file_c(submission_attributes, db_cursor, award_financial_frame):
     total_rows = award_financial_frame.shape[0]
     start_time = datetime.now()
     awards_touched = []
+    awards = Award.objects  # this stops the property lookup each iteration, saving 3+ seconds every 100 rows!
```
catherinedevlin: Huh, I guess I vaguely realized that was happening, but it hadn't occurred to me that it could add up to measurable time.  :the-more-you-know:

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -642,11 +642,41 @@ def load_file_c(submission_attributes, db_cursor, award_financial_frame):
             'program_activity': row.get('program_activity'),
         }
 
+        # individual FILE C D linkage
+        if not created and total_rows < 100000:
+            kwargs = {}
+            kwargs['recipient_id__isnull'] = False
+            if row.get('piid') is not None:
```
catherinedevlin: I'm trying to come up with a more elegant way for a series of "if not none then add to dict" steps like this, but I don't want to hold up your PR for it.  I'll let you know later if I come up with one.

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -642,11 +642,41 @@ def load_file_c(submission_attributes, db_cursor, award_financial_frame):
...
+                kwargs['piid'] = row.get('piid')
+            if row.get('parent_award_id') is not None:
+                kwargs['parent_award__piid'] = row.get('parent_award_id'),
+            if row.get('fain') is not None:
+                kwargs['fain'] = row.get('fain')
+            if row.get('uri') is not None:
+                kwargs['uri'] = row.get('uri')
+
+            award_queryset = awards.filter(**kwargs).values("id")[:2]
```
catherinedevlin: So... is it ever possible for this to return >1 row?  In which case, the query isn't ordered, so the one assigned will be picked arbitrarily... is that OK?

russellbodine: If the size of the queryset is not one, then it fails line 662 and we dont use the queryset.   Therefore the queryset doesnt have to be ordered.

PR #899, opened 2018-01-23
Feature transactions category counts

miahunsicker:
https://github.com/fedspendingtransparency/usaspending-api/wiki#spending-by-transaction-count

- Adding endpoint that gets counts for keyword search
- Adding function for IDs for "baby" download
- Changing piid and fain for displaying on the front end for award id

#TODO - fix sorting for these fields


usaspending_api/search/v2/views/search.py
```diff
@@ -838,3 +842,20 @@ def post(self, request):
 
         # build response
         return Response({"results": results})
+
+
+class SpendingByTransactionCountvisualizaitonViewSet(APIView):
```
tony-sappe: Visualization should start with uppercase "V"

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -66,8 +71,76 @@ def search_transactions(filters, fields, sort, order, lower_limit, limit):
...
+            'query_string': {
+                'query': keyword
+                }
+            }
+    }
+    try:
+        response = CLIENT.search(index=index_name, body=query)
+        return response['hits']['total']
+    except Exception:
+        return -1
```
tony-sappe: It would be better to log the exception instead of simply returning -1. I would discuss with @ebdabbs to see if a -1 for some of the results would be acceptable. I gut tell me it is better to return a value of 0

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -66,8 +71,76 @@ def search_transactions(filters, fields, sort, order, lower_limit, limit):
...
+
+def spending_by_transaction_count(filters):
+    keyword = filters['keyword']
+    response = {}
+    for category in award_categories:
+        response[category] = get_total_results(keyword, category)
+    return response
+
+
+def search_keyword_id_list_all(keyword):
```
tony-sappe: Would it be easier for both this function and the download function if this was a generator function instead of iteratively appending the transaction ids to a (potentially very large) python list?

usaspending_api/search/v2/views/search.py
```diff
@@ -838,3 +842,20 @@ def post(self, request):
...
+
+    @cache_response()
+    def post(self, request):
+        """
+            TODO
```
tony-sappe: Either fill in the doc string or remove it

usaspending_api/search/v2/elasticsearch_helper.py
MelvinDunn: Yep, Daniel wants an obnoxiously large python list returned. It's not a great method but as long as the heap memory on the server can handle a request like that whenever someone wants to potentially download something that's fine. Can you let me know what the runtime is for this function.

tony-sappe: @dpb-bah Can you test this function and comment about runtime?

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -66,8 +71,65 @@ def search_transactions(filters, fields, sort, order, lower_limit, limit):
...
+    query = {
+        "query": {
+            "query_string": {
+                "query": keyword
+            }
+        }, "size": DOWNLOAD_QUERY_SIZE}
+    try:
+        responses = CLIENT.search(index=index_name, body=query, timeout='3m')
+    except Exception as es1:
+        return es1
```
tony-sappe: Returning the exception isn't a good practice. Instead could the exception be logged and a None or empty list be returned?

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -66,8 +71,65 @@ def search_transactions(filters, fields, sort, order, lower_limit, limit):
...
+    transaction_id_list = []
+    try:
+        for response in (responses['hits']['hits']):
+            list_item = (response['_source']['transaction_id'])
+            transaction_id_list.append(list_item)
+        print(len(transaction_id_list))
```
tony-sappe: Remove print statement

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -5,21 +5,25 @@
...
-        import TRANSACTIONS_LOOKUP, award_type_mapping
+        import TRANSACTIONS_LOOKUP, award_type_mapping, award_categories
 
 
 logger = logging.getLogger('console')
 ES_HOSTNAME = settings.ES_HOSTNAME
 TRANSACTIONS_INDEX_ROOT = settings.TRANSACTIONS_INDEX_ROOT
+DOWNLOAD_QUERY_SIZE = settings.DOWNLOAD_QUERY_SIZE
 CLIENT = Elasticsearch(ES_HOSTNAME)
+TRANSACTION_ID_SIZE = settings.TRANSACTION_ID_SIZE
```
tony-sappe: Wasn't `TRANSACTION_ID_SIZE` removed from settings.py?

usaspending_api/search/v2/elasticsearch_helper.py
MelvinDunn: Noted, empty list returned and logged.

usaspending_api/search/v2/elasticsearch_helper.py
MelvinDunn: removed

usaspending_api/search/v2/elasticsearch_helper.py
MelvinDunn: yup

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -91,8 +91,9 @@ def get_total_results(keyword, index_name):
     try:
         response = CLIENT.search(index=index_name, body=query)
         return response['hits']['total']
-    except Exception as es1:
-        return es1
+    except Exception:
+        logging.exception("There was an error connecting to the ElasticSearch instance.")
```
tony-sappe: will the ES client raise descriptive error messages? If yes, it would be good to log the exception msg for tracking down problems on deployed instances

PR #907, opened 2018-01-25
transaction sum

MelvinDunn:
making a new api endpoint for the transaction sum


MelvinDunn:
https://github.com/fedspendingtransparency/usaspending-api/wiki#transaction-sum - gist

PR #909, opened 2018-01-26
Feature delete fabs rows (DEV-336)

catherinedevlin:
This one-time fixer command should close DEV-336.  It took about 45 minutes to run to completion on sandbox.

I'm not sure how running it through Jenkins can work, since it requires manually editing usaspending_api/database_scripts/broker_matviews/broker_server.sql


usaspending_api/common/management/commands/resave_instances.py
```diff
@@ -29,10 +29,11 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument('model', nargs='+', help='app and model to be re-saved')
+        parser.add_argument(
```
catherinedevlin: OK, I hadn't specifically planned to include these changes in this PR.  Oops.  But!  I do want them anyway - they just add an optional --id argument to `resave_instances`, which is handy for trying it out against just a row or a few rows.

usaspending_api/common/helpers.py
```diff
@@ -225,6 +227,24 @@ def fy(raw_date):
     return result
 
 
+@contextlib.contextmanager
```
shinson: This will help with future scripts.

usaspending_api/broker/management/commands/fix_deleted_fabs.py
```diff
@@ -0,0 +1,83 @@
...
+
+from django.core.management.base import BaseCommand
+from django.db import connections, transaction
+from django.db.utils import ProgrammingError
+
+from usaspending_api.awards.models import TransactionNormalized
+from usaspending_api.common.helpers import timer
+
+logger = logging.getLogger('console')
+exception_logger = logging.getLogger("exceptions")
```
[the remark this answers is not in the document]
catherinedevlin: Thank you, you're right!  Fixing that now.

PR #911, opened 2018-01-29
Feature bulk download keyword

dpb-bah:
AC:
- Update bulk download endpoint to include overriding `keyword` argument and provide defaults for optional arguments. Keyword argument uses elastic search to find specific transactions related to that keyword.

Note that this modifies the bulk download process to allow longer queries to be done via file instead of pasting the query in the command line.

- [x] [Bulk Download Gist](https://gist.github.com/dpb-bah/d64726d3a1555552736adf8c84651dcd)  updated and approved
- [x] [Transaction Sum PR](https://github.com/fedspendingtransparency/usaspending-api/pull/910) merged
- [x] Backend Reviewed


usaspending_api/bulk_download/v2/views.py
```diff
@@ -176,26 +197,24 @@ def post(self, request):
...
-        award_levels = json_request['award_levels']
-        award_types = json_request['filters']['award_types']
-        agency = json_request['filters']['agency']
-        sub_agency = json_request['filters'].get('sub_agency', None)
-        start_date = json_request['filters']['date_range'].get('start_date', None)
-        end_date = json_request['filters']['date_range'].get('end_date', None)
-        date_type = json_request['filters']['date_type']
 
-        for award_level in award_levels:
+        for award_level in json_request['award_levels']:
```
nmonga91: +1 for the simplification here

usaspending_api/bulk_download/v2/views.py
```diff
@@ -416,13 +435,33 @@ class BulkDownloadAwardsViewSet(BaseDownloadViewSet):
...
+            while not elastic_search_complete:
+                try:
+                    transaction_ids = elasticsearch_helper.get_transaction_ids(keyword=keyword, size=size)
+                    # flatten ids
+                    transaction_ids = list(itertools.chain.from_iterable(transaction_ids))
+                    elastic_search_complete = True
+                except (TransportError, ConnectionError) as e:
+                    logger.error(e)
+                    transaction_ids = []
+                    size = size//10
```
nmonga91: So if this errors out for any reason (timeout, connection issues but not a generic exception), we'll decrease the request by a factor of 10 and retry. We should make the relevant stake holders aware of this but this is probably a good policy to employ.

usaspending_api/bulk_download/v2/views.py
```diff
@@ -456,19 +495,19 @@ def process_filters(self, filters, award_level):
         # Get the date ranges
         try:
             date_range_filters = {}
-            if 'start_date' in filters['date_range'] and filters['date_range']['start_date']:
+            if filters['date_range']['start_date']:
                 date_range_filters['{}__gte'.format(date_attribute)] = filters['date_range']['start_date']
-            if 'end_date' in filters['date_range'] and filters['date_range']['end_date']:
+            if filters['date_range']['end_date']:
```
nmonga91: What happens when a nested field isn't populated? For example, the API request has `date_range` but doesn't have the nested `start_date` and `end_date` values? Currently the values are set to `None`, that would mean the date range filters wouldn't be applied and the user could potentially get everything, right?

usaspending_api/bulk_download/v2/views.py
dpb-bah: So this is not decreasing the max size of the transactions, but rather the size of the chunks elastic is returning. So if this errors out, the elastic search will not ask for 500,000 once but in ten requests of 50,000, and so on.

usaspending_api/bulk_download/v2/views.py
dpb-bah: Sure, I moved the argument handling beforehand in lines 232-251 (data range specifically at lines 249-251). If `date_range` isn't provided, it'll set date_range to `{'start_date': None, 'end_date': None}`. Similarly, if `date_range` is provided as an empty dictionary, it'll set it to the same. If either `start_date` or `end_date` are provided, they'll be populated accordingly. Otherwise, they're set to `None`.

usaspending_api/search/v2/urls_search.py
```diff
@@ -17,4 +17,5 @@
     url(r'^spending_by_transaction_count', SpendingByTransactionCountVisualizaitonViewSet.as_view()),
     url(r'^spending_by_transaction', SpendingByTransactionVisualizationViewSet.as_view()),
     url(r'^transaction_spending_summary', TransactionSummaryVisualizationViewSet.as_view()),
+    url(r'^transaction_sum', SpendingByTransactionSumVisualizaitonViewSet.as_view())
```
tony-sappe: Please remove all this extraneous code pulled from an earlier commit of the `transaction_sum` PR

usaspending_api/bulk_download/v2/views.py
nmonga91: Gotcha, my original concern was they'd be able to get _everything_ without any dates set, but that isn't a possibility since we're always limiting to the first 500k records.

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -114,24 +114,65 @@ def search_keyword_id_list_all(keyword):
...
+        return None
+
+
+def spending_by_transaction_sum(filters):
+    keyword = filters['keyword']
+    return get_sum_aggregation_results(keyword)
+
+
+def get_transaction_ids(keyword, size=500):
+    '''returns a generator that
```
nmonga91: we should keep the docstrings for functions consistent with using `"""` instead of `'''`

PR #913, opened 2018-01-29
Bugfix dev 221 agency list - include FREC agencies

shinson:
[Related bug](https://federal-spending-transparency.atlassian.net/browse/DEV-221)

Updating the agency list dropdown to include FREC agencies.

Example Agencies:
- DOL
- Inter-American Foundation
- Pension Benefit Guaranty Corporation

FREC agencies and their subtiers are now visible in bulk download dropdown. 
Path: `/api/v2/bulk_download/list_agencies/`
Same endpoint also allows user to view FREC agencies and their monthly downloads.

Future development: We will move away from modified list once a central list is agreed upon.


usaspending_api/bulk_download/v2/views.py
```diff
@@ -255,19 +255,23 @@ def verify_requested_columns_available(sources, requested):
 
 class BulkDownloadListAgenciesViewSet(APIView):
     modified_agencies_list = os.path.join(settings.BASE_DIR,
-                                          'usaspending_api', 'data', 'agency_list_broker_s3.csv')
+                                          'usaspending_api', 'data', 'modified_authoritative_agency_list.csv')
```
dpb-bah: If the point of bringing back the `modified_authoritative_agency_list` is to include the FREC and IS_FREC columns, could we just use `authoritative_agency_list` so no agencies are left out?

shinson: I'm bringing back the `modified_authoritative_agency_list` until we come to a consensus on an alternative.

PR #926, opened 2018-02-01
Bulk Download Fixes (filters, deleting on failures)

dpb-bah:
- Refactored filters to catch cases when arguments are of the wrong type
- Included `date_range` filter constraint to 365 days
- Bulk Download deletes messages upon failure


usaspending_api/bulk_download/v2/views.py
```diff
@@ -219,31 +219,64 @@ def create_bulk_download_job(self, json_request, file_name):
...
-            json_request['filters'][field] = json_request['filters'].get(field, fields_defaults[field])
-        json_request['filters']['date_range'] = json_request['filters'].get('date_range', {})
-        json_request['filters']['date_range']['start_date'] = (json_request['filters']['date_range']
-                                                               .get('start_date', None))
-        json_request['filters']['date_range']['end_date'] = json_request['filters']['date_range'].get('end_date', None)
+            filters[field] = filters.get(field, fields_defaults[field])
+            if not isinstance(filters[field], type(fields_defaults[field])):
+                type_name = type(fields_defaults[field]).__name__
+                raise InvalidParameterException('{} parameter not provided as a {}'.format(field, type_name))
+        filters['date_range']['start_date'] = filters['date_range'].get('start_date', '')
```
nmonga91: is it intentional to overwrite what's in `filters` for the `date_range`? For the date_range key, we should also enforce they provide _something_. It shouldn't be empty since that could fail the checks below for the day difference calculation.

Empty strings would also pass this check `if not isinstance(filters['date_range'][date_range_param], str)` which we probably don't want it to.

nmonga91: This will still check the year difference below so it will fail if over 1 year between dates. If not dates provided, assumed user wanted all of time, and then the check for 1 year difference fails as a result.

PR #927, opened 2018-02-01
DEV-366 Subtier list bulk download bugfix

shinson:
Update '/api/v2/bulk_download/list_agencies/' to show missing subtiers. Some agencies contain multiple subtiers with the same name but different codes. The distinct method was removing these dupe subtiers that were listed in the modified csv list. 

It should not display duplicates since it mirrors all the subagencies in the `modified_authoratative_agency_list` which does not contain duplicates when I tested it.

bug link: https://federal-spending-transparency.atlassian.net/browse/DEV-366

- [x] Check for duplicate subtiers
- [ ] Backend review


dpb-bah:
So this is definitely doable. It definitely resolves the bug though it still doesn't address the core issue of the `modified_agencies_list`. Also something to note, we've already made the change for the bulk download to merge the data based on subtier **name** and not the **subtier code** as before. So if in the future there are two subtier agencies with the same name but different code, their data will be merged in the bulk download.

PR #941, opened 2018-02-05
Dev 416 allow quarter in spending explorer

alburde1:
- Add optional quarter filter to spending explorer so frontend can specify quarter as well as fiscal year
- Adjust how we're filtering so it actually accounts for fy not just quarter.
- Apply 45 day buffer to everything so users can't request a specific fy+quarter if the quarter hasn't been finished for over 45 days or a specific fy if it hasn't been started for a full quarter + 45 days.

- [x] Reviewed by backend
- [x] Frontend updated/checked so nothing breaks when this is merged
- [x] Gist reviewed by frontend https://gist.github.com/alburde1/102b6d851502086421c6762b63871bfa


usaspending_api/spending_explorer/v2/filters/type_filter.py
```diff
@@ -28,22 +29,34 @@ def type_filter(_type, filters):
     # Get fiscal_date and fiscal_quarter
     for key, value in filters.items():
         if key == 'fy':
-            if value is not None:
-                fiscal_date, fiscal_quarter = generate_last_completed_fiscal_quarter(fiscal_year=value)
-            else:
+            try:
+                fiscal_year = int(value)
```
shinson: Should this throw an error when a request fiscal year param is 'YY' such as `fy: '17'`?

alburde1: It never did before so I didn't want to add it now. If they want to search the year 17 I say we should let them. It'll give them no data

nmonga91: If we're allowing it, this portion of the gist is inaccurate then (under Error Messages):
```
{
    "details": "Incorrect fiscal year format, should be four digit year: YYYY"
}
```

nmonga91: I don't see that error actually in the code, but I do think we should enforce a 4 digit year.

PR #945, opened 2018-02-06
Feature transaction spending

MelvinDunn:


usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -224,3 +224,38 @@ def get_transaction_ids_agg(keyword, size=10000):
         for result in response['aggregations']['results']['buckets']:
             results.append(result['key'])
         yield results
+
+
+def get_sum_and_count_aggregation_results(keyword):
+    index_name = '{}-'.format(TRANSACTIONS_INDEX_ROOT.replace('_', ''))+'*'
```
tony-sappe: Why is the string replace needed? The index root is the prefix for all transaction indices

MelvinDunn: it's not

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -224,3 +224,38 @@ def get_transaction_ids_agg(keyword, size=10000):
...
+                 }
+              },
+              "prime_awards_count": {
+                 "value_count": {
+                    "field": "transaction_id"
+                 }
+              }
+              }, "size": 0}
+    found_result = False
+    while not found_result:
```
tony-sappe: Why is there a while loop?

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -241,11 +241,11 @@ def get_sum_and_count_aggregation_results(keyword):
                  }
               }
               }, "size": 0}
-    found_result = False
-    while not found_result:
+    found_result = 0
+    while not found_result == 10:
```
tony-sappe: Better to have `found_result >= 10`

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -224,3 +224,38 @@ def get_transaction_ids_agg(keyword, size=10000):
...
+                 "value_count": {
+                    "field": "transaction_id"
+                 }
+              }
+              }, "size": 0}
+    found_result = 0
+    while not found_result == 10:
+        try:
+            response = CLIENT.search(index=index_name, body=query)
+            found_result += 1
```
tony-sappe: Need to move this outside of the try:except block. it is possible the first statement errors and found_result never gets incremented

PR #949, opened 2018-02-07
Hotfix: daily loader updates to support updated field mappings

nmonga91:


usaspending_api/references/models.py
```diff
@@ -569,364 +569,8 @@ class LegalEntity(DataSourceTrackedModel):
                                              verbose_name="Transaction Unique ID")
 
     def save(self, *args, **kwargs):
```
russellbodine: you can just delete this save method, its overwriting the super now.

usaspending_api/broker/helpers.py
```diff
@@ -0,0 +1,556 @@
...
+    if data_type == 'fabs':
+        business_types = row.get('business_types')
+        if business_types in ('R', '23'):
+            business_category_set.add('small_business')
+
+        if business_types in ('Q', '22'):
+            business_category_set.add('other_than_small_business')
+
+        if business_category_set & {'small_business', 'other_than_small_business'}:
+            business_category_set.add('category_business')
```
russellbodine: not the same as the old file because it doesnt contain le.for_profit_organization == "1" or:
```
 le.for_profit_organization == "1" or
-
-            "small_business" in categories or
-            "other_than_small_business" in categories
-        ):
-            categories.append("category_business")

```

usaspending_api/broker/helpers.py
```diff
@@ -0,0 +1,556 @@
...
+    if business_type == 'V':
+        return 'ALASKA NATIVE AND NATIVE HAWAIIAN SERVING INSTITUTIONS'
+    if business_type == 'W':
+        return 'NON-DOMESTIC (NON-US) ENTITY'
+    if business_type in ('X', '25'):
+        return 'OTHER'
+    return 'UNKNOWN TYPES'
+
+
+def get_business_categories(row, data_type):
```
russellbodine: all these business types dont take into account the LE variables like the old code did.  May not be right

nmonga91: This will in turn use the row's data as opposed to the values set in the object so it's addressed before calling "save"

usaspending_api/broker/helpers.py
nmonga91: The old method combined processing FABS and FPDS business categories. This method keep it separated. So FABS doesn't have that column available, therefore it's not checked. But it is checked below for FPDS.

russellbodine: your right!

usaspending_api/broker/helpers.py
russellbodine: you're right!

PR #951, opened 2018-02-08
Hotfix: submission loader performance

nmonga91:


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -564,6 +564,74 @@ def load_file_b(submission_attributes, prg_act_obj_cls_data, db_cursor):
...
+        for i in [(piid, "piid"), (fain, "fain"), (uri, "uri")]:
+            lookup_kwargs[i[1]] = i[0]
+            if parent_award_id:
+                # parent_award__piid
+                lookup_kwargs["parent_award_piid"] = parent_award_id
+
+        # Look for an existing award record
+        summary_award = Award.objects \
+            .filter(Q(**lookup_kwargs)) \
+            .filter(awarding_agency=awarding_agency) \
```
russellbodine: The awarding agency shouldnt be part of the find (for finding a file C row from a file D)

But it should be used if there is a create.

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -621,19 +689,19 @@ def load_file_c(submission_attributes, db_cursor, award_financial_frame):
...
-            parent_award_id=row.get('parent_award_id'),
-            use_cache=False)
+            parent_award_id=row.get('parent_award_id'))
 
         awards_touched += [award]
 
         award_financial_data = FinancialAccountsByAwards()
 
         value_map_faba = {
+            'award': award,
```
russellbodine: change award to  `award if created else None`

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -643,9 +711,8 @@ def load_file_c(submission_attributes, db_cursor, award_financial_frame):
         }
 
         # individual FILE C D linkage
```
russellbodine: move line 713 to 733 to inside new get or create summary award.  this way we dont have to do 2 db calls

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -662,21 +729,14 @@ def load_file_c(submission_attributes, db_cursor, award_financial_frame):
...
                 file_d_award = award_queryset[0]
                 value_map_faba['award_id'] = file_d_award['id']
-                logger.info('individual award id mapped: {}'.format(file_d_award['id']))
+                del value_map_faba['award']  # remove a potentially bad reference to the related award
+                # logger.info('individual award id mapped: {}'.format(file_d_award['id']))
 
         # Still using the cpe|fyb regex compiled above for reverse
         load_data_into_model(award_financial_data, row, value_map=value_map_faba, save=True, reverse=reverse)
 
     awards_cache.clear()
```
russellbodine: i added this check bulk because it is way faster than using django.

usaspending_api/etl/management/commands/load_submission.py
russellbodine: lines to remove
591
593-611

nmonga91: Keeping the old filters to ensure we avoid creating duplicate awards when the row in question gets reloaded and still doesn't find a matching award that came in from the authoritative source

usaspending_api/etl/management/commands/load_submission.py
nmonga91: Since all checks moved into `get_or_create_summary_award` that will handle returning `None` if it couldn't find an award so this check isn't necessary.

usaspending_api/etl/management/commands/load_submission.py
nmonga91: The bulk operation didn't seem like it was limiting on only the rows in question, but instead after each submission, operate on the entire data set. This would inevitably run into the same issue as `update_description_fields` where the data set was too large for it to operate efficiently.

PR #959, opened 2018-02-09
Hotfix elasticsearch stability refactor

tony-sappe:
- Added better error responses and removed potential infinite loops in code.
- Started the code changes to use the new `display_award_id` if it exists in the documents


usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -53,79 +62,92 @@ def search_transactions(filters, fields, sort, order, lower_limit, limit):
...
-def get_total_results(keyword, index_name):
-    index_name = '{}-{}'.format(TRANSACTIONS_INDEX_ROOT,
-                                index_name.replace('_', ''))+'*'
+    for index, award_types in indices_to_award_types.items():
+        if sorted(award_types) == sorted(filters['award_type_codes']):
+            index_name = '{}-{}-*'.format(TRANSACTIONS_INDEX_ROOT, index)
+            transaction_type = index
+            break
+    else:
+        logger.exception('Bad/Missing Award Types requested')
```
miahunsicker: Can we please direct users where to get the award type code/ requirements

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -44,7 +53,7 @@ def search_transactions(filters, fields, sort, order, lower_limit, limit):
     '''
     keyword = filters['keyword']
     query_fields = [TRANSACTIONS_LOOKUP[i] for i in fields]
-    query_fields.extend(['piid', 'fain'])
+    query_fields.extend(['piid', 'fain', 'uri', 'display_award_id'])
```
miahunsicker: I ran the latest `es_loader` for 2018 and don't have `display_award_id` in any documents or in the index mappings.

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -53,79 +62,92 @@ def search_transactions(filters, fields, sort, order, lower_limit, limit):
...
             'query_string': {
                 'query': keyword
-                }
-            },
+            }
+        },
         'sort': [{
             query_sort: {
                 'order': order}
-            }]
```
miahunsicker: #todo change the look ups so we're sorting on `display_award_id`. Are we doing that with this PR or is it non-critical?

usaspending_api/search/v2/elasticsearch_helper.py
tony-sappe: We should. But that is outside the scope of this sprint

usaspending_api/search/v2/elasticsearch_helper.py
tony-sappe: That is not an issue with this code. I can show you the new indices with `display_award_id` in the documents and help you figure out the loader script

usaspending_api/search/v2/elasticsearch_helper.py
tony-sappe: I don't think that should occur until after we switch to the new indices. It would involve over-complicated code to handle both old and new documents

tony-sappe: That's why the PR description has "Started the code changes to use the new display_award_id if it exists in the documents"

miahunsicker:
👍

PR #972, opened 2018-02-12
Hotfix: Fix award unique id generation

nmonga91:


usaspending_api/broker/management/commands/fabs_nightly_loader.py
```diff
@@ -144,8 +144,8 @@ def insert_new_fabs(self, to_insert, total_rows):
             # Generate the unique Award ID
             # "ASST_AW_" + awarding_sub_tier_agency_c + fain + uri
             generated_unique_id = 'ASST_AW_' +\
-                (row['awarding_sub_tier_agency_c'] if row['awarding_sub_tier_agency_c'] else '-NONE-') +\
-                (row['fain'] if row['fain'] else '-NONE-') +\
+                (row['awarding_sub_tier_agency_c'] if row['awarding_sub_tier_agency_c'] else '-NONE-') + '_' + \
```
catherinedevlin: We may in the future want to refactor this out into a function of its own, since it's well-defined and self-containable, but that's not necessary for a hotfix.

PR #973, opened 2018-02-12
refactoring

MelvinDunn:


usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -217,19 +215,18 @@ def get_sum_and_count_aggregation_results(keyword):
...
             results["prime_awards_obligation_amount"] = \
                 round(response['aggregations']["prime_awards_obligation_amount"]["value"], 2)
             return results
-        except (TransportError, ConnectionError) as e:
-            logger.error(e)
-            logger.error('Error retrieving ids. Retrying connection.')
+        except KeyError:
+            logger.error('Unexpected Response')
+    else:
+        return None
```
tony-sappe: There should be a check in search.py to handle this return value and provide the correct exception

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -171,7 +170,6 @@ def get_download_ids(keyword, field, size=10000):
     total = get_total_results(keyword, '*', max_iterations)
     if not total:
         logger.error('Error retrieving total results. Max number of attempts reached')
-        return None
```
tony-sappe: What is the behavior you desire by not returning here? Wouldn't line 174 be a python exception if total is `None`?

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -171,7 +170,6 @@ def get_download_ids(keyword, field, size=10000):
     total = get_total_results(keyword, '*', max_iterations)
     if not total:
```
tony-sappe: This should probably check `if total is not None:` to handle a total value of 0

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -169,10 +168,8 @@ def get_download_ids(keyword, field, size=10000):
 
     max_iterations = 10
     total = get_total_results(keyword, '*', max_iterations)
-    if not total:
+    if total is not None:
```
tony-sappe: My oringinal comment might have been incorrect, we need to check if total is None for the error message

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -169,10 +168,8 @@ def get_download_ids(keyword, field, size=10000):
...
         logger.error('Error retrieving total results. Max number of attempts reached')
-        return None
-
     n_iter = min(max(1, total // size), n_iter)
```
tony-sappe: What happens on this line when total is None?

MelvinDunn: It breaks if total is None. Still testing.

MelvinDunn: This has been resolved with a yield, then return. it's been tested.

PR #990, opened 2018-02-15
Use @timer context manager 

catherinedevlin:
Refactoring change to cut down bulk of logging / timing statements in loader scripts.  The @timer context manager has been in use for a couple weeks now, so this PR applies it to scripts it hadn't been used in yet.


usaspending_api/broker/management/commands/bulk_load_fabs.py
```diff
@@ -1,5 +1,4 @@
 import logging
-import timeit
 from datetime import datetime
 from django.core.management.base import BaseCommand
 from django.db import connections, connection, transaction as db_transaction
```
dpb-bah: Looks like the timer import statement for this file got removed?

dpb-bah:
Also flake8 issues.

usaspending_api/broker/management/commands/bulk_load_fabs.py
tony-sappe: That is correct. I assume it is no longer necessary and instead needs `from usaspending_api.common.helpers import timer`

PR #995, opened 2018-02-16
Corrected method of adding timezone info to datetime in Python 3.5

tony-sappe:
Original code works for Python 3.6. Unfortunately Jenkins runs Python 3.5.

### DIFF
--- a/usaspending_api/etl/management/commands/es_loader.py
+++ b/usaspending_api/etl/management/commands/es_loader.py
@@ -38,7 +38,7 @@ class Command(BaseCommand):
             help='If only one fiscal year is desired')
         parser.add_argument(
             '--since',
-            default='2001-01-01',
+            default=None,
             type=str,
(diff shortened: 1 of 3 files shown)


usaspending_api/etl/management/commands/es_loader.py
```diff
@@ -68,9 +68,9 @@ def handle(self, *args, **options):
         self.config['directory'] = options['dir'] + os.sep
         self.config['provide_deleted'] = options['deleted']
         self.config['recreate'] = options['recreate']
-        self.config['starting_date'] = datetime.strptime(options['since'], "%Y-%m-%d")
+        self.config['starting_date'] = datetime.strptime(options['since'] + '+0000', '%Y-%m-%d%z')
 
-        if self.config['recreate'] and self.config['starting_date'] != datetime(2001, 1, 1):
+        if self.config['recreate'] and self.config['starting_date'] > datetime(2007, 1, 1):
```
dpb-bah: Any reason for the year changes?

PR #999, opened 2018-02-20
Refactored Elasticsearch Client and new API validator

tony-sappe:
**High level description:**
1. Moved Elasitcsearch client functions into a new centralized location in the codebase `usaspending_api/core`
2. Created a slim API request data validator for the Keyword search API (TinyShield)

**Technical details:**
The reason for moving the Elasticsearch client functions to a centralized location in the codebase was primarily to begin the process of having a solid and fully integrated component of the project similar to how Django connects to a SQL-based datastore (in our case, Postgres)

The API request validator is present to lay a foundation for future API data validation across. It is only being used for the keyword search endpoints since there is a need to have them locked down.

Notes:
- Lacks thorough documentation. 
- Not 100% complete. But a majority of functionality is present and ready for testing as an MVP

**Link to JIRA Ticket:**
[Jira: DEV-504](https://federal-spending-transparency.atlassian.net/browse/DEV-504)

- [x] Matview impact assessment completed **N/A**
- [x] Frontend impact assessment completed
- [x] Data validation completed **N/A**
- [x] API Performance evaluation completed and present **N/A**


tony-sappe:
Lines 11-12 in `usaspending_api/core/elasticsearch/client.py` raises an error if the ES_HOSTNAME isn't set due to previous issues encountered. Is there a way to set that env var for Travis CI?

usaspending_api/search/v2/views/search.py
```diff
@@ -746,58 +746,43 @@ def __eq__(self, other):
...
+        metadata = get_simple_pagination_metadata(len(response), validated_payload['limit'], validated_payload['page'])
+
         results = []
-        for transaction in response:
+        for transaction in response[:validated_payload['limit']]:
             transaction["internal_id"] = transaction["Award ID"]
             if 'display_award_id' in transaction:
                 transaction["Award ID"] = transaction['display_award_id']
                 del transaction['display_award_id']
-            elif award_type != 'contracts':
```
dpb-bah: Why are these being commented out?

tony-sappe: We no longer need to pull from different fields to get the "Award ID".

I commented out instead of removed because there is still a need to fix the Award ID sorting to use this new field (currently the sorting ordering for this column doesn't make sense to a human). Comment can be removed once that is fixed

PR #1000, opened 2018-02-20
Added additional code climate config Radon params

tony-sappe:
Changing threshold might reduce the number of issues reported. Setting python to version 3 may or may not improve the issue reports.

Also corrected the Postgres version in the README

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -10,6 +10,9 @@ engines:
         enabled: false
   radon:
     enabled: true
+    config:
+      python_version: 3
+      threshold: "D"
 
 ratings:
   paths:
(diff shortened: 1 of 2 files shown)


tony-sappe:
Removed many of the reported "Cyclomatic complexity" issues since we aren't focused on tackling those. We can gradually address the existing issues and step up the threshold as we progress

.codeclimate.yml
```diff
@@ -10,6 +10,9 @@ engines:
         enabled: false
   radon:
     enabled: true
+    config:
+      python_version: 3
+      threshold: "D"
```
nmonga91: This is just the lowest "grade" code climate can give us?

tony-sappe: Good question. I intended to share this info.

Default is "B". D isn't the lowest, F is the lowest grade.
http://radon.readthedocs.io/en/latest/commandline.html#the-cc-command

PR #1001, opened 2018-02-20
Added new SQL file to drop the view

tony-sappe:
Useful for running at the end of the ES etl process

### DIFF
--- a/usaspending_api/database_scripts/etl/drop_view.sql
+++ b/usaspending_api/database_scripts/etl/drop_view.sql
@@ -1,0 +1,1 @@
+DROP VIEW IF EXISTS transaction_delta_view;


usaspending_api/database_scripts/etl/drop_view.sql
```diff
@@ -0,0 +1 @@
+DROP VIEW IF EXISTS transaction_delta_view;
```
nmonga91: To make sure I understand correctly, the ETL process would be as follows:
1. Create the `transaction_delta_view`
2. Run through the delta ETL process for the affected data
3. Drop the `transaction_delta_view` once ETL has successfully finished

So this would just primarily be for cleanup of an unused view once the delta ETL is done.

tony-sappe: Correct. That is what @willkjackson and I discussed. Originally this was going to be performed by the etl script. But that was not doable if we wanted to use read_only account or a read replica

PR #1002, opened 2018-02-20
Feature fix county data

alburde1:
Derive county codes where the derivations had failed before in `transaction_fabs` and `transaction_fpds`. Apply those derivations to `references_location`


usaspending_api/references/management/commands/fix_county_data.py
```diff
@@ -0,0 +1,395 @@
...
+                self.update_fpds_ppop(sess)
+                self.update_fabs_le(sess)
+                self.update_fabs_ppop(sess)
+                self.update_location(sess)
+
+                if options['all_matview']:
+                    self.delete_matviews(sess)
+            else:
+                if options['matview']:
+                    self.create_matviews(sess)
```
nmonga91: If we're picking a specific option here, would we want to run this matview creation _every_ time?

alburde1: Not really. If you create the matviews once and don't delete them, you can then run each section individually if you want (run just `fix_county_data -fpdsle` then wait until you have time before running `fix_county_data -fpdsppop` if needed). If you're starting from scratch you just have to call `fix_county_data -m -fpdsle` for example. It'll get to the matview first so no problems will be had.

PR #1050, opened 2018-02-28
Hotfix: Update spending explorer to group recipients by name

nmonga91:
**High level description:**
Updating spending explorer to group recipients by name and return any awards that match that recipient name when drilling down from recipients to awards.

**Technical details:**
Updating spending explorer logic to group by recipient name and to return awards based on the recipient name instead of ID. Also removing the "limit" since the grouping should handle the large amount of duplication previously present.

**Link to JIRA Ticket:**
N/A

- [x] Frontend impact assessment completed:
  - No frontend impact as confirmed by @kevinli-work 
    ```
    Timing changes (example of going from FY 18, Q1 Budget Function -> Social Security -> Recipient):
        Before =  11.7 seconds & 5.3 mb response (500 error b/c of cache size limit)
        After = ~7.8 seconds & 269.1 kb response
    ```

### DIFF
--- a/usaspending_api/spending_explorer/v2/filters/explorer.py
+++ b/usaspending_api/spending_explorer/v2/filters/explorer.py
@@ -81,8 +81,8 @@ class Explorer(object):
     def recipient(self):
         # Recipients Queryset
         alt_set = self.alt_set.filter(~Q(transaction_obligated_amount=Decimal('NaN')), award__recipient__isnull=False).\
-            annotate(id=F('award__recipient__legal_entity_id'), type=Value('recipient', output_field=CharField()),
-                     name=F('award__recipient__recipient_name'), code=F('award__recipient__recipient_unique_id')).\
+            annotate(id=F('award__recipient__recipient_name'), type=Value('recipient', output_field=CharField()),
+                     name=F('award__recipient__recipient_name'), code=F('award__recipient__recipient_name')).\
             values('id', 'type', 'name', 'code', 'amount').\
             annotate(total=Sum('transaction_obligated_amount')).\
             order_by('-total')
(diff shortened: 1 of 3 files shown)


usaspending_api/spending_explorer/v2/views/spending_explorer.py
```diff
@@ -17,6 +17,6 @@ def post(self, request):
         filters = json_request.get('filters', None)
 
         # Returned filtered queryset results
-        results = type_filter(_type, filters, limit=SPENDING_EXPLORER_LIMIT)
```
tony-sappe: Just for clarification, this is to remove the initial work to put a cap on the spending explore until the stakeholders make their decision?

usaspending_api/spending_explorer/v2/filters/explorer.py
```diff
@@ -81,8 +81,8 @@ def object_class(self):
     def recipient(self):
         # Recipients Queryset
         alt_set = self.alt_set.filter(~Q(transaction_obligated_amount=Decimal('NaN')), award__recipient__isnull=False).\
-            annotate(id=F('award__recipient__legal_entity_id'), type=Value('recipient', output_field=CharField()),
-                     name=F('award__recipient__recipient_name'), code=F('award__recipient__recipient_unique_id')).\
+            annotate(id=F('award__recipient__recipient_name'), type=Value('recipient', output_field=CharField()),
+                     name=F('award__recipient__recipient_name'), code=F('award__recipient__recipient_name')).\
```
tony-sappe: Why is `award__recipient__recipient_name` used for both `id` and `code`?

nmonga91: it's done this way to that the grouping still happens on the recipient name, but also to maintain the response parameters necessary by the frontend. By keeping `id` and `code` as the same column, the grouping isn't affected by varying values.

PR #1057, opened 2018-03-01
Hotfix: log cache information within headers

michaeldhess:


tony-sappe:
From original PR:

High level description:
Added a new Header field which will be useful for logging and troubleshooting

Technical details:
New header field: Cache-Trace
Values:

hit-cache : Response was retrieved from the cache
set-cache : Response was a cache-miss, Django thinks the response was cached
no-cache : Response was a cache-miss and not cached

PR #1058, opened 2018-03-01
Hotfix: order request before creating cache key

michaeldhess:


usaspending_api/common/cache_decorator.py
```diff
@@ -9,6 +9,7 @@ class CustomCacheResponse(CacheResponse):
     def process_cache_response(self, view_instance, view_method, request, args, kwargs):
         key = self.calculate_key(view_instance=view_instance, view_method=view_method,
                                  request=request, args=args, kwargs=kwargs)
+        print(key)
```
tony-sappe: Remove before merge

usaspending_api/common/cache_decorator.py
```diff
@@ -23,6 +24,8 @@ def process_cache_response(self, view_instance, view_method, request, args, kwar
             response.render()  # should be rendered, before picklining while storing to cache
 
             if not response.status_code >= 400 or self.cache_errors:
+                if self.cache_errors:
+                    logger.exception(self.cache_errors)
```
tony-sappe: Why not logger.error instead?

PR #1068, opened 2018-03-02
Hotfix: fix cache key generation

michaeldhess:
**High level description:**
Overwrite the Django rest framework default cache key generation to ensure we always generate the same key.

**Technical details:**
- Override the DefaultKeyConstructor `prepare_key` function
- Sort and "stringify" the request dict, then also sort the cache key dict before creating the key
- Update server logging to only log the `key` and `cache-trace` values, instead of their objects
- Fixed the `order_nested_object` function to no longer assume lists of dicts are the end of the line

**Link to JIRA Ticket:**
N/A

- [x] Ops team testing and approval
- [x] API Performance evaluation completed (N/A)


usaspending_api/common/logging.py
```diff
@@ -85,10 +85,10 @@ def process_response(self, request, response):
         self.log["response_ms"] = self.get_response_ms()
         self.log["traceback"] = None
         if response._headers:
-            if 'key' in response._headers:
-                self.log["key"] = response._headers['key']
-            if 'cache-trace' in response._headers:
-                self.log["cache_trace"] = response._headers['cache-trace']
+            if 'key' in response._headers and len(response._headers['key']) >= 2:
```
tony-sappe: Why the length check?

PR #1079, opened 2018-03-05
Fix query, updating mapping file & add reindex script

miahunsicker:
**High level description:**
This PR updates elastic field types for fields we are sorting on from text to keyword and adds their raw types as text to be searched on. This adds a script used to reindex ES and rename the new to the old indexes. Reindexing is less favorable to reloading IMO but throwing this in here for future reference. 

**Technical details:**
Update fields in ES mapping template
Grab nested fields for query to query string from multi_match, allowing Lucene queries on all fields
Change sorting from internal_id to display_award_id 

Link to JIRA Ticket:
[DEV-523] https://federal-spending-transparency.atlassian.net/browse/DEV-523

 Backend review completed
- [X] Matview impact assessment completed (N/A)
- [X] Frontend impact assessment completed (N/A)
- [ ] API Performance evaluation completed and present:
<>


usaspending_api/etl/es_transaction_template.json
```diff
@@ -117,22 +123,17 @@
           "index": false
         },
         "award_amount": {
-          "type": "scaled_float",
-          "scaling_factor": 100
+          "type": "float"
```
tony-sappe: These should remain `scaled_float`

usaspending_api/etl/management/commands/reindex_elastic.py
```diff
@@ -0,0 +1,157 @@
...
+        try:
+            response = ES_CLIENT.search(index=index_name)
+            return response["hits"]["total"]
+        except TransportError as e:
+            logging.error('[CONNECTION TIMEOUT] with ElasticSearch cluster: {e}'.
+                          format(e=str(e)))
+            return -1
+
+    def verify(self, source, target):
+        difference = self.get_total_count(source) - self.get_total_count(target)
```
tony-sappe: What will be the behavior if both of the get_total_count functions have errors and return -1?

usaspending_api/etl/management/commands/reindex_elastic.py
```diff
@@ -0,0 +1,157 @@
...
+            logging.warn('Still have {} documents to reindex'.format(difference))
+        return difference, difference/INGEST_RATE
+
+    def controller(self):
+        logging.warn('Reindexing the following \n')
+        pprint(self.reindex_dict)
+        print('\n\n')
+        for source, dest in self.reindex_dict.items():
+            if ES_CLIENT.indices.exists(dest):
+                if not self.wipe:
```
tony-sappe: rename to "delete_original" or "drop_original" to be more intuitive

usaspending_api/etl/management/commands/reindex_elastic.py
```diff
@@ -0,0 +1,157 @@
...
+                self.size, self.sleep_time = self.verify(source, dest)
+                logging.warn('Reindexing {} -- {} documents'.format(source, self.size))
+                try:
+                    self.single_reindex(source, dest)
+                    logging.info(source)
+                    ES_CLIENT.indices.flush('*')
+                except (ConnectionTimeout) as e:
+                    logging.error('[CONNECTION TIMEOUT] with ElasticSearch cluster: {e}'.
+                                  format(e=str(e)))
+                    for attempt in range(MAX_RETRIES):
```
tony-sappe: Is all of this code necessary in the except-clause?

usaspending_api/etl/management/commands/reindex_elastic.py
```diff
@@ -0,0 +1,157 @@
...
+                        self.size, self.sleep_time = self.verify(source, dest)
+                        if self.sleep_time > 0:
+                            logging.warn('---------- Sleep_time of index >>> {} ------------------'
+                                         .format(str(self.sleep_time)))
+                            time.sleep(self.sleep_time)
+        print('\n\n')
+        index_name = self.dest_prefix + '*'
+        ES_CLIENT.indices.flush('*')
```
tony-sappe: Would it be better to flush only the indices being modified in this script?

usaspending_api/etl/management/commands/reindex_elastic.py
```diff
@@ -0,0 +1,157 @@
...
+        time.sleep(5)
+        count = self.get_total_count(index_name)
+        logging.warn('Reindexed a total of {} documents'.format(count))
+        logging.warn('****   FINISHED REINDEXING  ****')
+
+        for source, dest in self.reindex_dict.items():
+            transaction_category = next((s for s in INDICES_SUB_NAMES if dest.find(s) != -1), None)
```
tony-sappe: Based on the script title I didn't expect it to be tightly coupled to the current index naming convention. Is there an expectation to run this on a regular bases or only when necessary?

usaspending_api/etl/management/commands/reindex_elastic.py
miahunsicker: yes... its if connection timeout... lets the code sleep so we don't start another reindex and have multiple concurrent reindexing jobs. It's an expected timeout.

usaspending_api/etl/management/commands/reindex_elastic.py
miahunsicker: It is coupled with current. We need to discuss this vs reloading. We shouldn't have to run this on a regular basis.

usaspending_api/etl/es_transaction_template.json
miahunsicker: Got reindexing errors for this field only. It works and does not effect performance. We've been at odds about this for weeks but I have yet to see what you are talking about. @MelvinDunn

usaspending_api/etl/management/commands/reindex_elastic.py
miahunsicker: non-trivial

usaspending_api/etl/management/commands/reindex_elastic.py
miahunsicker: If it's a transport error then they'd both be -1..wouldn't sleep and pass to the next index to reindex.

usaspending_api/etl/management/commands/reindex_elastic.py
miahunsicker: don't see `delete_original` in the script unless you mean the logging on line 116

tony-sappe: Sorry I wasn't more clear. `self.wipe` isn't intuitive. I recommend changing the variable name. The first comment has two suggestions

usaspending_api/etl/management/commands/reindex_elastic.py
tony-sappe: It is non-trival to change the code to flush only the touched indexes?
Or the impact of flushing all is trivial? 

If the latter, is there a possibility this will no longer hold true? Would more data/more indexes make this expensive at times?

usaspending_api/etl/management/commands/reindex_elastic.py
tony-sappe: Since this isn't planned for regular execution, I would argue this script will be more helpful if it isn't tied to the current award category descriptions. That way it could be used for any reindex job

usaspending_api/etl/es_transaction_template.json
tony-sappe: Since we were currently using `scaled_float`, the burden of proof lies with the developer to show and explain why the change is necessary if asked.

usaspending_api/etl/es_transaction_template.json
```diff
@@ -1,10 +1,10 @@
 {
-  "index_patterns": ["*"],
+  "index_patterns": ["transactions*"],
```
tony-sappe: Is there a reason this is changed here? `es_configure` already modified this value based upon `settings.TRANSACTIONS_INDEX_ROOT`

usaspending_api/etl/es_transaction_template.json
```diff
@@ -1,10 +1,10 @@
...
   "settings": {
     "index.mapping.ignore_malformed": true,
     "index.max_result_window" : 500000,
     "index": {
-      "number_of_shards" : 5,
+      "number_of_shards" : 3,
```
tony-sappe: What is the desired impact of reducing the shard count? Would it be beneficial to counter the reduced shards by increasing the replicas?

usaspending_api/etl/management/commands/reindex_elastic.py
miahunsicker: Its not expensive- it'd only be expensive on the first call. It's to try to maintain performance in the background while this is going on.

usaspending_api/etl/es_transaction_template.json
miahunsicker: I didn't touch replicas. This speeds up the job. We have too many shards as is. This is still a WIP and dependent on stress testing our baseline shard size - yet to do.

usaspending_api/etl/es_transaction_template.json
miahunsicker: I presented cases where the results were similar and had 0 effect on API calls and results. Would you like to see errors when I try to reindex it?

usaspending_api/etl/es_transaction_template.json
miahunsicker: Yes, as we discussed Friday. Naming conventions will be consistent and settings.TRANSACTIONS_INDEX_ROOT for ingest and querying will NOT start with transactions.

tony-sappe: If this is a different prefix, then:

1. Please set in settings.py instead of buried in django management commands
2. Alter `es_configure.py` so that is uses the new alias properly

usaspending_api/etl/management/commands/reindex_elastic.py
tony-sappe: So to summarize:
Calling flush on the entire cluster is an expensive operation the first time it is called. Subsequent calls are trival.
Correct?

It appears this is only called once in the script.

usaspending_api/etl/es_transaction_template.json
tony-sappe: Logically it would speed up writes and reindexing. The queries might be negatively impacted

miahunsicker: @tony-sappe

usaspending_api/etl/es_transaction_template.json
```diff
@@ -142,15 +147,23 @@
...
           "type": "text"
         },
         "awarding_subtier_agency_name": {
-          "type": "text",
-          "fielddata": true
+          "type": "keyword",
+          "fields": {
+            "raw": { 
+              "type": "text"
+            }
```
MelvinDunn: This is good in case we want to aggregate via "awarding_subtier_agency_name". But we'll have to append ".raw" to the search fields:

https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html

usaspending_api/etl/es_transaction_template.json
```diff
@@ -1,10 +1,10 @@
 {
-  "index_patterns": ["*"],
+  "index_patterns": ["*transactions*"],
```
tony-sappe: This will be overwritten when the template is loaded into an ES cluster using the `es_configure` management command

miahunsicker: ok

PR #1091, opened 2018-03-07
Promote Dev to Staging

tony-sappe:
Documentation improvements and adding concurrent refresh functionality to the rest of the materialized views


usaspending_api/api_docs/markdown/using_the_api.md
```diff
@@ -11,25 +11,24 @@
   </ul>-->
```
nmonga91: @russellbodine this file still has a lot of v1 documentation present. is there a reason this is still here?

russellbodine: @nmonga91 No reason besides that its necessary information for a user who has to use v1 endpoints.

PR #1099, opened 2018-03-08
adding stemmer to several fields

MelvinDunn:
**High level description:**
There was a substring search in Postgres that allowed "plane" to be returned in "planes". However, this was not bidirectional, searching "planes" would not return the same as "plane" I've been told. In this PR a stemmer has been used to strip plurals in the following examples provided:

`poptart`: 0 contracts vs `poptarts`: 2 contracts
`chair`: 37 other vs `chairs` 5 other
`plane` : 4000 contracts vs `planes` 217 contracts

Locally, the search with plurals now returns the same results (DATA ACT data was used with a few sample cases):

`poptart`: 2 results vs `poptarts`: 2 results
`chair`: 2 results vs `chairs` 2 results
`plane` : 4 results vs `planes` 4 results

**Technical details:**
A minimal_stemmer was used that picks up the plurals for these words. However, the minimal_stemmer fails for the use case of "fox" vs "foxes". We can talk about using other stemmers when we have other use cases we want to include. The minimal stemmer was used because we cared about plurals, not tenses of words.

To test performance and API we need to confirm that this is the stemmer that we would like to use by addressing more use cases. Data will have to be reloaded because this changes the mapping. Unsure if this will work with reindexing - because the analyzer is being changed, not the specific field type.

**Link to JIRA Ticket:**
[DEV-603](https://federal-spending-transparency.atlassian.net/browse/DEV-603)

- NA Matview impact assessment completed
- NA Frontend impact assessment completed
- [ ] API Performance evaluation completed and present:
    ```
    Timing changes:
        Before = X seconds
        After = Y seconds
    
    Data volume changes:
        Before = X rows
        After = Y rows


usaspending_api/etl/es_transaction_template.json
```diff
@@ -61,24 +75,28 @@
...
         "naics_description": {
-          "type": "text"
+          "type": "text",
+          "analyzer": "stemmer_analyzer"
         },
         "type_description": {
           "type": "text",
-          "fielddata": true
+          "fielddata": true,
+          "analyzer": "stemmer_analyzer"
```
tony-sappe: I don't believe this field needs to be changed. Example values are `NULL`, PO, DELIVERY ORDER,

tony-sappe:
Should this analyzer be used for additional fields as well?
- recipient_name
- awarding_toptier_agency_name
- awarding_subtier_agency_name
- funding_toptier_agency_name
- funding_subtier_agency_name
- pop_country_name
- recipient_location_country_name

usaspending_api/etl/es_transaction_template.json
miahunsicker: What branch did you base from? `type_description` is a keyword field now in dev. Please update.

tony-sappe:
Closing since the requirements are not defined well enough

PR #1141, opened 2018-03-22
Feature update es ingest

miahunsicker:
**High level description:**
Consolidating multiple indices in ES to one index. Users were querying for singular terms and were expecting results for their plural versions. Fixing the singularity issue in this PR.

**Technical details:**
* Update Elasticsearch ETL to ingest and use one index
* Add alias structures for filtering that we currently take advantage of on on the frontend
* Ingest with different settings for index refresh and replicas, reset those back to our defaults at the end
* Removing aliases for current indices and closing them once ingesting for new index is complete opposed to deleting them (this will be changed once we get snapshots set up in prod)
* Update mapping to for award_description to stem for singularity so users can get results for "poptart" includes award_descriptions with terms "poptarts"

**Link to JIRA Ticket:**
[NONE]

- [ ] Frontend impact assessment completed N/A
- [ ] API Performance evaluation completed and present


usaspending_api/etl/es_transaction_mapping.json
```diff
@@ -0,0 +1,270 @@
+{
```
tony-sappe: Why is this a new file instead of modifying the existing file?

usaspending_api/etl/es_transaction_mapping.json
```diff
@@ -0,0 +1,270 @@
...
+  "settings": {
+    "index.mapping.ignore_malformed": true,
+    "index.max_result_window": 50000,
+    "index.refresh_interval": -1,
+    "index": {
+      "number_of_shards": 5,
+      "number_of_replicas": 0
```
tony-sappe: Having 0 replicas is a bad idea

usaspending_api/etl/es_transaction_mapping.json
```diff
@@ -0,0 +1,270 @@
+{
+  "settings": {
+    "index.mapping.ignore_malformed": true,
+    "index.max_result_window": 50000,
+    "index.refresh_interval": -1,
```
tony-sappe: Is this modified to a better value after the indexing is complete?

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -85,6 +95,15 @@ def handle(self, *args, **options):
...
         self.config['stale'] = options['stale']
         self.config['keep'] = options['keep']
+        self.config['index_name'] = options['index_name']
+        self.config['previous_index'] = options['previous_index']
+
+        mappingfile = os.path.join(settings.BASE_DIR, 'usaspending_api/etl/es_transaction_mapping.json')
+        with open(mappingfile) as f:
+            mapping_dict = json.load(f)
+            self.config['mapping'] = json.dumps(mapping_dict)
+        self.config['doc_type'] = str(list(mapping_dict['mappings'].keys())[0])
```
tony-sappe: Is there a better way to do this? It seems fragile, I could see someone modifying the json file without understanding the ramifications.

If not please add a comment of intended behavior.

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -249,6 +252,44 @@ def streaming_post_to_es(client, chunk, index_name, job_id=None):
...
+        put_alias(client, new_index, alias_name, award_type_codes)
+
+    old_index_patterns = old_index + '*'
+    old_indices = client.indices.get(old_index_patterns).keys()
+    for index in old_indices:
+        try:
+            client.indices.delete_alias(index, '_all')
+            client.indices.close(index)
+            printf({'msg': 'Closing index "{}"'.format(index), 'job': '', 'f': 'ES Index Close'})
+        except (TransportError, Exception) as e:
```
tony-sappe: Does it make a difference to catch `TransportError` is `Exception` is already being handled?

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -249,6 +252,44 @@ def streaming_post_to_es(client, chunk, index_name, job_id=None):
...
+            print(e)
```
tony-sappe: Clean up print and/or replace with SystemExit

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -50,6 +50,16 @@ def add_arguments(self, parser):
             default=os.path.dirname(os.path.abspath(__file__)),
             type=str,
             help='Set for a custom location of output files')
+        parser.add_argument(
+            '--index_name',
+            type=str,
+            help='Set an index name to ingest data into',
+            default='staging_transactions')
+        parser.add_argument(
+            '--previous_index',
```
tony-sappe: Are you expecting the user running this script to know what the previous index name is? How do you propose we automate this?

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -161,6 +178,11 @@ def controller(self):
         download_proccess.join()
         es_index_process.join()
 
+        if self.config['recreate']:
```
tony-sappe: if self.config['recreate'] is `true`, won't it keep deleting the single destination index?

usaspending_api/etl/es_transaction_mapping.json
miahunsicker: It's a faster ingest, this is in the event recreate is true and index doesn't already exist.

usaspending_api/etl/es_transaction_mapping.json
miahunsicker: It is https://github.com/fedspendingtransparency/usaspending-api/blob/6ca492d196d9b416a93afb767f30cbc4520bea23/usaspending_api/etl/es_etl_helpers.py#L286

usaspending_api/etl/management/commands/es_rapidloader.py
miahunsicker: Then they shouldn't modify the json file. Single type mappings for our version of ES .. but will modify code to check. https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html

usaspending_api/etl/management/commands/es_rapidloader.py
miahunsicker: We could add date prefixes and pick the min date for the index pattern and assume that is the one we want to delete. Yes this assumes.

usaspending_api/etl/management/commands/es_rapidloader.py
miahunsicker: recreate is default set to false if flag isn't there. no, there are no deletes, it would close index patterns of previous_index. no deletes until we have a snapshot. someone would need to run this explicitly with --r or -recreate for indices to be closed

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -249,6 +252,44 @@ def streaming_post_to_es(client, chunk, index_name, job_id=None):
...
+            # raise SystemExit
+
+    index_settings = dict(refresh_interval='1s', number_of_replicas=1)
+    index_settings = dict(index=index_settings)
```
tony-sappe: It isn't obvious what is being done here. Could these settings be in a JSON file or as globals at the top to the page instead of buried in code?

usaspending_api/etl/es_etl_helpers.py
tony-sappe: I created a `printf` function for logger-like output to the console. The elasticsearch library liked to add junk to the logger output and I didn't find a quick resolution to that.

usaspending_api/etl/es_transaction_mapping.json
tony-sappe: Correct, we are currently using `es_transaction_template.json` for the index settings. The ETL process is currently using a template to guarantee that these settings are applied to a new index. This new mapping file is used by the python library when creating a new index. AFAIK there is only one difference in structure between the two files, which is the `index_patterns` key. If @miahunsicker  wants to add a new file then I would request `es_transaction_template.json` is updated to match this new file.

usaspending_api/etl/management/commands/es_rapidloader.py
tony-sappe: I verified locally using a single FY of data. If the `-r` flag is provide the script will continually delete the only index and recreate it for each file download aka "job". This will ultimately lead to the index with only data from the final job.

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -14,9 +15,9 @@
 from usaspending_api.etl.es_etl_helpers import deleted_transactions
 from usaspending_api.etl.es_etl_helpers import download_db_records
 from usaspending_api.etl.es_etl_helpers import es_data_loader
+from usaspending_api.etl.es_etl_helpers import swap_aliases
+from usaspending_api.etl.es_etl_helpers import take_snapshot
 from usaspending_api.etl.es_etl_helpers import printf
-
-
 # SCRIPT OBJECTIVES and ORDER OF EXECUTION STEPS
```
tony-sappe: Please update this comment to be correct with the recent changes

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -249,6 +252,45 @@ def streaming_post_to_es(client, chunk, index_name, job_id=None):
...
+        printf({'msg': 'ERROR: no aliases found for {}'.format(alias_patterns),
+                'f': 'ES Alias Drop'})
+
+    for award_type, award_type_codes in indices_to_award_types.items():
+        alias_name = '{}-{}'.format(settings.TRANSACTIONS_INDEX_ROOT, award_type)
+        printf({'msg': 'Putting alias "{}" with award codes {}'.format(alias_name, award_type_codes),
+                'job': '', 'f': 'ES Alias Put'})
+        put_alias(client, index, alias_name, award_type_codes)
+
+    index_settings = dict(refresh_interval='1s', number_of_replicas=1)
```
tony-sappe: Can these settings be moved to another location to make it easy to see what the "final settings" will be for the index? One idea is to have an "index-settings" which are applied to the index before indexing (initial and reload) and then revert to the final settings once the data has been loaded in and the script is moving on

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -70,6 +76,15 @@ def add_arguments(self, parser):
...
             help='CSV files are not deleted after they are uploaded')
+        parser.add_argument(
+            '-w',
+            '--swap',
+            action='store_true',
+            help='Flag allowed to put aliases to index and close all indices with aliases associated')
+        parser.add_argument(
+            '--save',
+            action='store_true',
+            help='Flag allowed to put aliases to index and close all indices with aliases associated')
```
tony-sappe: Update help string

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -111,26 +106,27 @@ def handle(self, *args, **options):
...
         if not os.path.isdir(self.config['directory']):
             printf({'msg': 'Provided directory does not exist'})
             raise SystemExit
 
-        does_index_exist = ES.indices.exists(self.config['index_name'])
-
-        if does_index_exist and self.config['recreate']:
-            ES.indices.delete(self.config['index_name'])
-            printf({'msg': 'Deleting index {}'.format(self.config['index_name'])})
+        if does_index_exist and (self.config['starting_date'] == datetime.strptime('2007-10-01+0000', '%Y-%m-%d%z')):
```
tony-sappe: Easier:
`if does_index_exist and options['since'] is None:`

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -82,25 +93,41 @@ def handle(self, *args, **options):
...
+        with open(mappingfile) as f:
+            mapping_dict = json.load(f)
+            self.config['mapping'] = json.dumps(mapping_dict)
+        self.config['doc_type'] = str(list(mapping_dict['mappings'].keys())[0])
+        self.config['max_query_size'] = mapping_dict['settings']['index.max_result_window']
+
+        does_index_exist = ES.indices.exists(self.config['index_name'])
+
+        if not does_index_exist:
+            self.config['provide_deleted'] = False
```
tony-sappe: Add printf line for notifying that `provide_deleted` is overwritten

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -82,25 +93,41 @@ def handle(self, *args, **options):
...
-            if self.config['recreate']:
-                print('Bad mix of parameters! An index should not be dropped if only a subset of data will be loaded')
-                raise SystemExit
             self.config['starting_date'] = datetime.strptime(options['since'] + '+0000', '%Y-%m-%d%z')
 
         if not os.path.isdir(self.config['directory']):
             printf({'msg': 'Provided directory does not exist'})
             raise SystemExit
 
+        if does_index_exist and (self.config['starting_date'] == datetime.strptime('2007-10-01+0000', '%Y-%m-%d%z')):
```
tony-sappe: Easier:
`if does_index_exist and options['since'] is None:`

PR #1143, opened 2018-03-23
minimal english stemmer

MelvinDunn:
Thing you asked for @miahunsicker


miahunsicker:
Just change to dev & were GTG @MelvinDunn

PR #1160, opened 2018-03-29
Feature sum summary loans include

MelvinDunn:
High level description:
Adding loan subsidies to the total of transaction sums.

Technical details:

Since the original_loan_subsidies do not have values greater than 0 (below is the main elasticsearch index), I do not need to filter for only loans to get the subsidy costs then add them. Other award types do not have subsidy costs, nor do loans have transaction amounts.

See below:

`curl 52.222.76.101:9200/2018-03-21-transactions/_search?q=original_loan_subsidy_cost:>0 AND transaction_amount:>0`

Simply having 2 aggregations then adding them in the code to get the total is sufficient for the problem.

Link to JIRA Ticket:
DEV-669

The following are ALL required for the PR to be merged:

[ ] - Backend review completed
[ ] - Matview impact assessment completed
[ ] - Frontend impact assessment completed
[X] - Data validation completed
[X] - API Performance evaluation completed and present:

Timing changes:
    Before = X seconds
    After = Y seconds

Data volume changes:
    Before = X rows
    After = Y rows

### DIFF
--- a/usaspending_api/search/v2/elasticsearch_helper.py
+++ b/usaspending_api/search/v2/elasticsearch_helper.py
@@ -145,7 +145,6 @@ def get_sum_aggregation_results(keyword, field='transaction_amount'):
                 }
             }
         }
-        
     }
 
     response = es_client_query(index=index_name, body=query, retries=10)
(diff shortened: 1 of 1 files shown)


MelvinDunn:
total subsidy cost needs to actually be loaded into elasticsearch.

PR #1173, opened 2018-04-05
Bugfix: Program Activity Data from Remote

dpb-bah:
**High level description:**
Updating Program Activity script to pull from the same file Broker uses on the S3 bucket.

**Technical details:**
* Lowercased the incoming headers in case we wanted to run it locally again later
* Uppercased the program activity name to prevent duplicates when filtering on the frontend

**Link to JIRA Ticket:**
[Dev-126](https://federal-spending-transparency.atlassian.net/browse/DEV-126)

- Matview impact assessment completed (N/A)
- Frontend impact assessment completed (N/A)
- API Performance evaluation completed and present (N/A)


usaspending_api/references/management/commands/load_reference_data.py
```diff
@@ -29,7 +29,7 @@ def handle(self, *args, **options):
             call_command('loadtas', 'gtas-sf133')
 
         self.logger.info("Loading program_activity.csv")
-        call_command('loadprogramactivity', 'usaspending_api/data/program_activity.csv')
+        call_command('loadprogramactivity')
```
nmonga91: We should do a check to see if this is running locally, similar to the TAS loader above.

usaspending_api/references/management/commands/loadprogramactivity.py
```diff
@@ -1,30 +1,68 @@
...
     def add_arguments(self, parser):
-        parser.add_argument('file', nargs=1, help='the file to load')
+        parser.add_argument('file', nargs='?', help='the file to load')
 
     def handle(self, *args, **options):
 
         # Create the csv reader
-        csv_file = options['file'][0]
+        csv_file = options['file']
+        if not csv_file:
```
nmonga91: Here we should support both local and non-local runs

dpb-bah: The change here supports both. The reason for the removed `[0]` is regarding line 22, the `?` only has one value or none whereas the nargs provides a list.

usaspending_api/references/management/commands/load_reference_data.py
dpb-bah: Agreed. Fixing.

PR #1175, opened 2018-04-06
add speedup to update_total_funding_amount

russellbodine:
**High level description:**
total funding amount sql script was too slow.  Added a simple optimization.

**Technical details:**
total funding amount is only influenced by fabs data, so removing all contracts provides huge speedups. 60% speedup on dev, and the speedup should be exponential because of the group by.

**Link to JIRA Ticket:**
NA

- [x] Matview impact assessment completed NA
- [x] Frontend impact assessment completed NA
- [x] Data validation completed NA
    ```
    Timing changes:
        Before = 800 ms
        After = 290 ms

### DIFF
--- a/usaspending_api/etl/management/commands/update_total_funding_amount_sql.py
+++ b/usaspending_api/etl/management/commands/update_total_funding_amount_sql.py
@@ -47,7 +47,7 @@ class Command(BaseCommand):
             join awards a
             on tn.award_id = a.id
             where a.is_fpds = false
-            and tn.funding_amount is not null 
+            and tn.funding_amount is not null
             GROUP BY award_id
         )
         update awards as a
--- a/usaspending_api/etl/management/commands/update_total_funding_amount_sql.py
+++ b/usaspending_api/etl/management/commands/update_total_funding_amount_sql.py
@@ -42,8 +42,13 @@ class Command(BaseCommand):
 
     SUM_TRANSACTION_TFA = """
         With sum_table as (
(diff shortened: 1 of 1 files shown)


usaspending_api/etl/management/commands/update_total_funding_amount_sql.py
```diff
@@ -42,8 +42,13 @@ def handle(self, *args, **options):
...
     SUM_TRANSACTION_TFA = """
         With sum_table as (
-        SELECT award_id, SUM(tn.funding_amount) sum_val
-        FROM transaction_normalized as tn join awards a on tn.award_id = a.id WHERE a.id = tn.award_id GROUP BY award_id
+            SELECT award_id, SUM(tn.funding_amount) sum_val
+            FROM transaction_normalized as tn
+            join awards a
+            on tn.award_id = a.id
+            where a.is_fpds = false
+            and tn.funding_amount is not null
```
nmonga91: what're the situations where `funding_amount` would be null here? 

also, `SUM` already ignores nulls. Reference: http://www.postgresqltutorial.com/postgresql-sum-function/

russellbodine: There are about 10% of transaction normalized rows that have nulls.  I can dig into why that is, but it is most likely a schema question.

Also, Sum skips nulls, but the join doesnt.  By adding it, we may add speedup/reduce memory.  That being said, I'm playing devils advocate, and it's probably not as valuable as I thought it originally was.

PR #1195, opened 2018-04-12
Hotfix: Nightly Load Deletes Fix

dpb-bah:
**High level description:**
By updating downloads to use matviews instead of tables, the nightly loads have gotten confused with Django. This substitutes Django's call to SQL.

**Technical details:**
(see above)

**Link to JIRA Ticket:**
 N/A

- * Matview impact assessment completed
- * Frontend impact assessment completed  (N/A)
- * API Performance evaluation completed (N/A)

### DIFF
--- a/usaspending_api/broker/management/commands/fabs_nightly_loader.py
+++ b/usaspending_api/broker/management/commands/fabs_nightly_loader.py
@@ -81,9 +81,9 @@ class Command(BaseCommand):
         db_cursor = connections['default'].cursor()
 
         # Financial Accounts by Awards
-        fa = 'DELETE ' \
-             'FROM "financial_accounts_by_awards" fa ' \
-             'WHERE fa."award_id" IN ({});'.format(delete_award_str_ids)
+        fa = 'UPDATE "financial_accounts_by_awards" ' \
+             'SET "award_id" = null '\
+             'WHERE "award_id" IN ({});'.format(delete_award_str_ids)
         # Subawards
         sub = 'UPDATE "awards_subaward" ' \
               'SET "award_id" = null ' \
@@ -103,7 +103,7 @@ class Command(BaseCommand):
             award_update_id_list.extend(update_award_ids)
             update_awards_query = 'UPDATE "awards" ' \
                                   'SET "latest_transaction" = null ' \
-                                  'WHERE a."id" IN ({});'.format(','.join(update_award_str_ids))
+                                  'WHERE a."id" IN ({});'.format(update_award_str_ids)
             queries.append(update_awards_query)
         if delete_award_ids:
             delete_awards_query = 'DELETE ' \
--- a/usaspending_api/broker/management/commands/fabs_nightly_loader.py
+++ b/usaspending_api/broker/management/commands/fabs_nightly_loader.py
@@ -80,14 +80,6 @@ class Command(BaseCommand):
 
         db_cursor = connections['default'].cursor()
 
-        # Financial Accounts by Awards
-        fa = 'UPDATE "financial_accounts_by_awards" ' \
-             'SET "award_id" = null '\
-             'WHERE "award_id" IN ({});'.format(delete_award_str_ids)
-        # Subawards
-        sub = 'UPDATE "awards_subaward" ' \
-              'SET "award_id" = null ' \
-              'WHERE "award_id" IN ({});'.format(delete_award_str_ids)
         # Transaction FABS
         fabs = 'DELETE ' \
                'FROM "transaction_fabs" tf ' \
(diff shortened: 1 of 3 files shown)


usaspending_api/broker/management/commands/fabs_nightly_loader.py
```diff
@@ -65,8 +65,44 @@ def delete_stale_fabs(ids_to_delete=None):
...
+        award_ids_list = ','.join(award_ids_delete)
+
+        db_cursor = connections['default'].cursor()
+
+        # Financial Accounts by Awards
+        fa = 'DELETE ' \
+             'FROM "financial_accounts_by_awards" fa ' \
+             'WHERE fa."award_id" IN ({});'.format(award_ids_list)
+        # Subawards
+        sub = 'DELETE ' \
```
nmonga91: Here, i don't think we want to _delete_ the subaward but instead set their `award_id` to null? Since we're still ingesting subawards even if we can't link them to an award

usaspending_api/broker/management/commands/fabs_nightly_loader.py
```diff
@@ -65,8 +65,44 @@ def delete_stale_fabs(ids_to_delete=None):
...
+              'FROM "awards_subaward" sub ' \
+              'WHERE sub."award_id" IN ({});'.format(award_ids_list)
+        # Transaction FPDS
+        fpds = 'DELETE ' \
```
nmonga91: I don't think this FPDS delete is necessary since we're only dealing with FABS in this specific loader.

usaspending_api/broker/management/commands/fpds_nightly_loader.py
```diff
@@ -107,7 +107,44 @@ def delete_stale_fpds(ids_to_delete=None):
...
+        award_ids_list = ','.join(award_ids_delete)
+
+        db_cursor = connections['default'].cursor()
+
+        # Financial Accounts by Awards
+        fa = 'DELETE ' \
+             'FROM "financial_accounts_by_awards" fa '\
+             'WHERE fa."award_id" IN ({});'.format(award_ids_list)
+        # Subawards
+        sub = 'DELETE ' \
```
nmonga91: Same as above

usaspending_api/broker/management/commands/fpds_nightly_loader.py
```diff
@@ -107,7 +107,44 @@ def delete_stale_fpds(ids_to_delete=None):
...
+              'FROM "awards_subaward" sub '\
+              'WHERE sub."award_id" IN ({});'.format(award_ids_list)
+        # Transaction FPDS
+        fpds = 'DELETE ' \
+               'FROM "transaction_fpds" tf '\
+               'WHERE tf."transaction_id" IN ({});'.format(transaction_id_list)
+        # Transaction FABS
+        fabs = 'DELETE ' \
```
nmonga91: I don't think this FABS delete is necessary since we're only dealing with FPDS in this specific loader.

usaspending_api/broker/management/commands/fabs_nightly_loader.py
```diff
@@ -65,8 +65,44 @@ def delete_stale_fabs(ids_to_delete=None):
...
+        # Transaction FABS
+        fabs = 'DELETE ' \
+               'FROM "transaction_fabs" tf ' \
+               'WHERE tf."transaction_id" IN ({});'.format(transaction_id_list)
+        # Transaction Normalized
+        tn = 'DELETE ' \
+             'FROM "transaction_normalized" tn ' \
+             'WHERE tn."id" IN ({});'.format(transaction_id_list)
+        # Awards
+        awards = 'DELETE ' \
```
nmonga91: Would we want to delete the award every time? or just when the _only_ transaction for that award is deleted?

usaspending_api/broker/management/commands/fabs_nightly_loader.py
dpb-bah: Agreed. Removing.

usaspending_api/broker/management/commands/fpds_nightly_loader.py
dpb-bah: Agreed. Removing.

usaspending_api/broker/management/commands/fabs_nightly_loader.py
dpb-bah: Agreed. Slightly worried the script was still doing this for a month now...

usaspending_api/broker/management/commands/fabs_nightly_loader.py
tony-sappe: If I understand the original logic, using Django the script would delete the award, then recreate it later if (and only if) there is a new transaction which belongs with that award.

I guess we really only want to delete the award if there are no other transactions associated, if not would need to update the award record instead of deleting. But does that fall within the scope of getting the nightly loader to work with the minor changes to the matview models?

PR #1217, opened 2018-04-20
Changed default CodeClimate config params

tony-sappe:
[DEV-851](https://federal-spending-transparency.atlassian.net/browse/DEV-851)

### High level description
Changed default CodeClimate configuration parameters for ones closer to where our code base currently is, so it becomes easier to chip away at CC issues. As a development team we would commit to not merging a PR which contains code introducing new CC issues. After there are 0 reported issues the configuration can be modified to be more strict which would facilitate a gradual improvement to our codebase.

### Technical details
None to speak of

### The following are ALL required for the PR to be merged
- [x] Matview impact assessment completed **N/A**
- [x] Frontend impact assessment completed **N/A**
- [x] Data validation completed **N/A**
- [x] API Performance evaluation completed and present **N/A**


.codeclimate.yml
```diff
@@ -1,5 +1,44 @@
...
+    config:
+      threshold: 45
+  nested-control-flow:
+    enabled: true
+    config:
+      threshold: 5
+  return-statements:
+    enabled: true
+    config:
+      threshold: 3
```
nmonga91: may want to revisit this one. sometimes it's better to short circuit the function with a return statement rather than adding additional control flow that just continues to the bottom of the function. May need to be a case by case basis for this.

.codeclimate.yml
```diff
@@ -1,5 +1,44 @@
+version: "2"
+
 languages:
   Python: true
+  Javascript: true
+
```
nmonga91: is the consensus to put a threshold on _everything_? for some things that the team agrees we don't need, would it make sense to exclude them from being checked? Similar to what frontend does

tony-sappe: I thought "why not" since there isn't much JS. I did exclude this dir from the tests: "usaspending_api/static_doc_files/js/*"

.codeclimate.yml
tony-sappe: I agree that sometimes it is better/cleaner to "short-circuit" a function. My thought process is that if the logic requires 4 or more return statements perhaps the function should be refactored.

.codeclimate.yml
nmonga91: Apologies, I meant to ask about the checks listed below

tony-sappe:
@kwhickey No, there are still a large number of outstanding issues. See https://federal-spending-transparency.atlassian.net/browse/DEV-851

PR #1233, opened 2018-05-07
Increase usage of TinyShield on v2 API, API versioning framework

IsaacRay:
**High level description:**
Validation for all search endpoints using TinyShield, and a framework for API versioning.

**Technical details:**

API versioning is handled by a decorator. To version an endpoint, you would create a function that takes the current form of the endpoint request, and transforms it to match the new endpoint structure depending on the version.  You would also allow it to accept the new endpoint structure. Then return the request to the decorator. The idea is that we should be able to version the API simply by bumping the API_VERSION constant in settings.py, and then go through the transformation functions for the prior API version and either remove them completely, or remove the portion of the function that handles backwards compatibility.

TinyShield is now implemented on all search endpoints. It enforces strict adherence to the request structures as outlined in the API docs, but also provides helpful response codes if the request does not conform. TinyShield uses award_filter.py in the core/validators directory to create models of the API requests that it uses to validate. You can also pass in ad-hoc models inline. TinyShield validates the request using `TinyShield(models).block(request.data)`.

**Link to JIRA Ticket:**
[DEV-784](https://federal-spending-transparency.atlassian.net/browse/DEV-784)

- [x] Frontend impact assessment completed


usaspending_api/core/validator/award_filter.py
```diff
@@ -13,8 +14,9 @@
     {'name': 'place_of_performance_scope', 'type': 'enum', 'enum_values': ['domestic', 'foreign']},
     {'name': 'program_numbers', 'type': 'array', 'array_type': 'text', 'text_type': 'search'},
     {'name': 'psc_codes', 'type': 'array', 'array_type': 'text', 'text_type': 'search'},
-    {'name': 'recipient_scope', 'type': 'enum', 'enum_values': ('domestic', 'foreign')},
```
tony-sappe: Is there a difference between using a set vs a list?

usaspending_api/settings.py
```diff
@@ -629,3 +629,5 @@
     "original_loan_subsidy_cost": "original_loan_subsidy_cost",
     "business_funds_indicator": "business_funds_indicator"
 }
+
+API_VERSION = 2
```
tony-sappe: Please move close to the top of the file.

usaspending_api/core/validator/award_filter.py
```diff
@@ -50,4 +52,8 @@
     a['optional'] = a.get('optional', True)  # future TODO: want to make time_period required
     a['key'] = 'filters{sep}{name}'.format(sep=TINY_SHIELD_SEPARATOR, name=a['name'])
     if a['type'] == 'array':
-        a['min'] = a.get('min', 1)
+        a['array_min'] = a.get('array_min', 1)
+        a['array_max'] = a.get('array_max', 0)
```
tony-sappe: Originally, a max value is applied to a model if it is not provided. It is probably a good idea to set an array and object max values. Even if the max is quite large (e.g. 5000) it is still providing an upper boundary for API users

usaspending_api/core/validator/helpers.py
```diff
@@ -170,7 +176,9 @@ def validate_text(rule):
         ord('\r'): None,  # carriage return
         ord('\n'): None,  # newline
     }
-    text_type = rule['text_type']
+    text_type = rule.get('text_type', None)
```
tony-sappe: `text_type` is already checked in TinyShield.check_models()

usaspending_api/search/v2/views/search.py
```diff
@@ -145,7 +154,14 @@ def post(self, request):
...
+            {'name': 'category', 'key': 'category', 'type': 'enum',
+                'enum_values': ["awarding_agency", "funding_agency", "recipient", "cfda_programs", "industry_codes"],
+                'optional': False}
+        ]
+        models.extend(copy.deepcopy(AWARD_FILTER))
+        models.extend(copy.deepcopy(PAGINATION))
+        json_request = TinyShield(models).block(request.data)
         category = json_request.get("category", None)
         scope = json_request.get("scope", None)
         filters = json_request.get("filters", None)
```
tony-sappe: (lines 168-169) `limit` and `page` default values should have their defaults applied in the TinyShield models

Also, for non-optional fields (for example, `category`) TinyShield should give confidence that those values exist, so instead of category = json_request.get("category", None) It could just be used as json_request["category"] in the code.

usaspending_api/search/v2/views/search.py
```diff
@@ -885,9 +921,9 @@ class SpendingByTransactionCountVisualizaitonViewSet(APIView):
     @cache_response()
     def post(self, request):
 
-        models = [{'name': 'keyword', 'key': 'filters|keyword', 'type': 'text', 'text_type': 'search', 'min': 3}]
+        models = [{'name': 'keyword', 'key': 'filters|keyword', 'type': 'array', 'array_type': 'text',
+                  'text_type': 'search', 'optional': False}]
```
tony-sappe: lost the `min` value of 3. That is a desired constraint

usaspending_api/core/validator/award_filter.py
```diff
@@ -14,7 +15,8 @@
     {'name': 'program_numbers', 'type': 'array', 'array_type': 'text', 'text_type': 'search'},
     {'name': 'psc_codes', 'type': 'array', 'array_type': 'text', 'text_type': 'search'},
     {'name': 'recipient_scope', 'type': 'enum', 'enum_values': ('domestic', 'foreign')},
-    {'name': 'recipient_search_text', 'type': 'array', 'array_type': 'text', 'text_type': 'search', 'max': 1, 'min': 1},
+    {'name': 'recipient_search_text', 'type': 'array', 'array_type': 'text',
+     'text_type': 'search', 'array_max': 1, 'array_min': 1},
```
tony-sappe: (Adding discussed comments to document and track)
Originally, the `min` and `max` were "top-level" attributes while the type-specific min and max (`array_max`, `object_min`, ...) were intended for the sub-level. If that can be maintained it will probably allow more consistency across all models and different places in the code.

However, if it requires a non-trival amount of effort to revert back then keep as-is for now and we can reevaluate in the future.

usaspending_api/core/validator/helpers.py
```diff
@@ -25,11 +25,14 @@
 def _check_max(rule):
     value = rule['value']
     if rule['type'] in ('integer', 'float'):
-        if value > rule['max']:
+        if value > rule['max'] and rule['max'] != 0:
```
tony-sappe: Unsure it is a good idea to have a unenforced max value as 0. Would it be easier to apply the max defaults by type instead of setting the value to 0? They are already very large values and wouldn't require special logic.

IsaacRay: This is an artifact and should have been removed, good catch

usaspending_api/search/v2/views/search.py
IsaacRay: Good catch here, this should have been separated out to the other ticket.

usaspending_api/core/validator/award_filter.py
IsaacRay: Also, for the sake of getting this ticket out, I'd like to push this particular change towards a later, phase 2 assessment of tinyshield once we've had it in the wild a little bit.

tony-sappe:
Matview impact assessment completed **(N/A)**
Frontend impact assessment completed **(N/A)**
Data validation completed **(N/A)**
API Performance evaluation completed  **(N/A)**

PR #1236, opened 2018-05-09
Advanced Search Page API Performance TLC

tony-sappe:
[DEV-909](https://federal-spending-transparency.atlassian.net/browse/DEV-909)

### High level description
A number of changes over the past few months took a toll on Advanced Search Page API performance. Noteworthy changes include:
- Using matviews for downloads of data pre-FY2008
- Including loan subsidy costs in the aggregation amounts for visualization API endpoints

This PR primarily makes changes to materialized view indexes and ORM code to craft slightly different SQL select statements.

### Technical details
1. All matviews which are not necessary for downloads were changed back to FY2008 and forward only (`summary_*` views)
2. The matviews which do include FY2001-FY2007 data had substantial rework to their indexes (`universal_*` views)
    - All indexes are partial indexes using action date (existing indexes are now only for >= '2007-10-01')
    - New indexes which cover older data for downloads
3. `universal_award_matview` is now ordered by `total_obligation DESC NULLS LAST` instead of `action_date DESC`
4. No longer using fiscal_year for queries
5. Using `generated_pragmatic_obligation` instead of federal_action_obligation or total_subsidy_cost depending on the award type
6. Capped Advanced Search Page API date range minimum start date to 2007-10-01
7. Combined action_date ranges when requested ranges overlap or are adjacent (e.g. {2008-10-01, 2009-09-30} and {2009-10-01, 2010-09-30} are merged into {2008-10-01, 2010-09-30} 
8. Explicit ORM->SQL ordering for spending by award to be DESC NULLS LAST which matches existing indexes

### The following are ALL required for the PR to be merged
- [x] Frontend impact assessment completed **N/A**
- [x] API Performance evaluation completed and present

#### Matview impact assessment
All materialized views should be recreated as all have significant changes to either the view, or indexes, or both. Creation time should be shorter and total storage needs are reduced. Unfortunately neither are back to the original (FY2008+ only) metrics. Functionality won't break if the code is deployed before DB changes, but performance will degrade even more than currently.

#### Data validation completed
Aggregated obligation values need to be verified just to confirm accuracy, even thought the logic is the same, simply moved from API SQL to matview create SQL.


usaspending_api/core/validator/helpers.py
```diff
@@ -16,9 +16,9 @@
 
 TINY_SHIELD_SEPARATOR = '|'
 
-INVALID_TYPE_MSG = 'Invalid value in \'{key}\'. \'{value}\' is not a valid {type}'
-ABOVE_MAXIMUM_MSG = 'Field \'{key}\' value \'{value}\' is above max {max}'
-BELOW_MINIMUM_MSG = 'Field \'{key}\' value \'{value}\' is below min {min}'
+INVALID_TYPE_MSG = 'Invalid value in \'{key}\'. \'{value}\' is not a valid ({type})'
```
dpb-bah: `[Value] is not a valid (string)`. Is that supposed to be `[Value] is not a valid type (string)`?

usaspending_api/database_scripts/matview_generator/universal_award_matview.json
```diff
@@ -131,7 +131,7 @@
     "  (awards.category IS NOT NULL or contract_data.pulled_from='IDV') AND",
     "  latest_transaction.action_date >= '2000-10-01'",
     "ORDER BY",
-    "  latest_transaction.action_date DESC"
+    "  awards.total_obligation DESC NULLS LAST"
```
dpb-bah: Asking more for documentation: why the change of order?

usaspending_api/search/tests/test_mock_data_search.py
```diff
@@ -31,7 +31,7 @@ def all_filters():
         "place_of_performance_scope": "domestic",
         "place_of_performance_locations": [{"country": "USA"},
                                            {"country": "PQR"}],
-        "award_type_codes": ["A", "B", "02", '05', 'S'],
+        "award_type_codes": ["A", "B"],
```
dpb-bah: Why the dropping of the "02", '05', 'S'?

usaspending_api/search/v2/views/search.py
```diff
@@ -81,20 +84,28 @@ def post(self, request):
...
+                .values('award_type') \
+                .annotate(
+                    month=ExtractMonth('action_date'),
+                    year=ExtractYear('action_date'),
+                    transaction_amount=Sum('amount')) \
+                .values('month', 'year', 'transaction_amount')
         else:
-            data_set = queryset.values('fiscal_year')
-            if not (group == 'fy' or group == 'fiscal_year'):
+            queryset = queryset.values('fiscal_year')
```
dpb-bah: Can we include the `for Awards we use sum_transaction_amount` down here?

usaspending_api/settings.py
```diff
@@ -19,6 +19,10 @@
 # User-specified timeout limit for streaming downloads
 DOWNLOAD_TIMEOUT_MIN_LIMIT = 10
 
+API_MAX_DATE = '2020-09-30'
```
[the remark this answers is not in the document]
tony-sappe: Yeah, I thought it was far enough into the future for now. It could be 2099-09-30 and it wouldn't change anything. It is for the date range combine logic

usaspending_api/search/v2/views/search.py
tony-sappe: What is `sum_transaction_amount`?

usaspending_api/core/validator/helpers.py
dpb-bah: It's just a grammatical error:
`' \'{value}\' is not a valid ({type})'`
will come out like `'A' is not a valid (int)'`
`\'{value}\' is not a valid type ({type})'` will come out like `'A' is not a valid type (int)'`

usaspending_api/search/v2/views/search.py
dpb-bah: Sorry, this should just be `#  for Awards we use transaction_amount`.

dpb-bah: Transaction amount you can see below on lines 100 and 107.

PR #1255, opened 2018-05-16
Extra SQL Log for Download Exceptions

dpb-bah:
Requested by the Operations team for whenever downloads fail.


nmonga91:
@willkjackson @kwhickey @hsharma96 @sharb 

Are the error messages added here enough to provide additional info for the `subprocess` call?

kwhickey:
If this is just the short-term fix to surface some additional logs, then yes it should be fine. We have a tech debt item for a longer-term investigation into logging sub-processes: DEV-1024

PR #1271, opened 2018-05-23
DEV-979: Adding statement_timeout to Django database connections

nmonga91:
**High level description:**
Added optional use of setting a statement timeout for long running database queries, at the Django level. The setting does not persist across connections, and since each API request is a new connection, no reset of the value is necessary.

**Technical details:**
Added a decorator to set a statement_timeout on a Django database connection. Defaults to 5 min but can be overridden to different times for different endpoints. The setting does not persist across connections, and since each API request is a new connection, no reset of the value is necessary.

**Link to JIRA Ticket:**
[DEV-979](https://federal-spending-transparency.atlassian.net/browse/DEV-979)

  - Tested manually but did not see a good way to test the timeout without having the unit/integration tests sending a sleep request to the database to test the statement timeout value.


kwhickey:
@nmonga91 It'd be good to see some automated integration tests that verify the behavior. An integration tests that runs a query that just spins, and spins longer than the timeout, catches the exception, and confirms the the duration elapsed (from a python timer) is only slightly greater than the db timeout, but less than the sql query sleep. 

Postgres has a pg_sleep: https://www.if-not-true-then-false.com/2010/postgresql-sleep-function-pg_sleep-postgres-delay-execution/

-- You may need to use django's feature to call raw sql, or find a way to run sql in test prep that puts a read-lock on a table that a django model wants to query

Another test we need is concurrent queries can run with different timeouts. E.g. have a master test function that calls 2 others with decorators on them, but different timeouts. Ensure each uses its own timeout and the decorator doesn't end up editing some _global_ timeout.
- may need to do a few riffs on this:
a) sequential calls
b) parallel calls in 2 different threads
c) parallel calls in 2 different sub-processes

(I'm worried the connection and this statement_timeout may be shared by django across db calls)

... I'll look more at the code in a bit.

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,40 @@
+import logging
+
+from django.db import connection
+from django.utils.decorators import method_decorator
+
+DEFAULT_DB_TIMEOUT_IN_MS = 300000
+logger = logging.getLogger(__name__)
+
+
+def set_db_timeout(timeout_in_ms=DEFAULT_DB_TIMEOUT_IN_MS):
```
tony-sappe: Can this value be in seconds instead of ms? The value can be multiplied by 1000 when it is setting `statement_timeout` in postgres

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,40 @@
...
+    """ Decorator used to set the database statement timeout within the Django app scope
+
+        Args:
+            timeout_in_ms: timeout value to set, in milliseconds
+
+        NOTE:
+            The statement_timeout is only set for this specific connection. Any subsequent request will have a default
```
tony-sappe: From my understanding, Django/uwsgi will maintain an idle connection to the DB as long as the worker exists. That would mean the setting does persist the specific API request handling.

Running `select * from pg_stat_activity;` in a production DB will typically show over 100 idle connections.

usaspending_api/common/helpers/decorators.py
nmonga91: I can change that

usaspending_api/common/helpers/decorators.py
nmonga91: I've added a reset

nmonga91:
@kwhickey Tests have been added

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,72 @@
...
+from usaspending_api.common.exceptions import EndpointTimeoutException
+
+logger = logging.getLogger(__name__)
+
+
+def set_db_timeout(timeout_in_seconds=settings.DEFAULT_DB_TIMEOUT_IN_SECONDS):
+    """ Decorator used to set the database statement timeout within the Django app scope
+
+        Args:
+            timeout_in_seconds: timeout value to set, in milliseconds
```
kwhickey: Says `timeout_in_seconds` but docs say milliseconds. Clarify.

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,72 @@
...
+from django.conf import settings
+from django.db import connection
+from django.db.utils import OperationalError
+
+from usaspending_api.common.exceptions import EndpointTimeoutException
+
+logger = logging.getLogger(__name__)
+
+
+def set_db_timeout(timeout_in_seconds=settings.DEFAULT_DB_TIMEOUT_IN_SECONDS):
```
kwhickey: same as below. clarify sec or ms

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,72 @@
...
+            the database, there is no need to reset the statement_timeout value.
+
+        Examples:
+            @set_db_timeout(test_timeout_in_ms)
+            def func_running_db_call(...):
+                ...
+
+            OR
+
+            @set_db_timeout()  # This will use the default value in settings
```
kwhickey: Can that default timeout not be applied globally, _as the default_? Or must every call that needs to timeout have this decorator (with or without args)?

Maybe there's a global patch that can be applied to set a default `statement_timeout` on anything within the app that calls the db?

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,72 @@
...
+            def func_running_db_call(...):
+                ...
+    """
+
+    timeout_in_ms = int(timeout_in_seconds * 1000)
+
+    def wrap(func):
+        def wrapper(*args, **kwargs):
+            with connection.cursor() as cursor:
+                cursor.execute("show statement_timeout")
```
kwhickey: Are you sure that our app db user will have permissions to query the timeout in prod?

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,72 @@
...
+                logger.warning('DB TIMEOUT DECORATOR: Old Django statement_timeout value = %s' %
```
kwhickey: You should say old "Postgres" timeout here, since it is actually printing the value set on the Postgres db, correct? Maybe add "_on this connection_" if the value is just for one db connection.

usaspending_api/common/helpers/decorators.py
kwhickey: Does this actually require a call to the DB? 
Does Django store the timeout as any kind of attribute of a connection object, that can be looked at (and assumed default if null)?

Multiple round-trips to the db on every db call to get and then set a db timeout might hurt our overall performance, especially for chatty interfaces/apis

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,72 @@
...
+                    'DB TIMEOUT DECORATOR: Setting Django statement_timeout to %dms' % timeout_in_ms)
+                cursor.execute("set statement_timeout={0}".format(timeout_in_ms))
+
+                cursor.execute("show statement_timeout")
+                logger.warning('DB TIMEOUT DECORATOR: New Django statement_timeout value = %s' %
+                               str(cursor.fetchall()[0][0]))
+
+            try:
+                func_response = func(*args, **kwargs)
+            except OperationalError:
```
kwhickey: You sure the timeout is the only thing that might throw this exception?

usaspending_api/common/tests/integration/test_decorators.py
```diff
@@ -0,0 +1,60 @@
+# Stdlib imports
```
kwhickey: Can a test be added for using the default timeout in settings? 

To avoid a 5min test, you'd have to feed the test runner different settings. Can you override settings for a single test?

kwhickey: It'd also be good to have a inverse test with a pg_sleep and no decorator, and prove that the time it took to run was >= to the pg_sleep amount.

usaspending_api/common/helpers/decorators.py
```diff
@@ -0,0 +1,72 @@
...
+
+def set_db_timeout(timeout_in_seconds=settings.DEFAULT_DB_TIMEOUT_IN_SECONDS):
+    """ Decorator used to set the database statement timeout within the Django app scope
+
+        Args:
+            timeout_in_seconds: timeout value to set, in milliseconds
+
+        NOTE:
+            The statement_timeout is only set for this specific connection. Any subsequent request will have a default
+            value of 0. The alteration DOES NOT persist across connections and since each request is a new connection to
```
kwhickey: This seems to contradict the finally, where you do set it back to 0. Maybe just reword, and say "while this won't persist across connections, we set it back to 0 as a precaution after the decorated function completes", in case anything reuses that connection.

usaspending_api/common/helpers/decorators.py
nmonga91: Will update

usaspending_api/common/helpers/decorators.py
nmonga91: The timeout won't be set unless the decorator is used. So it won't affect anything globally.

usaspending_api/common/helpers/decorators.py
nmonga91: This is a good thought, thinking back, i'm not sure if we're even able to set the statement timeout on a read replica. the django app manages switching between the replica and main, so that'll need to be investigated further

usaspending_api/common/helpers/decorators.py
nmonga91: This is the lowest level for which we can handle the timeout, so if other situations can cause an operational error, it'll be difficult to decipher without checking the actual log message

usaspending_api/common/tests/integration/test_decorators.py
nmonga91: You can, i can add a test that mocks the settings value to test the default

usaspending_api/common/helpers/decorators.py
nmonga91: From what I researched on the statement_timeout options, it looks like setting requires a DB call and it doesn't store it anywhere else, so it would require a call to the DB to retrieve the latest value for this. 

These should be sub-second queries and I would not foresee them adversely affecting performance on a call compared to the execution of other components of the business logic.

kwhickey:
Approving based on answers to questions and further code review. 

I agree that this may not be good for production yet without some time to trial this on a particular endpoint that gets a lot of activity. Let's discuss what endpoint we want to put this decorator on for next sprint's work.

We also want to determine the frontend behavior when it times out. 
_Do we want to let the frontend timeout be slightly > backend db statement timeout, to allow the frontend to give some indication that the search was too complex on certain pages?_ Or continue to just drop at 1min with "an error occurred".

And last possible enhancement is to ensure we can find the timed-out web requests in the logs. We definitely will want to search specifically for this condition. Can we do that by response code? By the error logged when the timeout occurs?

PR #1296, opened 2018-06-08
Hotfix: Subaward loader dups

dpb-bah:
**High level description:**
Resolving the duplicate bug created from the subaward loader after modifying it to handle subawards without linked awards. 

**Technical details:**
* Problem was the delete function wasn't mapping the `internal_id`s correctly as broker has them lowercased whereas the website has them uppercased. This includes the fix to match them correctly.
* Also discovered the broken award linkages base their data off the `broker_award_id`s, which, upon learning, get deleted whenever the award/subaward data gets updated from FSRS. This now uses `internal_id` as that doesn't change and what the broker uses identify subawards. 
* Updated Subaward Loader tests
    * Using positive mock ID integers as the subaward ignores ids < 1 to delete/update
    * Using lowercased mock internal IDs to match the broker
    * Added test for deletes/updates
    * Added test for fixing broken award links

**Link to JIRA Ticket:**
[DEV-894](https://federal-spending-transparency.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=DEV&modal=detail&selectedIssue=DEV-894)

- [X] Unit & integration tests updated with relevant test cases
- Matview impact assessment completed (N/A)
- Frontend impact assessment completed (N/A)
- [X] Data validation completed
- API Performance evaluation completed and present (N/A)


usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -23,7 +23,7 @@ class Command(BaseCommand):
     help = "Load new FSRS data from broker."
 
     @staticmethod
-    def get_award_data(db_cursor, award_type, max_id, ids=None):
+    def get_award_data(db_cursor, award_type, max_id, internal_ids=None):
```
nmonga91: so when we're re-associating unlinked subawards, we look them up by their report id (aka internal_id) now instead of the broker PKs?

dpb-bah: Correct. The broker grants/procurements get deleted and then readded per update based incoming FSRS data.

PR #1301, opened 2018-06-13
DEV-979: Configure default DB statement timeout

nmonga91:
**High level description:**
Removing default timeout settings var (to be set by ops in configs) making `timeout_in_seconds` param required for `set_db_timeout` decorator.

**NOTE:** Download endpoints have been overridden with a statement timeout of `0` which means there is no timeout.

**Technical details:**
Removing default timeout settings var (to be set by ops in configs) making `timeout_in_seconds` param required for `set_db_timeout` decorator. 

**NOTE:** Download endpoints have been overridden with a statement timeout of `0` which means there is no timeout.

@kwhickey @willkjackson OPs will need to utilize the following command when deploying the application to set the default statement timeout for all endpoints (which can be overridden using the provided decorator):
```
export PGOPTIONS='-c statement_timeout=5min'
```

**Link to JIRA Ticket:**
[DEV-979](https://federal-spending-transparency.atlassian.net/browse/DEV-979)


nmonga91:
@kwhickey in response to your review, we would still need to explicitly set the timeout on download endpoints to `0`, yes? if running locally, we won't be able to set different launch time env vars as we can with an actual deploy.

kwhickey:
@nmonga91  - 
Yes, this approach doesn't change anything about the need to still have operations team work done to the deployment files, so that an environment variable is set with the desired timeout for each deployed component (where the download deployment may use a different env var value).

This has the added benefit of:
1) It's easier to see what's going on with the timeout, and trace back to where it is being set. Easier to reason about, which is always an important thing with configs. 
2) If we _don't_ for some reason explicitly set the  env var, it will fall back to the file's default of 300 seconds

This also means that we should keep the decorators on the download endpoints too, so that in #2 above, 300 gets overridden to 0.

Developers locally would have 2 options: 
1) Edit the `settings.py` file to change the config, but be sure not to commit the change
2) Set an environment variable in their terminal when they run the app, similar to how we recommend setting the `DATA_BROKER_DATABASE_URL` env var for local dev.

PR #1303, opened 2018-06-14
Update spending explorer unreported data to only be displayed at top level entry points

nmonga91:
**High level description:**
Updating to only show unreported data if at the top level of the spending explorer, therefore had to check that the type wasn't used again with additional filters.

**Technical details:**
Updating to only show unreported data if at the top level of the spending explorer. Had to check that the type wasn't used again with additional filters and then set the `expected_total` to the `actual_total` if unreported data wasn't being utilized.

**Link to JIRA Ticket:**
(PENDING ACTUAL TICKET, WILL BE UPDATED)
[DEV-123](https://federal-spending-transparency.atlassian.net/browse/DEV-123)

  - Data validated to no longer show unreported data at nested levels even if the `type` matches one of the top entry points

### DIFF
--- a/usaspending_api/spending_explorer/v2/filters/type_filter.py
+++ b/usaspending_api/spending_explorer/v2/filters/type_filter.py
@@ -11,9 +11,10 @@ from usaspending_api.spending_explorer.v2.filters.spending_filter import spendin
 
 UNREPORTED_DATA_NAME = 'Unreported Data'
 VALID_UNREPORTED_DATA_TYPES = ['agency', 'budget_function', 'object_class']
+VALID_UNREPORTED_FILTERS = ['fy', 'quarter']
 
 
-def get_unreported_data_obj(queryset, limit, spending_type, actual_total, fiscal_year, fiscal_quarter) -> \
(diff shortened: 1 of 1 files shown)


usaspending_api/spending_explorer/v2/filters/type_filter.py
```diff
@@ -45,7 +47,7 @@ def get_unreported_data_obj(queryset, limit, spending_type, actual_total, fiscal
         values_list('total_obligation', flat=True). \
         first()
 
-    if spending_type in VALID_UNREPORTED_DATA_TYPES:
+    if spending_type in VALID_UNREPORTED_DATA_TYPES and list(filters.keys()).sort() == VALID_UNREPORTED_FILTERS.sort():
```
tony-sappe: Comparing equality between two `set`s might be better. A quick check in the REPL showed that sets do not need to be ordered for equality

`... and set(filters.keys()) == set(VALID_UNREPORTED_FILTERS):`

nmonga91: Updated to use set comparison for completeness

PR #1324, opened 2018-07-09
USASPENDING_AWS_REGION env var consolidation

willkjackson:
**High level description:**
Over the history of this API, many environment variable names have been used. Formerly, this repo used:
`CSV_AWS_REGION `
`BULK_DOWNLOAD_AWS_REGION`
and in many ETL scripts:
`AWS_REGION`

(There is also `CFDA_REGION`, which needs to exist separately as it comes from a different source).

This consolidates all of these into one environment variable, `USASPENDING_AWS_REGION`.

**Link to JIRA Ticket:**
[link not provided]

- [ ] Operations review completed (for scripts/jobs that use these env vars)

### DIFF
--- a/usaspending_api/download/v2/views.py
+++ b/usaspending_api/download/v2/views.py
@@ -41,7 +41,7 @@ from usaspending_api.references.models import ToptierAgency
 
 @api_transformations(api_version=settings.API_VERSION, function_list=API_TRANSFORM_FUNCTIONS)
 class BaseDownloadViewSet(APIDocumentationView):
-    s3_handler = S3Handler(name=settings.BULK_DOWNLOAD_S3_BUCKET_NAME, region=settings.BULK_DOWNLOAD_AWS_REGION)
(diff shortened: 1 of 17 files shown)


usaspending_api/download/v2/views.py
```diff
@@ -41,7 +41,7 @@
 
 @api_transformations(api_version=settings.API_VERSION, function_list=API_TRANSFORM_FUNCTIONS)
 class BaseDownloadViewSet(APIDocumentationView):
-    s3_handler = S3Handler(name=settings.BULK_DOWNLOAD_S3_BUCKET_NAME, region=settings.BULK_DOWNLOAD_AWS_REGION)
+    s3_handler = S3Handler(name=settings.BULK_DOWNLOAD_S3_BUCKET_NAME)
```
nmonga91: Will this fall back to the env var?

willkjackson: Yes, if you look at S3Handler's `init` (not sure why we are using such a simple wrapper for boto) it is set there.

PR #1336, opened 2018-07-18
OPS-257 qualifying the schema in which the enum type is declared for pg_dump restore

kwhickey:
**High level description:**
Cannot restore dumped databases without qualifying where (what schema) pg_restore or psql should find the `total_obligation_bins` enum type. This adds that schema-qualification

**Technical details:**
See: See this postgres bug report for details: https://www.postgresql.org/message-id/flat/153184074037.1405.15097715315257989847%40wrigleys.postgresql.org

**Link to JIRA Ticket:**
[OPS-257](https://federal-spending-transparency.atlassian.net/browse/OPS-257) discovered during working through this Operations Epic

**Links to Dependent/Related PRs**
N/A

**Special Deployment Steps**

1. Will need to explicitly run this sql script on each database. Note, if not run, it will run as part of the nightly pipeline where that jenkins job is run

- [ ] Unit & integration tests updated with relevant test cases N/A
- [ ] Frontend impact assessment completed N/A
- [ ] Data validation completed N/A
- [ ] API Performance evaluation completed and present: N/A

### DIFF
--- a/usaspending_api/database_scripts/matviews/functions_and_enums.sql
+++ b/usaspending_api/database_scripts/matviews/functions_and_enums.sql
@@ -5,10 +5,10 @@ CREATE EXTENSION IF NOT EXISTS intarray;
 
 -- The function(s) and enum(s) below were originally created for the materialized views
 
-CREATE TYPE total_obligation_bins AS ENUM ('<1M', '1M..25M', '25M..100M', '100M..500M', '>500M');
+CREATE TYPE public.total_obligation_bins AS ENUM ('<1M', '1M..25M', '25M..100M', '100M..500M', '>500M');
 
 
-CREATE OR REPLACE FUNCTION obligation_to_enum(award NUMERIC) RETURNS total_obligation_bins AS $$
+CREATE OR REPLACE FUNCTION obligation_to_enum(award NUMERIC) RETURNS public.total_obligation_bins AS $$
   DECLARE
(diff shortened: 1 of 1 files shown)


usaspending_api/database_scripts/matviews/functions_and_enums.sql
```diff
@@ -5,10 +5,10 @@ CREATE EXTENSION IF NOT EXISTS intarray;
 
 -- The function(s) and enum(s) below were originally created for the materialized views
 
-CREATE TYPE total_obligation_bins AS ENUM ('<1M', '1M..25M', '25M..100M', '100M..500M', '>500M');
+CREATE TYPE public.total_obligation_bins AS ENUM ('<1M', '1M..25M', '25M..100M', '100M..500M', '>500M');
```
tony-sappe: Is this line change necessary? If yes, I would say we want the functions to match this explicitness too to be consistent

kwhickey: Answer is, it depends. Depends on what the `search_path` is for the session executing the query. In a normal psql session, it will stay as default, which would create the object in `public` schema, unless a schema for the user's name exists.

I think though to be explicit, we should add it to all, so it will work regardless of a person's set `search_path` and  any schemas they have on their machine. Will make the change.

PR #1404, opened 2018-07-25
Stg -> Master

michaeldhess:


willkjackson:
Not ready according to @tony-sappe

PR #1431, opened 2018-08-11
Hotfix: Moving c to d linkage call to end of the submission load

nmonga91:
Moving c to d linkage call to end of the submission load for each submission with some additional code cleanup


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -119,17 +116,14 @@ def signal_handler(signal, frame):
                     .format(submission_id, award_financial_frame.shape[0]))
         logger.info('Loading File C data')
         start_time = datetime.now()
-        awards_touched = load_file_c(submission_attributes, db_cursor, award_financial_frame)
+        load_file_c(submission_attributes, db_cursor, award_financial_frame)
         logger.info('Finished loading File C data, took {}'.format(datetime.now() - start_time))
 
+        # update C to D linkage using SQL scripts
+        call_command('update_file_c_linkages')
```
kwhickey: @nmonga91 what is the usual runtime of this sql? 
Would it be simple to create a version of this that takes in a submission ID and somehow narrows the awards that need to be linked, rather than attempting to fill _all_ gapped links? Or is this pretty efficient as-is, in that it targets just those `financial_accounts_by_award` records with a `null` `award_id`?

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -119,17 +116,14 @@ def signal_handler(signal, frame):
...
+
         if not options['nosubawards']:
-            try:
-                start_time = datetime.now()
-                logger.info('Loading subaward data...')
-                load_subawards(submission_attributes, awards_touched, db_cursor)
-                logger.info('Finshed loading subaward data, took {}'.format(datetime.now() - start_time))
-            except Exception:
-                logger.warning("Error loading subawards for this submission")
+            logger.warning("Loading subaward during a submission load is no longer supported")
```
kwhickey: @nmonga91 What is the impact on data completeness and quality if we are not attempting to load subawards any longer? 
Do we need to compensate with backfilling subawards later? Does your SQL somehow load them too? If not, when will they be loaded? 

Also, might it be better to leave this code here and just have the jenkins job stage pass in the `nosubawards` option?

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -661,32 +654,10 @@ def load_file_c(submission_attributes, db_cursor, award_financial_frame):
             update_skipped_tas(row, skipped_tas)
             continue
 
-        # Find a matching transaction record, so we can use its subtier agency information to match to (or create) an
-        # Award record.
-
-        # Find the award that this award transaction belongs to. If it doesn't exist, create it.
-        filters = {}
```
kwhickey: @nmonga91 If we have a way already to pass in options, like the `nosubaward` option, it may be better to just guard this block with another new option, leaving the code there to be optimized, but have jenkins run this stage with a `link_c_to_d_on_load` flag set to false.

If you were to guard this, the use of the sql linkage above would need to be activated by the same guard.

usaspending_api/etl/management/commands/load_submission.py
nmonga91: It targets just those `financial_accounts_by_award` records with a `null` `award_id`. It'll still operate on _all_ unlinked records so it's not the most efficient method here. I think it's best to move to a submission specific SQL set for this.

usaspending_api/etl/management/commands/load_submission.py
nmonga91: The jenkins job already passes `nosubawards` and this method doesn't load it correctly to begin with, so I just essentially "deprecated" it

usaspending_api/etl/management/commands/load_submission.py
nmonga91: The linkage should be attempted every time, I don't think we should allow the option to exclude it

PR #1442, opened 2018-08-13
C to D Linkage Mods

IsaacRay:
**High level description:**
Adjusts the logic for C to D linkages

**Technical details:**
Modifies the SQL to run two queries depending on whether the parent PIID is null or not, and matches on PIID only if it is null, and matches on both PIID and parent PIID if its not.

Link to JIRA Ticket:
[DEV-1374](https://federal-spending-transparency.atlassian.net/browse/DEV-1374)

The following are ALL required for the PR to be merged:

- [x] Unit & integration tests updated with relevant test cases
- [NA] Frontend impact assessment completed

When tested against dev database, PIIDs listed in ticket are successfully linked by the modifications. In the dev run of the linkage script, an additional ~200k records were linked.


usaspending_api/etl/management/sql/c_file_linkage/update_file_c_linkages_piid.sql
```diff
@@ -10,7 +10,7 @@ SET
 		WHERE
 			UPPER(aw.piid) = UPPER(faba.piid)
 			AND
-			UPPER(aw.parent_award_piid) IS NOT DISTINCT FROM UPPER(faba.parent_award_id)
+			UPPER(aw.parent_award_piid) = UPPER(faba.parent_award_id)
```
tony-sappe: Does the predicate ever need to be true when both `aw.parent_award_piid` and `faba.parent_award_id` are NULL? If so the original SQL handled that appropriately

PR #1546, opened 2018-09-14
Fixing minor SpendingByAward bugs

tony-sappe:
**Description:**
Some request edge cases were not being handled properly by the API logic.

**Technical details:**
When "no intersection" is provided in `award_type_codes` then there will never be results returned. So it is better to explicitly return 0 results early in the API logic rather than let the request processing continue to allow exceptions to occur in this edge case.

	- [ ] Backend
8. [x] Jira Ticket [DEV-1560](https://federal-spending-transparency.atlassian.net/browse/DEV-1560):

```
API documentation updated: No edits to API contract, just better enforcement of existing rules
Matview impact assessment completed: No matview edits
Data validation completed: No data edits or data logic edits
Appropriate Operations ticket(s) created: No data changes
```


usaspending_api/search/v2/views/search.py
```diff
@@ -277,8 +277,8 @@ def post(self, request):
                 m['optional'] = False
 
         json_request = TinyShield(models).block(request.data)
-        fields = json_request.get("fields", None)
-        filters = json_request.get("filters", None)
+        fields = json_request["fields"]
+        filters = json_request.get("filters", {})
```
dpb-bah: Should this logic also apply to SpendingByAwardCount (line 434)

usaspending_api/search/v2/views/search.py
```diff
@@ -287,6 +287,14 @@ def post(self, request):
         lower_limit = (page - 1) * limit
         upper_limit = page * limit
 
+        if "no intersection" in filters["award_type_codes"]:
```
dpb-bah: Should this check also be applied to `SpendingByAwardCount`?

usaspending_api/search/v2/views/search.py
tony-sappe: I thought about this, but I didn't want to make those changes without testing. It would also be necessary for SpendingByTransactionCount

dpb-bah: The SpendingByTransaction count point I agree would be more. I just bring this up considering SpendingByAward and SpendingByAwardCount go hand-in-hand and their numbers should match.

tony-sappe: I'll add it to SpendingByAwardCount since I have time.
I checked SpendingByTransactionCount and it doesn't support that filter so it is a moot point

PR #1558, opened 2018-09-20
NAICS duplicate code fix (branched from Staging)

justinle:
**Description:**
Fix for the duplicate NAICS codes

**Technical details:**
Found when NAICS codes appeared multiple times in auto-complete. This was caused by NAICS codes with trailing spaces. Issue resolved by importing NAICS codes as integers.

	- [ ] Backend
3. [X] Data validation completed

### DIFF
--- a/usaspending_api/references/management/commands/load_naics.py
+++ b/usaspending_api/references/management/commands/load_naics.py
@@ -4,7 +4,7 @@ import re
 import glob
 
 from django.conf import settings
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 from openpyxl import load_workbook
 from django.db import transaction
 
@@ -32,7 +32,7 @@ class Command(BaseCommand):
         load_naics(path=options['path'], append=options['append'])
 
 
-def populate_naics_fields(ws, naics_year):
+def populate_naics_fields(ws, naics_year, path):
     for current_row, row in enumerate(ws.rows):
         if current_row == 0:
             continue
@@ -43,7 +43,7 @@ def populate_naics_fields(ws, naics_year):
         try:
             naics_code = int(row[0].value)
         except ValueError:
-            logging.warning('could not parse to string: %s', row[0].value)
+            raise CommandError('Invalid NAICS code: {0}. Please review file {1}'.format(naics_code, path))
 
         naics_desc = row[1].value.strip()
 
@@ -76,4 +76,4 @@ def load_naics(path, append):
         ws = wb.active
 
         naics_year = p_year.search(ws["A1"].value).group()
-        populate_naics_fields(ws, naics_year)
+        populate_naics_fields(ws, naics_year, path)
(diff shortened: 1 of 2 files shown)


usaspending_api/references/management/commands/load_naics.py
```diff
@@ -40,8 +40,12 @@ def populate_naics_fields(ws, naics_year):
         if not row[0].value:
             break  # Reads file only until a blank line
 
-        naics_code = row[0].value
-        naics_desc = row[1].value
+        try:
+            naics_code = int(row[0].value)
+        except:
```
tony-sappe: "bare" except is a flake8 error

usaspending_api/references/management/commands/load_naics.py
```diff
@@ -40,8 +40,12 @@ def populate_naics_fields(ws, naics_year):
...
+            logging.warning('could not parse to string: %s', row[0].value)
```
tony-sappe: What does the try-except block do for the script? if it fails the type cast it will produce a python exception below since `naics_code` won't be defined. If you want to skip you will need to add a `continue` here. But I would discourage that. if int() is failing on any entries, perhaps you should use .strip() like you do for the description.

usaspending_api/references/management/commands/load_naics.py
```diff
@@ -40,8 +40,12 @@ def populate_naics_fields(ws, naics_year):
...
+
+        naics_desc = row[1].value.strip()
```
tony-sappe: Are there descriptions with trailing whitespace?

usaspending_api/references/tests/test_load_naics.py
```diff
@@ -14,8 +14,12 @@ def test_naics_existing_and_new_files():
 
     call_command('load_naics')
 
-    naics_2002_data_total = NAICS.objects.all().filter(year=2002).count()
+    naics_cascade_count = (NAICS.objects.all().filter(year=2017).count()
```
tony-sappe: I recommend not adding all of the SQL search results together. You can have an `assert` for one or two years.

There are other test cases you could add if you want. Some ideas are checking there are zero rows in the model before the `load_naics` command is executed; searching for a specific NAICS, verifying all rows have years only in 2002,2007,2012,2017.

usaspending_api/references/management/commands/load_naics.py
justinle: Not that I've found so far, that's just there for data validation in case of leading and trailing whitespaces.

usaspending_api/references/management/commands/load_naics.py
justinle: For validation purposes, the idea is to stop the script. The reasoning being that if the NAICS code cannot be parsed into an integer, the source file would need to be reviewed for integrity/security.

tony-sappe: If that is the case, you should raise an exception with a helpful and useful message in the except block to explicitly stop the script there.

PR #1560, opened 2018-09-21
Added NAICS description lookup to spending_by_category

tony-sappe:
**Description:**

**Technical details:**

	- [ ] Backend
8. [x] Jira Ticket [DEV-1109](https://federal-spending-transparency.atlassian.net/browse/DEV-1109):

```
DB models, matviews, and API structure remain unchanged in this PR.
No operations tasks are required.
```

### DIFF
--- a/usaspending_api/search/v2/views/spending_by_category.py
+++ b/usaspending_api/search/v2/views/spending_by_category.py
@@ -17,7 +17,7 @@ from usaspending_api.core.validator.award_filter import AWARD_FILTER
 from usaspending_api.core.validator.pagination import PAGINATION
 from usaspending_api.core.validator.tinyshield import TinyShield
 from usaspending_api.recipient.models import RecipientLookup, StateData
-from usaspending_api.references.models import Agency, Cfda, PSC, LegalEntity, RefCountryCode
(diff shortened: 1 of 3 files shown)


usaspending_api/search/v2/views/spending_by_category.py
```diff
@@ -404,3 +405,12 @@ def fetch_state_name_from_code(state_code):
         logger.warning("{} not found for state_code: {}".format(",".join(columns), state_code))
         return None
     return result[columns[0]]
+
+
+def fetch_naics_description_from_code(naics_code, passthrough=None):
```
IsaacRay: The function here is architected more flexibly than the name implies... i.e., you could add additional attributes to the "column" dict, and be fetching more than "description", so it might make sense to name the function something like fetch_naics_attributes_from_code and also potentially pass in the "columns" dict as an arg... not something that has to be changed, just a thought.

tony-sappe: I made from a copy of other similar functions in this file. The `columns` list is more useful for lines 414 and 416 than for 412 (if that makes sense).
I agree it could be more capable and it wouldn't require much effort, but there isn't any need.

PR #1566, opened 2018-09-26
CSV names now inherit archive name on Award Data Archive

justinle:
**Description:**
**Data Award Archive:** Archived files will lose its useful naming convention when unzipped, particularly the fiscal year, the download type (Full vs. Delta), creation date.

The contents of the archive files of Data Award Archive downloads are now inherited by the children `.csv` files.

**Technical details:**
Additional optional parameter named `file_name` has been added to `split_csv` function. When defined, the resulting archived `.csv` file inherits the archive name. 

1. [x] Necessary PR reviewers:
	- [x] Backend
2. [x] Jira Ticket [DEV-1367](https://federal-spending-transparency.atlassian.net/browse/DEV-1367):

### DIFF
--- a/usaspending_api/download/helpers.py
+++ b/usaspending_api/download/helpers.py
@@ -198,6 +198,7 @@ def split_csv(file_path, delimiter=',', row_limit=10000, output_name_template='o
         `output_name_template`: A %s-style template for the numbered output files.
         `output_path`: Where to stick the output files.
         `keep_headers`: Whether or not to print the headers in each output file.
+        `file_name`: Alternative naming convention for output files (optional)
     Example usage:
         >> from toolbox import csv_splitter;
         >> csv_splitter.split('/home/ben/input.csv');
--- a/usaspending_api/download/helpers.py
+++ b/usaspending_api/download/helpers.py
@@ -123,7 +123,7 @@ def multipart_upload(bucketname, regionname, source_path, keyname, headers={}, g
         bytes = min([bytes_per_chunk, remaining_bytes])
         part_num = i + 1
         pool.apply_async(_upload_part, [bucketname, regionname, mp.id,
-                         part_num, source_path, offset, bytes])
+                                        part_num, source_path, offset, bytes])
     pool.close()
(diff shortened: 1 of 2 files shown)


usaspending_api/download/helpers.py
```diff
@@ -198,17 +198,29 @@ def split_csv(file_path, delimiter=',', row_limit=10000, output_name_template='o
...
     current_piece = 1
-    current_out_path = os.path.join(
-        output_path,
-        output_name_template % current_piece
-    )
+    if file_name is not None:
+        zip_idx = file_name.find(".zip")
+
+        if(zip_idx > 0):
+            file_name = "%s.csv" % file_name[:zip_idx]
```
dpb-bah: Can you generate the `file_name` outside of `split_csv`? It seems a little counter-intuitive for a `split_csv` function to take in an argument that's expected to be a `.zip`. Also would recommend using `os` instead of string's `find` for getting the filename sans extension like so (it's your call, just my preference):

```
import os
...
file_name = os.path.splitext("path_to_file")[0]
```

usaspending_api/download/helpers.py
```diff
@@ -198,17 +198,29 @@ def split_csv(file_path, delimiter=',', row_limit=10000, output_name_template='o
...
+
+        current_out_path = os.path.join(
```
dpb-bah: Do these still have numbers on the end of them? That's kinda a crucial feature and the premise of the bulk downloads. Basically for csvs with more than a million rows, we want to split those out to other csvs with the order preserved. All to say, it should end up something like, 

`download_results.zip`
 -> `download_results_1.csv`
 -> `download_results_2.csv`
etc.

justinle: Oh, I understand now. The new commit now utilizes the numbering convention argument.

usaspending_api/download/helpers.py
justinle: Simplified the filename preservation to be one level up, also changed the method to be less rigid.

PR #1597, opened 2018-10-22
DEV-1668, part 1: Converted various IDs to 64 bit integers to match database

kbard:
**Description:**

**Technical details:**
This particular pull request deals with resolving primary keys that are 64 bit integers in the database but Django thinks should be 32 bit.  The fields in question had not broken the 32 bit barrier yet, so there were no observable issues.  The resolution was to update Django to 64 bit.  Because of the way Python handles integers, there should be no additional changes necessary in the API.

This could potentially cause issues in the website as the largest integer JavaScript can handle is significantly smaller than a traditional 64 bit integer.

	- [ ] Backend
	- [ ] Frontend
7. [X] Appropriate Operations ticket(s) created  (N/A)

Unit & integration tests passed but not changes were necessary.
This should change nothing for operations.
Should not affect performance in any way since we're just updating Django to mirror what's already in the database.
```
```


tony-sappe:
For Javascript:
`> Number.MAX_SAFE_INTEGER`
**9007199254740991**

This should be large enough for the foreseeable future on this project. Plus, there has been an effort to not use the BigInt PK keys but deterministic and permanent UID strings for the API. The risk to the frontend JS is minimal

kbard:
Re-opening.

PR #1598, opened 2018-10-22
DEV-1668: Convert data_source mixin field to TextField to match desired data type

kbard:
**Description:**
Issue DEV-1668 deals with resolving schema discrepancies between the actual production database and what Django thinks should be in the database.  Because there are hundreds of differences, we will be performing several pull requests to keep changes more manageable.

**Technical details:**
This particular pull request deals with updating the `DataSourceTrackedModel` Django model mixin `data_source` field to TextField to match the desired behavior of the database.  Django TextField and CharField translate to TEXT and VARCHAR in PostgreSQL.  TEXT and VARCHAR in PostgreSQL are practically identical with the only difference being the field size limitation imposed by VARCHAR.  This means that overall impact should be minimal.

**Requirements for PR merge:**
1. [X] Unit & integration tests updated  (N/A)
2. [X] API documentation updated  (N/A)
3. [x] Necessary PR reviewers:
	- [x] Backend
4. [X] Matview impact assessment completed  (N/A)
5. [X] Frontend impact assessment completed  (N/A)
6. [X] Data validation completed  (N/A)
7. [X] Appropriate Operations ticket(s) created  [OPS-396](https://federal-spending-transparency.atlassian.net/browse/OPS-396)
8. [X] Jira Ticket [DEV-1668](https://federal-spending-transparency.atlassian.net/browse/DEV-1668):
	- [X] Link to this Pull-Request
	- [X] Performance evaluation of affected (API | Script | Download)  (N/A)
	- [X] Before / After data comparison  (N/A)

**Area for explaining above N/A when needed:**
```
Unit & integration tests passed.  No changes were necessary.
Saw no documentation that needed updating.
Matviews are created directly from the database so the database remains unchanged for this PR.
Python treats TextField and CharField the same way so there should be no impact on the API and, subsequently, the front end app.
Data has not changed.  Validation is not needed.
Should not affect performance in any way since we're just updating Django to mirror what's already in the database and updating the database to use a 100% compatible data type (from VARCHAR to TEXT).
Before / After data comparison - No data was changed as a result of this update.
```

PR #1601, opened 2018-10-24
DEV-1668: Bring legal_entity into sync with database

kbard:
Depends on https://github.com/fedspendingtransparency/usaspending-api/pull/1598 being merged first.

**Description:**

**Technical details:**
This particular pull request deals with bringing NULLABLE/NON-NULLABLE legal_entity fields into alignment.  Most of the fields in the Django model were updated to be NULLABLE since the actual database columns contained NULLs.

	- [ ] Backend
5. [X] Frontend impact assessment completed  (N/A)

```
Unit & integration tests passed.  No changes were necessary.
The Django code already handles the case where these fields could be NULL so this shouldn't impact the front end at all.
No matching operations ticket required.
Should not affect performance in any way since we're just updating Django to mirror what's already in the database and updating the database to use a 100% compatible data type (from VARCHAR to TEXT).
```


usaspending_api/references/models.py
```diff
@@ -462,122 +462,117 @@ class LegalEntity(DataSourceTrackedModel):
 
     individuals
     '''
-    business_categories = ArrayField(models.TextField(), default=list)
+    business_categories = ArrayField(models.TextField(), default=list, null=True)
```
dpb-bah: Why is this allowed to be null? I think we ran into issues in the past where business_categories was `null` instead of an empty list.

kbard: In the actual database, roughly 2.4% of legal_entity records are NULL and 6.7% are '{}'.  This was just changed to match the actual data (obviously).  We can change the database to not allow nulls, but we'll have to track down how this column is populated and ensure no more NULLs are inserted.  Who decides if this is a new/separate issue or part of DEV-1668?

kbard: Added [DEV-1759](https://federal-spending-transparency.atlassian.net/browse/DEV-1759) to deal with `business_entities` issue separately.  @dpb-bah ok to sign off on this pull request as is?

PR #1624, opened 2018-11-19
DEV-1668: Remaining Django to DB differences

kbard:
**IMPORTANT:  The migrations in this pull request should be faked (--fake) when applied to an existing database.  There are [complementary SQL scripts](https://federal-spending-transparency.atlassian.net/browse/OPS-402) that will bring the database into sync.  For fresh installs, do not fake and do not run SQL scripts.**

**Description:**

**Technical details:**
This particular pull request deals with bringing remaining differences (from other pull requests) into sync with the database.

**NOTE:** In order to simplify pull requests for DEV-1668, we will not generate migration files until the final pull request.  It was just getting too messy continually fixing migrations every merge.

	- [x] Backend
5. [X] Frontend impact assessment completed  (N/A)

```
Unit & integration tests passed.  No changes were necessary.
The Django code already handles the case where these fields could be NULL so this shouldn't impact the front end at all.
No matching operations ticket required.
Should not affect performance in any way since we're just updating Django to mirror what's already in the database.
```


kbard:
Found an issue.  Will remove `do not merge` flag once fixed.

kbard:
Issue resolved.  Merge away.

usaspending_api/awards/models.py
```diff
@@ -211,13 +211,13 @@ class Award(DataSourceTrackedModel):
     certified_date = models.DateField(blank=True, null=True, help_text="The date this record was certified")
     create_date = models.DateTimeField(auto_now_add=True, blank=True, null=True,
                                        help_text="The date this record was created in the API")
-    update_date = models.DateTimeField(auto_now=True, null=True,
+    update_date = models.DateTimeField(auto_now=True, null=True, db_index=True,
```
tony-sappe: I was about to leave a comment that this index might not be necessary. But it has been used 227 times since the postgres stats were last reset.

However, in the DB it is sorted DESC. We might need to use the Index object in the meta class method to add the existing ordering

e.g. `Index(fields=['-update_date'])`

usaspending_api/awards/models.py
```diff
@@ -824,10 +824,12 @@ class TransactionFABS(models.Model):
     ultimate_parent_legal_enti = models.TextField(blank=True, null=True)
     ultimate_parent_unique_ide = models.TextField(blank=True, null=True)
     uri = models.TextField(blank=True, null=True, db_index=True)
+    legal_entity_foreign_descr = models.TextField(blank=True, null=True)
```
tony-sappe: Please move between the fields they are between in the DB (`legal_entity_city_code` and `place_of_perfor_state_code`)

usaspending_api/awards/serializers_v2/serializers.py
```diff
@@ -235,6 +235,7 @@ class Meta:
 
 
 class TransactionFPDSSerializerV2(LimitableSerializerV2):
+    small_business_competitive = serializers.CharField()
```
tony-sappe: What will this do?

usaspending_api/recipient/models.py
```diff
@@ -62,12 +62,12 @@ class HistoricParentDUNS(models.Model):
     Model representing DUNS data (imported from the broker)
     """
 
-    awardee_or_recipient_uniqu = models.TextField(primary_key=True)
-    legal_business_name = models.TextField()
-    ultimate_parent_unique_ide = models.TextField()
-    ultimate_parent_legal_enti = models.TextField()
+    awardee_or_recipient_uniqu = models.TextField(primary_key=True, db_index=True)
```
tony-sappe: Neither this index nor the index on `year` has been used. Might be better to drop from the DB

usaspending_api/recipient/models.py
```diff
@@ -81,7 +81,7 @@ class RecipientProfile(models.Model):
     recipient_unique_id = models.TextField(null=True, db_index=True)
     recipient_name = models.TextField(null=True, db_index=True)
     recipient_affiliations = ArrayField(base_field=models.TextField(), default=list, size=None)
-    award_types = ArrayField(base_field=models.TextField(), default=list, size=None)
+    award_types = ArrayField(base_field=models.TextField(), default=list, size=None, db_index=True)
```
tony-sappe: Unsure why, but the current index is a GIN index. Most likely need to add index definition in the Meta class

usaspending_api/recipient/models.py
```diff
@@ -100,8 +100,8 @@ class Meta:
 class RecipientLookup(models.Model):
     recipient_hash = models.UUIDField(unique=True, null=True)
     legal_business_name = models.TextField(null=True, db_index=True)
-    duns = models.TextField(null=True)
-    parent_duns = models.TextField(null=True)
+    duns = models.TextField(unique=True, null=True, db_index=True)
```
tony-sappe: This index can only be a partial unique index, there are thousands of NULLs. See the Meta class for the existing indexes for both of these fields

usaspending_api/references/models.py
```diff
@@ -367,7 +367,7 @@ def load_city_county_data(self):
 
 class LegalEntity(DataSourceTrackedModel):
     legal_entity_id = models.BigAutoField(primary_key=True, db_index=True)
-    recipient_name = models.TextField(blank=True, verbose_name="Recipient Name", null=True)
+    recipient_name = models.TextField(blank=True, verbose_name="Recipient Name", null=True, db_index=True)
```
tony-sappe: Current index only used 526 times compared to the 13160 times for the next-most used (and 1724359751 using the PK index)

It might be better to drop the DB index instead of defining the index here.

usaspending_api/awards/models_matviews.py
```diff
@@ -655,7 +655,7 @@ class Meta:
 
 
 class AwardCategory(models.Model):
-    type_code = models.TextField()
+    type_code = models.TextField(db_index=True)
```
tony-sappe: Index unused in the DB. There are 14 rows in this table. I don't even know where this would be used by the API, I only see old reload scripts using it.

usaspending_api/awards/models.py
tony-sappe: Your call if you want to make it ordered DESC or simply drop since it is hardly used

usaspending_api/awards/models.py
kbard: I'm a bit confused by this one.  Pretty much none of the columns line up between the DBs I'm looking at and the Django model.  Should I rearrange all of the columns so they match?

usaspending_api/awards/serializers_v2/serializers.py
kbard: So converting `small_business_competitive` from Text to Boolean altered the API contract a bit.  Instead of returning "True"/"False", the API was returning True/False which was breaking a unit test.  This line of code causes the unit test to pass.  A couple alternatives would be to change the field back to Text and update the database or update the unit test to handle the change in the API return type.This one really got me wondering, though, how many other contracts were altered by these Django changes but were not caught by unit tests...

usaspending_api/awards/models.py
tony-sappe: I wouldn't rearrange all of them. But when you are adding new columns you might as well order them to match the DB.

My guess is I can predict the table ordering by looking at the script used to reload the data.

kbard: Well, I moved it to beneath `legal_entity_city_code` as it is in the database.  Just know that `legal_entity_city_code` in the database is nowhere near where it is in the Django model.

usaspending_api/awards/migrations/0047_auto_20181126_1528.py
```diff
@@ -0,0 +1,811 @@
...
+            field=models.NullBooleanField(),
+        ),
+        migrations.AlterField(
+            model_name='transactionfpds',
+            name='detached_award_proc_unique',
+            field=models.TextField(null=True, unique=True),
+        ),
+        # Drop "_like" indexes since we're converting from text to integer.
+        migrations.RunPython(
+            drop_like_pattern_ops_index('awards', 'transactionfpds', 'detached_award_procurement_id')
```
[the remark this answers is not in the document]
kbard: Yeah.  I only move forward.  There is no going back.  CHAAAARGE!!!!!!

No, but for serious, is an oversight.  Will correct.  Will probably also just drop back to SQL since we've fully embraced it at this point.

PR #1638, opened 2018-12-07
[DEV-1916] Part I: New Parent Award Model

tony-sappe:
**Description:**
A new model to contain IDV aggregations

**Technical details:**
In addition to storing aggregations of children, this model also contains a self-referencing key for faster relationship lookups

New Model: _parent_award_
Fields:
- `award_id` (awards PK)
- `parent_award_id` (self-referencing FK)
- `generated_unique_award_id`
- `direct_idv_count`
- `direct_contract_count`
- `direct_total_obligation`
- `direct_base_and_all_options_value`
- `direct_base_exercised_options_val`
- `rollup_contract_count`
- `rollup_total_obligation`
- `rollup_base_and_all_options_value`
- `rollup_base_exercised_options_val`

Updates to _awards_
Removed `parent_award_id`
Added `fpds_agency_id`, `fpds_parent_agency_id`, `base_exercised_options_val`

	- [ ] Backend
8. [x] Jira Ticket [DEV-1916](https://federal-spending-transparency.atlassian.net/browse/DEV-1916):
	- [x] Performance evaluation of affected script (N/A)

```
Only ORM edits in this PR
```


usaspending_api/awards/models.py
```diff
@@ -158,8 +158,8 @@ class Award(DataSourceTrackedModel):
                                       "blanket purchase agreement. It is used to track the contract, and any "
                                       "modifications or transactions related to it. After October 2017, it is "
                                       "between 13 and 17 digits, both letters and numbers.")
-    parent_award = models.ForeignKey('awards.Award', related_name='child_award', null=True,
-                                     help_text="The parent award, if applicable")
+    fpds_agency_id = models.TextField(db_index=True, blank=True, null=True)
```
kbard: Agency is a low cardinality field.  Think `fpds_agency_id` and `fpds_parent_agency_id` are worth indexing?

usaspending_api/awards/models.py
```diff
@@ -960,3 +960,16 @@ class Subaward(DataSourceTrackedModel):
     class Meta:
         managed = True
         db_table = 'subaward'
+
+
+class ParentAward(models.Model):
+    award = models.OneToOneField(Award, primary_key=True)
```
kbard: Because we're keeping dupes "as is" for now, we can't use award_id as the primary key.  Some options are

1. no primary key - Django doesn't like this
2. surrogate primary key - just another dumb number, but more space efficient than # 3
3. primary key that is a list of [award_id, parent_award_id] - don't know if it's worth the hassle

Thoughts?

usaspending_api/awards/models.py
tony-sappe: Might not be, I'll drop the index from these fields and add a new index for `parent_award_piid`

usaspending_api/broker/management/sql/update_transaction_ids.sql
```diff
@@ -106,7 +106,6 @@ CREATE TABLE transaction_fpds_new_2 AS (
         transaction_fpds_new.awarding_sub_tier_agency_n AS awarding_sub_tier_agency_n,
         transaction_fpds_new.awarding_agency_code AS awarding_agency_code,
         transaction_fpds_new.awarding_agency_name AS awarding_agency_name,
-        transaction_fpds_new.parent_award_id AS parent_award_id,
```
dpb-bah: Curious why this is being removed, these changes shouldn't be involved with the transactionfpds/transactionfabs tables? Just the awards table?

tony-sappe: Good catch, this should remain.

PR #1672, opened 2019-02-05
DEV-1935/DEV-2049: Create new endpoint for IDV children

kbard:
**Description:**
The goal of this pull request is to create a new endpoint to support the Contracts/IDV Contracts tabs on the new IDV pages on USAspending.

**Technical details:**
New endpoint: `/api/v2/awards/idvs/awards/`.  See `awards.md` document included with this pull request for more details.

	- [x] Backend
	- [x] Frontend
8. [X] Jira Ticket [DEV-2049](https://federal-spending-transparency.atlassian.net/browse/DEV-2049):

```
The services of our outstanding Operations team is not required for this ticket/pull request.
```


usaspending_api/core/validator/tinyshield.py
```diff
@@ -225,11 +264,12 @@ def enforce_rules(self):
...
         if rule.get('allow_nulls', False) and rule['value'] is None:
-            return rule['value']
-        elif rule['type'] not in ('array', 'object'):
+            _return = rule['value']
+        elif rule['type'] not in ('array', 'object', 'any'):
             if rule['type'] in VALIDATORS:
-                return VALIDATORS[rule['type']]['func'](rule)
+                _return = VALIDATORS[rule['type']]['func'](rule)
             else:
                 raise Exception('Invalid Type {} in rule'.format(rule['type']))
```
tony-sappe: Perhaps you should pull the exception code into a function so the messages are uniform with the nicer output you created in lines 321-325

usaspending_api/core/validator/tinyshield.py
```diff
@@ -267,7 +307,23 @@ def apply_rule(self, rule):
...
                 object_result[k] = self.apply_rule(child_rule)
-            return object_result
+            _return = object_result
+        # Any is a "special" type since it is is really a collection of other rules.
+        elif rule['type'] == 'any':
+            for child_rule in rule['models']:
+                child_rule['value'] = rule['value']
+                try:
+                    # First successful rule wins.
+                    return self.apply_rule(child_rule)
```
tony-sappe: Great job removing the multiple return statements from this function. Any desire to set `_return` and break the loop here instead of returning?

usaspending_api/core/validator/tinyshield.py
```diff
@@ -302,3 +358,15 @@ def recurse_append(self, struct, mydict, data):
             else:
                 mydict[level] = {}
                 self.recurse_append(struct, mydict[level], data)
+
+    @staticmethod
+    def get_model_by_name(models, name):
```
tony-sappe: Handy function. I'm unsure I would include it in the TinyShield class since it is a utility

usaspending_api/awards/v2/views/idvs/awards.py
```diff
@@ -0,0 +1,151 @@
...
+        sql = sql.format(
+            award_id_column=Identifier(award_id_column),
+            award_id=Literal(award_id),
+            sort_column=Identifier(request_data['sort']),
+            sort_direction=SQL(request_data['order']),
+            limit=Literal(request_data['limit'] + 1),
+            offset=Literal((request_data['page'] - 1) * request_data['limit']),
+        )
+        with connection.cursor() as cursor:
+            cursor.execute(sql, [award_id])
```
tony-sappe: I encountered a 500 with this request (AttributeError: 'Composed' object has no attribute 'lower')

**URL**: `/api/v2/awards/idvs/awards/`
**BODY**: `{"award_id":"CONT_AW_9700_9700_71T0_SPM30008D3155"}`

Edit:
This was only when DEBUG=True in settings.py. Confusing why it occurs, should be investigated

usaspending_api/awards/award_id_helper.py
```diff
@@ -0,0 +1,81 @@
+from enum import Enum
+
+
+# Award id is currently defined as bigint in the database.  These are minimum
+# and maximum values for bigint.  While using sys.maxsize may look appealing,
+# it is based on the word size of the machine which will cause problems on
+# anything other than a 64 bit machine thus making it incorrect for this purpose.
```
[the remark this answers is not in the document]
kbard: Quantum computing is just around the corner...

usaspending_api/awards/award_id_helper.py
```diff
@@ -0,0 +1,81 @@
...
+MIN_INT = -2**63     # -9223372036854775808
```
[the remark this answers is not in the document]
kbard: Up for debate.  While the sequence starts at 1, it's still a BIGINT field.  On a project many moons ago, a client had a manual process for inserting data into one of their tables.  The manually entered data received negative IDs to indicate that they were manually entered.  ¯\_(ツ)_/¯

Anyhow, I'm not opposed to changing it if that's how we wanna go.

usaspending_api/awards/award_id_helper.py
```diff
@@ -0,0 +1,81 @@
...
+MAX_INT = 2**63 - 1  # +9223372036854775807
+
+
+class AwardIdType(Enum):
+    """
+    Multi-state return value for detect_award_id_type.
+    """
+    internal = 1   # award_id is probably an internal (integer) award id
+    generated = 2  # award_id is probably a generated_unique_award_id
+    unknown = 3    # type is unknown
```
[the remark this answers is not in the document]
kbard: Which is part of the reason I didn't go with that.  I suppose we could accept the `award_id` column name as an optional parameter.  Anyhow, I'm not terribly sold on the solution I've implemented.  In fact, I pretty much don't like it, but it seemed the lesser of many evils.

I am totally open to suggestions.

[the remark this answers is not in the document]
kbard: Also, in response to "Based on other new code it's just treated the same as generated", that's not entirely true.  In those cases, the input type is either constrained via a regex on the URL or by running it through TinyShield, so I was confident that the data type I was receiving was going to be either an integer or a string.  There should have been no "unknown".

But on the other hand, you're right in that I should have programmed that more defensively.  Who knows what future developers (including future me) might do with my assumptions.

usaspending_api/core/validator/tinyshield.py
```diff
@@ -225,11 +264,12 @@ def enforce_rules(self):
                 self.recurse_append(struct, self.data, self.apply_rule(item))
 
     def apply_rule(self, rule):
+        _return = None
         if rule.get('allow_nulls', False) and rule['value'] is None:
```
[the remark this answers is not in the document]
kbard: Erm.  I'm not understanding.  Which conditional?  And what appears backwards?

usaspending_api/core/validator/tinyshield.py
kbard: You thinking a standalone function in the TinyShield file or would you want that in a separate file?

usaspending_api/awards/v2/views/idvs/awards.py
kbard: Man.  Good catch.  This is a problem with `django-debug-toolbar`.  It is trying to show off the SQL which it has incorrectly assumed is a string instead of a `psycopg2` SQL object.  I don't usually run with `django-debug-toolbar` enabled (mainly because it's slow - but now I have another reason!).

We can simply materialize the SQL before passing it into Django.  Bingo bango, problem solved!

usaspending_api/core/validator/tinyshield.py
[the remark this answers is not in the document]
tony-sappe: .... Nevermind. I was referencing `if rule.get('allow_nulls', False) and rule['value'] is None:`, however in the morning light I am reading it correctly

usaspending_api/core/validator/tinyshield.py
tony-sappe: We can place it in `utils.py` in the same directory

usaspending_api/core/validator/tinyshield.py
kbard: The exception occurs if it runs through the loop and doesn't find a match.  I'm not sure how to circumvent the return without introducing yet another variable tracking whether or not it found an answer.  Which is worse?  A spurious `return` or an unnecessary artifact?

tony-sappe: Leave as-is. a "unnecessary artifact" is worse

usaspending_api/core/validator/tests/unit/test_award.py
```diff
@@ -0,0 +1,175 @@
...
+
+    models = [get_generated_award_id_rule(key='my_award_id', name='your_award_id', optional=True)]
+    r = TinyShield(models).block({'my_award_id': 'abcd'})
+    assert r == {'my_award_id': 'abcd'}
+
+    models = [get_generated_award_id_rule(key='my_award_id', name='your_award_id', optional=True)]
+    r = TinyShield(models).block({})
+    assert r == {}
+
+    # Bad rules.
```
tony-sappe: I think it's "bad requests" not rules

usaspending_api/core/validator/tinyshield.py
```diff
@@ -165,37 +185,56 @@ def block(self, request):
...
+
+        if model['type'] == 'any':
+            if in_any is True:
+                raise Exception('Nested "any" rules are not supported.')
+            # "any" child models are always optional and inherit their parent's name and key.
+            for sub_model in model['models']:
+                sub_model['name'] = model['name']
+                sub_model['key'] = model['key']
+                sub_model['optional'] = True
+                TinyShield.check_model(sub_model, True)
```
tony-sappe: What is the benefit of creating staticmethods and calling a new class instead of using a class method and `self`?

usaspending_api/core/validator/tinyshield.py
tony-sappe: ignore

usaspending_api/awards/v2/views/idvs/awards.py
```diff
@@ -0,0 +1,141 @@
...
+        # By this point, our award_id has been validated and cleaned up by
+        # TinyShield.  We will either have an internal award id that is an
+        # integer or a generated award id that is a string.
+        award_id = request_data['award_id']
+        award_id_column = 'award_id' if type(award_id) is int else 'generated_unique_award_id'
+
+        sql = GET_IDVS_SQL if request_data['idv'] else GET_CONTRACTS_SQL
+        sql = sql.format(
+            award_id_column=Identifier(award_id_column),
+            award_id=Literal(award_id),
```
tony-sappe: Dumb question: since you are using Literal here, do you even need to pass `award_id` as a parameter on line 127?

If we want to keep the parameter, then is `award_id` here redundant?

usaspending_api/core/validator/tinyshield.py
kbard: No benefit.  Probably just an artifact of refactoring.  Good catch.

usaspending_api/awards/v2/views/idvs/awards.py
kbard: Man.  You're determined not to let anything through, are you!

This is another good catch.  Also probably an artifact from refactoring.  I vaguely remember experimenting to see if it mattered which way I passed in the award id.

PR #1684, opened 2019-02-19
Fixing "Recipient Hash" value returned for /v2/awards/

tony-sappe:
**Description:**
Recipient Profile pages are contingent on the "level" of the recipient (Parent, Child, or Recipient). Since the API response value was not including the level in the string, the links were broker.

**Technical details:**
A little more complicated than desired since there isn't an easy way to lookup the recipient level from a transaction

	- [x] Backend
8. [x] Jira Ticket [DEV-2034](https://federal-spending-transparency.atlassian.net/browse/DEV-2034):

```
Minor edit to API response values for a specific field
```

### DIFF
--- a/usaspending_api/common/recipient_lookups.py
+++ b/usaspending_api/common/recipient_lookups.py
@@ -4,8 +4,7 @@ from usaspending_api.recipient.models import RecipientLookup
 def obtain_recipient_uri(recipient_name, recipient_unique_id, parent_recipient_unique_id=None):
     if not recipient_unique_id:
         return create_recipient_uri_without_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id)
-
-    return fetch_recipient_hash_using_name_and_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id)
+    return fetch_recipient_uri_with_duns(recipient_unique_id, parent_recipient_unique_id)
 
 
 def create_recipient_uri_without_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id=None):
@@ -14,13 +13,13 @@ def create_recipient_uri_without_duns(recipient_name, recipient_unique_id, paren
     return combine_recipient_hash_and_level(recipient_hash, recipient_level)
 
 
-def fetch_recipient_hash_using_name_and_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id):
+def fetch_recipient_uri_with_duns(recipient_unique_id, parent_recipient_unique_id):
     recipient_hash = fetch_recipient_hash_using_duns(recipient_unique_id)
     recipient_level = obtain_recipient_level({"duns": recipient_unique_id, "parent_duns": parent_recipient_unique_id})
     return combine_recipient_hash_and_level(recipient_hash, recipient_level)
 
 
-def generate_missing_recipient_hash(recipient_unique_id, recipient_name):
+def generate_missing_recipient_hash(recipient_name, recipient_unique_id):
     # SQL: MD5(UPPER(CONCAT(awardee_or_recipient_uniqu, legal_business_name)))::uuid
     import hashlib
     import uuid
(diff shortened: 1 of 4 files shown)


usaspending_api/common/recipient_lookups.py
```diff
@@ -0,0 +1,59 @@
...
+
+    return recipient["recipient_hash"] if recipient else None
+
+
+def obtain_recipient_level(recipient_record: dict) -> str:
+    if recipient_is_standalone(recipient_record):
+        level = "R"
+    elif recipient_is_child(recipient_record):
+        level = "C"
+    return level
```
kbard: So `recipient_is_child` is the boolean opposite of `recipient_is_standalone`.  Why not just have an `else`?  It also makes my linter scream because `level` is not initialized.

usaspending_api/common/recipient_lookups.py
```diff
@@ -0,0 +1,59 @@
...
+    return fetch_recipient_hash_using_name_and_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id)
+
+
+def create_recipient_uri_without_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id=None):
+    recipient_hash = generate_missing_recipient_hash(recipient_name, recipient_unique_id)
+    recipient_level = obtain_recipient_level({"duns": recipient_unique_id, "parent_duns": parent_recipient_unique_id})
+    return combine_recipient_hash_and_level(recipient_hash, recipient_level)
+
+
+def fetch_recipient_hash_using_name_and_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id):
```
kbard: `recipient_name` isn't used in this function which also makes the function name a bit misleading.

usaspending_api/common/recipient_lookups.py
```diff
@@ -0,0 +1,59 @@
...
+
+def obtain_recipient_uri(recipient_name, recipient_unique_id, parent_recipient_unique_id=None):
+    if not recipient_unique_id:
+        return create_recipient_uri_without_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id)
+
+    return fetch_recipient_hash_using_name_and_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id)
+
+
+def create_recipient_uri_without_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id=None):
+    recipient_hash = generate_missing_recipient_hash(recipient_name, recipient_unique_id)
```
kbard: `generate_missing_recipient_hash` parameters appear to be out of order from the actual function signature.

usaspending_api/common/recipient_lookups.py
kbard: Why `fetch_recipient_HASH_using_name_and_duns` and its sister function called `create_recipient_URI_without_duns`?  Should they both be URI?

usaspending_api/common/recipient_lookups.py
tony-sappe: Good find

usaspending_api/common/recipient_lookups.py
tony-sappe: fixed

usaspending_api/common/recipient_lookups.py
tony-sappe: > FUNCTIONS SHOULD DO ONE THING. THEY SHOULD DO IT WELL. THEY SHOULD DO IT ONLY.

https://www.goodreads.com/book/show/3735293-clean-code, chapter 3

I'm slowly learning. Open to discussions

kbard: Not sure I understand the reference, but the code should work.  It just seems weird to say:
```
if my_thing is one of two values:
    do this
else if my_thing is the other of two values:
    do something else
```

rather than

```
if my_thing is one of two values:
    do this
else:
    do something else
```

This also absolves you of having to initialize `level`.

¯\\\_(ツ)\_/¯

PR #1685, opened 2019-02-19
Stg -> dev downmerge

tony-sappe:
Downmerge quick fix for Recipient Hash values returned by /v2/awards/


tony-sappe:
Closing since #1691 resolves the merge conflict

PR #1688, opened 2019-02-21
Fix for inappropriate TinyShield usage

kbard:
**Description:**
Resolves bug introduced by TinyShield usage misunderstanding.

**Technical details:**
In several locations in code, TinyShield objects were being reused.  In its current incarnation, this does not work.  A new TinyShield must be instantiated per use.

	- [x] Backend

### DIFF
--- a/usaspending_api/awards/v2/views/idvs/awards.py
+++ b/usaspending_api/awards/v2/views/idvs/awards.py
@@ -79,20 +79,16 @@ GET_CONTRACTS_SQL = SQL("""
 """)
 
 
-def _prepare_tiny_shield_rules():
-    """
-    Our TinyShield rules never change.  Encapsulate them here and store them
-    once in TINY_SHIELD_RULES.
-    """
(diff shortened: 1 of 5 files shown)


usaspending_api/awards/v2/views/idvs/awards.py
```diff
@@ -79,20 +79,16 @@
...
 
 
-def _prepare_tiny_shield_rules():
-    """
-    Our TinyShield rules never change.  Encapsulate them here and store them
-    once in TINY_SHIELD_RULES.
-    """
-    models = customize_pagination_with_sort_columns(SORTABLE_COLUMNS, DEFAULT_SORT_COLUMN)
-    models.extend([
+def _prepare_tiny_shield_model():
```
tony-sappe: It looks like there are multiple models being defined by this function

IsaacRay:
@kbard I found that reusing models at all can be problematic. In search/v2/views/search.py, for instance, I ended up needing to deepcopy the AWARD_FILTER model because otherwise I was getting inconsistent results from the API endpoints from one call to another. It looks like you are doing something similar, however I'm wondering if you still may have some subtle bugginess in this approach.

PR #1704, opened 2019-03-08
New Multiprocess design for download service

tony-sappe:
**Description:**
New functionality to monitor the queue, kick off a new process to handle the processing, and allow better exception handling when issues occur.

**Technical details:**
- Use multiprocessing to start a process for each download job
- Allows a shorter visibility timeout since the monitor can send "heartbeats" during processing
- Created new DownloadJobStatus values for future improvements
- More logging
- Two new management commands for processing "frozen" downloads `restart_download` and `resolve_frozen_downloads`

    - [x] Backend
    - [x] Operations
8. [x] Jira Ticket (N/A)

```
No API behavior was modified to warrant doc or frontend edits
Sadly no existing tests to update. Currently no good testing framework
Original work started under bug tickets, spun into it's own entity
```


usaspending_api/download/filestreaming/csv_generation.py
```diff
@@ -190,15 +196,21 @@ def split_and_zip_csvs(zipfile_path, source_path, source_name, download_job=None
...
-
+        write_to_log(message='Beginning the CSV file partition', download_job=download_job)
         list_of_csv_files = partition_large_csv_file(source_path, row_limit=EXCEL_ROW_LIMIT,
                                                      output_name_template=output_template)
 
         if download_job:
-            write_to_log(message='Splitting csvs took {:.4f} seconds'.format(time.perf_counter() - log_time),
-                         download_job=download_job)
+            write_to_log(
+                message='Partitioning CSV file into {} files took {:.4f} seconds'.format(
```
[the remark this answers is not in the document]
tony-sappe: Ah, yes. I've used that before. The `:.4f` is for significant digits after the decimal.

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
+from usaspending_api.download.download_exceptions import FatalError
 from usaspending_api.download.filestreaming import csv_generation
 from usaspending_api.download.helpers import write_to_download_log as write_to_log
 from usaspending_api.download.lookups import JOB_STATUS_DICT
 from usaspending_api.download.models import DownloadJob
 
-DEFAULT_VISIBILITY_TIMEOUT = 60 * 30
-
-logger = logging.getLogger("console")
+CURRENT_MESSAGE = None
```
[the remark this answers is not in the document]
tony-sappe: You're correct. I was thinking that was for all global vars.....

usaspending_api/download/filestreaming/csv_generation.py
```diff
@@ -130,7 +138,7 @@ def get_csv_sources(json_request):
     return csv_sources
 
 
-def parse_source(source, columns, download_job, working_dir, start_time, message, zipfile_path, limit):
+def parse_source(source, columns, download_job, working_dir, start_time, zipfile_path, limit):
```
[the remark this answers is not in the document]
tony-sappe: Yep, I would love to refactor all of this old code

usaspending_api/download/management/commands/resolve_frozen_downloads.py
```diff
@@ -0,0 +1,102 @@
...
+            filter_set &= Q(update_date__gte=kwargs["start"].replace(tzinfo=pytz.UTC))
+        if kwargs["end"]:
+            filter_set &= Q(update_date__lt=kwargs["end"].replace(tzinfo=pytz.UTC))
+
+        return filter_set
+
+    def search_for_downloads(self, filters):
+        return DownloadJob.objects.filter(filters).order_by("update_date").values_list("download_job_id", flat=True)
+
+    def filter_downloads(self, **kwargs):
```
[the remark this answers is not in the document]
tony-sappe: Must have been copy-paste

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
+        message.change_visibility(VisibilityTimeout=new_visibility)
+    except botocore.exceptions.ClientError as e:
+        write_to_log(message="Unable to set VisibilityTimeout. Message might have previously been released or removed")
+
+
+def release_sqs_message(message):
+    set_sqs_message_visibility(message, 0)
+
+
+def retrive_download_job_from_db(download_job_id):
```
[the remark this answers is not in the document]
tony-sappe: facepalm

usaspending_api/download/filestreaming/csv_generation.py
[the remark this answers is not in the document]
kbard: Sorry.  Was referring to the earlier `{}` that was listing the file count.

usaspending_api/download/filestreaming/csv_generation.py
```diff
@@ -52,16 +53,21 @@ def generate_csvs(download_job, sqs_message=None):
...
+        exc_msg = "InvalidParameterException was raised while attempting to process the DownloadJob"
+        download_job.error_message = '{}:\n{}'.format(exc_msg, stack_trace)
+        download_job.save()
+        raise InvalidParameterException(e)
     except Exception as e:
         # Set error message; job_status_id will be set in download_sqs_worker.handle()
-        download_job.error_message = 'An exception was raised while attempting to write the file:\n{}'.format(str(e))
+        stack_trace = "".join(traceback.format_exception(etype=type(e), value=e, tb=e.__traceback__))
+        exc_msg = "An exception was raised while attempting to process the DownloadJob"
+        download_job.error_message = '{}:\n{}'.format(exc_msg, stack_trace)
```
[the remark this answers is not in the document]
tony-sappe: They were a last-minute addition today. I still have goals to refactor the download code (everything in `/filestreaming/`)

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
+    try:
+        csv_generation.generate_csvs(download_job=download_job)
+    except Exception as e:
+        write_to_log(message="Caught exception", download_job=download_job, is_error=True)
+        return 11  # arbitrary positive integer
+
+    return 0
+
+
+def remove_message_from_sqs(message):
```
kwhickey: doc string

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
+    if message is None:
+        return
+    try:
+        message.delete()
+    except Exception as e:
+        write_to_log(message="Unable to delete SQS message. Message might have previously been released or removed")
+
+
+def set_sqs_message_visibility(message, new_visibility):
```
kwhickey: doc string

(e.g. clarify that this will essentially reset the timeout to the given time, starting at the time it is changed)

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
+    if message is None:
+        write_to_log(message="No SQS message to modify. Message might have previously been released or removed")
+        return
+    try:
+        message.change_visibility(VisibilityTimeout=new_visibility)
+    except botocore.exceptions.ClientError as e:
+        write_to_log(message="Unable to set VisibilityTimeout. Message might have previously been released or removed")
+
+
+def release_sqs_message(message):
```
kwhickey: doc string
(i.e. clarify that this will make the message visible in the queue again, immediate, for any other consumer to consume)

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
+
+
+def retrive_download_job_from_db(download_job_id):
+    download_job = DownloadJob.objects.filter(download_job_id=download_job_id).first()
+    if download_job is None:
+        raise FatalError("Download Job {} record missing in DB!".format(download_job_id))
+    return download_job
+
+
+def update_download_record(download_job_id, status, message=None):
```
kwhickey: doc string

If the `message` attribute is only used as an error message, rename it to `error_message` to make it more clear to callers of the function

usaspending_api/download/management/commands/resolve_frozen_downloads.py
```diff
@@ -0,0 +1,102 @@
...
+            default=False,
+            help="If true: include and restart monthly download jobs",
+        )
+
+    def handle(self, *args, **options):  # used by parent class
+        logger.info("Beginning management command")
+
+        filters = self.parse_arguments_to_queryset_filter(**options)
+        list_of_downloads = self.search_for_downloads(filters)
+        logger.info("Found {} Download jobs to restart".format(len(list_of_downloads)))
```
kwhickey: In other code, `len(some_list_of_orm_objects)` causes all of them to be lazily loaded from the db just to get a count. If all you're looking for is the count to log, use `.count()` on the filter.

usaspending_api/download/management/commands/resolve_frozen_downloads.py
```diff
@@ -0,0 +1,102 @@
...
+
+            download = DownloadAdministrator()
+            download.search_for_a_download(download_job_id=download_job_id)
+            download.restart_download_operation()
+
+    def parse_arguments_to_queryset_filter(self, **kwargs):
+        filter_set = Q(job_status=JOB_STATUS_DICT[kwargs["status"]])
+        filter_set &= Q(monthly_download=kwargs["monthly"])
+        if kwargs["start"]:
+            filter_set &= Q(update_date__gte=kwargs["start"].replace(tzinfo=pytz.UTC))
```
kwhickey: Wouldn't the created date be more appropriate here? Date since the download was created. It could have recently had its status updated causing the update date to be more recent than the created date.

usaspending_api/download/management/commands/resolve_frozen_downloads.py
```diff
@@ -0,0 +1,102 @@
...
+            help="Earliest datetime to begin restarting downloads",
+        )
+        parser.add_argument(
+            "-e",
+            "--end",
+            "--end-date",
+            "--end-datetime",
+            dest="end",
+            type=dateutil.parser.parse,
+            help="Most recent datetime to restart downloads",
```
kwhickey: Might be useful to say that the current date (now) will be used if not set (i.e. no upper limit)

usaspending_api/download/management/commands/restart_download.py
```diff
@@ -0,0 +1,63 @@
+import logging
```
kwhickey: Is it possible just to combine these two commands into one since they both restart "frozen" jobs. For instance just have another `download_job_id` param that, when passed in, targets only that job. 

From a "operator" perspective, this would be easier to get the same job done with one command

usaspending_api/references/management/commands/load_download_static_data.py
```diff
@@ -14,6 +14,7 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
 
         logger = logging.getLogger('console')
+        JobStatus.objects.all().delete()
```
kwhickey: What does this do? Seems a bit of a rogue line of code to be here without a comment explaining it (or I just don't the context of this management command).

usaspending_api/settings.py
```diff
@@ -164,24 +165,27 @@
...
-    DATABASES['db_source'] = dj_database_url.parse(os.environ.get('DB_SOURCE'), conn_max_age=10)
-    DATABASES['db_r1'] = dj_database_url.parse(os.environ.get('DB_R1'), conn_max_age=10)
+    DATABASES['db_source'] = dj_database_url.parse(os.environ.get('DB_SOURCE'), conn_max_age=CONNECTION_MAX_SECONDS)
+    DATABASES['db_r1'] = dj_database_url.parse(os.environ.get('DB_R1'), conn_max_age=CONNECTION_MAX_SECONDS)
     DATABASE_ROUTERS = ['usaspending_api.routers.replicas.ReadReplicaRouter']
 
 # import a second database connection for ETL, connecting to the data broker
 # using the environemnt variable, DATA_BROKER_DATABASE_URL - only if it is set
 if os.environ.get('DATA_BROKER_DATABASE_URL') and not sys.argv[1:2] == ['test']:
-    DATABASES['data_broker'] = dj_database_url.parse(os.environ.get('DATA_BROKER_DATABASE_URL'), conn_max_age=600)
```
kwhickey: This will effectively change it from 600 seconds to 10 seconds. That seems like it will cause issues if someone deliberately put it to 600 for some reason.

usaspending_api/urls.py
```diff
@@ -58,4 +58,5 @@
     import debug_toolbar
     urlpatterns += [
         url(r'^__debug__/', include(debug_toolbar.urls)),
+        url(r'^plate/', include('django_spaghetti.urls')),
```
kwhickey: What does this do? Adds the ER diagram.

usaspending_api/download/v2/download_admin.py
```diff
@@ -0,0 +1,56 @@
+from django.conf import settings
+from django.db.models import Q
+
+from usaspending_api.common.sqs_helpers import get_sqs_queue_resource
+from usaspending_api.download.filestreaming import csv_generation
+from usaspending_api.download.lookups import JOB_STATUS_DICT
+from usaspending_api.download.models import DownloadJob
+
+
+class DownloadAdministrator:
```
kwhickey: This may be a good candidate to centralize some of our Data Access code/logic for a model (`DownloadJob` in this case) per the dev meeting topic here: https://github.com/fedspendingtransparency/data-act-documentation/blob/master/dev_meetings/WeeklyDevMeeting_2019-02-04.md#model-managers

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
 from usaspending_api.download.lookups import JOB_STATUS_DICT
 from usaspending_api.download.models import DownloadJob
 
-DEFAULT_VISIBILITY_TIMEOUT = 60 * 30
-
-logger = logging.getLogger("console")
+CURRENT_MESSAGE = None
+DEFAULT_VISIBILITY_TIMEOUT = 60
+LONG_POLL_SECONDS = 10
+MONITOR_SLEEP_TIME = 5  # must be significantly less than DEFAULT_VISIBILITY_TIMEOUT
```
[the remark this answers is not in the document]
tony-sappe: Not a bad idea. The function will be arbitrary but can be improved in the future if needed

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
-                    WaitTimeSeconds=10, MessageAttributeNames=["All"], VisibilityTimeout=DEFAULT_VISIBILITY_TIMEOUT
+        # route signal handling to custom function
+        for sig in [signal.SIGINT, signal.SIGQUIT, signal.SIGTERM]:
+            signal.signal(sig, signal_handler)
+
+        download_service_manager()
+
+
+def signal_handler(signum, frame):
+    write_to_log({"message": "Received signal ({}). Gracefully stopping Download Job".format(signum)})
```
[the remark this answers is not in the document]
tony-sappe: Done

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,160 @@
...
+    release_sqs_message(CURRENT_MESSAGE)
+    raise SystemExit
```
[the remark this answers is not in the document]
tony-sappe: We don't need to describe what `SystemExit` is.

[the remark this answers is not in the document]
tony-sappe: Yes, there are other places in the codebase we return with a positive integer

usaspending_api/references/management/commands/load_download_static_data.py
tony-sappe: This line allows the static data loading script to be "stateless" i.e. the script can be run over and over and the table will always be left in the same state.

usaspending_api/settings.py
tony-sappe: Very true. I cannot think of a good reason why the connection should remain for 600s after the last transaction completes

usaspending_api/urls.py
tony-sappe: Yeah. Unrelated to the downloads

usaspending_api/download/v2/download_admin.py
tony-sappe: This is the starting block to refactoring the current Download logic in `views.py` and `csv_generation.py`. I thought about using a manager, but I was running out of time and it would only be a subset of the current class functionality

usaspending_api/download/management/commands/resolve_frozen_downloads.py
tony-sappe: Count is just for logging. The next code line loops through the results

usaspending_api/download/management/commands/resolve_frozen_downloads.py
tony-sappe: That's technically not correct. If end isn't provided, the date range is "open". There are some notes in the command help documentation.

usaspending_api/download/management/commands/resolve_frozen_downloads.py
tony-sappe: Arguments can be made for both. When I'm looking in the download records I usually use `update_date` in the SQL instead of `create_date`.
It's just to provide the ability to restart downloads not being processed but should have been processed.

usaspending_api/download/filestreaming/csv_generation.py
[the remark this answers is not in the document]
tony-sappe: I don't think I've seen more than a few files. Each file is up to one million rows

usaspending_api/download/management/commands/restart_download.py
tony-sappe: It's possible. But making them separate kept them smaller and easier to understand

usaspending_api/download/management/commands/download_sqs_worker.py
tony-sappe: function name is accurate and descriptive. Code is straightforward

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,182 @@
...
+    8: "SIGFPE [8] (Floating point exception)",
+    9: "SIGKILL [9] (non-catchable, non-ignorable kill)",
+    11: "SIGSEGV [11] (Invalid memory reference)",
+    14: "SIGALRM [12] (Timer signal from alarm(2))",
+    15: "SIGTERM [15] (software termination signal)",
+}
+current_message = None
+DEFAULT_VISIBILITY_TIMEOUT = 60
+LONG_POLL_SECONDS = 10
+MONITOR_SLEEP_TIME = DEFAULT_VISIBILITY_TIMEOUT / 20  # This needs to be less than DEFAULT_VISIBILITY_TIMEOUT
```
kwhickey: This could yield a decimal number, possibly less than 1 second, yes?

e.g. if timeout got set to 15 seconds in this case. 

Thinking about it again, my concern was more that we don't end up in an erroneous situation, and less-so that the proportion is right. 

How about leaving the values hard-coded, but adding some defensive coding that throws a `FatalError` at the beginning of the job starting if `MONITOR_SLEEP_TIME >= DEFAULT_VISIBILITY_TIMEOUT`?

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,59 +1,185 @@
...
-        queue = get_sqs_queue_resource(queue_name=settings.BULK_DOWNLOAD_SQS_QUEUE_NAME)
-
-        write_to_log(message="Starting SQS polling")
-        while True:
-            second_attempt = True
-            try:
-                # Grabs one (or more) messages from the queue
-                messages = queue.receive_messages(
-                    WaitTimeSeconds=10, MessageAttributeNames=["All"], VisibilityTimeout=DEFAULT_VISIBILITY_TIMEOUT
+        if MONITOR_SLEEP_TIME >= DEFAULT_VISIBILITY_TIMEOUT:
```
kwhickey: Looks good.

PR #1733, opened 2019-04-03
DEV-2023: Fixed Bulk Download API bug for sub-awards

tony-sappe:
**Description:**
Added an explicit reference to the subaward materialized view to resolve a bug when trying to use the endpoint to generate downloads.

**Technical details:**
A line to add a SQL fragment caused ambiguity in the column reference.

    - [x] Backend
8. [x] Jira Ticket [DEV-2023](https://federal-spending-transparency.atlassian.net/browse/DEV-2023):
    - [x] Performance evaluation of affected Download

```
No API behavior changes except correcting an error causing downloads to fail.
No impact to operations or frontend
```

### DIFF
--- a/usaspending_api/awards/v2/filters/sub_award.py
+++ b/usaspending_api/awards/v2/filters/sub_award.py
@@ -91,7 +91,7 @@ def subaward_filter(filters, for_downloads=False):
             transaction_ids = [str(transaction_id) for transaction_id in transaction_ids]
             queryset = queryset.filter(latest_transaction_id__isnull=False)
             queryset &= queryset.extra(
-                where=['"latest_transaction_id" = ANY(\'{{{}}}\'::int[])'.format(','.join(transaction_ids))])
+                where=['"subaward_view"."latest_transaction_id" = ANY(\'{{{}}}\'::int[])'.format(','.join(transaction_ids))])
 
         elif key == "time_period":
             min_date = API_SEARCH_MIN_DATE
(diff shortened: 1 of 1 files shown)


usaspending_api/awards/v2/filters/sub_award.py
```diff
@@ -90,8 +90,9 @@ def keyword_parse(keyword):
             logger.info('Found {} transactions based on keyword: {}'.format(len(transaction_ids), keyword))
             transaction_ids = [str(transaction_id) for transaction_id in transaction_ids]
             queryset = queryset.filter(latest_transaction_id__isnull=False)
-            queryset &= queryset.extra(
-                where=['"latest_transaction_id" = ANY(\'{{{}}}\'::int[])'.format(','.join(transaction_ids))])
+
+            sql_fragment = '"subaward_view"."latest_transaction_id" = ANY(\'{{{}}}\'::int[])'
```
BrianZito: I'm not that experienced with Postgres, but `'{{{}}}'::int[]` seems like code that justifies an explanatory comment.

usaspending_api/awards/v2/filters/sub_award.py
```diff
@@ -90,8 +90,9 @@ def keyword_parse(keyword):
...
+            queryset &= queryset.extra(where=[sql_fragment.format(','.join(transaction_ids))])
```
BrianZito: I see no protection against SQL injection here, and Django's documentation on .extra clearly states it DOESN'T protect against that. https://docs.djangoproject.com/en/2.1/ref/models/querysets/

tony-sappe: Good eye on using .extra(). The list of `transaction_ids` are from the ES index, not user-provided. So risk of SQL injection is very low

usaspending_api/awards/v2/filters/sub_award.py
tony-sappe: The curly brackets are python string format syntax: https://docs.python.org/3.3/library/string.html#format-string-syntax

The inner `{}` is for the list of transaction ids, and the `{{` ..`}}` are for including a singe `{}` around the transaction ids in the final string.

I can add a comment

PR #1734, opened 2019-04-04
DEV-2192: Django-Filters prototype for v1 migration

IsaacRay:
**Description:**
This takes a very stripped down approach to creating django-filter based endpoints that restrict the use of complex ORM joins

**Technical details:**
django-filters integrates fairly well with DRF, however using viewsets requires a bit of a fancy override to get it to play nicely. You'll need to run pip install requirements to get the latest version of django-filters if you want to try this out. A sample request would be:

http://localhost:8000/api/v1/awards/faba/?treasury_account__agency_id=080

This request filters on a join to the treasury account table and uses the agency id from that table to filter FinancialAccountsByAwards records that have that treasury account object as a related object.

1. [NA] Unit & integration tests updated
    - [ ] Backend
8.  [X] Jira Ticket [DEV-2192](https://federal-spending-transparency.atlassian.net/browse/DEV-2192):
    - [NA] Performance evaluation of affected (API | Script | Download)
    - [NA] Before / After data comparison

```
This is a prototype that won't ultimately be merged, its just for showing possibilities for future migration approaches.
```


usaspending_api/awards/v1/views.py
```diff
@@ -182,3 +185,57 @@ def get_queryset(self):
...
         ordered_queryset = self.order_records(self.request, queryset=filtered_queryset)
         return ordered_queryset
+
+
+class FABAFilter(filters.FilterSet):
+
+    class Meta:
+        model = FinancialAccountsByAwards
+        # You can create any relationships you want here using ORM notation
+        fields = ('piid', 'treasury_account__agency_id')
```
kwhickey: Can you you showcase surfacing a joined field by an aliased name, rather than having it surface via the dunder syntax?

usaspending_api/awards/v1/views.py
```diff
@@ -182,3 +185,57 @@ def get_queryset(self):
...
+        I spent a lot of time investigating the serializers here trying to come up with a better solution. However
+        ultimately I determined that the LimitableSerializer class which we base all our serializers on is actually
+        fairly well put together. It handles prefetching well, so that the nested serializers don't prompt new queries,
+        and it doesn't really mess with the core functionality of the serializer as written in DRF. My main objection
+        to them is having to specify default_fields unless verbose is set. If I was ultimately going to rewrite this,
+        I might switch that around so that verbose is the default. I did not feel that that was worth investing time in
+        for this investigation however.
+    '''
+
+    serializer_class = FinancialAccountsByAwardsSerializer
```
kwhickey: Where is the class definition of this Serializer in this PR?

Seeing that is part of AC 2 and 4
> 2. A developer can successfully use serializers adapting a model object in DRF
> 4. Not all fields of the model are returned to the API user (they can be limited)

It'd be good to see how you manage only allowing explicit declared fields to be returned.

usaspending_api/awards/v1/views.py
kwhickey: Once this is converted to the more concise version using `filterset_fields` in the View class, can you see you can showcase some precision in being able to select _which kinds_ of filters ("lookups") are allowed using the dictionary syntax seen here: https://django-filter.readthedocs.io/en/latest/ref/filterset.html#declaring-filterable-fields?
```
fields = {
            'username': ['exact', 'contains'],
            'last_login': ['exact', 'year__gt'],
        }
```

usaspending_api/awards/v1/views.py
kwhickey: Once added, I was hoping to see the ability to filter on a related field, who was added to the serializer class and given an alias. See:
https://www.django-rest-framework.org/api-guide/serializers/#specifying-fields-explicitly
or 
https://www.django-rest-framework.org/api-guide/relations/#primarykeyrelatedfield
```
class AlbumSerializer(serializers.ModelSerializer):
    tracks = serializers.PrimaryKeyRelatedField(many=True, read_only=True)

    class Meta:
        model = Album
        fields = ('album_name', 'artist', 'tracks')
```

In your case, maybe use something like that to wrap around the `treasury_account__agency_id` relation and have it returned as a field in the response just named `agency`.

tony-sappe:
Closing since this was never intended to be merged and the ticket for the effort was marked as "Done"

PR #1747, opened 2019-04-16
Update fpds_nightly_loader.py

tony-sappe:
Minor edits to a datetime in `fpds_nightly_loader` on line 347.

Unsure if I should have left line 345 as-is or the change it an improvement.


kbard:
Potato potato... `datetime.utcnow()` and `datetime.now(timezone.utc)` are functionally equivalent once they are fed into `update_last_load_date`.  It was the whole casting things to strings that was messing everything up.

Thanks for cleaning up my mess.

kbard:
Not having unit tests for this code is a real risk.

PR #1773, opened 2019-04-25
DEV-2336: Added readme and renamed files

kbard:
**Description:**
Added a readme file to the download zip that describes the constituent files.  Also renamed zip and constituent files to contain the award id.

2. [x] No API documentation updates required
    - [ ] Backend
4. [x] No materialized views are affected by these changes
5. [x] Front end is unaffected by these changes
7. [x] No Operations ticket(s) required
8. [x] Jira Ticket [DEV-2336](https://federal-spending-transparency.atlassian.net/browse/DEV-2336):

### DIFF
--- a/usaspending_api/common/tests/test_text_helpers.py
+++ b/usaspending_api/common/tests/test_text_helpers.py
@@ -1,0 +1,51 @@
+from usaspending_api.common.helpers.text_helpers import slugify_text_for_file_names
+
+
+def test_slugify_text_for_file_names():
+    assert slugify_text_for_file_names("test") == "test"
+    assert slugify_text_for_file_names("test", "default") == "test"
+    assert slugify_text_for_file_names("test", "default", 50) == "test"
+    assert slugify_text_for_file_names("test", "default", 2) == "te"
+
+    assert slugify_text_for_file_names(None) is None
+    assert slugify_text_for_file_names(None, None) is None
+    assert slugify_text_for_file_names(None, "default") == "default"
+    assert slugify_text_for_file_names(None, "default", 50) == "default"
+    assert slugify_text_for_file_names(None, "default", 2) == "default"
+
+    assert slugify_text_for_file_names("---") is None
+    assert slugify_text_for_file_names("---", "default") == "default"
+    assert slugify_text_for_file_names("---", "default", 50) == "default"
+    assert slugify_text_for_file_names("---", "default", 2) == "default"
+
+    garbage = r")*(&*()THIS(#@$@^*IS<,>.?/:;\"'{[}]|\+=_-)(*&^%$#@!A><?>\":}TEST()(*&(*&*(*"
+    assert slugify_text_for_file_names(garbage) == "THIS_IS_A_TEST"
+    assert slugify_text_for_file_names(garbage, "default") == "THIS_IS_A_TEST"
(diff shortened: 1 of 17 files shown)


usaspending_api/common/tests/test_text_helpers.py
```diff
@@ -0,0 +1,51 @@
...
+    assert slugify_text_for_file_names("test") == "test"
+    assert slugify_text_for_file_names("test", "default") == "test"
+    assert slugify_text_for_file_names("test", "default", 50) == "test"
+    assert slugify_text_for_file_names("test", "default", 2) == "te"
+
+    assert slugify_text_for_file_names(None) is None
+    assert slugify_text_for_file_names(None, None) is None
+    assert slugify_text_for_file_names(None, "default") == "default"
+    assert slugify_text_for_file_names(None, "default", 50) == "default"
+    assert slugify_text_for_file_names(None, "default", 2) == "default"
```
tony-sappe: Seems like either option of:
- returning the full default string when it exceeds the "max_length" of characters, or
- a substring of the default string with length "max_length"

could be counter-intuitive. Any ideas?

usaspending_api/download/filestreaming/include_file_description.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def include_file_description_in_zip(file_description_text, destination_file_name, working_dir, zipfile_path):
+    write_to_log(message="Adding file description")
+
+    destination_file_path = os.path.join(working_dir, destination_file_name)
+    with open(destination_file_path, "w") as f:
+        f.write(file_description_text)
+
+    zip_file = zipfile.ZipFile(zipfile_path, 'a', compression=zipfile.ZIP_DEFLATED, allowZip64=True)
```
tony-sappe: I need to investigate (or be shown) how this works to produce the desired results.

usaspending_api/common/tests/test_text_helpers.py
kbard: Not sure why returning the default string if it exceeds max_length is counter-intuitive.  I have to trust that the caller knows what they're doing.  If they give me a default, I'm going to want to return the default.  If they need the default to follow the same rules, they're more than welcome to call this function on the default value as well.

    slugify_text_for_file_names(
        'my string',
        slugify_text_for_file_names('my default')
    )

usaspending_api/download/filestreaming/include_file_description.py
kbard: 1) Write the provided text to the provided destination file.
2) Append the newly written file to the provided zip file.

This logic was basically lifted from the existing code.  It's how the download builds up the zip file.  It just continually appends newly generated csv files.

usaspending_api/common/tests/test_text_helpers.py
tony-sappe: Good point on the sentinel value

usaspending_api/download/filestreaming/include_file_description.py
tony-sappe: I simply needed to run it myself. It works, of course.

It would be better to consolidate the zip file commands to share the same parameters instead of having to duplicate the same settings across different files.

kbard: I am cleaning this up a bit.  I'm not happy with it in general.  Also adding unit tests.  GO ME!

PR #1780, opened 2019-04-30
DEV-2417: Provide a means by which to capture additional Transaction delta changes

kbard:
**Description:**
As a data re-purposer, I need monthly delta files to also include transactions which were added/modified by appropriate data change operation scripts so that the delta files keep me in sync with USAspending.gov data.

**Technical details:**
We have created a table that is used by the monthly delta file generation process where we can record one off changes and fixes.

1. [x] Found no tests
2. [x] Found no API documentation that required updating
    - [x] Backend
    - [x] Operations
4. [x] No materialized views are affected
5. [x] Front end is unaffected
7. [x] No operations ticket required
8. [x] Jira Ticket [DEV-2417](https://federal-spending-transparency.atlassian.net/browse/DEV-2417):
    - [x] Query now pulls from two tables, but should not take appreciably longer as one of the tables will be quite tiny


usaspending_api/awards/models/transaction_delta.py
```diff
@@ -0,0 +1,72 @@
...
+                for chunk_start in range(0, len(transaction_ids), CHUNK_SIZE):
+                    chunk_of_ids = transaction_ids[chunk_start: chunk_start + CHUNK_SIZE]
+                    chunk_of_inserts = tuple(
+                        TransactionDelta(transaction_id=transaction_id, created_at=created_at)
+                        for transaction_id in chunk_of_ids
+                    )
+
+                    # Fake an upsert by first deleting transactions then inserting.
+                    TransactionDelta.objects.filter(pk__in=chunk_of_ids).delete()
+                    TransactionDelta.objects.bulk_create(chunk_of_inserts)
```
BrianZito: When would you need to upsert a delta? Shouldn't all changes be new and unique?

usaspending_api/awards/tests/unit/test_transaction_delta.py
```diff
@@ -0,0 +1,67 @@
...
+
+class TransactionDeltaTestCase(TransactionTestCase):
+
+    def setUp(self):
+        for _id in range(1, 5):
+            mommy.make('awards.TransactionNormalized', id=_id)
+            mommy.make('awards.TransactionFPDS', transaction_id=_id)
+            mommy.make('awards.Award', id=_id, latest_transaction_id=_id)
+
+    def test_stuff(self):
```
BrianZito: Better title needed. I would rather see this split out into a couple tests with more precise names.

usaspending_api/awards/tests/unit/test_transaction_delta.py
```diff
@@ -0,0 +1,67 @@
...
+        # Get the created_at for transaction 1.  Sleep half a tick.
+        # Update transaction 1.  Make sure new timestamp is newer.
+        created_at = TransactionDelta.objects.get(pk=1).created_at
+        sleep(0.1)
+        TransactionDelta.objects.update_or_create_transaction(1)
+        new_created_at = TransactionDelta.objects.get(pk=1).created_at
+        assert new_created_at > created_at
+
+        TransactionDelta.objects.update_or_create_transactions([3] * (CHUNK_SIZE * 2))
+        assert TransactionDelta.objects.count() == 3
```
BrianZito: Why should this be 3? Shouldn't it be CHUNK_SIZE?

usaspending_api/awards/models/transaction_delta.py
tony-sappe: It is conceivable that multiple data change operations would be run between monthly delta file generations. Therefore we need to allow an existing record to be updated.

kbard: Correct.

usaspending_api/awards/tests/unit/test_transaction_delta.py
kbard: My apologies as this was not intended to be the final version.  I checked in my work in progress just in case I suffered a fatal skydiving accident while on vacation.

usaspending_api/awards/tests/unit/test_transaction_delta.py
kbard: It is three because of the two previous insertions and this one (where we only added the unique transaction... coincidentally with a transaction id of "3").

usaspending_api/awards/tests/unit/test_transaction_delta.py
kbard: No such luck.

usaspending_api/download/management/commands/populate_monthly_delta_files.py
```diff
@@ -284,12 +302,29 @@ def add_arguments(self, parser):
...
+                            'database indicated by the TRANSACTION_DELTA_URL environment variable. '
+                            'This will typically be a staging or production database.')
 
     def handle(self, *args, **options):
         """ Run the application. """
         agencies = options['agencies']
         award_types = options['award_types']
         last_date = options['last_date']
+        remove_transaction_deltas = options['remove_transaction_deltas']
+        transaction_delta_url = os.environ.get('TRANSACTION_DELTA_URL')
```
tony-sappe: There have been a movement away from individual scripts directly accessing env vars. All configs should either be provided by cli args or settings.py

PR #1788, opened 2019-05-02
set static files to our hosted location

willkjackson:
**Description:**
When testing locally, the static assets need to be collected and deployed, and we perform a substitution during deployments. This simplifies that by allowing a dev/user to point to the assets URL that we host on files.usaspending.gov.


tony-sappe:
@willkjackson the static files are generated and served by the django test server when DEBUG=True

PR #1790, opened 2019-05-03
Mod/dev-2432/ 2430 mark deprecated endpoints

emily-brents:
**Description:**
Adds comment/decorator to endpoint views that mark those endpoints as deprecated. Also adds warning header in responses to v1 endpoints to warn users that the endpoint is deprecated
**Technical details:**
Adds class-level decorator to views for v1 endpoints that adds 'X-API-Warn': 'Deprecated API' header to the response.

    - [x] Backend
6. [N/A] Data validation completed
8. [X] Jira Ticket [DEV-2432](https://federal-spending-transparency.atlassian.net/browse/DEV-2432):
    - [N/A] Performance evaluation of affected (API | Script | Download)
    - [N/A] Before / After data comparison

No change to any functionality of endpoints or data changes, just adding a header.


usaspending_api/decorators.py
```diff
@@ -0,0 +1,14 @@
+import logging
```
tony-sappe: Please store this in `usaspending-api/usaspending_api/common/api_versioning.py`

And change the header value to 
```WARNING! You are using a deprecated version of the API which maybe unstable and will be removed in the future.```

File: usaspending_api/decorators.py [hunk 2]
```diff
@@ -0,0 +1,14 @@
...
+
+
+logger = logging.getLogger('console')
+
+
+def deprecated(function):
+    """Add deprecation warning to endpoint"""
+    def wrap(request, *args, **kwargs):
+        response = function(request, *args, **kwargs)
+        logger.warning('Endpoint "{}" is deprecated. Please move to v2 endpoint.'.format(request.path))
```
BrianZito: Are vs endpoints always `v2/<v1endpoint>`? Otherwise we might want to add the url of the new function as a second parameter.

tony-sappe: Not all v1 endpoints have a complementary v2 version

BrianZito: With the less specific message you suggested this works better, but a reference to the v2 endpoint list might be an improvement. Future implementers might only know about our API based on legacy systems' calls to it.

kwhickey:
Only a cursory review from me, but I like the approach. Maintainable and light-weight.

PR #1808, opened 2019-05-10
Es mock

BrianZito:
Creates mock output for elasticsearch without pulling in any fragile dependencies

### DIFF
--- a/usaspending_api/common/elasticsearch/mock_elasticsearch.py
+++ b/usaspending_api/common/elasticsearch/mock_elasticsearch.py
@@ -2,31 +2,36 @@ class MockElasticSearch:
     def __init__(self):
         pass
 
-    def search(self):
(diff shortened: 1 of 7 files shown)


usaspending_api/common/elasticsearch/mock_elasticsearch.py
```diff
@@ -0,0 +1,45 @@
...
+                    }
+                }
+            ]
+        }
+    }
+
+def connection_error():
+    return ConnectionError
+
+stored_values = {"simple_search_by_city": simple_city_search,"connection_error":connection_error}
```
BrianZito: This was set up as an array of functions so that you can have tests which intentionally take a long time or other strange behavior. The actual functions will probably need to be moved to multiple modules for readability.

PR #1809, opened 2019-05-10
Es mock

BrianZito:
Creates mock output for Elasticsearch without pulling in any fragile dependencies.

### DIFF
--- a/usaspending_api/common/elasticsearch/mock_elasticsearch.py
+++ b/usaspending_api/common/elasticsearch/mock_elasticsearch.py
@@ -2,31 +2,36 @@ class MockElasticSearch:
     def __init__(self):
         pass
 
-    def search(self):
-        return {
-    "took": 46,
-    "timed_out": false,
-    "_shards": {
-        "total": 5,
-        "successful": 5,
-        "skipped": 0,
-        "failed": 0
-    },
-    "hits": {
-        "total": 34800,
-        "max_score": 4.362885,
-        "hits": [
-            {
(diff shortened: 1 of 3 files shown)


usaspending_api/common/elasticsearch/mock_elasticsearch.py
```diff
@@ -0,0 +1,45 @@
...
+                    }
+                }
+            ]
+        }
+    }
+
+def connection_error():
+    return ConnectionError
+
+stored_values = {"simple_search_by_city": simple_city_search,"connection_error":connection_error}
```
BrianZito: This is an array of functions instead of just json objects so that more complex behavior like timing out can be simulated. The functions themselves may need to be moved to multiple external files as they grow.

usaspending_api/common/elasticsearch/mock_elasticsearch.py
```diff
@@ -0,0 +1,48 @@
...
+            ]
+        }
+    }
+
+
+def connection_error():
+    return ConnectionError
+
+
+stored_values = {"simple_search_by_city": simple_city_search, "connection_error": connection_error}
```
BrianZito: This is an array of functions instead of just json objects so that more complex behavior like timing out can be simulated. The functions themselves may need to be moved to multiple external files as they grow.

tony-sappe:
Closing since we have a new pytest fixture for #1820 and can use that for future tests

PR #1804, opened 2019-05-09
[dev-2661] adding new mock endpoint -- CitySearchAutoComplete

MaxwellKendall:
**Description:**
Mocks out the new autocomplete endpoint for city search per [DEV-2261](https://federal-spending-transparency.atlassian.net/browse/DEV-2261) and adds optional city property to LocationObject per [DEV-2632](https://federal-spending-transparency.atlassian.net/browse/DEV-2632)

**Details for new endpoint:**

```javascript
// Url:
/api/v2/autocomplete/city/
// Request: 
{
	"search_text": "Springfield",
	"limit": 40,
	"filter": {
		"country_code": "USA",
		"scope": "recipient_location",
		"state_code": ""
	}
}
// Response:
{
    "count": 10,
    "results": [
        {
            "city_name": "Springfield",
            "state_code": "VA"
        },
        {
            "city_name": "Springfield",
            "state_code": "MA"
        },
        {
            "city_name": "Springfield",
            "state_code": "UT"
        },
        {
            "city_name": "Springfield",
            "state_code": "NV"
        },
        {
            "city_name": "Springfield",
            "state_code": "KY"
        },
        {
            "city_name": "Springfield",
            "state_code": "GA"
        },
        {
            "city_name": "Springfield",
            "state_code": "SC"
        }
    ]
}
```
**Requirements for PR merge:**

1. ~[ ] Unit & integration tests updated~
2. [ ] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
    - [ ] Domain Expert <OPTIONAL>
4. ~[ ] Matview impact assessment completed~
5. ~[ ] Frontend impact assessment completed~
6. ~[ ] Data validation completed~
7. ~[ ] Appropriate Operations ticket(s) created~
8. ~[ ] Jira Ticket [DEV-123](https://federal-spending-transparency.atlassian.net/browse/DEV-123)~:
    - [ ] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #1826, opened 2019-05-17
DEV-2617: Autocomplete for City Search

IsaacRay:
**Description:**
An autocomplete endpoint for returning cities found in the data based on a partial search string. Leverages elasticsearch to boost performance.

**Technical details:**
Uses Lucene query_string syntax to complete a wildcard search against the ES index. Uses an aggregation query to return distinct results.

2. [X] API documentation updated (#1804)
    - [ ] Backend
8. [X] Jira Ticket [DEV-2617](https://federal-spending-transparency.atlassian.net/browse/DEV-2617):
    - [ NA] Before / After data comparison

Performance for various search strings was always under 2s.

```
No impacted matviews for this work, no operations work required. New endpoint, so no before and after. Integration tests forthcoming.
```


requirements/requirements.txt
```diff
@@ -2,6 +2,7 @@ appdirs==1.4.3
 attrs==19.1.0
 boto3==1.9.116
 botocore==1.12.116
+certifi==2019.3.9
```
BrianZito: what is this?

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -203,6 +205,9 @@ def db_rows_to_dict(cursor):
 
 
 def download_db_records(fetch_jobs, done_jobs, config):
+    print("Queue has items: {}".format(not fetch_jobs.empty()))
+    sleep(5)  # just because
```
BrianZito: ...just because?

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -309,7 +314,7 @@ def put_alias(client, index, alias_name, award_type_codes):
 def create_aliases(client, index, silent=False):
     for award_type, award_type_codes in INDEX_ALIASES_TO_AWARD_TYPES.items():
         alias_name = "{}-{}".format(settings.TRANSACTIONS_INDEX_ROOT, award_type)
-        if silent is False:
+        if silent is True:
```
BrianZito: Why does this print when silent is true? I would assume that variable suppresses logs.

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -35,7 +36,12 @@
 # python manage.py es_rapidloader --index-name NEWINDEX --reload-all all
 
 DEFAULT_DATETIME = datetime.strptime("2007-10-01+0000", "%Y-%m-%d%z")
-ES = Elasticsearch(settings.ES_HOSTNAME, timeout=300)
+
+if 'https' in settings.ES_HOSTNAME:
+    ES = Elasticsearch(settings.ES_HOSTNAME, timeout=300, use_ssl=True, verify_certs=True, ca_certs=certifi.where())
+else:
+    ES = Elasticsearch(settings.ES_HOSTNAME, timeout=300)
```
BrianZito: Is the timeout for creating the ES or for each query? At 5 minutes no user will still be waiting for a response.

usaspending_api/references/v2/views/city.py
```diff
@@ -0,0 +1,113 @@
...
+    },
+    {"key": "search_text", "name": "search_text", "type": "text", "text_type": "search", "optional": False},
+    {"key": "limit", "name": "limit", "type": "integer", "max": 500, "optional": True, "default": 10},
+]
+
+
+@validate_post_request(models)
+class CityAutocompleteViewSet(APIDocumentationView):
+    """
+    endpoint_doc:
```
BrianZito: Fill in endpoint doc location

usaspending_api/etl/es_etl_helpers.py
kbard: This is going to be a merge conflict I think.  This was fixed.

usaspending_api/references/v2/views/city.py
```diff
@@ -0,0 +1,113 @@
...
+                    "aggs": {
+                        "max_score": {"max": {"script": "_score"}},
+                        "states": {"terms": {"field": return_fields[1], "size": 50000}}
+
+                        },
+                }
+            }
+        }
+
+        hits = es_client_query(index="{}*".format(settings.TRANSACTIONS_INDEX_ROOT), body=query, retries=10)
```
BrianZito: retries already defaults to 10. I would remove that to DRY up code, unless there's a special reason this should ALWAYS be 10

usaspending_api/references/v2/views/city.py
BrianZito:
```suggestion
    endpoint_doc: usaspending_api/api_contracts/autocomplete/City.md
```

requirements/requirements.txt
tony-sappe: https://pypi.org/project/certifi/

usaspending_api/etl/management/commands/es_rapidloader.py
tony-sappe: This is an ETL script, not API

requirements/requirements.txt
BrianZito: I was less asking that to understand the library personally, and more to check what it's being used for and if this dependency was intended to be added. It seems odd that this project would just need to start supporting SSL now.

usaspending_api/etl/es_etl_helpers.py
kbard: This needs to go back to `if silent is False:`.

usaspending_api/database_scripts/etl/transaction_delta_view.sql
```diff
@@ -75,14 +75,16 @@ SELECT
...
 
   UTM.recipient_location_country_code,
   UTM.recipient_location_country_name,
   UTM.recipient_location_state_code,
   UTM.recipient_location_county_code,
   UTM.recipient_location_county_name,
   UTM.recipient_location_zip5,
-  UTM.recipient_location_congressional_code
+  UTM.recipient_location_congressional_code,
+  TRIM(TRAILING FROM COALESCE(FPDS.legal_entity_city_name, FABS.legal_entity_city_name)) AS recipient_location_city_name
```
kbard: Why just trailing?  Will leading whitespace not mess things up?

requirements/requirements.txt
tony-sappe: This was intentionally added for the elasticsearch etl script to support planned changes in the Elsticsearch architecture.

usaspending_api/database_scripts/etl/transaction_delta_view.sql
tony-sappe: I didn't notice any leading whitespace

usaspending_api/references/v2/views/city.py
BrianZito: @tony-sappe quick thoughts? If there's no complaints I can just do this myself

usaspending_api/references/v2/views/city.py
```diff
@@ -0,0 +1,113 @@
...
```
kwhickey: @IsaacRay - what is the name of the index here, and the alias to be used for it? In this case, shouldn't the alias cover wildcard-names of indexes it hits? ... such that we should be using the alias here and not need wildcards at the end?

BrianZito:
```suggestion
        hits = es_client_query(index="{}*".format(settings.TRANSACTIONS_INDEX_ROOT), body=query)
```

tony-sappe: @kwhickey The index names does not matter for the API view logic.
The ES alias prefix is used by the API view logic and is set in `settings.TRANSACTIONS_INDEX_ROOT` with a default value of `future-transactions`. The wildcard is necessary since there are several aliases for the different award types

usaspending_api/references/v2/views/city.py
```diff
@@ -0,0 +1,127 @@
...
+@validate_post_request(models)
+class CityAutocompleteViewSet(APIDocumentationView):
+    """
+    endpoint_doc: usaspending_api/api_contracts/autocomplete/City.md
+    """
+
+    @cache_response()
+    def post(self, request, format=None):
+
+        search_text = es_sanitize(request.data["search_text"])
```
tony-sappe: Good idea to upper these input strings:
- search_text
- country_code
- state_code

PR #1838, opened 2019-05-23
DEV-2739: Elasticsearch Query Speed Optimizations to City Search Autocomplete 

kwhickey:
**Description:**
Make the search faster by cutting out the need for analysis and scoring and reducing buckets.

Profiled against the ~staging `nonprod`~ local single-node cluster, 5 shard index with 1M docs. _(AWS `nonprod` cluster does not have the profiler because it is an `X-Pack` feature)_
_**Before**_: `108-125ms` range per-shard for query, `32-95ms` range per-shard for aggregation
_**After**_:  `3-6ms` range per-shard for query, `9-19ms` range per-shard for aggregation

**Technical details:**
Using a pure filter wrapped in a `constant_score` allows no scoring to take place, cutting out some time. Also using the `wildcard` term-level query allows it to match completely on the keyword type of the multi-field. It will not provide false-matches based on tokenized terms from the overall city name, because it looks at the entirety of the text, not analyzed tokens produced by the standard analyzer. With this approach the python-side removal of entries that did not match the initial search text could be removed.

Also, propagating the `limit` from the request down to `cities` buckets `size`, since we end up reducing results to only the top-`limit` buckets anyway. No need to ask the cluster to work on more buckets nor to parse/sort them on the API side.

    - [ ] Backend
    - N/A Frontend <OPTIONAL>
    - N/A Operations <OPTIONAL>
8. Jira Ticket [DEV-2739](https://federal-spending-transparency.atlassian.net/browse/DEV-2739):

```
```

- Related to [DEV-2617](https://federal-spending-transparency.atlassian.net/browse/DEV-2617) and PR #1826

### DIFF
--- a/.env
+++ b/.env
@@ -13,3 +13,8 @@ POSTGRES_CLUSTER_DIR=../docker_db
 
 # MATVIEW_SQL_DIR has to be inside of the project (check the docker-compose file)
 MATVIEW_SQL_DIR=matview_sql
+
+# Host of an your elasticsearch cluster node
+# Defaulted to a cluster running on localhost (resolvable from within docker by host.docker.internal) until
+# elasticsearch is incorporated in this compose file
+ES_HOSTNAME=https://host.docker.internal:9200
--- a/.env
+++ b/.env
@@ -14,7 +14,7 @@ POSTGRES_CLUSTER_DIR=../docker_db
 # MATVIEW_SQL_DIR has to be inside of the project (check the docker-compose file)
 MATVIEW_SQL_DIR=matview_sql
 
-# Host of an your elasticsearch cluster node
+# Host of an elasticsearch cluster node
 # Defaulted to a cluster running on localhost (resolvable from within docker by host.docker.internal) until
 # elasticsearch is incorporated in this compose file
 ES_HOSTNAME=https://host.docker.internal:9200
--- a/.env
+++ b/.env
@@ -13,8 +13,3 @@ POSTGRES_CLUSTER_DIR=../docker_db
 
 # MATVIEW_SQL_DIR has to be inside of the project (check the docker-compose file)
 MATVIEW_SQL_DIR=matview_sql
-
-# Host of an elasticsearch cluster node
-# Defaulted to a cluster running on localhost (resolvable from within docker by host.docker.internal) until
-# elasticsearch is incorporated in this compose file
-ES_HOSTNAME=https://host.docker.internal:9200
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -66,4 +66,3 @@ services:
     command: /bin/sh -c "python -u manage.py runserver 0.0.0.0:8000"
     environment:
       DATABASE_URL: postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/data_store_api
-      ES_HOSTNAME: ${ES_HOSTNAME}
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -66,3 +66,4 @@ services:
     command: /bin/sh -c "python -u manage.py runserver 0.0.0.0:8000"
     environment:
       DATABASE_URL: postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/data_store_api
+      ES_HOSTNAME: ${ES_HOSTNAME}
--- a/usaspending_api/common/elasticsearch/client.py
+++ b/usaspending_api/common/elasticsearch/client.py
@@ -22,8 +22,9 @@ def create_es_client():
     try:
         # If the connection string is using SSL with localhost, disable verifying
         # the certificates to allow testing in a development environment
-        if settings.ES_HOSTNAME.startswith("https://localhost"):
-            logger.warn("SSL cert verification is disabled. Safe only for local development")
+        if settings.ES_HOSTNAME.startswith("https://localhost") or settings.ES_HOSTNAME.startswith(
+                "https://host.docker.internal"):
+            logger.warning("SSL cert verification is disabled. Safe only for local development")
             import urllib3
             urllib3.disable_warnings()
             ssl_context = create_ssl_context()
--- a/usaspending_api/common/elasticsearch/client.py
+++ b/usaspending_api/common/elasticsearch/client.py
@@ -22,8 +22,8 @@ def create_es_client():
     try:
         # If the connection string is using SSL with localhost, disable verifying
         # the certificates to allow testing in a development environment
-        if settings.ES_HOSTNAME.startswith(("https://localhost", "https://host.docker.internal")):
-            logger.warning("SSL cert verification is disabled. Safe only for local development")
+        if settings.ES_HOSTNAME.startswith("https://localhost"):
+            logger.warn("SSL cert verification is disabled. Safe only for local development")
             import urllib3
             urllib3.disable_warnings()
             ssl_context = create_ssl_context()
--- a/usaspending_api/common/elasticsearch/client.py
+++ b/usaspending_api/common/elasticsearch/client.py
@@ -22,8 +22,7 @@ def create_es_client():
     try:
         # If the connection string is using SSL with localhost, disable verifying
         # the certificates to allow testing in a development environment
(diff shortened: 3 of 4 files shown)


usaspending_api/references/v2/views/city.py
```diff
@@ -65,26 +65,23 @@ def prepare_search_terms(request_data):
     return [es_sanitize(field).upper() if isinstance(field, str) else field for field in fields]
 
 
-def create_elasticsearch_query(return_fields, scope, search_text, country, state):
-    query_string = create_es_search("wildcard", scope, search_text, country, state)
+def create_elasticsearch_query(return_fields, scope, search_text, country, state, limit):
+    query_string = create_es_search(scope, search_text, country, state)
```
kbard: Super duper minor thing, but this is no longer returning a string.

usaspending_api/references/v2/views/city.py
```diff
@@ -95,29 +92,58 @@ def create_elasticsearch_query(return_fields, scope, search_text, country, state
...
-            ["query_string"]["fuzzy_prefix_length"] to a value like 1
+        Args:
+            scope: which city field was chosen for searching `pop` (place of performance) or `recipient_location`
+            search_text: the text the user is typing in and sent to the backend
+            country: optional country selected by user
+            state: optional state selected by user
+            term_level_query_method: Supports `wildcard` or `fuzzy`. Defaults to wildcard
     """
-    method_char = "~" if method == "fuzzy" else "*"
+    method_char = "" if term_level_query_method == "fuzzy" else "*"
```
kbard: Removing the tilde changed how the fuzzy match behaves.  I'm not sure I understand the difference.  Both seem to be returning fuzzy matches, but with the tilde we don't get anything for "LOND" whereas we do without.  I think I like the sans tilde results better.

Either way, the fuzzy matching is not being used anywhere.  Is it worth keeping?

usaspending_api/references/v2/views/city.py
```diff
@@ -134,11 +160,10 @@ def query_elasticsearch(query, search_text):
 def parse_elasticsearch_response(hits, search_text):
```
kbard: Again, not your code, but `search_text` doesn't seem to be used.  Should we remove it?

tony-sappe: It was used before these changes. Removing makes sense

docker-compose.yaml
```diff
@@ -66,3 +66,4 @@ services:
     command: /bin/sh -c "python -u manage.py runserver 0.0.0.0:8000"
     environment:
       DATABASE_URL: postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/data_store_api
+      ES_HOSTNAME: ${ES_HOSTNAME}
```
tony-sappe: You are the third developer attempting to add this env var to the docker-compose file. Let's wait for Will/Joe to add an ES container to the yaml file first.

usaspending_api/common/elasticsearch/client.py
```diff
@@ -22,8 +22,9 @@ def create_es_client():
     try:
         # If the connection string is using SSL with localhost, disable verifying
         # the certificates to allow testing in a development environment
-        if settings.ES_HOSTNAME.startswith("https://localhost"):
-            logger.warn("SSL cert verification is disabled. Safe only for local development")
+        if settings.ES_HOSTNAME.startswith("https://localhost") or settings.ES_HOSTNAME.startswith(
+                "https://host.docker.internal"):
```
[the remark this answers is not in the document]
tony-sappe: @kbard wow, I learn something new every day.

Perhaps there is a better way to disable SSL cert verification without looking for specific hostname patterns. It would be outside the scope of this ticket to solve.

.env
```diff
@@ -13,3 +13,8 @@ POSTGRES_CLUSTER_DIR=../docker_db
 
 # MATVIEW_SQL_DIR has to be inside of the project (check the docker-compose file)
 MATVIEW_SQL_DIR=matview_sql
+
+# Host of an your elasticsearch cluster node
```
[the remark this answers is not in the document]
kwhickey: Done

usaspending_api/common/elasticsearch/client.py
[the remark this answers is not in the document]
kwhickey: Done

usaspending_api/references/v2/views/city.py
```diff
@@ -95,29 +92,58 @@ def create_elasticsearch_query(return_fields, scope, search_text, country, state
...
+        "must": [
+            {
+                term_level_query_method: {
+                    "{}_city_name.keyword".format(scope): search_text + method_char
+                }
+            }
+        ]
+    }
+
+    def build_country_match(country_match_scope, country_match_country):
```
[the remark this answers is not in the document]
kwhickey: Done

usaspending_api/references/v2/views/city.py
```diff
@@ -95,29 +92,58 @@ def create_elasticsearch_query(return_fields, scope, search_text, country, state
...
     elif country == "FOREIGN":
-        query_string = ("NOT (({scope}_country_code:USA) OR "
-                        "({scope}_country_code:UNITED STATES)) AND ").format(scope=scope)
+        # Create a "Should Not" query with a nested bool, to get everything non-USA
+        query["should"] = {
+            "bool": {
+                "must_not": [build_country_match(scope, "USA"), build_country_match(scope, "UNITED STATES")]
+            }
+        }
     elif country and country != "USA":
```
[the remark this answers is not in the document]
kwhickey: Done

usaspending_api/references/v2/views/city.py
kwhickey: Done

usaspending_api/references/v2/views/city.py
kwhickey: Done. Removed fuzzy. It was cool but extraneous and not SOLID

usaspending_api/references/v2/views/city.py
kwhickey: Done

docker-compose.yaml
kwhickey: It would be helpful to keep this in. If it were _just_ a step in Dockerizing local elasticsearch, then yes, holding off makes sense. However, this is needed to make the standard setup (i.e. running API in docker) work with elasticsearch features -- _wherever_ that elasticsearch may be. 

i.e. you _could_ set this to a dockerized elasticsearch container, but you could also set it to the `sandbox` or `nonprod` cluster endpoints, or your own standalone non-dockerized ES cluster.

Not having this is like not setting the `DATABASE_URL` env var (albeit a much smaller impact).

usaspending_api/references/v2/views/city.py
```diff
@@ -65,26 +65,23 @@ def prepare_search_terms(request_data):
...
-    query_string = create_es_search("wildcard", scope, search_text, country, state)
+def create_elasticsearch_query(return_fields, scope, search_text, country, state, limit):
+    query_string = create_es_search(scope, search_text, country, state)
     query = {
         "_source": return_fields,
         "size": 0,
         "query": {
-            "bool": {
-                "must": query_string,
+            "constant_score": {
```
IsaacRay: The ES docs are a little unclear on this, but I'm not certain that this constant_score result is even being looked at by the aggregation part of the query, which is all we care about.  What I mean is, I don't think the aggregation cares that the results all have the same score, because its still organizing around doc count. Do you have an example where this is actually making a noticeable impact?

kwhickey: @IsaacRay - I tried out a few finicky ways to get these queries to run only in the _filter context_, and this one was one of the few that worked: https://www.elastic.co/guide/en/elasticsearch/guide/current/combining-filters.html

However, I am just now trying it out with a `bool` instead of `constant_score` and it seems to work the same. I'll report back.

docker-compose.yaml
tony-sappe: `ES_HOSTNAME` can easily be set in settings.py, unlike `DATABASE_URL`

usaspending_api/references/v2/views/city.py
kwhickey: @IsaacRay not a definitive source, but some evidence that they're the same and `bool` should be fine: 

I also tested it locally with `bool` and ran the profiler and performance was on par (maybe better?). Changing now.

PR #1870, opened 2019-06-12
DEV-2504: Promote unique_award_key to awards

kbard:
**Description:**

At present, the new Broker `unique_award_key` field can only be found in transactions.  It needs to be promoted into the `awards` table so that we may start taking advantage of it.

**Technical details:**

We are repurposing the `awards.generated_unique_award_id` as this will be much less disruptive to the overall code base.  In a future ticket, we will rename the `generated_unique_award_id` throughout the project in order to help alleviate confusion.

2. [x] API documentation should be unaffected
    - [ ] Backend
4. [x] Materalized views have been updated where necessary
5. [x] This change should be transparent to the front end
7. [x] Operations ticket: [DEV-731](https://federal-spending-transparency.atlassian.net/browse/DEV-731)
8. [x] Jira Ticket [DEV-2504](https://federal-spending-transparency.atlassian.net/browse/DEV-2504):


usaspending_api/broker/management/commands/load_fsrs.py
```diff
@@ -28,123 +28,83 @@ def get_award_data(db_cursor, award_type, max_id, internal_ids=None):
...
+        if not award:
+            msg = "[Internal ID {}] Award not found for unique_award_key '{}'"
+            logger.warning(msg.format(row['internal_id'], row['unique_award_key']))
+            return None
 
-            # We don't have a matching award for this subgrant, log a warning and continue to the next row
-            if not award:
-                msg = "[Internal ID {}] Award not found for FAIN '{}'"
-                logger.warning(msg.format(row['internal_id'], row['fain']))
-                return None
```
BrianZito: Can you briefly explain what's happening in this part?

kbard: I believe it's attempting to find USAspending awards for Broker subawards.  This is a supporting function.  The documentation for it is in `gather_shared_award_data` function.

> Creates a dictionary with internal IDs as keys that stores data for each award that's shared between all
> its subawards so we don't have to query the DB repeatedly
> Note: subawards without a matching award will still be processed, just without a matching id which may be matched later

PR #1872, opened 2019-06-12
Hotfix: update congressional justifications

BrianZito:


kbard:
Very few of these new CJ URLs seem to actually work.  Is that ok?  Also, `http://www.nbrc.gov/cj` was removed.  Just checking that this was actually intentional since all the others were additions.

Is there a ticket associated with this work?  Should we link it here?

BrianZito:
This is an update to https://federal-spending-transparency.atlassian.net/browse/DEV-2714

Yes, I should have linked that.

PR #1889, opened 2019-06-24
Mod/Dev-2643 Update IDV download zip contents

sethstoudenmier:
**Description:**
Updated the readme.txt inside of the IDV download zip and also added the Data_Dictionary_Crosswalk.xlsx. 

**Technical details:**
Added the ability to copy a file from an HTTP URL to a specified file path. This is used to copy the data dictionary from "https://files.usaspending.gov/docs/DATA+Transparency+Crosswalk.xlsx" to the zip folder.

    - [x] Backend
8. [ ] Jira Ticket [DEV-2643](https://federal-spending-transparency.atlassian.net/browse/DEV-2643):
    - [X ] Link to this Pull-Request

```
```


usaspending_api/download/filestreaming/csv_generation.py
```diff
@@ -388,3 +392,13 @@ def fail_download(download_job, exception, message):
...
     download_job.job_status_id = JOB_STATUS_DICT['failed']
     download_job.save()
+
+
+def add_data_dictionary_to_zip(working_dir, zip_file_path):
+    write_to_log(message="Adding data dictionary to zip file")
+    data_dictionary_file_name = "Data_Dictionary_Crosswalk.xlsx"
+    data_dictionary_file_path = os.path.join(working_dir, data_dictionary_file_name)
+    # URL needs to be updated if filename changes
+    data_dictionary_url = "https://files.usaspending.gov/docs/DATA+Transparency+Crosswalk.xlsx"
```
BrianZito: Future developers won't find this out until it's too late. We should look into some way to add this to the settings file, read it from a constant where the download endpoint is defined, make it an environment variable, or maybe even read it off of the endpoint contract itself.

kbard: We sort of have something like this in the ROSETTA_DICT_S3_PATH setting, but it is used to retrieve the file from S3 in `load_rosetta.py`.

BrianZito: I would be in favor of putting the two side-by-side then

BrianZito: If you take this suggestion, please also remap the `s3_public_url` local variable found in load_rosetta.py to whatever solution you use, as the exact same url is established there.

sethstoudenmier: Good point. I will make the change.

usaspending_api/common/retrieve_file_from_uri.py
```diff
@@ -60,6 +61,8 @@ def _handle_http(self):
         f = tempfile.SpooledTemporaryFile()
         f.write(r.content)
         f.seek(0)  # go to beginning of file for reading
+        if self.dest_file_path:
```
kbard: Is it difficult to implement this for local and s3 files as well?

Also, what are your thoughts on, instead of a __init__ parameter, what if there was a `copy_file_object` (or maybe just `copy`) method that accepted the file name and copied that-a-way?  In my head, I want the __init__ to be just about the file object.

And if you really just wanted to deal with http for now, maybe raise a NotImplementedError exception for the other two so other developers don't spend time trying to figure out why their files aren't copying.

usaspending_api/data/idv_download_readme.txt
```diff
@@ -22,3 +24,7 @@ This file contains award-level data (data that sums obligated amounts and ceilin
 File: IDV_[Award PIID]_TransactionHistory_#.csv
 
 This file contains transaction-level data representing all of the modifications made to this IDV. This data is also available from the Advanced Search or Custom Award Download sections of the site.
+
+File: Data_Dictionary_Crosswalk.xlsx
+
+This file contains the data dictionary covering all elements available for download from USASpending.gov. You can find an online and up-to-date version of the data dictionary here: https://www.usaspending.gov/#/download_center/data_dictionary
```
kbard: Very minor, but it's USAspending (lower case s in "spending").

usaspending_api/api_contracts/contracts/downloadCenter/DownloadCenter.md
```diff
@@ -103,4 +103,4 @@ This endpoint returns data corresponding to the latest data dictionary csv file.
 + `total_rows`: 393 (required, number)
 + `total_columns`: 12 (required, number)
 + `total_size`: `119.32KB` (required, string)
-+ `download_location`: `http://files-nonprod.usaspending.gov/docs/DATA+Transparency+Crosswalk.xlsx` (required, string)
++ `download_location`: `http://files.usaspending.gov/docs/DATA+Transparency+Crosswalk.xlsx` (required, string)
```
BrianZito: Should we be concerned at all about this changing? Personally I don't see an issue with just not using the nonprod version of the key, but this might cause automated tests to fail if the crosswalk changes.

kbard: This is `https` I do believe.

usaspending_api/references/management/commands/load_rosetta.py
```diff
@@ -40,7 +40,7 @@
 class Command(BaseCommand):
     help = "Loads an Excel spreadsheet of DATA Act/USAspending data names across the various systems into <>"
 
-    s3_public_url = "https://files{}.usaspending.gov/docs/DATA+Transparency+Crosswalk.xlsx"
+    s3_public_url = settings.DATA_DICTIONARY_DOWNLOAD_URL
```
kbard: I believe the changes in this file need to be reverted.  The deployment scripts update the `DOWNLOAD_ENV` setting which is used to determine which URL to pull from.  In fact, maybe DATA_DICTIONARY_DOWNLOAD_URL could be updated in the settings file to use the same logic so everywhere it's used it's pointing to the correct location.  That sounded better in my head.  Let me know if what I said didn't make sense and I'll try to esplain more betterer.

usaspending_api/api_contracts/contracts/downloadCenter/DownloadCenter.md
BrianZito:
```suggestion
+ `download_location`: (required, string)
```

In retrospect, I don't think we really need to be testing the exact URL. I would advise NOT specifying value here, as it's just a contract.

sethstoudenmier: I like that more. Will add that into my changes.

sethstoudenmier: On that note actually, it might be weird that everything else says the value and type next to it. Thoughts?

BrianZito: We've had this debate for a while at this point, and TBH I forgot the resolution, but I would drop all the values. We aren't promising consumers that the file will be exactly 119.32 Kilobytes. @ebdabbs, you agree?

sethstoudenmier: For the time being I have put it back except correcting it to `https` instead of `http`. I agree it is weird having exact values, but don't want just download_location to be empty.

usaspending_api/api_contracts/contracts/downloadCenter/DownloadCenter.md
```diff
@@ -103,4 +103,4 @@ This endpoint returns data corresponding to the latest data dictionary csv file.
 + `total_rows`: 393 (required, number)
 + `total_columns`: 12 (required, number)
 + `total_size`: `119.32KB` (required, string)
-+ `download_location`: `http://files-nonprod.usaspending.gov/docs/DATA+Transparency+Crosswalk.xlsx` (required, string)
++ `download_location`: `https://files-nonprod.usaspending.gov/docs/DATA+Transparency+Crosswalk.xlsx` (required, string)
```
BrianZito: This will be a problem if we run these in the production environment one day, but we don't do that currently.

usaspending_api/common/retrieve_file_from_uri.py
```diff
@@ -45,6 +45,15 @@ def get_file_object(self):
         else:
             raise NotImplementedError("No handler for scheme: {}!".format(self.parsed_url_obj.scheme))
 
+    def copy(self, dest_file_path):
+        """
+            create a copy of the file and place at "dest_file_path"
+        """
+        if self.parsed_url_obj.scheme.startswith("http"):
```
BrianZito: this still works for https, right?

sethstoudenmier: Yep. Tested locally with https and startswith should work for anything that begins with "http".

PR #1891, opened 2019-06-26
Mod/Dev-2967: Return human-readable business categories

sethstoudenmier:
**Description:**
Return human-readable names for the business categories within the recipient object when accessing /api/v2/awards/<award_id>.

**Technical details:**
The [API Contract](https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/api_contracts/contracts/awards/AwardProfile.md#recipient-object) for /api/v2/awards/<award_id> says that the business_categories list as part of the recipient object should all be human-readable names. I referenced [get_business_categories.py](https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/broker/helpers/get_business_categories.py) to determine all of the business categories and create a list of objects that map each to their display name. 

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
8. [ ] Jira Ticket [DEV-2967](https://federal-spending-transparency.atlasasian.net/browse/DEV-2967):

```
```


usaspending_api/common/helpers/business_categories_helper.py
```diff
@@ -0,0 +1,143 @@
...
+    {"display_name": "U.S. Territory Government", "field_name": "us_territory_or_possession"},
+    {"display_name": "U.S. Local Government", "field_name": "local_government"},
+    {"display_name": "Native American Tribal Government", "field_name": "indian_native_american_tribal_government"},
+    {"display_name": "U.S. Government Authorities", "field_name": "authorities_and_commissions"},
+    {"display_name": "Interstate Entity", "field_name": "interstate_entity"},
+    {"display_name": "Council of Governments", "field_name": "council_of_governments"},
+
+    # Individuals
+    {"display_name": "Individuals", "field_name": "individuals"},
+]
```
BrianZito: Have we checked that these strings don't appear elsewhere in the codebase? I would like to see one source of truth for these.

sethstoudenmier: I did not see anywhere that had the display name. Seemed that everything used the underscore name only. Even the API Contract says that it will return the display name, but returns the underscore name.

BrianZito: Well I would like to see the underscore names, even if not connected to the display name, all come from the same list.

sethstoudenmier: Gotcha, I am following now. I would really like to get a complete list of these strings and then go and make a lookup for them. The get_business_categories.py file seemed to be where the strings originated so I used those, but there does seem to be different sets of strings out there.

BrianZito: I trust you to take targets of opportunity as you see fit.

usaspending_api/common/helpers/business_categories_helper.py
```diff
@@ -0,0 +1,143 @@
+# Dictionary of Business Categories that pair them with their human readable name
```
kbard: Dumb question.... would a dictionary with the field name as the key and the display name as the value not work here?

```
{
    'category_business': 'Category Business',
    'small_business': 'Small Business',
    ...
}
```

Seems like it might simplify the code a bit.

sethstoudenmier: That was the direction I first went, but I wasn't sure if we would ever want to look them up by display name. Figured having a list of dictionaries would allow a lookup either direction. If it will rarely happen then I agree it would be much simpler to go with a dictionary where the key is the field_name and the value is the display_name and can make that change.

kbard: I mean, you can still look up in the other direction in a dictionary pretty much the same way you're doing it here.  And if it's something you do regularly, you can just make an inverse dictionary based on the original dictionary.

Issue #1890, opened 2019-06-26
500 error on spending endpoint


ece-jacob-scott:
Hi, thank you for making this api the idea is amazing!

I'm currently just messing around with the endpoints and get an internal server error whenever I try and access the `/spending` endpoint. Regardless of the `type` specified.

![postman screenshot](https://user-images.githubusercontent.com/36082020/60199878-803bac00-980a-11e9-929f-109f91a0a7e9.png)

The only header that was set was for "Content-Type". I read the documentation again and I think I'm making the request correctly, if you think I am making an incorrect request please tell me where I'm going wrong.

Thank you!

lisaam:
The documentation needs to be updated to better reflect when filters are required, like with budget_function

Parameters like this work: {"type":"budget_function","filters":{"fy":"2019","quarter":"2"}}

ece-jacob-scott:
Okay that worked. Thank you! Why does it give me a 500 error though? That's kind of misleading it should be a 400 error since it's a client error. If I have time I will work on a merge request for this. Thank you again for a quick response.

PR #1894, opened 2019-06-26
Added a check for filters and throw a 400 if filter is None

ece-jacob-scott:
**Description:**
Just added a check in the spending_view filter. The check makes sure that if `filters` is None then the function raises a 400 error informing the user that they are missing a mandatory parameter. Previously the server would throw a 500 error.

Raised the issue in [this issue](https://github.com/fedspendingtransparency/usaspending-api/issues/1890).

**Technical details:**
Adds an if statement after the checks for `_type` in the spending_view filter function.

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>

```
```


ece-jacob-scott:
Someone already changed the docs for this change 

e4504673866f25d0fefff21dbf3c9788d45707c4

BrianZito:
@kbard Is this something that should be done via TinyShield?

usaspending_api/spending_explorer/v2/filters/type_filter.py
```diff
@@ -90,6 +90,9 @@ def type_filter(_type, filters, limit=None):
             'Valid Types: budget_function, budget_subfunction, federal_account, program_activity,'
             'object_class, recipient, award, award_category agency, agency_type, agency_sub')
 
+    if filters is None:
+        raise InvalidParameterException('Missing Required Request Parameter, "fiters": { "filter_options" }')
+
```
BrianZito: Thanks for your contribution! Did you check that this raises the exception before the point in the code where the 500 error comes up? Ideally there should be an integration test to prove this functionality in usaspending_api/spending_explorer/tests/integration/test_spending_explorer.py

ece-jacob-scott: I did not test this out I just assumed that this works because when "types" is missing the exception gets raised and the "filters" property isn't used before that point. However, I can test it this weekend probably when I have time and add a test.

tony-sappe:
> @kbard Is this something that should be done via TinyShield?

@BrianZito Optimally yes, however it is currently not used in that view.

PR #1901, opened 2019-07-01
change to plain keyword, country not county

emily-brents:
**Description:**
Change to keyword type, fixed `pop_county_code` to `pop_country_code`

**Technical details:**

    - [ ] Backend
7. [X] Appropriate Operations ticket(s) created - https://federal-spending-transparency.atlassian.net/browse/DEV-2870#icft=DEV-2870
8. [X] Jira Ticket [DEV-2806](https://federal-spending-transparency.atlassian.net/browse/DEV-2806):

```
no change to matviews, no change to frontend, all tests already updated
```

**Data Validation / Performance
Before - Time: 33 ms
```
{
    "count": 1,
    "results": [
        {
            "city_name": "BRISTOL",
            "state_code": "GBR"
        }
    ]
}
```

After - 15 ms
```
{
    "count": 1,
    "results": [
        {
            "city_name": "BRISTOL",
            "state_code": "GBR"
        }
    ]
}
```

### DIFF
--- a/usaspending_api/etl/es_transaction_mapping.json
+++ b/usaspending_api/etl/es_transaction_mapping.json
@@ -214,12 +214,7 @@
           "index": false
         },
         "pop_country_code": {
-          "type": "text",
-          "fields": {
-            "keyword": {
-              "type": "keyword"
-            }
-          }
+          "type": "keyword"
         },
         "pop_country_name": {
           "type": "text"
@@ -249,12 +244,7 @@
         },
 
         "recipient_location_country_code": {
-          "type": "text",
-          "fields": {
-            "keyword": {
-              "type": "keyword"
-            }
-          }
+          "type": "keyword"
         },
         "recipient_location_country_name": {
           "type": "text"
--- a/usaspending_api/etl/es_transaction_template.json
+++ b/usaspending_api/etl/es_transaction_template.json
@@ -225,7 +225,7 @@
           "index": false
         },
         "pop_country_code": {
-          "type": "text"
+          "type": "keyword"
         },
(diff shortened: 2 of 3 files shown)


usaspending_api/etl/es_transaction_mapping.json
```diff
@@ -249,12 +244,7 @@
         },
 
         "recipient_location_country_code": {
-          "type": "text",
-          "fields": {
-            "keyword": {
-              "type": "keyword"
-            }
-          }
```
BrianZito: Previously we had discussed the merits of having keyword as a field rather than changing type. What changed?

emily-brents: Tony said so

BrianZito: This is an open source project, whereas unfortunately (perhaps fortunately) Tony's brain is not. For my own education if nothing else, can we note why this change was wrong to make? "We" can be Tony if needed.

emily-brents: `it's returning the value in .keyword which is identical to the default and misses the original purpose of searching exact matches on keyword.` -according to Tony

tony-sappe: To close the loop: the original purpose was to change the type from `text` to `keyword` since the queries on those fields are only exact match and keyword type should increase performance.

Another option would be to create the keyword subtype and then use the .keyword on the field for exact match queries. IIRC we leaned towards not subtyping unless we had the immediate need for both types

PR #1905, opened 2019-07-03
Fix/Dev-2740: Handle parent recipients differently and cover missed cases

sethstoudenmier:
**Description:**
Missed some cases when originally implementing this API change for DEV-2740. This PR should handle when a parent recipient does not exist for a provided recipient.

**Technical details:**
Use a boolean to determine when a recipient is a parent and also check if the recipient has a name and unique id before creating a hash when the recipient does not exist. If the parent recipient does not exist return None.

    - [ ] Backend
8. [ ] Jira Ticket [DEV-2740](https://federal-spending-transparency.atlassian.net/browse/DEV-2740):

```
```

### DIFF
--- a/usaspending_api/common/recipient_lookups.py
+++ b/usaspending_api/common/recipient_lookups.py
@@ -5,24 +5,40 @@ def obtain_recipient_uri(
         recipient_name,
         recipient_unique_id,
         parent_recipient_unique_id=None,
-        child_recipient_unique_id=None):
-    if not recipient_unique_id:
-        return create_recipient_uri_without_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id)
-    return fetch_recipient_uri_with_duns(recipient_unique_id, parent_recipient_unique_id, child_recipient_unique_id)
+        is_parent_recipient=False):
+    # No need to create a hash when the parent recipient does not exist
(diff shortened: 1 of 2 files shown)


usaspending_api/common/recipient_lookups.py
```diff
@@ -5,24 +5,40 @@ def obtain_recipient_uri(
...
-        child_recipient_unique_id=None):
-    if not recipient_unique_id:
-        return create_recipient_uri_without_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id)
-    return fetch_recipient_uri_with_duns(recipient_unique_id, parent_recipient_unique_id, child_recipient_unique_id)
-
-
-def create_recipient_uri_without_duns(recipient_name, recipient_unique_id, parent_recipient_unique_id=None):
+        is_parent_recipient=False):
+    # No need to create a hash when the parent recipient does not exist
+    if is_parent_recipient and not (recipient_name and recipient_unique_id):
```
tony-sappe: The parent-child definition requires a parent DUNS but not a name. As reference: we have a few thousand parent recipients with no associated name (aka DUNS-only).

This should be sufficient:
```if is_parent_recipient and not recipient_unique_id```

PR #1921, opened 2019-07-10
Fix/dev 2740: Return None when the recipient name and id are both null

sethstoudenmier:
**Description:**
The API response should be null for the recipient hash when both the name and id are null.

**Technical details:**
There are three possible outcomes for the recipient hash:
- A recipient name and id are both provided
- A recipient name is provided, but no id
- A recipient name and id are both not provided

Currently, the first two are handled, but an invalid hash is provided when there are id and no name for the recipient. In this instance, "None" is going to be returned so that the recipient hash is null on the API response.

    - [x] Backend
    - [x] Frontend <OPTIONAL>
8. [X] Jira Ticket [DEV-2740](https://federal-spending-transparency.atlassian.net/browse/DEV-2740):

```
```


usaspending_api/common/recipient_lookups.py
```diff
@@ -6,10 +6,9 @@ def obtain_recipient_uri(
         recipient_unique_id,
         parent_recipient_unique_id=None,
         is_parent_recipient=False):
-    # No need to create a hash when the parent recipient does not exist
-    if is_parent_recipient and not recipient_unique_id:
```
BrianZito: Was this parent logic not important? I didn't see it in the ticket description.

BrianZito:
How tall of an order would it be to get tests for this? Specifically, I'd like to see this added into existing integration tests, so that it will trip if the orm implementation changes and stops calling the library code.

usaspending_api/common/recipient_lookups.py
sethstoudenmier: Previously I added this logic to prevent altering the way that the functionality for the recipient hash worked and only affects the parent recipient. There is a bug where even the recipient hash runs into an issue when there is not an id and name so I opened this up to return None for the hash regardless of parent or not.

usaspending_api/awards/tests/test_awards_idv_v2.py
```diff
@@ -274,6 +402,13 @@ def test_award_endpoint_different_ids(client, awards_and_transactions):
     assert json.loads(resp.content.decode("utf-8")) == expected_response_idv
 
 
+@pytest.mark.django_db
+def test_award_endpoint_for_null_recipient_information(client, awards_and_transactions):
+    resp = client.get("/api/v2/awards/3/", content_type="application/json")
+    assert resp.status_code == status.HTTP_200_OK
+    assert json.loads(resp.content.decode("utf-8")) == expected_response_idv_without_recipient_id_and_name
```
BrianZito: Very good choice making this a variable, but it looks like a lot of that object is filler. Could we do the asserts on specific sub-objects that are important to the logic in order to make the test clearer? It would also allow us to make the objects smaller and easier to modify for new tests.
```suggestion
    assert json.loads(resp.content.decode("utf-8")).recipient == expected_response_idv_without_recipient_id_and_name.recipient
```
(note: I didn't test this suggestion. Just wanted to show an example)

usaspending_api/common/recipient_lookups.py
```diff
@@ -1,49 +1,41 @@
...
-        "parent_duns": parent_recipient_unique_id,
-        "is_parent_recipient": is_parent_recipient
-    })
-    return combine_recipient_hash_and_level(recipient_hash, recipient_level)
-
+    elif recipient_unique_id:
+        recipient_hash = fetch_recipient_hash_using_duns(recipient_unique_id)
+
+    if recipient_hash is None:
+        recipient_hash = generate_missing_recipient_hash(recipient_unique_id, recipient_name)
```
BrianZito: I'm confused: wasn't the whole point of this ticket to have the recipient hash sometimes be none?

sethstoudenmier: The first case will return None if the necessary information is not there.

usaspending_api/common/recipient_lookups.py
```diff
@@ -1,49 +1,41 @@
...
-    recipient_level = obtain_recipient_level({
-        "duns": recipient_unique_id,
-        "parent_duns": parent_recipient_unique_id,
-        "is_parent_recipient": is_parent_recipient
-    })
     return combine_recipient_hash_and_level(recipient_hash, recipient_level)
 
 
-def generate_missing_recipient_hash(recipient_name, recipient_unique_id):
+def generate_missing_recipient_hash(recipient_unique_id, recipient_name):
```
BrianZito: strange you didn't have to change any usages of this somewhere else. Did you change this order just for convenience?

sethstoudenmier: It changed the order in the above section to match the new order here. Everywhere else the id is referenced prior to the name and so I changed it for simplicity. The only place it is not is on the actually obtain function at the top, but that would cause additional changes outside of that file that I did not want to worry about for this fix.

usaspending_api/common/recipient_lookups.py
```diff
@@ -1,49 +1,41 @@
...
-    recipient_hash = generate_missing_recipient_hash(recipient_name, recipient_unique_id)
-    recipient_level = obtain_recipient_level({
-        "duns": recipient_unique_id,
-        "parent_duns": parent_recipient_unique_id,
-        "is_parent_recipient": is_parent_recipient
-    })
-    return combine_recipient_hash_and_level(recipient_hash, recipient_level)
-
+    elif recipient_unique_id:
+        recipient_hash = fetch_recipient_hash_using_duns(recipient_unique_id)
```
BrianZito: This logic could be arranged more cleanly.

Ex: if recpient_unique_id return fetch...
   if parent return none
   if not recipient_name return none

There's other ways to do it but it looks like you don't need that recipient_hash value to be used again if recipient_unique_id is present

tony-sappe: @BrianZito If I understand your example, there is an assumption that a pre-calculated recipient hash will always be found by the DUNS. This isn't true 100% of the time, especially in testing environments

BrianZito: Ok, I didn't take that into account. In that case, I would group the attempt to get hash from duns with the lines below that catch substitute with generate_missing_recipient_hash (which I assume never returns none), and put the special cases above by themselves. The recipient_hash variable is really only used to mark if the dun fetch worked, which confused me the first few times I looked at this.

BrianZito: new Ex:
if yada yada case 1:
return none because we don't have enough info

duns_fetch = try_to_fetch_on_duns
if not duns_fetch:
do backup thing

tony-sappe: In that recent example it is possible to search RecipientLookup with a `null` in `recipient_unique_id`, which won't provide desirable behavior.

Adding the if statement back in, it's exactly like the code in the PR

PR #1924, opened 2019-07-11
Mod/Dev-2923: Executive Compensation Refactor

sethstoudenmier:
**Description:**
New executive compensation fields (name and amount for five executives) were added to FABS and FPDS tables. This story is refactoring the current work surrounding executive compensation.  

**Technical details:**
This story is broken down into multiple parts that were split between Tony and me:
* Migration to add new fields to FPDS and FABS tables
* Migration to drop legal entity officers table
* Update FPDS/FABS loaders
* Write script to backfill executive compensation from broker for both FPDS and FABS
* Update /v2/awards endpoint to pull executive compensation data from new location

Since this was a refactor current functionality should not change.

    - [X] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [X] Jira Ticket [DEV-2923](https://federal-spending-transparency.atlassian.net/browse/DEV-2923):

```
Unit and Integration tests were only updated to test that the executive compensation data
was actually being pulled. Previously the tests expected an empty response which did not
test the actual functionality.
```


usaspending_api/awards/models/transaction_fabs.py
```diff
@@ -93,6 +93,16 @@ class TransactionFABS(models.Model):
...
+    officer_1_name = models.TextField(null=True, blank=True)
+    officer_1_amount = models.DecimalField(max_digits=23, decimal_places=2, blank=True, null=True)
+    officer_2_name = models.TextField(null=True, blank=True)
+    officer_2_amount = models.DecimalField(max_digits=23, decimal_places=2, blank=True, null=True)
+    officer_3_name = models.TextField(null=True, blank=True)
+    officer_3_amount = models.DecimalField(max_digits=23, decimal_places=2, blank=True, null=True)
+    officer_4_name = models.TextField(null=True, blank=True)
+    officer_4_amount = models.DecimalField(max_digits=23, decimal_places=2, blank=True, null=True)
+    officer_5_name = models.TextField(null=True, blank=True)
+    officer_5_amount = models.DecimalField(max_digits=23, decimal_places=2, blank=True, null=True)
```
BrianZito: Is there any reason we don't normalize this and the fields in fpds to a shared object?

BrianZito: In hindsight, I can think of a couple good reasons, probably the first one being DAIMS rules.

usaspending_api/database_scripts/job_archive/backfill_per_transaction_exec_comp.py
```diff
@@ -0,0 +1,322 @@
...
+        update_query = build_spending_update_query(transaction_sql, broker_data)
+        with Timer() as t:
+            update_cursor.execute(
+                update_query,
+                [col for row in broker_data for col in row]
+            )
+        row_count = update_cursor.rowcount
+        progress = determine_progress()
+        print(
+            "[{} - {:.2%}] {:,} => {:,}: {:,} rows updated in {} with an estimated remaining run time of {}"
```
BrianZito: ?

usaspending_api/awards/models/transaction_fabs.py
tony-sappe: Django almost supports shared object in concept but in practice it requires class inheritance. It might be a confusing process and the final state won't be that clean.

Personally I like having all of the DB columns in one list even if they are sometimes repeated.

usaspending_api/database_scripts/job_archive/backfill_per_transaction_exec_comp.py
```diff
@@ -0,0 +1,322 @@
...
+            _min = min_id
+            while _min <= max_id:
+                _max = min(_min + CHUNK_SIZE - 1, max_id)
+                broker_fabs_data = run_broker_select_query(BROKER_FABS_SELECT_SQL)
+                if broker_fabs_data:
+                    updated_row_count = run_spending_update_query(SPENDING_FABS_UPDATE_SQL, "FABS", broker_fabs_data)
+                    fabs_row_count += updated_row_count
+                else:
+                    print_no_rows_to_update("FABS")
+                _min = _max + 1
```
BrianZito: min is higher than max?

sethstoudenmier: It is a trick I took from some work that Kirk did with a large update. The _min and _max are essentially a sliding window over the range of ids between min and max. That is why above the _max is updated to be the _min + CHUNK_SIZE + 1

kbard: Oh lawd... never steal ideas from me... they always lead to ruin.

BrianZito:
```suggestion
               # move to next chunk
                _min = _max + 1
```
Not important, just an idea

kbard: Actually, in newer throwaway code I've abstracted away the iterating of the ids in a generator.  I'm not saying do this since this is one-time-use code, but just an FYI thing.

https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/database_scripts/job_archive/reset_awards_subaward_rollups.py#L212-L217

usaspending_api/awards/models/transaction_fabs.py
kbard: Yeah.  I go back and forth on this myself.  On the one hand, using mixins makes the core of the models more concise and practically guarantees conformity.  On the other hand, now I have to go tracing through a bunch of mixins and super classes to figure out what fields a model is supposed to have and what they look like.  Plus you almost always end up needed to make minor field tweaks which causes a proliferation of the original mixins with minor variances.

usaspending_api/database_scripts/job_archive/backfill_per_transaction_exec_comp.py
```diff
@@ -0,0 +1,322 @@
...
+        high_comp_officer5_amount is not null or
+        high_comp_officer5_full_na is not null
+    )
+"""
+
+SPENDING_FABS_UPDATE_SQL = """
+update
+    transaction_fabs as fabs set
+        officer_1_name = broker.high_comp_officer1_full_na,
+        officer_1_amount = cast(broker.high_comp_officer1_amount as double precision),
```
kbard: So these `double precision`s are technically incorrect.  They probably won't cause any problems since all of our numbers have only two decimal places, but you should get in the habit of using `numeric` when dealing with financial values.

https://www.postgresql.org/docs/10/datatype-numeric.html

- numeric - user-specified precision, **exact**
- double precision - variable-precision, **inexact**

usaspending_api/database_scripts/job_archive/backfill_per_transaction_exec_comp.py
sethstoudenmier: I can make this change to improve readability and then go the route of Kirk's abstracted _min and _max id generator in any future throwaway scripts.

usaspending_api/database_scripts/job_archive/backfill_per_transaction_exec_comp.py
sethstoudenmier: I will make this change in my next commit along with the comment Brian mentioned above.

BrianZito: If a budget item were large enough, it might make double format imprecise even to the second decimal point. We might want to look at this more broadly for things like aggregations where our multi-trillion dollar budget might actually cause problems.

kbard:
These two SQL files still reference `references_legalentityofficers`:

- `purge_legal_entities.sql`
- `restock_exec_comp.sql`

usaspending_api/database_scripts/job_archive/backfill_per_transaction_exec_comp.py
tony-sappe: We have standardized on `numeric` for the stored values and the aggregations. What did you want to look into?

BrianZito: Mostly if we have standardized on numeric for the stored values and the aggregations. I know I haven't been checking so far.

BrianZito: I looked at a couple of our DB tables, and it looks like everything is numeric.

BrianZito:
I'm not merging since your description leaves slots for other reviewers

BrianZito:
aaaand Kirk merged.

PR #1931, opened 2019-07-16
DEV-2116 HOTFIX: Update to long running SQL query

kbard:
This hotfix is designed to resolve an issue with a sub-optimal query that was taking entirely too long to complete.  Has been tested on both DEV and STAGING and runs significantly faster than the previous version.

### DIFF
--- a/usaspending_api/awards/management/commands/load_subawards_sql/150_reset_unlinked_award_subaward_fields.sql
+++ b/usaspending_api/awards/management/commands/load_subawards_sql/150_reset_unlinked_award_subaward_fields.sql
@@ -12,8 +12,13 @@ set
     total_subaward_amount = null,
     subaward_count = 0
 
+from
+    awards a
+    left outer join subaward s on s.award_id = a.id
+
 where
-    id not in (select award_id from subaward where award_id is not null) and (
-        total_subaward_amount is not null or
-        subaward_count is distinct from 0
-    );
+    s.award_id is null and (
+        a.total_subaward_amount is not null or
+        a.subaward_count is distinct from 0
+    ) and
+    awards.id = a.id;


usaspending_api/awards/management/commands/load_subawards_sql/150_reset_unlinked_award_subaward_fields.sql
```diff
@@ -12,8 +12,13 @@ set
     total_subaward_amount = null,
     subaward_count = 0
 
+from
+    awards a
+    left outer join subaward s on s.award_id = a.id
```
BrianZito: Is the fundamental difference that this filters on the join instead of the where?

usaspending_api/awards/management/commands/load_subawards_sql/150_reset_unlinked_award_subaward_fields.sql
```diff
@@ -12,8 +12,13 @@ set
...
+
 where
-    id not in (select award_id from subaward where award_id is not null) and (
-        total_subaward_amount is not null or
-        subaward_count is distinct from 0
-    );
+    s.award_id is null and (
```
BrianZito:
```suggestion
    s.award_id is null 
    and (
```

usaspending_api/awards/management/commands/load_subawards_sql/150_reset_unlinked_award_subaward_fields.sql
```diff
@@ -12,8 +12,13 @@ set
...
+        a.total_subaward_amount is not null or
+        a.subaward_count is distinct from 0
+    ) and
```
BrianZito:
```suggestion
    ) 
    and
```

usaspending_api/awards/management/commands/load_subawards_sql/150_reset_unlinked_award_subaward_fields.sql
BrianZito: Actually, won't this just select every award, since there will always be a sub-award which isn't one of that award's subs?

kbard:
@BrianZito,

I'll attempt to answer all of your questions in a single post.

The goal of this query is to ensure that awards with no subawards have their default rollup values (null/0).  Awards can fall out of sync if all of their subawards are deleted (which can and does happen on occasion).  This piece of code was sparked by observations of our actual data where we had quite a few incorrect subaward rollup values for awards.

The previous query used a `NOT IN` to attempt to achieve this.  I knew `NOT IN` was usually frowned upon in Postgres but testing showed it not to be a problem, so I didn't bother further optimization.  Once we started having problems with this script, I rewrote it to what you see here now.

```
from
    awards a
    left outer join subaward s on s.award_id = a.id
```

The `from` is going to return all awards and only those subawards that have a matching `award_id`.

```
    s.award_id is null
```

The first bit of the `where` is immediately going to filter out awards where we have subawards.  This leaves us with awards that have not been linked to a subaward.

```
        a.total_subaward_amount is not null or
        a.subaward_count is distinct from 0
```

The next bit is going to filter out awards where the subaward rollup pieces are not already set to their defaults.

```
    awards.id = a.id;
```

The final bit is then going to only update awards where:
- there is no subaward
- and their subaward rollup amounts are not already set to the default

Which should match the expected behavior.

BrianZito:
I still feel like that join is too inclusive, but I'm running some queries now to make sure I'm not just making a fool of myself.

BrianZito:
Running `select count(*) from  awards aa where aa.id in (select a.id from awards a left outer join subaward s on s.award_id = a.id where s.id is null);` vs `select count(*) from  awards aa where aa.id not in (select a.id from awards a left outer join subaward s on s.award_id = a.id where s.id is null);` in my local returned 154k results that would be wiped, and 500 that would not. That's not all of them like I expected, but do those densities make sense to you? I don't have a sense of the nature of the data.

kbard:
Unfortunately, that's not going to be a good comparison.  

There seemed to be some confusion as to what the default rollup values should be.  One script nulled everything and the other zeroed the count and nulled the dollar figure.  We ran a pre-script that reset those values to the Django default which was zero for count and null for dollar figure.  Unless you run that pre-script, your count is going to show a much higher count than it should.

But yes, those figures are very close to what I'm seeing in my developer database (which I recently restored and, as such, does not have the new subaward stuff in it yet).

BrianZito:
My edit suggestions stand though.

PR #1942, opened 2019-07-25
Executive compensation is pulled from the latest transaction with data

BrianZito:
**Description:**
Modifies EC logic to provide data whenever possible.

**Technical details:**

    - [ ] Backend
8. [x] Jira Ticket [DEV-3039](https://federal-spending-transparency.atlassian.net/browse/DEV-3039):

```
```


usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -41,6 +41,9 @@ def construct_assistance_response(requested_award_dict):
         return None
     response.update(award)
 
+    # TODO: Add this to the response and update the contract.
```
sethstoudenmier: Should this be a TODO? I know you added the ability to get executive compensation for either FPDS or FABS, but are we sure that this will get added to FABS? If not, then I would be worried that this comment might get confused for expected behavior not implemented.

BrianZito: The impression I got from talking to @tony-sappe was that we know we want to add it at some point, but I have not seen any requirements for it in the backlog. Tony, can you clarify?

tony-sappe: @BrianZito did you talk with @ebdabbs about updating the assistance response in this PR instead of the future?

tony-sappe: @BrianZito please update the assistance response to include executive compensation details.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -41,6 +41,8 @@ def construct_assistance_response(requested_award_dict):
         return None
     response.update(award)
 
+    response["executive_details"] = create_officers_object(award, FABS_ASSISTANCE_FIELDS, "fabs")
```
tony-sappe: Wouldn't it make more sent to include in the list of `response["<key>"]` lines? What is the justification for segregating at the top?

BrianZito: Consistency with the other responses. It struck me as strange as well, but I assumed it was better to minimize changes once in flight.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -233,15 +238,20 @@ def create_place_of_performance_object(db_row_dict):
     )
 
 
-def create_officers_object(db_row_dict):
+def create_officers_object(award, mapper, transaction_type):
+
+    transaction = fetch_latest_ec_details(award["id"], mapper, transaction_type)
```
tony-sappe: If the only field necessary is the award.id field, would it be better for the parameter to be the award PK value like `fetch_fabs_details_by_pk()`?

BrianZito: It's very subjective, but I would argue that taking in the whole award object follows single responsibility principle a little better by abstracting the exact logic to deciding what transaction maps to what award. On a practical level though I don't think it matters. I'm happy to change it for consistency.

BrianZito:
The structure of the EC object didn't change. Do you want examples to show that?

usaspending_api/awards/v2/data_layer/orm.py
BrianZito: Sorry, I answered the wrong question before. I seperated this line because it no longer makes use of the transaction object like everything else does (since under current business logic the "latest transaction" can mean something different for EC and other fields), so that future devs don't glance over this line and assume that if just pulls a field from the transaction variable like everything else.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -134,13 +137,16 @@ def construct_idv_response(requested_award_dict):
...
+    response["executive_details"] = create_officers_object(award, mapper, "fpds")
+
     transaction = fetch_fpds_details_by_pk(award["_trx"], mapper)
 
     response["parent_award"] = parent_award
     response["parent_generated_unique_award_id"] = parent_award["generated_unique_award_id"] if parent_award else None
-    response["executive_details"] = create_officers_object(transaction)
     response["latest_transaction_contract_data"] = transaction
     response["funding_agency"] = fetch_agency_details(response["_funding_agency"])
+    print(transaction)
```
kbard: This seems like a debug line.  Do we want to keep it for rollout?

emily-brents: please remove this

tony-sappe:
> The structure of the EC object didn't change. Do you want examples to show that?

@BrianZito It should be easy to find IDVs which return different Executive Compensation values using the new logic. Also including notes on the expected performance of different scenarios (awards with many transactions without EC values, awards with few transactions, ...) is helpful.

PR #1946, opened 2019-07-27
Reorganize the API contracts

sethstoudenmier:
**Description:**
Following up on discussions from the API contract meetings, the contracts have been reorganized to match their URL instead of the page they effect. While making these changes I tried to avoid making any edits to the actual contents that were not needed for the purpose of reorganization. ~I still need to fix the Dredd tests, but wanted to go ahead and create a PR for feedback.~

**Technical details:** 
Here is a list of the notable changes:
*Note: This is slightly out of date with the most recent round of cleaning up.*

* /contracts/awards/AwardProfile.md was moved to /contracts/v2/award.md so that the path would read similar to the route
* /contracts/awards/Transactions.md was moved to /contracts/v2/transactions.md so that the path would read similar to the route
* /contracts/downloadCenter/DownloadCenter.md was split between /contracts/v2/download/accounts.md and /v2/references/data_dictionary.md
* /contracts/federal_accounts/FederalAccountProfile.md and FederalAccounts.md were consolidated into one file /contracts/v2/federal_accounts.md
* /contracts/federal_oblications/ObjectClass.md was moved to /contracts/v2/federal_obligations.md
* /contracts/recipient/RecipientProfile.md was split between multiple files:
   * /contracts/v2/recipient/duns.md (for the GET and POST)
   * /contracts/v2/recipient/children.md
   * /contracts/v2/search/new_award_over_time.md
* /contracts/search/AdvancedSearch.md was split between multiple files:
   * /contracts/v2/search/spending_by_award
   * /contracts/v2/search/spending_by_category
   * /contracts/v2/search/spending_over_time
   * /contracts/v2/download/count
   * /contracts/v1/references/filter
   * /contracts/v1/references/hash
* /contracts/spendingExplorer/SpendingExplorer.md was moved to /contracts/v2/spending.md
* /contracts/state/StateProfile.md was split between multiple files:
   * /v2/recipient/state.md (for both GET) ***This route is an odd case since we pass in two different variables following "/state/"*
   * /v2/recipient/state/awards.md
* /contracts/subawards/SubAwards.md was moved to /contracts/v2/subawards.md

To aid me in reviewing the before and after I made four different ".txt" files of the contract directory with and without the endpoints.

[Old_API_Contract_Structure_With_Endpoints.txt](https://github.com/fedspendingtransparency/usaspending-api/files/3437623/Old_API_Contract_Structure_With_Endpoints.txt)
[Old_API_Contract_Structure.txt](https://github.com/fedspendingtransparency/usaspending-api/files/3437620/Old_API_Contract_Structure.txt)
[New_API_Contract_Structure_With_Endpoints.txt](https://github.com/fedspendingtransparency/usaspending-api/files/3437621/New_API_Contract_Structure_With_Endpoints.txt)
[New_API_Contract_Structure.txt](https://github.com/fedspendingtransparency/usaspending-api/files/3437622/New_API_Contract_Structure.txt)

    - [X] Backend
    - [X] Frontend
8. N/A Jira Ticket 

```
~The documentation on the API Contracts should probably be updated to match this new
structure. For example, most endpoints have a section that calls out the endpoints as 
being used for specific pages. This is useful information, but with it being 
now scattered around it has lost some of its usefulness in my opinion. I can make 
the changes as a result of this PR but was uncertain the best path forward for that.~

UPDATES: Made a pass-through and updated formatting of .md files to be similar to 
each other trying to remove unnecessary headers. Also, all Dredd tests were passing 
locally at the time of me making the most recent commits.

UPDATES 8/14/19: Made another pass with the recent comments on the PR. After 
consideration and reaching out to get opinions, I have added backticks around all 
items that would be represented as a string by JSON. This way we can avoid the 
issues that special characters cause for Dredd and also create a sense of consistency 
across our API Contracts.

UPDATES 8/21/19: Added a template for future API contract reference, updated the README
to match some of the decisions with the new approach, and made another pass of cleanup.

```


usaspending_api/api_contracts/contracts/v2/federal_accounts.md
```diff
@@ -37,6 +37,30 @@ This endpoint returns a list of federal accounts, their number, name, managing a
         + `next`: 2 (number, required, nullable)
         + `fy`: `2018` (string, required)
         + `results` (array[FederalAccountListing], fixed-type)
+ 
+# Federal Account Profile
```
tony-sappe: This should be in its own file.

sethstoudenmier: I was thinking that even though you pass a variable "accountNumber" on the endpoint after "federal_accounts" that it is essentially just the GET for "/api/v2/federal_accounts" that returns a specific one instead of a list. Creating a file called /v2/federal_accounts/account_number.md would lead me to believe that there is an actual route with that name and not that "account_number" is a variable.

usaspending_api/api_contracts/contracts/v1/references/filter.md
```diff
@@ -0,0 +1,131 @@
...
+
++ Response 200 (application/json)
+    + Attributes
+        + hash : `5703c297b902ac3b76088c5c275b53f9` (required, string)
+
+# Data Structures
+
+## Filter Objects
+### FilterObject (object)
++ `keywords` : poptarts (optional, array[string])
```
ebdabbs: The request is formatted like the filter object in the `v1/references/hash/` contract response. 
```
"awardAmounts": {},
"awardType": [],
"extentCompeted": [], ...
```

usaspending_api/api_contracts/contracts/v2/download/count.md
```diff
@@ -0,0 +1,132 @@
...
+    Availability Type Code - X or null
++ `main` (required, string)
+    Main Account Code - four digits
++ `sub` (optional, string, nullable)
+    Sub-Account Code - three digits
+
+### FilterObjectAwardTypes (array)
+List of filterable award types
+
+### Sample
```
ebdabbs: Are `Sample` and `Default` data structures? Are they being used in this file?

usaspending_api/api_contracts/contracts/v2/recipient/children.md
```diff
@@ -0,0 +1,39 @@
+FORMAT: 1A
+HOST: https://api.usaspending.gov
+
+# Recipient Profile
+
+This endpoint is used to power USAspending.gov's recipient profile pages. This data can be used to visualize the government spending that pertains to a specific recipient.
+
+# Group Profile Page
```
ebdabbs: Since we're moving each endpoint to a separate file, we could get rid of the "Group" heading and keep just a  heading/ description specific to this file. (Similar elsewhere)

sethstoudenmier: Agreed. I wasn't certain on what to do. Can do that on my next pass.

usaspending_api/api_contracts/contracts/v2/federal_accounts.md
sethstoudenmier: @tony-sappe What are your thoughts on my response to your comment?

usaspending_api/api_contracts/contracts/v1/references/filter.md
sethstoudenmier: Neither of them actually use the FilterObject mentioned in the "Data Structures" section?

usaspending_api/api_contracts/contracts/v2/download/count.md
sethstoudenmier: I believe they are used as part of the `FilterObjectAwardTypes`. I am going to change the header value level to make it more obvious.

usaspending_api/api_contracts/contracts/v1/references/filter.md
BrianZito: This is an old version of the API

usaspending_api/api_contracts/contracts/v2/autocomplete/city.md
```diff
@@ -0,0 +1,30 @@
...
++ Response 200 (application/json)
+    + Attributes (object)
+        + `count`: `10` (required, number)
+        + `results` (required, array[AutocompleteCityResult], fixed-type)
+
+# Data Structures
+
+## AutocompleteCityResult (object)
++ `city_name`: `Springfield` (required, string)
++ `state_code`: `VA` (required, string, nullable)
```
BrianZito:
```suggestion
+ `state_code`: `SC` (required, string, nullable)
```
Might confuse folks if we say SC returns VA

usaspending_api/api_contracts/contracts/v2/financial_balances/agencies.md
```diff
@@ -1,30 +1,28 @@
...
 These endpoints are used to power USAspending.gov's agency profile pages. This data can be used to better understand the different ways that a specific agency spends money.
 
-## Financial Balances [/api/v2/financial_balances/agencies/{?fiscal_year,funding_agency_id}]
+## Get Financial Balances [GET /api/v2/financial_balances/agencies/{?fiscal_year,funding_agency_id}]
 
 This endpoint returns aggregated balances for a specific government agency in a given fiscal year.
 
 + Parameters
-    + fiscal_year: 2017 (required, number)
+    + `fiscal_year`: `2017` (required, number)
```
BrianZito: Do we need the tics if these are numbers?

sethstoudenmier: No, I can go through and remove the tics for everything that is not a string. I was going to go through and try to make the use of tics consistent but noticed that we used them differently all throughout the API contracts. Personally, think it helps readability to have them formatted as:

+ Parameters
   + name: `value` (required, type)
   + another_name (optional, type, nullable)

BrianZito: As of the last meeting we've set standard for the keys but not the values, but keep in mind some of these pages are V1 endpoints that we aren't really maintaining anymore.

usaspending_api/api_contracts/contracts/v2/recipient/children.md
```diff
@@ -0,0 +1,35 @@
...
+# Data Structures
+
+## ChildRecipient (object)
++ name: Child of ABC Corporation (required, string, nullable)
+    Name of the child recipient. `null` if the child recipient's name is not provided.
++ duns: `001006360` (required, string, nullable)
+    Child recipient's DUNS. `null` if the child recipient's DUNS is not provided.
++ `recipient_id`: `0036a0cb-0d88-2db3-59e0-0f9af8ffef57-C` (required, string)
+    A unique identifier for the child recipient.
++ `state_province`: New Jersey (required, string, nullable)
```
BrianZito:
```suggestion
+ `state_province`: `New Jersey` (required, string, nullable)
```

usaspending_api/api_contracts/contracts/v2/recipient/state/awards.md
```diff
@@ -0,0 +1,31 @@
...
+
+This endpoint is used to power USAspending.gov's state profile pages. This data can be used to visualize the government spending that occurs in a specific state or territory.
+
+## Get State Award Breakdown [GET /api/v2/recipient/state/awards/{fips}/{?year}]
+
+This endpoint returns the award amounts and totals, based on award type, of a specific state or territory, given its USAspending.gov `id`.
+
++ Parameters
+
+    + fips: 51 (required, string)
```
BrianZito:
```suggestion
    + fips: `51` (required, string)
```
Feel free to ask around and not take these if more people think it's better the other way. This is not a hill I'll die on.

usaspending_api/api_contracts/contracts/v2/recipient/state/awards.md
```diff
@@ -0,0 +1,31 @@
...
+        The FIPS code for the state you want to view. You must include leading zeros.
+    + year: 2017 (optional, string)
```
BrianZito:
```suggestion
    + year: `2017` (optional, string)
```

usaspending_api/api_contracts/contracts/v2/recipient/state/awards.md
```diff
@@ -0,0 +1,31 @@
...
+        The fiscal year you would like data for. Use `all` to view all time or `latest` to view the latest 12 months.
+        
++ Response 200 (application/json)
+
+    + Attributes (array[StateBreakdown], fixed-type)
+
+# Data Structures
+
+## StateBreakdown (object)
++ type: contracts (required, string)
```
BrianZito:
```suggestion
+ type: `contracts` (required, string)
```

usaspending_api/api_contracts/contracts/v2/search/spending_by_category.md
```diff
@@ -0,0 +1,170 @@
...
++ `recipient_scope` (optional, enum[string])
+    + domestic
+    + foreign
++ `recipient_locations` (optional, array[LocationObject])
++ `recipient_type_names`: `category_business` (optional, array[string])
+    See options at https://github.com/fedspendingtransparency/usaspending-api/wiki/Recipient-Business-Types
++ `award_type_codes` (optional, FilterObjectAwardTypes)
+    See use at
+    https://github.com/fedspendingtransparency/usaspending-api/wiki/Search-Filters-v2-Documentation#award-type
++ `award_ids`: SPE30018FLGFZ, SPE30018FLJFN (optional, array[string])
```
BrianZito:
```suggestion
+ `award_ids`: `SPE30018FLGFZ`, `SPE30018FLJFN` (optional, array[string])
```

usaspending_api/api_contracts/contracts/v2/search/spending_by_category.md
```diff
@@ -0,0 +1,170 @@
...
++ `award_amounts` (optional, array[AwardAmounts])
++ `program_numbers`: `10.331` (optional, array[string])
++ `naics_codes`: 311812 (optional, array[string])
++ `psc_codes`: 8940, 8910 (optional, array[string])
```
BrianZito:
```suggestion
+ `psc_codes`: `8940`, `8910` (optional, array[string])
```

usaspending_api/api_contracts/contracts/v2/search/spending_over_time.md
```diff
@@ -0,0 +1,160 @@
...
++ `fiscal_year`: `2018` (required, string)
++ `quarter`: `1` (optional, string)
+    Excluded when grouping by `fiscal_year` or `month`.
++ `month`: `1` (optional, string)
+    Excluded when grouping by `fiscal_year` or `quarter`.
+
+
+## Filter Objects
+### FilterObject (object)
++ `keywords` : poptarts (optional, array[string])
```
BrianZito:
```suggestion
+ `keywords` : `transport` (optional, array[string])
```
Ross has said that he's okay with "silly" searches so long as they return results, but it might be best to avoid brand names just in case.

usaspending_api/api_contracts/contracts/v2/search/spending_over_time.md
```diff
@@ -0,0 +1,160 @@
...
++ `recipient_scope` (optional, enum[string])
+    + domestic
+    + foreign
++ `recipient_locations` (optional, array[LocationObject])
++ `recipient_type_names`: `category_business` (optional, array[string])
+    See options at https://github.com/fedspendingtransparency/usaspending-api/wiki/Recipient-Business-Types
++ `award_type_codes` (optional, FilterObjectAwardTypes)
+    See use at
+    https://github.com/fedspendingtransparency/usaspending-api/wiki/Search-Filters-v2-Documentation#award-type
++ `award_ids`: SPE30018FLGFZ, SPE30018FLJFN (optional, array[string])
```
BrianZito:
```suggestion
+ `award_ids`: `SPE30018FLGFZ`, `SPE30018FLJFN` (optional, array[string])
```

usaspending_api/api_contracts/contracts/v2/autocomplete/accounts/a.md
```diff
@@ -3,16 +3,16 @@ HOST: https://api.usaspending.gov
 
 # Availability Type Code [/api/v2/autocomplete/accounts/a/]
 
-This endpoint powers the USAspending.gov Availability Type Code (A) autocomplete in the Advanced Search -> Program Source -> Treasury Account filter.  It returns the list of potential Availability Type Codes narrowed by other components supplied in the Treasury Account filter.
+This endpoint powers the USAspending.gov Availability Type Code (A) autocomplete in the Advanced Search -> Program Source -> Treasury Account filter.
 
-## POST
+## List of potential Availability Type Codes [POST /api/v2/autocomplete/accounts/a/]
```
kbard: So this is a debate for the larger team (@tony-sappe, @ebdabbs, @BrianZito, etc), but should we be putting the description in the method title or on the line below like other descriptive text?  For example:

```
## [POST]

List of potential availability type codes.

...
```

I get the desire to have a description per method since a contract can contain multiple.  Ours usually don't (unless you count OPTIONS), but on an updatable site it's very common to support multiple methods per URL.

Also, does it make sense to repeat the URL?  I was trying to move away from that since the URL is usually specified at the top of contract.

usaspending_api/api_contracts/contracts/v2/financial_balances/agencies.md
kbard: Yeah.  Don't get me started on backticks.  Everyone seems to have a different opinion on how we should use them.  At a minimum, [they need to enclose attributes or values that contain keywords or special characters](https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#6-reserved-characters--keywords).

usaspending_api/api_contracts/contracts/v2/autocomplete/accounts/a.md
sethstoudenmier: I am fine to not repeat the URL the only issue I noticed is on a route where there are parameters in the URL. It might get confusing to have a GET /this/endpoint/{var1}/
and a GET /this/endpoint/ on the same page without specifying somehow that one is a specific entity and the other is a list.

usaspending_api/api_contracts/contracts/v2/autocomplete/accounts/main.md
```diff
@@ -56,7 +56,7 @@ Each component listed here may be omitted, null, or a string value.  If omitted,
     Ending Period of Availability (4 characters). TAS only.
 + `a` (optional, string, nullable)
     Availability Type Code (1 character). Will either be 'X' or null. TAS only.
-+ `main` (optional, string, nullable)
++ `main`: `30` (optional, string, nullable)
```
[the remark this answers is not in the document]
sethstoudenmier: It probably was intentional, but they all kinda started to blur together a little. Will make sure this gets fixed.

usaspending_api/api_contracts/contracts/v2/autocomplete/accounts/a.md
kbard: Have we tested Dredd using different method URLs in the same contract where we have the primary URL listed at the top?  Is it able to figure which to call correctly?

Another option would be to only include the method URL if it differs from the document level URL.

Anyhow, the duplicated URL isn't a showstopper.

usaspending_api/api_contracts/contracts/v2/financial_balances/agencies.md
tony-sappe: This README doc needs work, but #5 and #6 mentions when to use backticks:
https://github.com/fedspendingtransparency/usaspending-api/tree/dev/usaspending_api/api_contracts

usaspending_api/api_contracts/contracts/v2/federal_accounts.md
tony-sappe: I don't have a good response to the URL/filename question. I suppose we need a convention to denote variables in the URL.

The obvious differentiator for me is the endpoints and responses are different which makes me believe they should be different files.

usaspending_api/api_contracts/contracts/v2/autocomplete/accounts/a.md
sethstoudenmier: Per a comment Tony had further up and this comment I have created a separate file if we use a variable in the URI of the endpoint, but not if we use a variable as query parameters. This way we can use simple GET, POST, etc. names for Header 2 with a short description underneath.

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
```diff
@@ -0,0 +1,219 @@
...
++ `date_type` (optional, enum[string])
+    + Members
+        + `action_date`
+        + `last_modified_date`
+
+### LocationObject (object)
++ `country`: `USA` (required, string)
++ `state`: `VA` (optional, string)
++ `county` (optional, string)
++ `city` (optional, string)
```
tony-sappe: `zip` and `district` are missing in many LocationObject definitions

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
```diff
@@ -0,0 +1,219 @@
...
++ `recipient_scope` (optional, enum[string])
+    + Members
+        + `domestic`
+        + `foreign`
++ `recipient_locations` (optional, array[LocationObject], fixed-type)
++ `recipient_type_names`: `category_business` (optional, array[string])
+    See options at https://github.com/fedspendingtransparency/usaspending-api/wiki/Recipient-Business-Types
++ `award_type_codes` (optional, FilterObjectAwardTypes)
+    See use at
+    https://github.com/fedspendingtransparency/usaspending-api/wiki/Search-Filters-v2-Documentation#award-type
```
tony-sappe: This page no longer exists

usaspending_api/api_contracts/contracts/v2/references/data_dictionary.md
```diff
@@ -0,0 +1,38 @@
...
+
+This endpoint powers USAspending.gov's data dictionary page.
+
+## GET
+
+This endpoint returns data corresponding to the latest data dictionary csv file.
+
++ Response 200 (application/json)
+    + Attributes
+        + `document` (DataDictionary)
```
tony-sappe: Typically, there isn't a benefit defining a data structure as the entire response. It is beneficial when the response is an array of objects, or could be reused.

In this case, the entire DataDictionary object could be listed as the Response. This doesn't need to change for this PR, but it shouldn't become the standard.

usaspending_api/api_contracts/contracts/search/AdvancedSearch.md
```diff
@@ -1,397 +0,0 @@
-FORMAT: 1A
```
tony-sappe: ⭐️ ⭐️ ⭐️

usaspending_api/api_contracts/README.md
```diff
@@ -8,11 +8,13 @@ All new API endpoint documentation must be in markdown that conforms to the [API
 See [Tools for Testing and Mocking](tools-for-contracts.md)
 
 ## Guidelines to Remember when Writing Contracts
-1. One endpoint per markdown file
```
tony-sappe: We are still writing one endpoint per file

usaspending_api/api_contracts/README.md
```diff
@@ -8,11 +8,13 @@ All new API endpoint documentation must be in markdown that conforms to the [API
...
-1. Markdown filenames should be in "CamelCase"
+1. Markdown filenames should be in snake_case
+1. Endpoints that are different HTTP methods can share one markdown file
 1. Folder structure should mimic the URL and use "snake_case"
-    - Don't include "api/v2" in the folder structure
```
tony-sappe: Perhaps alter the note to say mimic the URL path after `api/`

usaspending_api/api_contracts/template.md
```diff
@@ -0,0 +1,42 @@
+FORMAT: 1A
```
tony-sappe: yay. Nice start

PR #1955, opened 2019-08-05
Mod/Dev-2973: Add federal accounts for award to download

sethstoudenmier:
**Description:**
Add the Federal Account to the award and transaction download. Added so that the column will display similar to "075-0512; 075-0580; 075-8004" where "<agency_identifier>-<main_account_code>".

Screenshot showing the new column and an entry:
Screen Shot 2019-08-05 at 11 04 21 AM

**Technical details:**
Representing annotations for downloads as an entry into the ordered dictionary with (<alias_name>, None). The None allows for easy removal when gathering all of the query paths, but allows for the correct location to be used with respect to order. The annotations are created as part of a function that is unique to each table. 

The hope is that any future annotations can be added to the annotations function per table and an entry with a matching alias can be added to the ordered dictionary to achieve what is needed.

    - [X] Backend
8. [X] Jira Ticket [DEV-2973](https://federal-spending-transparency.atlassian.net/browse/DEV-2973):

```
```

### DIFF
--- a/usaspending_api/download/helpers/download_annotations_helper.py
+++ b/usaspending_api/download/helpers/download_annotations_helper.py
@@ -1,0 +1,37 @@
+from django.contrib.postgres.aggregates import StringAgg
+from django.db.models.functions import Concat
+from django.db.models import Value
+
+
+def universal_transaction_matview_annotations():
+    # transaction__award__financial_set__treasury_account__federal_account__agency_identifier
+    federal_account_query_path = "transaction__award__financial_set__treasury_account__federal_account"
+    annotation_fields = {
+        "federal_accounts_funding_this_award": StringAgg(
+            Concat(
+                "{}__agency_identifier".format(federal_account_query_path),
+                Value("-"),
+                "{}__main_account_code".format(federal_account_query_path),
+            ),
+            ";",
+            distinct=True,
+        )
+    }
+    return annotation_fields
+
+
+def universal_award_matview_annotations():
+    # transaction__award__financial_set__treasury_account__federal_account__agency_identifier
+    federal_account_query_path = "award__financial_set__treasury_account__federal_account"
+    annotation_fields = {
+        "federal_accounts_funding_this_award": StringAgg(
+            Concat(
+                "{}__agency_identifier".format(federal_account_query_path),
+                Value("-"),
+                "{}__main_account_code".format(federal_account_query_path),
+            ),
+            ";",
+            distinct=True,
+        )
+    }
+    return annotation_fields
--- a/usaspending_api/download/helpers/download_annotations_helper.py
+++ b/usaspending_api/download/helpers/download_annotations_helper.py
@@ -4,7 +4,6 @@ from django.db.models import Value
 
 
 def universal_transaction_matview_annotations():
-    # transaction__award__financial_set__treasury_account__federal_account__agency_identifier
     federal_account_query_path = "transaction__award__financial_set__treasury_account__federal_account"
     annotation_fields = {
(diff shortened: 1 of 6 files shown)


usaspending_api/download/helpers/download_annotations_helper.py
```diff
@@ -0,0 +1,35 @@
...
+            Concat(
+                "{}__agency_identifier".format(federal_account_query_path),
+                Value("-"),
+                "{}__main_account_code".format(federal_account_query_path),
+            ),
+            ";",
+            distinct=True,
+        )
+    }
+    return annotation_fields
```
BrianZito: can we rename this to "download_annotations_library" or have these stored in the dict or something? I've never been a big fan of "helper" classes, and this isn't really a utils library, it's part of the business logic.

BrianZito: I might actually advise putting it into the historical lookups file, since those two are so tightly coupled.

sethstoudenmier: I personally like the idea of keeping them separate but agree I should change the name. I think it will help as either of those two files continue to grow. They are tightly coupled, but not directly only through csv_source and csv_generation.

usaspending_api/download/lookups.py
```diff
@@ -65,6 +70,7 @@
         "assistance_data": "transaction__assistance_data",
         "filter_function": universal_transaction_matview_filter,
         "is_for_idv": False,
+        "annotations_function": universal_transaction_matview_annotations,
```
BrianZito: What if someone (possibly me) needs to add a second annotation for a different field of the same object?

sethstoudenmier: Do you mean having one for assistance_data and one for contract_data, or having the function return two annotations?

usaspending_api/download/v2/download_column_historical_lookups.py
```diff
@@ -1489,18 +1497,23 @@
...
 
 # Likewise, IDV Transactions start directly in TransactionFPDS instead of
 # UniversalTransactionView.
 query_paths["idv_transaction_history"] = {
     "d1": OrderedDict(
-        [(k, v[13:] if v.startswith("transaction__") else v) for k, v in query_paths["transaction"]["d1"].items()]
+        [
+            (k, v[13:] if v.startswith("transaction__") else v)
+            for k, v in query_paths["transaction"]["d1"].items()
+            if v is not None
```
BrianZito: @emily-brents pointed out that this might introduce risk compared to sentinel value (probably some singleton class). I'm not sure exactly how this will break on the loop unless we start deriving values for the map from functions, but it is possible.

usaspending_api/download/lookups.py
BrianZito: Nevermind, I failed to realize that these functions themselves distinguish between fields.

usaspending_api/download/v2/download_column_historical_lookups.py
sethstoudenmier: I have some ideas on how I can rework this and accomplish the same effect.

BrianZito: Total 180 here, but can we save that for a later PR? I want to branch off of this without conflicting so I would rather this branch be shorter-lived.

sethstoudenmier: Absolutely. If there are any minor issues I can fix those after we are done with this PR.

PR #1956, opened 2019-08-05
Mod/dev 1039 add fiscal year to download

BrianZito:
**Description:**
Added Fiscal Year of Action Date to downloads of Transaction and Subaward

**Technical details:**
"EXTRACT" was added to the list of keywords for derived fields. A safety was also added in case a function includes a field that isn't listed in the historical mapping (if, for example, the d1 and d2 downloads for the same thing ever had different columns)

    - [ ] Backend
4. [ N/A] Matview impact assessment completed
5. [ N/A] Frontend impact assessment completed
7. [ N/A] Appropriate Operations ticket(s) created
8. [ ] Jira Ticket [DEV-1038](https://federal-spending-transparency.atlassian.net/browse/DEV-1039):

```
```


BrianZito:
I modified some variable names in the last two commit specifically so that code climate wouldn't complain about duplicate code. I'm comfortable with this because that code is using the same column names but in two different tables. I am willing to defend the point that these should NOT be combined into a helper function, because that would involve coupling those column names.

usaspending_api/download/lookups.py
```diff
@@ -26,9 +26,12 @@
 from usaspending_api.awards.v2.filters.sub_award import subaward_download
 from usaspending_api.awards.v2.lookups.lookups import award_type_mapping
 from usaspending_api.financial_activities.models import FinancialAccountsByProgramActivityObjectClass
-from usaspending_api.download.helpers.download_annotations_helper import (
+from usaspending_api.download.helpers.download_annotation_functions import (
```
emily-brents: Is there a particular reason this file was renamed?

BrianZito: "helper" is not a very descriptive name. Besides, the file clearly isn't "helping" some function. It's a function library.

PR #1960, opened 2019-08-06
DEV-3212: Minor tweak to TAS autocomplete ticket

kbard:
**Description:**

One more minor tweak to remove an unnecessary `distinct` and change sort order.  Very minor.  Nothing should be noticeably affected.

### DIFF
--- a/usaspending_api/database_scripts/matview_generator/tas_search_matview.json
+++ b/usaspending_api/database_scripts/matview_generator/tas_search_matview.json
@@ -2,7 +2,7 @@
   "final_name": "tas_search_matview",
   "refresh": true,
   "matview_sql": [
-    "select distinct",
+    "select",
     "    min(faba.financial_accounts_by_awards_id) tas_search_id,",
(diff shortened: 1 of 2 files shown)


usaspending_api/database_scripts/matview_generator/tas_search_matview.json
```diff
@@ -2,7 +2,7 @@
   "final_name": "tas_search_matview",
   "refresh": true,
   "matview_sql": [
-    "select distinct",
+    "select",
```
BrianZito: Why isn't this nessicary? I thought that that was a key point of why the min function was okay.

kbard: The grouping/`min` achieves (roughly) the same thing as the `distinct` in this case so the `distinct` is superfluous.

BrianZito: that makes sense

PR #1968, opened 2019-08-10
Staging -> Dev Hotfix downmerge

tony-sappe:
Downmerge hotfix to update an outdated constraint on the FABS table


BrianZito:
Why is this needed?

tony-sappe:
> Why is this needed?

Every time a PR "skips" a lower env it needs to propagate down from the "higher env" branches into the "lower branches"

BrianZito:
My bad, I meant why do we need to be updating the unique together?

tony-sappe:
> My bad, I meant why do we need to be updating the unique together?

https://github.com/fedspendingtransparency/usaspending-api/pull/1966

PR #1975, opened 2019-08-15
DEV-3184: Do not return bad recipient hashes

kbard:
**Description:**

There are certain rare cases where recipients appear in the recipient lookup but for which we do not have profile information.  This ticket checks that we have profile information before returning the recipient hash and returns null if not.

    - [x] Backend
    - [ ] Frontend
8. [x] Jira Ticket [DEV-3184](https://federal-spending-transparency.atlassian.net/browse/DEV-3184):


usaspending_api/awards/v2/views/idvs/activity.py
```diff
@@ -46,10 +46,7 @@
         ca.period_of_performance_start_date,
         ca.piid,
         rl.legal_business_name                          recipient_name,
-        rl.recipient_hash || case
-            when tf.ultimate_parent_unique_ide is null then '-R'
-            else '-C'
```
BrianZito: Why was this logic removed? Didn't Seth have a PR that required this?

kbard: It was moved down into the FROM clause.

tony-sappe: It looks like it was modified and moved below, not removed

PR #1977, opened 2019-08-15
[DEV-3233] Split `universal_award_matview` into smaller models for TAS Performance

tony-sappe:
**Description:**
`universal_award_matview` **divided** into smaller matviews which in aggregate contain the same data:
- `mv_contract_award_search`
- `mv_directpayment_award_search`
- `mv_grant_award_search`
- `mv_idv_award_search`
- `mv_loan_award_search`
- `mv_other_award_search`
- `mv_pre2008_award_search`

A new view on top of those matviews is essentially the same model as `universal_award_matview`:
- `vw_award_search`

**Technical details:**
- Removed `universal_award_matview` and `summary_award_view`
- Started new naming convention for DB objects/models:
  - Singular names (accepted SQL best-practice outside of this project)
  - for Materialized Views:
    - Suffix Model name with "Matview"
    - Prefix DB object with "mv_
  - for Views:
    - Suffix Model name with "View"
    - Prefix DB object with "vw_

    - [ ] Backend
8. [x] Jira Ticket [DEV-3233](https://federal-spending-transparency.atlassian.net/browse/DEV-3233):

```
No API functionality should change other than performance
```


usaspending_api/awards/models_matviews/mv_search_award_idv.py
```diff
@@ -85,4 +85,4 @@ class UniversalAwardView(models.Model):
 
     class Meta:
         managed = False
-        db_table = "universal_award_matview"
+        db_table = "mv_search_award_idv"
```
BrianZito: Why is this just a rename from the origonal (and named "mv" instead of "my") when all the others seem to be copies?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+        self.order = json_request["order"]
+        self.limit = json_request["limit"]
+        self.page = json_request["page"]
+        self.lower_bound = (self.page - 1) * self.limit
+        self.upper_bound = self.page * self.limit + 1
+
+        if "no intersection" in filters["award_type_codes"]:
+            # "Special case": there will never be results when the website provides this value
+            return Response(
+                {"limit": self.limit, "results": [], "page_metadata": {"page": self.page, "hasNext": False}}
```
BrianZito: These key names are getting used a lot, as is the general json structure. I would suggest a factory method to return this dict given values, and using that rather than constants or duplicating stuff.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+        self.sort = json_request.get("sort") or self.fields[0]
+        if self.sort not in self.fields:
+            raise InvalidParameterException(
+                "Sort value '{}' not found in requested fields: {}".format(self.sort, self.fields)
+            )
+
+        if subawards:
+            return Response(self._handle_subawards(filters))
+
+        return Response(self._handle_awards(filters))
```
BrianZito: I would do an if/else here since current format makes it look like a guard clause rather than core logic.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def _handle_awards(self, filters):
+        """ Award logic for SpendingByAward"""
+        try:
+            obtain_view_from_award_group(filters["award_type_codes"])
+        except Exception:
+
```
BrianZito:
```suggestion
```

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+        if subawards:
+            return Response(self._handle_subawards(filters))
+
+        return Response(self._handle_awards(filters))
+
+    def _handle_awards(self, filters):
+        """ Award logic for SpendingByAward"""
+        try:
+            obtain_view_from_award_group(filters["award_type_codes"])
```
BrianZito: Nothing is set to this function's return. Are you just calling it for the side-effects?

BrianZito: There don't seem to be any (which is a very good thing, but confuses me more)

sethstoudenmier: The return is wrapped in a Response() for the POST above it.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+                        row["Award ID"] = award[id_type]
+                        break
+            results.append(row)
+
+        return {"limit": self.limit, "results": results, "page_metadata": {"page": self.page, "hasNext": has_next}}
+
+    def _handle_subawards(self, filters):
+        """ sub-award logic for SpendingByAward"""
+        # Test to ensure the award types are a proper subset of at least one group (contracts or grants)
+        if set(filters["award_type_codes"]) - set(procurement_type_mapping.keys()) and set(
```
sethstoudenmier: Personally a fan of using `set_a.difference(set_b)` instead of `set_a - set_b` because it reads more clear.

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: that's the return of `_handle_awards`, but I was talking about just `obtain_view_from_award_group`

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+                values.add(award_idv_mapping.get(field))
+
+        # Modify queryset to be ordered by requested "sort" in the request or default value(s)
+        if set(filters["award_type_codes"]) <= set(contract_type_mapping):  # contracts
+            sort_filters = [award_contracts_mapping[self.sort]]
+        elif set(filters["award_type_codes"]) <= set(loan_type_mapping):  # loans
+            sort_filters = [loan_award_mapping[self.sort]]
+        elif set(filters["award_type_codes"]) <= set(idv_type_mapping):  # idvs
+            sort_filters = [award_idv_mapping[self.sort]]
+        else:  # assistance data
```
BrianZito:
```suggestion
        elif set(filters["award_type_codes"]) <= set(assistance_type_mapping): # assistance data
```
I would add a sanity check here in case future changes to the code break this functionality. Having it default to sorting on assistance will be hard to spot in the future

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+        ) - set(assistance_type_mapping.keys()):
+            error_msg = (
+                '{{"message": "\'award_type_codes\' array can only contain types codes of one group.",'
+                '"award_type_groups": {{'
+                '"procurement": {},'
+                '"assistance": {}}}}}'
+            ).format(json.dumps(procurement_type_mapping), json.dumps(assistance_type_mapping))
+            raise UnprocessableEntityException(json.loads(error_msg))
+
+        subawards_values = list(contract_subaward_mapping.keys()) + list(grant_subaward_mapping.keys())
```
BrianZito: minor, but I feel like this should be `subawards_keys` or `subawards_fields`

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+
+        if set(filters["award_type_codes"]) <= set(procurement_type_mapping):
+            sort_filters = [contract_subaward_mapping[self.sort]]
+        elif set(filters["award_type_codes"]) <= set(assistance_type_mapping):
+            sort_filters = [grant_subaward_mapping[self.sort]]
+        else:
+            msg = "Sort key '{}' doesn't exist for the selected Award group".format(self.sort)
+            raise InvalidParameterException(msg)
+
+        if self.sort == "Award ID":
```
BrianZito:
```suggestion
        # Explictly set NULLS LAST in the ordering to encourage the usage of the indexes
        if self.sort == "Award ID":
```

usaspending_api/search/v2/views/spending_by_award_count.py
```diff
@@ -0,0 +1,77 @@
...
+    def post(self, request):
+        models = [{"name": "subawards", "key": "subawards", "type": "boolean", "default": False}]
+        models.extend(copy.deepcopy(AWARD_FILTER))
+        models.extend(copy.deepcopy(PAGINATION))
+        json_request = TinyShield(models).block(request.data)
+        filters = json_request.get("filters", None)
+        subawards = json_request["subawards"]
+
+        if filters is None:
+            raise InvalidParameterException("Missing required request parameters: 'filters'")
```
BrianZito: TinyShield can't handle this?

usaspending_api/search/v2/views/spending_by_award_count.py
```diff
@@ -0,0 +1,77 @@
...
+from usaspending_api.common.cache_decorator import cache_response
+from usaspending_api.common.exceptions import InvalidParameterException
+from usaspending_api.common.helpers.orm_helpers import category_to_award_materialized_views
+from usaspending_api.common.validator.award_filter import AWARD_FILTER
+from usaspending_api.common.validator.pagination import PAGINATION
+from usaspending_api.common.validator.tinyshield import TinyShield
+
+
+@api_transformations(api_version=settings.API_VERSION, function_list=API_TRANSFORM_FUNCTIONS)
+class SpendingByAwardCountVisualizationViewSet(APIView):
```
kbard: So the biggest difference I see here is that the existing code uses summary views to get counts when it can whereas this is trying to count from the materialized views directly.  I believe this pretty much single-handedly explains the performance difference on counts.

Just trying to count `mv_search_award_contract` by itself is taking 1 minute on staging for FY2019.

usaspending_api/awards/models_matviews/mv_search_award_idv.py
tony-sappe: First I should clarify the new naming convention, "mv" is a prefix for "materialized view". I wrote notes in the PR description, which aren't enough long term but for now are as good as it gets. Eventually all of the db matviews and views will follow the new convention.

Secondly, all of the `mv_` models are identical except for the model name and db object name. It doesn't matter which model was the copy since they are identical.

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Perhaps, but this reads clean. No unnecessary "else:" with indentation. IIRC it is becoming the preferred style.

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: obtain_view_from_award_group() will throw an error if the type list isn't a proper subset of a category of award type codes

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: I believe this is how it was previously. I wasn't planning to refactor the entire view code even if it now appears to be my intention.

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Previous name

usaspending_api/search/v2/views/spending_by_award_count.py
tony-sappe: Close. There was only one "summary award" materialized view, and it was a small subset of filters. TAS would have used universal_award_matview.

I know the current performance sucks, I foolishly switched back to the Django ORM without proper performance tests after the fact

usaspending_api/search/v2/views/spending_by_award_count.py
tony-sappe: Perhaps is can now. Originally there were edge cases which caused problems

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Sounds like a lot of work for only a little benefit.

Our API is terribly inconsistent. Which makes any attempt to standardize existing API input or responses difficult.

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: That's tricky logic to follow. I would advise putting the rest of the method in a try, then catching a specific exception obtain_view_from_award_group() throws afterwards and doing error handling there. That, or building a `award_types_valid` function. I would lean slightly towards the former.

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: reasonable. My suggestion stands, but I certainly won't block on it.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,259 @@
...
+
+            if award["award_type"] == "procurement":
+                for field in self.fields:
+                    row[field] = award.get(contract_subaward_mapping[field])
+            elif award["award_type"] == "grant":
+                for field in self.fields:
+                    row[field] = award.get(grant_subaward_mapping[field])
+            results.append(row)
+
+        return {"limit": self.limit, "results": results, "page_metadata": {"page": self.page, "hasNext": has_next}}
```
BrianZito: In response to the comment thread on line 73, here's another example with the same format. I wouldn't suggest trying to standardize the responses across every view, just handle complete duplicates. These views all used to be the same, so I can't imagine the JSON response has diverged that much at least between the new views. Another factor to consider is that the data contract assumes that each of these views will return responses of the same format since this seperation is abstracted.

tony-sappe:
> Are we skipping the tests because we can't multithread effectively, because we can't rely on timing, or some other reason? Do we realistically expect to put these tests back in a reasonable time frame, or might it be better to delete them and make it more clear this is a hole in the tests?

The current pytest fixtures load data and run tests within a DB transaction. This allows clean teardown by running a "rollback" at the end of the test file. The async logic uses a different library which creates a separate connection. Because of the other connection holding an open transaction the SQL queries are blocked, which locks the tests and the tests will hang indefinitely until an external action terminates the tests.

By removing these tests I've reduced code coverage. This isn't desirable. If people verify the current method is performant and accurate I can attempt to reinstate the tests. It doesn't make sense to spend time on the tests if the underlying code isn't good enough.

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: I'll get to work making this PR even larger

BrianZito: I'd be fine with seeing this in a separate PR, and I'd be happy to do the work myself, but I'm concerned that concerns of risk will come up, and nobody is going to want to retest this area. Would you be alright if I just made the changes on your branch, which would still up PR size but at least not burden you any more?

usaspending_api/awards/migrations/0058_reporting_matviews.py
```diff
@@ -0,0 +1,508 @@
...
+import django.contrib.postgres.fields
+import django.contrib.postgres.search
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('awards', '0057_auto_20190810_0246'),
```
emily-brents: There's been a few new migrations in dev, so this file will have to be renamed to `0061_reporting_matviews.py` and change the dependency to `0060_transactionfpds_solicitation_date`

tony-sappe: Only if this PR is merged into dev. Current target is staging, which would require the migrations in dev to be altered

usaspending_api/common/data_connectors/async_spending_by_award_count.py
```diff
@@ -0,0 +1,32 @@
...
+from usaspending_api.common.helpers.orm_helpers import generate_raw_quoted_query
+from usaspending_api.common.helpers.sql_helpers import get_database_dsn_string
+
+
+async def run(sql):
+    conn = await asyncpg.connect(dsn=get_database_dsn_string())
+    removeal_index = sql.find("GROUP BY")  # Django refuses to provide a viable option to exclude "GROUP BY ..."
+    sql_result = await conn.fetch(sql[:removeal_index])
+    await conn.close()
+    print(sql_result)
```
emily-brents: Probably should remove this print statement

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: I made changes. Might not perfectly follow the original design principles but should be a good start.

usaspending_api/search/v2/views/spending_by_award_count.py
tony-sappe: Re-implemented the async SQL calls. Performance is much better during my tests

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: renamed to `field_external_name_list`

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Refactored

usaspending_api/common/data_connectors/async_spending_by_award_count.py
tony-sappe: probably

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Refactored the check

usaspending_api/search/v2/views/spending_by_award_count.py
tony-sappe: Re-implemented the async SQL calls. Performance is much better during my tests

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,298 @@
...
+    endpoint_doc = "usaspending_api/api_contracts/contracts/search/AdvancedSearch.md"
+
+    @cache_response()
+    def post(self, request):
+        """Return all awards matching the provided filters and limits"""
+
+        json_request = self.validate_request_data(request.data)
+        filters = json_request["filters"]
+        subawards = json_request["subawards"]
+        self.fields = json_request["fields"]
```
BrianZito: Why is this a class object when filters is a method object?

tony-sappe:
> Bravo. I don't know if you did that last refactor because CC told you to or for the love of the art, but this is some very followable code.

@BrianZito see if my recent commit made things better or worse for readability

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,291 @@
...
+            self.populate_response(results=[], has_next=False)
+            return Response(self._response_dict)
+
+        self.raise_if_award_types_not_valid_subset(filters["award_type_codes"], subawards)
+        self.raise_if_sort_key_not_valid(self.pagination["sort_key"], self.fields, subawards)
+
+        if subawards:
+            self._construct_subaward_variables(filters)
+        else:
+            self._construct_award_variables(filters)
```
BrianZito: Is it unpythonic to have private functions return a value that can then be set as a variable, rather than change around class variables behind the scenes? Referential Transparency isn't a principle here, but as a future dev I wouldn't know when I would and wouldn't want to dive into this function.

tony-sappe: Did you see how many class attributes are being set in those methods?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,291 @@
...
+
+        self.type_code_to_field_map = {}
+        for award_type in loan_type_mapping:
+            self.type_code_to_field_map[award_type] = loan_award_mapping
+        for award_type in non_loan_assistance_type_mapping:
+            self.type_code_to_field_map[award_type] = non_loan_assistance_award_mapping
+        for award_type in award_idv_mapping:
+            self.type_code_to_field_map[award_type] = award_idv_mapping
+        for award_type in contract_type_mapping:
+            self.type_code_to_field_map[award_type] = award_contracts_mapping
```
BrianZito: Will more than one of these ever apply? If not, why not use some kind of enum to indicate which kind of award is being processed, and use that to sort which mapping to apply. Similar usages might be able to shorten this code by a lot, and increase adaptability if and when this splitting strategy changes.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,291 @@
...
+
+        self.raise_if_award_types_not_valid_subset(filters["award_type_codes"], subawards)
+        self.raise_if_sort_key_not_valid(self.pagination["sort_key"], self.fields, subawards)
+
+        if subawards:
+            self._construct_subaward_variables(filters)
+        else:
+            self._construct_award_variables(filters)
+
+        self.fetch_data()
```
BrianZito: I would make the same argument here as on line 85.

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: the section between 118 and 130 is an example of a bunch of things changing at once, some in scope of the function, some in scope of the class.

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: way ahead of you :)

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        self._response_dict = {
+            "limit": self.pagination["limit"],
+            "results": None,
+            "page_metadata": {"page": self.pagination["page"], "hasNext": None},
+        }
+
+        self.raise_if_award_types_not_valid_subset(filters["award_type_codes"], subawards)
+        self.raise_if_sort_key_not_valid(self.pagination["sort_key"], self.fields, subawards)
```
BrianZito:
```suggestion
        self.raise_if_sort_key_not_valid(self.pagination["sort_key"], self.fields, subawards)
        
```
I feel like there's a logical gap between the lines above and below

tony-sappe: I expected you to say that

BrianZito: You grow wiser my young padawan. And have you readied a response?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,276 @@
...
+    def populate_attributes_for_awards(self, filters):
+        """ Award logic for SpendingByAward"""
+        minimum_db_fields = {"award_id", "piid", "fain", "uri", "type"}
+        model = obtain_view_from_award_group(filters["award_type_codes"])
+        api_to_db_mapping_list = [
+            award_contracts_mapping,
+            award_idv_mapping,
+            loan_award_mapping,
+            non_loan_assistance_award_mapping,
+        ]
```
BrianZito: Does this need to be a function level variable? Seems like a constant to me. Maybe even something we could have in the mappings file/folder itself as an "all mappings" so that one isn't missed someday.

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Your original comment is OBE

usaspending_api/common/helpers/orm_helpers.py
```diff
@@ -26,8 +59,8 @@ class FiscalYear(Func):
 
 
 def generate_raw_quoted_query(queryset):
-    """
-    Generates the raw sql from a queryset with quotable types quoted.
+    """Generates the raw sql from a queryset with quotable types quoted.
```
BrianZito: black?

tony-sappe: no

BrianZito: Is it convention to start on the same line as the opening to the multi-line comment?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,184 @@
...
+    def if_no_intersection(self):
+        # "Special case" behavior: there will never be results when the website provides this value
+        return "no intersection" in self.filters["award_type_codes"]
+
+    def construct_queryset(self):
+        sort_by_fields = self.get_sort_by_fields()
+        database_fields = self.get_database_fields()
+        base_queryset = self.constants["filter_queryset_func"](self.filters)
+        queryset = self.custom_queryset_order_by(base_queryset, sort_by_fields, self.pagination["sort_order"])
+        return queryset.values(*list(database_fields))[self.pagination["lower_bound"] : self.pagination["upper_bound"]]
```
BrianZito: Does Django magic ensure that we only fetch the ones we want, or are we doing the whole query here and just slicing off the chunk we want to show?

usaspending_api/awards/models_matviews/mv_search_award_all_pre2008.py
```diff
@@ -0,0 +1,88 @@
+from django.contrib.postgres.fields import ArrayField
+from django.contrib.postgres.search import SearchVectorField
+from django.db import models
+
+from usaspending_api.awards.models import Award
+
+
+class MatviewSearchAwardAllPre2008(models.Model):
```
kwhickey: @tony-sappe For this and the other `SearchAward` matview models: I feel like there's some major opportunity here for inheritance. E.g. an Abstract Base Class, wherein the general structure of one of these matview awards models id defined. 

Since they can be unioned, and only differ by "type" of award, won't they all contain the same object attributes (and DB table columns)? 

That would make any future maintenance of these a lot simpler. Since there is potential for actually _re-combining_ these again if Postgres _partitioned_ tables can add value (after "tableview" refactor) ... having a single abstract base class with small derived models would again make that future work smaller.

usaspending_api/awards/models_matviews/mv_search_award_idv.py
kwhickey: If you might use an arbitrary model, the my previous comment about having an _abstact_ base class may be better suited just as a base class (non-abstract). Then the base class can be used in the general case to carry data, and specific sub-classes can be used to do the mapping to the specific db table.

usaspending_api/awards/models_matviews/vw_search_award.py
```diff
@@ -0,0 +1,88 @@
+from django.contrib.postgres.fields import ArrayField
+from django.contrib.postgres.search import SearchVectorField
+from django.db import models
+
+from usaspending_api.awards.models import Award
+
+
+class ViewSearchAward(models.Model):
```
kwhickey: While not a matview, it appears this model could also inherit from the same base class as the matviews re: https://github.com/fedspendingtransparency/usaspending-api/pull/1977/files#r316308611

kwhickey: For consideration - in the past working with ORMs, I a sweetspot in naming conventions seemed to be suffixing the classes with *what* they were (nothing for tables, `*View` for views, `*MatView` for materialized views, etc.)  ... but _prefixing_ database objects with what they were (e.g. `vw_*` or `mv_*` or `ix_*` ... ). 

I think it was just a subtle thing that
- when working through the code, you're trying to find that functional class, the thing that serves some business function (business model focused) 
- but when browsing through the db or writing sql, you often wanted to go to a type of object (get to the views, get to the matview, etc.). (physical model focused)

I'm fine either way though as long as we get to a consistent point.

usaspending_api/search/v2/views/spending_by_award_count.py
```diff
@@ -0,0 +1,60 @@
...
+        results = {"contracts": 0, "idvs": 0, "grants": 0, "direct_payments": 0, "loans": 0, "other": 0}
+        if subawards:
+            results = {"subcontracts": 0, "subgrants": 0}
+
+        if "award_type_codes" in filters and "no intersection" in filters["award_type_codes"]:
+            # "Special case": there will never be results when the website provides this value
+            return Response({"results": results})
+
+        if not subawards:
+            results = async_fetch_category_counts(filters, category_to_award_materialized_views())
```
kwhickey: Do we need to call `await async_fetch_category_counts(...)` here?

usaspending_api/common/data_connectors/async_spending_by_award_count.py
```diff
@@ -0,0 +1,25 @@
+import asyncio
+
+from django.db.models import Count
+
+from usaspending_api.awards.v2.filters.matview_filters import matview_search_filter
+from usaspending_api.common.helpers.orm_helpers import generate_raw_quoted_query
+from usaspending_api.common.data_connectors.async_sql_query import async_run_select
+
+
+def async_fetch_category_counts(filters, category_to_model_dict):
```
kwhickey: Nice! and good use of naming convention to demarcate an async function. Should this method be declared with `async def async_fetch_category_counts(...)`?
I haven't written code using this, but it may be recommended to use the `async`/`await` keywords for code like this: https://docs.python.org/3.5/library/asyncio-task.html

i.e. so that when this is called, it is _awaited_

e.g. 
```
def some_function()
    # ...
    await async_fetch_category_counts(filters, category_to_model_dict)
    # ...
```

If needed, it should be applied here: https://github.com/fedspendingtransparency/usaspending-api/pull/1977/files#r316344546

usaspending_api/database_scripts/matview_generator/mv_search_award_all_pre2008.json
```diff
@@ -0,0 +1,382 @@
+{
```
kwhickey: Similar to the suggestion of class inheritance for models, might there be a way to save on lines of code for all the JSON templates? 
Maybe a base template that gets interpolated to inject model-specific names/identifiers?

usaspending_api/database_scripts/matviews/vw_search_award.sql
```diff
@@ -0,0 +1,21 @@
+BEGIN;
```
kwhickey: Given there's no data being generated with this (like with a matview), it seems like it would make more sense to create this view as part of a Django migration. If alterations are needed, we would write another migration to alter the view.

usaspending_api/search/tests/test_spending_by_award_type_count.py
```diff
@@ -30,6 +74,7 @@ def test_spending_by_award_type_success(client, refresh_matviews):
     assert resp.status_code == status.HTTP_200_OK
 
 
+@pytest.mark.skip
```
kwhickey: Are these and subsequent skips something we think we can rectify (come up with a way to test the async code)? 
Problem identified and concept for a solution formed? 
Or is it looking like we may need to remove the tests?

usaspending_api/database_scripts/matviews/vw_search_award.sql
tony-sappe: Unfortunately this cannot be managed by Django since it depends on non-django managed materialized views. Tests fail since this migration would always fail

usaspending_api/common/helpers/orm_helpers.py
tony-sappe: As I recently learned: yes

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Queryset are lazy loaded, the DB still hasn't been queried yet. This is actually modifying the queryset to add `limit` and `offset` if necessary

usaspending_api/search/v2/views/spending_by_award_count.py
tony-sappe: No, all of the async/await is encapsulated elsewhere. This function will block until it returns all of the results. Which is what is desired

usaspending_api/common/data_connectors/async_spending_by_award_count.py
tony-sappe: That's what async_run_select() is in usaspending_api/common/data_connectors/async_spending_by_award_count.py.

Perhaps this function *shouldn't* say async since it calls the async function

usaspending_api/database_scripts/matview_generator/mv_search_award_all_pre2008.json
tony-sappe: It is certainly possible. However I would rather not since the JSON -> SQL is already convoluted enough

usaspending_api/search/tests/test_spending_by_award_type_count.py
tony-sappe: I haven't found a way yet to allow the async DB calls to run within the current pytest fixtures.

I'm sure there is a way to make new fixtures with new DB connections, but I have't found a good solution yet

tony-sappe:
> Well now the class names don't match the file names...

@BrianZito Can you elaborate? The model names and class names did not change in that commit or any commits this week until `d0f2e06` pushed a few minutes ago

usaspending_api/search/tests/test_spending_by_award_type_count.py
kwhickey: @tony-sappe this article gives a good synopsis of the problem and 2 reasonable solutions:
https://medium.com/ideas-at-igenius/testing-asyncio-python-code-with-pytest-a2f3628f82bc

> So, we need a way to tell pytest to use an event loop for executing tests defined as asyncio coroutines. 

Ultimately the recommendation is add in use of [`pytest-asyncio`](https://github.com/pytest-dev/pytest-asyncio) (which the article builds up to helpfully)

usaspending_api/database_scripts/matviews/vw_search_award.sql
kwhickey: I see... has to be run _after_ the matview schema objects are created. 
Takes us back to `USAspending DB Initialize and Update` Epic: https://federal-spending-transparency.atlassian.net/browse/DEV-982

usaspending_api/common/data_connectors/async_spending_by_award_count.py
kwhickey: Yes - I think you're right that this function _is not_ async (shouldn't be _awaited_) and should drop that naming convention. 

Brings up the other question:
[`results[k] = asyncio.ensure_future(async_run_select(sql[:remove_groupby_string_index]), loop=loop)`](https://github.com/fedspendingtransparency/usaspending-api/pull/1977/files#diff-954252da19751382babb8e78569beafbR19)

Should ^^ code have `await`? Or is the `asyncio.ensure_future(...)` an alternative way to _await_ the async result?

usaspending_api/awards/models_matviews/base_mv_model.py
```diff
@@ -5,7 +5,15 @@
 from usaspending_api.awards.models import Award
 
 
-class UniversalAwardView(models.Model):
+class BaseAwardSearchModel(models.Model):
```
kwhickey: I feel like this file should be named `base_award_search.py`
Reasons
- generally I don't think we have a convention of suffixing with `_model` (file) or `Model` (class), because the location of the module and class _within_ a `models` folder/package gives that indication
- I think you could take out `mv` for matview, because this particular class doesn't really ever become a matview. It's "baseness" is about it providing the model structure and fields to other models. How those other models manifest and are persisted in the db is more their issue
- should add `award_search` indicating what business purpose this model is providing a base definition of

usaspending_api/common/data_connectors/async_spending_by_award_count.py
tony-sappe: My understanding is
`asyncio.ensure_future()` creates the "promise" the task will be run in the future (like a queue)
`asyncio.gather()` will run all the Tasks _concurrently_
`run_until_complete()` is blocking until all Tasks complete and return

usaspending_api/search/tests/test_spending_by_award_type_count.py
tony-sappe: I'll read the medium post. *However* a quick glance makes me think it doesn't describe the issue here.

What is occurring are the pytest fixtures are running using one connection, and this asyncio function is creating a new connection using a different library. What happens is the pytest fixture holds an idle transaction and locks the new async connection from reading the DB models.

To resolve this issue, I have two potential ideas:
1. Create new fixtures to setup/teardown using the new `asyncpg` library for only these tests in question
2. Find a way to exchange the db connection used by the fixtures

usaspending_api/common/data_connectors/async_spending_by_award_count.py
kwhickey: I see... the example https://docs.python.org/3.5/library/asyncio-task.html#example-parallel-execution-of-tasks would seem to indicate that you don't necessarily need to call `asyncio.ensure_future(...)`, but could pass the collection of async callables (aka coroutines, that is the calls to `async_run_select`) directly into `asyncio.gather(...)`. 

But either way, it looks like the code will work. So all good.

tony-sappe: Good find. I tried several approaches to get the coroutines to A: run, B: block until all were done and C: obtain the results. I stopped once it worked without thinking of checking for extraneous code.

usaspending_api/awards/models_matviews/base_mv_model.py
tony-sappe: done

usaspending_api/common/helpers/orm_helpers.py
BrianZito: welp, if it's convention it's convention

usaspending_api/common/helpers/orm_helpers.py
```diff
@@ -51,9 +84,50 @@ def generate_raw_quoted_query(queryset):
...
+def obtain_view_from_award_group(type_list):
+    if not type_list:
+        raise AwardGroupsException("Invalid award type list: No types provided.")
+
+    type_set = set(type_list)
+    for category, values in CATEGORY_TO_MODEL.items():
+        if type_set <= values["types"]:
+            return values["model"]
+    else:
+        raise AwardGroupsException("Invalid award type list: Types cross multiple categories.")
```
BrianZito: Technically couldn't reaching here also mean that you picked a type not in any category?

BrianZito:
Not merging since you previously said you wanted to wait for more approvals

requirements/requirements.txt
```diff
@@ -44,4 +44,5 @@ urllib3==1.24.2
 xlrd==1.0.0
 jsonpickle==0.9.6
 filechunkio==1.8
-pip==19.1.1
+pip==19.2.2
```
BrianZito: was this intended? If so, why?

usaspending_api/common/data_connectors/spending_by_award_count_asyncpg.py
```diff
@@ -0,0 +1,25 @@
...
+def fetch_all_category_counts(filters, category_to_model_dict):
+    loop = asyncio.new_event_loop()
+    results = {}
+    for k, v in category_to_model_dict.items():
+        queryset = matview_search_filter(filters, v).annotate(count=Count("*")).values("count")
+        sql = generate_raw_quoted_query(queryset)
+
+        # Django refuses to provide a viable option to exclude "GROUP BY ..." so it is stripped before running the SQL
+        remove_groupby_string_index = sql.find("GROUP BY")
+        results[k] = asyncio.ensure_future(async_run_select(sql[:remove_groupby_string_index]), loop=loop)
```
BrianZito: How do we know the statement always ends with the `GROUP BY` clause?

usaspending_api/conftest_helpers.py
```diff
@@ -63,15 +63,12 @@ def _add_contents(self):
...
+        This materialized view is used by transaction_delta_view.sql, so
+        we will need to refresh it in order for transaction_delta_view to see
+        changes to the underlying tables.
         """
         with connection.cursor() as cursor:
-            cursor.execute(
-                "refresh materialized view universal_award_matview; "
-                "refresh materialized view universal_transaction_matview;"
-            )
+            cursor.execute("refresh materialized view universal_transaction_matview;")
```
BrianZito: why is awards gone?

usaspending_api/search/tests/test_spending_by_award_type_count.py
```diff
@@ -1,16 +1,70 @@
...
+
+# ===================================================
+# Below tests SKIPPED due to the introduction of asyncpg.
+# asyncpg is a different library to open DB connections to Postgres and supports async functions
+# The test data fixtures hold an open idle DB transaction which blocks SQL queries using the new connection
+# These test need to be re-implemented, but the team has allowed this feature to move forward
+# As the task is much larger than anticipated
+
+# Tony Aug 2019
+# ===================================================
```
BrianZito: # Tony 2019
seriously, good documentation though. That will come in handy in a year.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,184 @@
...
+            "lower_bound": (json_request["page"] - 1) * json_request["limit"],
+            "page": json_request["page"],
+            "sort_key": json_request.get("sort") or self.fields[0],
+            "sort_order": json_request["order"],
+            "upper_bound": json_request["page"] * json_request["limit"] + 1,
+        }
+
+        if self.if_no_intersection():  # Like an exception, but API response is a HTTP 200 with a JSON payload
+            return Response(self.populate_response(results=[], has_next=False))
+        raise_if_award_types_not_valid_subset(self.filters["award_type_codes"], self.is_subaward)
```
BrianZito:
```suggestion

        raise_if_award_types_not_valid_subset(self.filters["award_type_codes"], self.is_subaward)
```

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,184 @@
...
+            "limit": json_request["limit"],
+            "lower_bound": (json_request["page"] - 1) * json_request["limit"],
+            "page": json_request["page"],
+            "sort_key": json_request.get("sort") or self.fields[0],
+            "sort_order": json_request["order"],
+            "upper_bound": json_request["page"] * json_request["limit"] + 1,
+        }
+
+        if self.if_no_intersection():  # Like an exception, but API response is a HTTP 200 with a JSON payload
+            return Response(self.populate_response(results=[], has_next=False))
```
BrianZito: If you put in no_intersection, but you gave an invalid request, are you sure you want to return 200?

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: These sections are logically seperate

requirements/requirements.txt
tony-sappe: Yes.
The real question is: Why not?

It's good to keep pip up-to-date since installing compromised packages (or having the install hijacked mid-flight) is a potential attack vector

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -0,0 +1,184 @@
...
+        models.extend(copy.deepcopy(PAGINATION))
+        for m in models:
+            if m["name"] in ("award_type_codes", "fields"):
+                m["optional"] = False
+
+        return TinyShield(models).block(request_data)
+
+    def if_no_intersection(self):
+        # "Special case" behavior: there will never be results when the website provides this value
+        return "no intersection" in self.filters["award_type_codes"]
```
BrianZito: At one line, this was probably too small to be a function IMO

usaspending_api/common/data_connectors/spending_by_award_count_asyncpg.py
tony-sappe: Kirk and I dumped the SQL from the queryset

usaspending_api/conftest_helpers.py
tony-sappe: This entire PR revolves around removing `universal_award_matview`

BrianZito: ...

BrianZito: right

usaspending_api/common/data_connectors/spending_by_award_count_asyncpg.py
BrianZito: Don't we have LIMIT when we do pagination? That goes after group by iirc

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Yes, however I don't like it.

The problem is if the no intersection check is not performed at the beginning, then that value needs to be accounted for in the other code (either by removing it is it exists, or adding to too all of the award group sets.

The root issue is including "no intersection" in the first place. It is a terrible idea and terrible behavior to support.

BrianZito: fair enough

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Improves readability above. Devs can read the code and know exactly what the check is doing without caring about the details. The details are encapsulated in this function.

IMHO classes can be too small, but functions done properly should be small. Some well-respected developers recommend 2-4 lines. Including the comment (Because comments are lines) this is a 2-line function

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: I disagree

usaspending_api/common/data_connectors/spending_by_award_count_asyncpg.py
tony-sappe: This query is an aggregation of counts. No limit or pagination

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: While the no_intersection flag can be thought of as "exception" behavior,  the lower group are conditions under which errors are thrown, while the upper is just a case of special responses. We're entirely in subjective territory, but I would call that a difference worthy of a blank line.

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: Converted, this would equate to `if "no intersection" in self.filters["award_type_codes"]:`. If that's too complicated to not be a function, why isn't every if statement?

I don't wish to disrespect some developers, but I probably made a mistake bringing line counts into this at all.

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Both the "no intersection" and the exceptions thrown are special behavior checking for input data which doesn't meet the required conditions (even though the data are technically correct) and short-circuiting the response, preventing the rest of the view logic from executing.

BrianZito: It's not required to not be "no intersection", that's a valid, albeit special, case. If I saw an error thrown when trying to run this endpoint and decided to look in the code in this repo, I'm not looking for that line, and diving into that (two line) function would be a waste of my time.

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: At one point during this branch development, there were more lines in the function.

Because of repeated issues handling this special case, I would prefer to leave it as-is. I do realize the function is small and the logic is straight-forward. But the **significance** of the logic is not obvious. There have been numerous issues in the past with this value not being checked early enough in the view method and instead of a HTTP 200 response with the proper schema, a HTTP 500 is returned.

BrianZito: I absolutely agree that something this silly should come first, and tbh I would actually rather let the function thing go and save my energy for the fight over the blank line. But I would much rather get through you reviewing MY PR before code freeze...

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: There are no "errors" being thrown in this section of code. Exceptions are raised which are caught by the view method and return HTTP4xx responses. You wouldn't see a stack trace even with debug=True.

If the DRF exceptions didn't convert all JSON values to strings the no intersection check would also raise an exception.

PR #1985, opened 2019-08-19
DEV-3242 added batching to fpds loader

BrianZito:
**Description:**
Adds chunking logic to fpds loader to allow it to accommodate input of arbitrary size.

**Technical details:**
fabs was not similarly updated because it has a function that takes in two kinds of data at the same time, meaning chunking might be problematic. This does NOT protect against memory problems from files of excessive size.

    - [ ] Backend
8. [ ] Jira Ticket [DEV-3242](https://federal-spending-transparency.atlassian.net/browse/DEV-3242):

```
```

### DIFF
--- a/usaspending_api/broker/management/commands/fpds_nightly_loader.py
+++ b/usaspending_api/broker/management/commands/fpds_nightly_loader.py
@@ -397,7 +397,7 @@ class Command(BaseCommand):
     def next_batch_generator(file):
         while True:
             lines = file.readlines(BATCH_FETCH_SIZE)
-            lines = list(map(lambda bianary: bianary.decode(), lines))
+            lines = list(map(lambda binary: binary.decode(), lines))
             if len(lines) == 0:
                 break
             yield lines
--- a/usaspending_api/broker/management/commands/fpds_nightly_loader.py
+++ b/usaspending_api/broker/management/commands/fpds_nightly_loader.py
@@ -28,6 +28,7 @@ from usaspending_api.etl.award_helpers import (
 from usaspending_api.etl.broker_etl_helpers import dictfetchall
 from usaspending_api.etl.management.load_base import load_data_into_model, format_date, create_location
 from usaspending_api.references.models import LegalEntity, Agency
+from usaspending_api.common.retrieve_file_from_uri import RetrieveFileFromUri
 
 logger = logging.getLogger("console")
 
@@ -393,12 +394,13 @@ class Command(BaseCommand):
         logger.info("FPDS SPECIFIC (RE)LOAD COMPLETE")
 
     def load_fpds_from_file(self, afa_id_file_path):
-        with open(afa_id_file_path) as file:
+        with RetrieveFileFromUri(afa_id_file_path).get_file_object(True) as file:
             next_batch = []
             for line in file.readlines():
+                print(line)
                 next_batch.append(line)
                 if len(next_batch) >= BATCH_FETCH_SIZE:
-                    logger.info("Loading ids from file filled max batch size. Loading immediately...")
+                    logger.info("Loading batch from provided file...")
                     self.load_specific_transactions(next_batch)
                     next_batch.clear()
             #  Load final batch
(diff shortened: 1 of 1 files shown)


usaspending_api/broker/management/commands/fpds_nightly_loader.py
```diff
@@ -338,16 +339,25 @@ def insert_new_fpds(self, to_insert, total_rows):
 
     def perform_load(self, ids_to_delete, ids_to_insert):
 
-        if len(ids_to_delete) > 0:
-            with timer("deletion of all stale FPDS data", logger.info):
-                self.delete_stale_fpds(ids_to_delete=ids_to_delete)
-        else:
+        delete_index = 0
+        while delete_index < ids_to_delete.__len__():
```
tony-sappe: In Python it is `len(<list/dict/...>)` same for line 345, 354, 357.

If a dunder is accessed directly there is probably a better way.

tony-sappe: Did you look in self.fetch_fpds_data_generator()?  Using `for...` and `yield ...` appears to be a cleaner pattern to iterate through batches

usaspending_api/broker/management/commands/fpds_nightly_loader.py
```diff
@@ -401,6 +393,21 @@ def load_specific_transactions(self, detached_award_procurement_ids):
 
         logger.info("FPDS SPECIFIC (RE)LOAD COMPLETE")
 
+    @staticmethod
+    def next_batch_generator(file):
+        while True:
+            lines = file.readlines(BATCH_FETCH_SIZE)
+            lines = list(map(lambda bianary: bianary.decode(), lines))
```
tony-sappe:
```suggestion
            lines = list(map(lambda binary: binary.decode(), lines))
```

tony-sappe: map and lambda are older practices in Python. List comprehensions are often recommended.
Merging the previous line 41 and the new line would be:
`lines = [line.decode().rstrip() for line in lines]`

usaspending_api/broker/management/commands/fpds_nightly_loader.py
```diff
@@ -422,17 +429,22 @@ def add_arguments(self, parser):
...
     @transaction.atomic
     def handle(self, *args, **options):
-        if any([options["detached_award_procurement_ids"], options["id_file"]]):
-            ids_from_file = read_afa_ids_from_file(options["id_file"]) if options["id_file"] else set()
-            explicit_ids = set(options["detached_award_procurement_ids"]) if options["detached_award_procurement_ids"] else set()
-            detached_award_procurement_ids = list(explicit_ids | ids_from_file)
+        if options["id_file"]:
+            self.load_fpds_from_file(options["id_file"])
+
+        if options["detached_award_procurement_ids"]:
```
tony-sappe: Is it intentional to allow a file to be read and loaded in conjunction with a parameter list of IDs?

usaspending_api/broker/management/commands/fpds_nightly_loader.py
```diff
@@ -422,17 +429,22 @@ def add_arguments(self, parser):
...
+            explicit_ids = options["detached_award_procurement_ids"]
 
-            self.load_specific_transactions(detached_award_procurement_ids)
+            self.load_specific_transactions(explicit_ids)
         else:
             self.nightly_loader(options["date"])
+
+        # Check the linkages from file C to FPDS records and update any that are missing
+        with timer("updating C->D linkages", logger.info):
```
tony-sappe: The only change by moving this update command here is that it will always run, even if no transactions are loaded/deleted.

usaspending_api/broker/management/commands/fpds_nightly_loader.py
BrianZito: It seemed to not do any harm, and in theory someone might want to do it. This functionality was tested.

usaspending_api/broker/management/commands/fpds_nightly_loader.py
BrianZito: It also means that it will only run once if load_transactions is called many times, as is now the case when a file is read.

tony-sappe: I don't know how infrequent there are 0 FPDS transactions in a 24hrs period. What I would like to draw attention to is the possibility of needlessly increasing the time for those runs and when manually testing.

BrianZito: Worth looking at. I'll add the "DO NOT MERGE" tag and let this sit at the bottom of the PR list while I figure out if the refactor branch ends up the more practical option.

PR #1988, opened 2019-08-21
dev-3054 added mapping from state code to state name

BrianZito:
**Description:**
When displaying the state code or state name in v2 endpoints, the code will reach out for the other variable if the first one is not available.

**Technical details:**

    - [ ] Backend
8. [x] Jira Ticket [DEV-3054](https://federal-spending-transparency.atlassian.net/browse/DEV-3054):

```
```


BrianZito:
I don't have conformation from PO that this is actually what we want to do, hence "do not merge"

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -225,7 +226,10 @@ def create_place_of_performance_object(db_row_dict):
             ("country_name", db_row_dict["_pop_country_name"]),
             ("county_name", db_row_dict["_pop_county_name"]),
             ("city_name", db_row_dict["_pop_city_name"]),
-            ("state_code", db_row_dict["_pop_state_code"]),
+            ("state_code", db_row_dict["_pop_state_code"] if db_row_dict.get("_pop_state_code", None)
```
tony-sappe: No need to use .get() since the keys will be present in the dict
```suggestion
            ("state_code", db_row_dict["_pop_state_code"] if db_row_dict["_pop_state_code"]
```

BrianZito: In practice that has been shown not to be true. 45400393, the award in the ticket, is one such example, having only the state name.

usaspending_api/common/helpers/data_constants.py
```diff
@@ -0,0 +1,77 @@
+
```
tony-sappe: Perhaps rename to a more descriptive filename to denote these are State Name - State Code translators.

There are other functions dealing with other state data (FIPS) which could be moved to this file in the future

usaspending_api/awards/v2/data_layer/orm.py
tony-sappe: Using .get() is good to use when it is possible the key does not exist in the DB. If the key doesn't exist, instead of throwing a KeyError it will return None by default (or a provided value)

The original code guaranteed the key to be present in `db_row_dict`, so the value was obtained by `db_row_dict["_pop_state_code"]`. I didn't encounter any issues when I tested my recommended changes locally using `45400393`

usaspending_api/common/helpers/data_constants.py
BrianZito: I would like this file to expand beyond the scope of purely state codes and names, since we don't seem to have a library for this kind of generic real-world business logic.

tony-sappe: I fail to understand the vision

usaspending_api/awards/v2/data_layer/orm.py
BrianZito: where does the code guarantee that?

usaspending_api/common/helpers/data_constants.py
BrianZito: It was a vague vision, but I suppose it would also be a home for things like agency codes and names (if we ever did that), maybe budget category names if we ever did that, or stuff like the words for "dollars" and "cents". Just a grab bag of stuff that is too static to justify a reference table.

usaspending_api/awards/v2/data_layer/orm.py
emily-brents: afaik `db_row_dict` is pulled from the `transaction` response in the `construct_assistance/contract_response`, which gets pulled from `FPDS_CONTRACT_FIELDS/FABS_ASSISTANCE_FIELDS`, both of which will always contain a `_pop_state_code` field, even if the field is NULL/None.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -225,7 +226,10 @@ def create_place_of_performance_object(db_row_dict):
             ("country_name", db_row_dict["_pop_country_name"]),
             ("county_name", db_row_dict["_pop_county_name"]),
             ("city_name", db_row_dict["_pop_city_name"]),
-            ("state_code", db_row_dict["_pop_state_code"]),
+            ("state_code", db_row_dict["_pop_state_code"] if db_row_dict["_pop_state_code"]
+                else state_code_from_name(db_row_dict.get("_pop_state_name", None))),
```
tony-sappe: might as well remove all the `.get()` in lines 230-232

BrianZito:
```suggestion
                else state_code_from_name(db_row_dict["_pop_state_name"])),
```

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -225,7 +226,10 @@ def create_place_of_performance_object(db_row_dict):
             ("country_name", db_row_dict["_pop_country_name"]),
             ("county_name", db_row_dict["_pop_county_name"]),
             ("city_name", db_row_dict["_pop_city_name"]),
-            ("state_code", db_row_dict["_pop_state_code"]),
+            ("state_code", db_row_dict["_pop_state_code"] if db_row_dict["_pop_state_code"]
+                else state_code_from_name(db_row_dict["_pop_state_name"])),
+            ("state_name", db_row_dict["_pop_state_name"] if db_row_dict.get("_pop_state_name", None)
```
BrianZito:
```suggestion
            ("state_name", db_row_dict["_pop_state_name"] if db_row_dict["_pop_state_name"]
```

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -225,7 +226,10 @@ def create_place_of_performance_object(db_row_dict):
             ("country_name", db_row_dict["_pop_country_name"]),
             ("county_name", db_row_dict["_pop_county_name"]),
             ("city_name", db_row_dict["_pop_city_name"]),
-            ("state_code", db_row_dict["_pop_state_code"]),
+            ("state_code", db_row_dict["_pop_state_code"] if db_row_dict["_pop_state_code"]
+                else state_code_from_name(db_row_dict["_pop_state_name"])),
+            ("state_name", db_row_dict["_pop_state_name"] if db_row_dict["_pop_state_name"]
+                else state_name_from_code(db_row_dict.get("_pop_state_code", None))),
```
BrianZito:
```suggestion
                else state_name_from_code(db_row_dict["_pop_state_code"])),
```

PR #2013, opened 2019-09-05
DEV-3232: New agency loader and supporting models

kbard:
**Description:**

New Agency loader based on unified agency CSV file.  Supports new CGAC, FREC, Subtier, and Toptier agency models/tables which have not yet been rolled into production.

**Technical details:**

Supports new models while not breaking old models/tables.

2. [x] API documentation is unaffected
    - [ ] Backend
4. [x] Materialized views are unaffected
8. [x] Jira Ticket [DEV-3232](https://federal-spending-transparency.atlassian.net/browse/DEV-3232):


usaspending_api/agencies/management/commands/load_agencies_new.py
```diff
@@ -0,0 +1,91 @@
+"""
+! DO NOT USE ! DO NOT USE ! DO NOT USE ! DO NOT USE ! DO NOT USE ! DO NOT USE ! DO NOT USE !
```
BrianZito: ?

usaspending_api/common/helpers/text_helpers.py
```diff
@@ -38,3 +41,19 @@ def slugify_text_for_file_names(text, default=None, max_length=None):
         return value[:max_length] if max_length else value
 
     return default
+
+
+def standardize_whitespace(string):
```
tony-sappe: This is nice, it might be prudent to use this for the existing FABS and FPDS loaders too.

`canonicalize_string()` in  usaspending_api/references/helpers.py

usaspending_api/references/management/commands/load_cgacs.py
```diff
@@ -3,7 +3,6 @@
```
tony-sappe: Isn't the model being removed in this PR? Can we drop this loader?

usaspending_api/agencies/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+from django.utils.translation import ugettext_lazy as _
```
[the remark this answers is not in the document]
kbard: `_` is used by convention to represent localization functions in Python.

https://docs.python.org/3.5/library/gettext.html#gettext.gettext

Additionally, it's how it's used in the documentation and throughout Django.

https://docs.djangoproject.com/en/1.11/topics/i18n/translation/

Honestly, I'm not a fan myself since `_` is _also_ used by convention to represent throwaway values.

usaspending_api/agencies/management/commands/load_agencies_new.py
```diff
@@ -0,0 +1,91 @@
...
+
+    help = (
+        "Loads CGACs, FRECs, Subtier Agencies, and Toptier Agencies.  Load is all or nothing -- if "
+        "anything fails, nothing gets saved."
+    )
+
+    def add_arguments(self, parser):
+
+        parser.add_argument(
+            "agency_file",
```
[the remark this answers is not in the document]
kbard: I used `RetrieveFileFromUri`.  Our current incarnation does not directly support text files.

https://github.com/fedspendingtransparency/usaspending-api/blob/mod/dev-3232-new-agency-loader/usaspending_api/common/retrieve_file_from_uri.py#L55-L57

Apparently the Broker flavor does.  We should probably spend a couple minutes enhancing ours accordingly.

usaspending_api/agencies/models/cgac.py
```diff
@@ -0,0 +1,52 @@
+from django.db import transaction
+from django.db.models import Manager, Max, Model, TextField
+from usaspending_api.agencies.models.raw_agency_codes_csv import RawAgencyCodesCSV
+
+
+class CGACManager(Manager):
+
+    def perform_import(self):
+        """
+        Imports CGACs from the raw agency codes table.
```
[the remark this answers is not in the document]
kbard: Not gonna lie, I have a problem with being dinged for this.  Why?  Does it affect functionality?  Does it affect comprehension?  It feels totally arbitrary.  Sure it's inconsistent, but I feel like we have bigger fish to fry.

usaspending_api/agencies/models/raw_agency_codes_csv.py
```diff
@@ -0,0 +1,107 @@
...
+    mission = TextField(blank=True, null=True)
+    website = TextField(blank=True, null=True)
+    congressional_justification = TextField(blank=True, null=True)
+    icon_filename = TextField(blank=True, null=True)
+    comment = TextField(blank=True, null=True)
+
+    objects = RawAgencyCodesCSVManager()
+
+    class Meta:
+        db_table = "raw_agency_codes_csv"
```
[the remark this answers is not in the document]
kbard: It would be nice to agree to a naming convention for "raw"-ish data.

On a side note, I keep going back and forth on whether or not to even keep this file in the database.  On the one hand, it would be nice to have it for easy reference.   On the other, I'm not sure I want people glomming onto a raw table which is a risk whenever something is included in the database.

I tried putting it in its own schema, but Django really doesn't want to easily support that functionality.  I feel like this would give a cleaner separation and make it more obvious that it's a different kind of animal.

Anyhow, back to the original question, I'd like to differentiate between "raw" data and "reference" data.  "raw"-ish data is what this table is.  It contains warts and would not be something I'd want to query regularly.  Normally this would go into a "staging" schema or some place where we can work on it in isolation without risk of it become part of the production database.  This "raw" table is carved into a bunch of "reference" tables.

Anyhow, I'll rename it `raw_agency`... or maybe I'll remove it altogether.  AARRGGHH!!!

usaspending_api/agencies/tests/test_load_agencies.py
```diff
@@ -0,0 +1,78 @@
+import pytest
+import os
+
+from django.core.management import call_command
+from django.core.management.base import CommandError
+from usaspending_api.agencies.models import CGAC, FREC, SubtierAgency, ToptierAgency, RawAgencyCodesCSV
+
+
+AGENCY_FILE = os.path.join(os.path.split(__file__)[0], "test_load_agencies.csv")
```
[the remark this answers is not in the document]
kbard: Neat.

usaspending_api/references/management/commands/load_cgacs.py
kbard: I wasn't planning on putting the new loader into regular use yet since we're not really using any of the features yet, but I suppose we can.

Also, model got moved not removed.  I need to clean up how I'm doing that.  I don't like the  current recreate method.

kbard: I think I'm going to stick with my original thought and hold off on putting the new loader into production yet.  Everything's in a half state where we have two sets of toptier/subtier agencies.  It's just going to get confusing and create extra work updating the nightly pipeline multiple times.  We can remove the `load_cgacs` as part of the followup ticket where we start folding this new agency stuff into the rest of the app.

usaspending_api/agencies/apps.py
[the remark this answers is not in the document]
tony-sappe: I'll allow it, but I hate it. It doesn't make any sense

usaspending_api/agencies/management/commands/load_agencies_new.py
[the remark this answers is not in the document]
tony-sappe: If you have time it would be nice to do so now. Otherwise I'll leave as-is for future Kirk :)

usaspending_api/agencies/models/raw_agency_codes_csv.py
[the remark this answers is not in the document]
tony-sappe: Removal is better. If we don't want anybody accessing this table

usaspending_api/agencies/management/commands/load_agencies_new.py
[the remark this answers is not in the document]
kbard: Honestly, too much to do right now.  I'll add it to my TODO list.

usaspending_api/agencies/models/cgac.py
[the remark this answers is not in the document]
tony-sappe: All backend devs on this project can do a better job on docstrings. As I learn to be better, I'm encouraging others to move toward accepted best practices.

https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

I won't hold the PR over this issue, but if you need to make other changes I would appreciate changing all of the short docstrings into single lines.

usaspending_api/agencies/management/commands/load_agencies_new.py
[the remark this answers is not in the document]
kbard: Incorporated Broker's S3 text file handling code.

usaspending_api/agencies/management/commands/load_agencies_new.py
```diff
@@ -0,0 +1,131 @@
...
+                with transaction.atomic():
+                    self._execute_sql_file("create_temp_table.sql")
+                    self._import_agency_file()
+                    self._execute_sql_file("populate_dependent_tables.sql")
+                self._execute_sql("vacuum analyze cgac")
+                self._execute_sql("vacuum analyze frec")
+                self._execute_sql("vacuum analyze subtier_agency_new")
+                self._execute_sql("vacuum analyze toptier_agency_new")
+        except Exception:
+            logger.error("ALL CHANGES WERE ROLLED BACK DUE TO EXCEPTION")
```
BrianZito:
```suggestion
            logger.error("ALL CHANGES WERE ROLLED BACK DUE TO EXCEPTION: {}".format(str(Exception)))
```
Since we swallow the exception we might want to display it when rolling back...

usaspending_api/agencies/management/commands/load_agencies_new.py
```diff
@@ -0,0 +1,131 @@
...
+        logger.info("AGENCY FILE: {}".format(self.agency_file))
+
+        self.connection = get_connection(read_only=False)
+
+        try:
+            with Timer("Overall import"):
+                with transaction.atomic():
+                    self._execute_sql_file("create_temp_table.sql")
+                    self._import_agency_file()
+                    self._execute_sql_file("populate_dependent_tables.sql")
```
BrianZito: Why are these three atomic while the others aren't?

usaspending_api/agencies/management/commands/load_agencies_sql/populate_dependent_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+insert into
+    cgac (
+        cgac_code,
+        agency_name,
+        agency_abbreviation
+    )
+select
+    cgac_agency_code,
+    max(agency_name),
+    max(agency_abbreviation)
```
BrianZito: Is this because we assume the row returned to be unique? If so, why not just limit the select to one row and possibly improve performance?

BrianZito: nevermind, I see the `group by` now. Still an odd way to do this, but not one that will cause problems.

usaspending_api/agencies/management/commands/load_agencies_sql/populate_dependent_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+    )
+select
+    frec,
+    max(frec_entity_description),
+    max(frec_abbreviation)
+from
+    temp_raw_agency_csv
+where
+    frec is not null and
+    frec_entity_description is not null
```
BrianZito: So if there's no `frec_entity_description`, we can just not include it in the DB?

usaspending_api/agencies/management/commands/load_agencies_new.py
kbard: > raise
We're not swallowing the exception.

usaspending_api/agencies/management/commands/load_agencies_new.py
kbard: Vacuums can't be run in a transaction.

usaspending_api/agencies/management/commands/load_agencies_sql/populate_dependent_tables.sql
kbard: Since in most cases all we show is the entity name, an entity without a name is meaningless.

kbard:
> I'm confused what purpose the new tables serve (including where in the code they actually get used), but this seems like a smart way to go about it style-wise.

The new agency tables aren't being used yet.  They will be promoted as part of the follow-on ticket where they will lose their vestigial `_new` and  become first class members of USAspending.

usaspending_api/agencies/management/commands/load_agencies_sql/populate_dependent_tables.sql
BrianZito: In terms of errors, it seems like we are choosing between having the search for a frec return nothing when we mess up adding a description, and having a frec with no description be returned. The second seems a little more explicit of a failure to me.

BrianZito:
Thought: The four new tables are all generated from the same source, and presumably need to stay in sync since they store data like agency name. Is there a way to make these tables read-only? Would it be possible to have them as matviews?

kbard:
> Thought: The four new tables are all generated from the same source, and presumably need to stay in sync since they store data like agency name. Is there a way to make these tables read-only? Would it be possible to have them as matviews?

They're going to get rebuilt nightly.  Make changes at your own risk.  They could have been matviews had we left the raw table in the database, but that was not a popular option.

Good thought, though.

BrianZito:
I suppose nightly rebuilds are a security measure in their own right. Still, is there any way to make a comment about this on the table? These are some very unusual tables...

usaspending_api/agencies/management/commands/load_agencies_new.py
```diff
@@ -27,65 +32,101 @@ class Command(BaseCommand):
...
+                        website,
+                        congressional_justification,
+                        icon_filename
+                    ) values %s
+                """,
+                agencies,
+            )
+
+    def _vacuum_tables(self):
+        self._execute_sql("vacuum analyze cgac")
```
tony-sappe: `vacuum (full, analyze) <table>`?

small tables should take seconds

usaspending_api/agencies/management/commands/load_agencies_new.py
```diff
@@ -0,0 +1,132 @@
...
+                    self._execute_sql_file("populate_dependent_tables.sql")
+                self._vacuum_tables()
+        except Exception:
+            logger.error("ALL CHANGES WERE ROLLED BACK DUE TO EXCEPTION")
+            raise
+
+    def _execute_sql(self, sql):
+        with self.connection.cursor() as cursor:
+            cursor.execute(sql)
+            if cursor.rowcount > -1:
```
tony-sappe: I learn something new everyday... carry on!

usaspending_api/agencies/management/commands/load_agencies_new.py
```diff
@@ -0,0 +1,132 @@
...
+                logger.info("{:,} rows affected".format(cursor.rowcount))
+
+    def _execute_sql_file(self, filename):
+        file = self.sql_path / filename
```
tony-sappe: Isn't Pathlib great?!

PR #2021, opened 2019-09-06
[DEV-2484] Alter Data Dictionary download link

tony-sappe:
**Description:**
Request was to change the download filename from `DATA+Transparency+Crosswalk.xlsx` -> `Data_Dictionary_Crosswalk.xlsx`

**Technical details:**
- Refactored boto3 out of the load script to conform to the project's best practices.
- Changed response so a local load will populate the download location. (won't notice a difference when using the default URL or providing the public URL)

    - [ ] Backend
8. [x] Jira Ticket [DEV-2484](https://federal-spending-transparency.atlassian.net/browse/DEV-2484):

```
No matviews harmed in the making of this PR
No changes to API behavior
```


usaspending_api/references/management/commands/load_rosetta.py
```diff
@@ -41,36 +41,33 @@ class Command(BaseCommand):
...
 
     def handle(self, *args, **options):
+        script_start_time = perf_counter()
         logger.info("Starting load_rosetta management command")
+        logger.info("Loading data from {}".format(options["path"]))
 
-        script_start_time = perf_counter()
-        rosetta_object = extract_data_from_source_file(path=options["path"])
+        try:
+            local_filepath = Path("temp_local_data_dictionary.xlsx")
```
BrianZito:
```suggestion
            filepath = Path("temp_local_data_dictionary.xlsx")
```
and obviously all the other places too

usaspending_api/references/management/commands/load_rosetta.py
```diff
@@ -41,36 +41,33 @@ class Command(BaseCommand):
...
-        download_path = settings.ROSETTA_DICT_S3_PATH
-        bucket_name = download_path.split("/")[0]
-        filepath = "temp_data_act_crosswalk.xlsx"
-        logger.info("Using S3 file: {} from {}".format(filepath, bucket_name))
-        s3_client = boto3.client("s3", region_name=settings.USASPENDING_AWS_REGION)
-        s3_client.download_file(bucket_name, "/".join(settings.ROSETTA_DICT_S3_PATH.split("/")[1:]), filepath)
-
-    file_size = os.path.getsize(filepath)
-    wb = load_workbook(filename=filepath)
+def extract_data_from_source_file(filepath: str) -> dict:
```
BrianZito: This filepath parameter used to be optional, and the `-p` arg from the command still is. Is there some reason we know a path will always be provided now?

tony-sappe: The default value for the argument was changed from None to the global value. So it will always be a value

usaspending_api/references/management/commands/load_rosetta.py
tony-sappe: Why is that?
I can see an argument to change it to `temp_filepath`. It helps to let developers know the argument could be on a remote system, but openpyxl needs a local file. It can't accept a file object

BrianZito: I misread the line below. Now I see that this is the destination path, not a source value that can be overwritten by a uri

PR #2025, opened 2019-09-09
Staging -> Dev

tony-sappe:
Correcting oversights introduced by previous PRs, already in staging


tony-sappe:
Ignore CC total coverage report. This will be a false issue for downmerges due to the deleted files in dev

PR #2031, opened 2019-09-13
[HOTIX] To correct delta file issue

kbard:
**Description:**

A new feature was added to download code that did not play nicely with delta files causing exceptions to be thrown.

**Technical details:**

Two issues:
1. Because of the way annotations were being applied to querysets, if the queryset was accessed outside of the expected pathway (in this case a UNION statement), the annotations were not being applied correctly.
2. A piece of code that manually manipulates SQL was not prepared to handle postgres functions that accept parameters separated by commas.

This is a true HOTFIX.  I tested it as much as I could locally, but will create a followup ticket to perform a more thorough evaluation/cleanup after the fact.

### DIFF
--- a/usaspending_api/download/filestreaming/csv_source.py
+++ b/usaspending_api/download/filestreaming/csv_source.py
@@ -19,7 +19,7 @@ class CsvSource:
                 self.agency_code = agency.cgac_code
             else:
                 raise InvalidParameterException("Agency with that ID does not exist")
-        self.queryset = None
+        self._queryset = None
         self.file_name = None
         self.is_for_idv = VALUE_MAPPINGS[source_type].get("is_for_idv", False)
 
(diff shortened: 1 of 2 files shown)


kbard:
> I'm still not fully sure what all this SQL manipulation is doing, but if it doesn't work it should be obvious, and I don't forsee any side-effects

Yeah.  It was fun to walk through.  Bottom line is, it's applying aliases and reordering annotated columns.   I spent a little time trying to figure out if this is possible through the ORM and determined it would take a ton of work.  I walked the code all the way down to the Postgres adapter and it's just not something Django's going to want to do without significant coercion.

This is definitely a fragile piece of code.   Would love to find a better way.

tony-sappe:
> > I'm still not fully sure what all this SQL manipulation is doing, but if it doesn't work it should be obvious, and I don't forsee any side-effects
> 
> Yeah. It was fun to walk through. Bottom line is, it's applying aliases and reordering annotated columns. I spent a little time trying to figure out if this is possible through the ORM and determined it would take a ton of work. I walked the code all the way down to the Postgres adapter and it's just not something Django's going to want to do without significant coercion.
> 
> This is definitely a fragile piece of code. Would love to find a better way.

I suspect using database views for the downloads would greatly simply the code and make the process more reliable

PR #2038, opened 2019-09-17
Dev 3351 naics loader includes parents

BrianZito:
**Description:**
Updated naics master files to include parent and grandparent naics

**Technical details:**
Because of some strange naics codes, I modified the parser to parse some basic expressions of multiple codes per description. As before, if the parsing fails it will log the unparsable code.

    - [ ] Backend
7. [x] Appropriate Operations ticket(s) created (N/A?)
8. [x] Jira Ticket [DEV-3351](https://federal-spending-transparency.atlassian.net/browse/DEV-3351):

```
```


usaspending_api/references/management/commands/load_naics.py
```diff
@@ -73,5 +92,5 @@ def load_naics(path, append):
         wb = load_workbook(filename=path)
```
kbard: This sorts the files in the wrong order because the names are not uniform:

- naics2007.xlsx
- 2002 naics.xlsx
- 2-digit_2012_Codes.xlsx
- 2-6 digit_2017_Codes.xlsx

Need to either rename the files or come up with a more clever sorting algorithm.

usaspending_api/references/management/commands/load_naics.py
```diff
@@ -32,26 +32,45 @@ def handle(self, *args, **options):
...
+                "Unparsable NAICS range value: {0}. Please review file {1}".format(naics_range_string, path)
+            )
+    else:
+        raise CommandError("Unparsable NAICS range value: {0}. Please review file {1}".format(naics_range_string, path))
+
+
+def load_single_naics(naics_code, naics_year, naics_desc):
+
+    # crude way of ignoring naics of length 3 and 5
+    if len(str(naics_code)) not in {2, 4, 6}:
```
kbard: This is ok and will work, but traditionally we would use a tuple here `(2, 4, 6)`.   Tuples are lighter weight than sets.

usaspending_api/references/management/commands/load_naics.py
kbard: > This sorts the files in the wrong order because the names are not uniform:
> 
>     * naics2007.xlsx
> 
>     * 2002 naics.xlsx
> 
>     * 2-digit_2012_Codes.xlsx
> 
>     * 2-6 digit_2017_Codes.xlsx
> 
> 
> Need to either rename the files or come up with a more clever sorting algorithm so the files are loaded in chronological order.

Hmm.   After looking through the code s'more, I see there's a check later to not update newer values with older stuff.  Perhaps this is ok.  Still, though, those file names....

kbard:
>     2. We need to establish some sort of hierarchy in the NAICS table.  Children should point to their parents.

After sleeping on this.  If you're ok with inferring the hierarchy from the ids (e.g. `11` is the parent of `1111` because it starts from `11`) I think I'm ok letting that do the hierarchical work for us until such time as it proves to be untenable.  The way we're going to be using the hierarchy can be greatly trivialized using LIKE/startswith instead of actually trying to work with an adjacency list.

usaspending_api/references/management/commands/load_naics.py
BrianZito: I was a little miffed when I first looked at the code to just parse the year of regex, but I do kind of like that this preserves the names of the files as they come from the website (kinda). That said, upon re-reading the ticket the technical details do dictate the filenames, so I'll change them around.

PR #2039, opened 2019-09-18
DEV-2752: More agency cleanup

kbard:
**Description:**

- Put new unified agency file into production
- Cleaned up links to agency tables
- Tightened up models a bit

There's a never ending amount of work left to do for agencies but this is all I can get done this sprint.

2. [x] API should be unaffected
    - [ ] Backend
    - [ ] Frontend
8. [x] Jira Ticket [DEV-2752](https://federal-spending-transparency.atlassian.net/browse/DEV-2752):


usaspending_api/agencies/management/commands/load_agencies.py
```diff
@@ -0,0 +1,248 @@
...
+        "website",
+        "congressional_justification",
+        "icon_filename",
+    ],
+)
+
+
+# Set keep_count to True to update the overall affected row count.  Set skip_if_no_changes to True to not run
+# this step if the overall affected row count is zero.  Useful for not running expensive steps if nothing changed.
+ProcessingStep = namedtuple("ProcessingStep", ["description", "file", "function", "keep_count", "skip_if_no_changes"])
```
tony-sappe: nice

usaspending_api/references/tests/test_agency.py
```diff
@@ -8,27 +8,23 @@
...
-    Uses NIST and DOC (NIST is a subtier of DOC)
     """
 
-    # Make sure the subtier's top agency = the expected toptier agency
-    subtier = Agency.objects.get(toptier_agency__cgac_code="013", subtier_agency__subtier_code="1341")
-
-    department = Agency.objects.get(toptier_agency__cgac_code="013", toptier_flag=True)
+    # Can't run vacuums in a transaction.  Since tests are run in a transaction, we'll NOOP the
+    # function that performs the vacuuming.
+    monkeypatch.setattr(
```
tony-sappe: Lifehack

usaspending_api/agencies/README.md
```diff
@@ -0,0 +1,77 @@
+# A Little Background
+
+Whenever an award or subaward is made, there are several federal agencies tied to the award
+including, but not limited to:
+
+* Awarding Agency - The agency that interacted with the recipient to make the award.
+* Funding Agency - The agency that provided the bulk of the funding for the award.
+* Reporting Agency - The agency that reported the award to the system of record (FSRS, DABS, etc).
+
+These agencies may overlap or be missing entirely from the system of record.
```
BrianZito: `or be missing entirely from the system of record.` suggests that they can point to an agency which doesn't exist, as opposed to just being null

usaspending_api/agencies/README.md
```diff
@@ -0,0 +1,77 @@
...
+An agency can have many subtiers.  Oftentimes (but not always), one of those subtiers
+represents the agency as a whole.  For example, 020 is Department of the Treasury.  As of
+this writing, Treasury has 45 subtiers.  One of those subtiers, 2000, represents the entirety
+of Treasury.  It is roughly equivalent to 020 but at the subtier level.  2004, on the other
+hand, represents the Office of Inspector General within Treasury and, as such, is not equivalent
+to Treasury as a whole.  `toptier_flag` is used to identify those subtiers that are essentially
+equivalent to their toptier counterpart.  In this example, `toptier_flag` would be True for 2000
+but False for 2004.
+
+## Armed Forces
```
sethstoudenmier: First of all, this README is extremely helpful! For Armed Forces are coast guard and marines their own toptier agency and not in this grouping?

usaspending_api/agencies/README.md
```diff
@@ -0,0 +1,77 @@
...
+## Subtier Agency
+
+Subtier Agencies are entities that fall within the purview of a Toptier Agency, for example Office
+of Inspector General within the Department of the Treasury (TREAS) or National Institutes of Health
+(NIH) within the Department of Health and Human Services (HHS).  There are about 1,500 of these.
+
+Subtier agencies are identified by a four digit Subtier code.  The Subtier code list is maintained
+by the product owners of USAspending.
+
+An agency can have many subtiers.  Oftentimes (but not always), one of those subtiers
```
BrianZito:
```suggestion
An agency can have many subtiers.  Often, but not always, one of those subtiers
```

usaspending_api/agencies/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Armed Forces
+
+There are currently four toptier agencies that get rolled up into 097 Department of Defense for
+reporting purposes:
+
+* 017 Department of the Navy
+* 021 Department of the Army
+* 057 Department of the Air Force
+* 096 Corps of Engineers - Civil Works
```
BrianZito: Is this one subtier? I thought @sethstoudenmier or @emily-brents  did a thing about that.

usaspending_api/agencies/management/commands/load_agencies.py
```diff
@@ -0,0 +1,276 @@
...
+                "part of the nightly pipeline.".format(MAX_CHANGES)
+            ),
+        )
+
+    def handle(self, *args, **options):
+        self.agency_file = options["agency_file"]
+        logger.info("AGENCY FILE: {}".format(self.agency_file))
+
+        self.force = options["force"]
+        logger.info("FORCE: {}".format(self.force))
```
BrianZito:
```suggestion
        logger.info("CHANGE LIMIT: {}".format(str(MAX_CHANGES) if self.force else "unlimited"))
```
it will be easy to miss what "FORCE" is in the future, and accidentally putting in too many agencies for one night might be a source of confusion in the future.

usaspending_api/agencies/README.md
kbard: I believe The Marines fall under Navy.  U.S. Coast Guard apparently falls under Department of Homeland Security.

BrianZito: Coast Guard leaves and joins the military every few years. The've been out for a while now; might jump back in at some point.

usaspending_api/agencies/management/commands/load_agencies.py
BrianZito: Is there no way to have the function be an actual function as a first-class citizen instead of a string you have to use reflection on?

usaspending_api/agencies/management/commands/load_agencies.py
```diff
@@ -0,0 +1,276 @@
...
+                        include_toptier_without_subtier
+                    ) values %s
+                """,
+                self.agencies,
+                page_size=len(self.agencies),
+            )
+            if cursor.rowcount > -1:
+                logger.info("{:,} rows affected".format(cursor.rowcount))
+
+    def _run_steps(self):
```
BrianZito: This whole function looks like a great candidate for reuse; I would look into moving this and the ProccesingStep class into a common area for anything we are using the sql approach for.

usaspending_api/agencies/management/commands/load_agencies.py
kbard: There is, and I really wanted to do that, but it made the code a bit messier.  I will take another look at that and see if I can find a cleaner way.  I especially don't like having both a file and a function.  It's only ever going to be one or the other.

usaspending_api/agencies/management/commands/load_agencies_sql/clean_up.sql
```diff
@@ -0,0 +1,4 @@
+drop table if exists temp_load_agencies_agency;
+drop table if exists temp_load_agencies_raw_agency;
+drop table if exists temp_load_agencies_subtier_agency;
+drop table if exists temp_load_agencies_toptier_agency;
```
BrianZito: This will fail silently if I forget to add a temp table here. It would be nice if it was programmatically pulling it off of a list that was populated every time a sql file made a temp table.

usaspending_api/agencies/management/commands/load_agencies.py
kbard: While I would love to do this, I'm going to have to pass for this ticket.  I tinkered a little and I think it's bigger than a breadbox.

usaspending_api/agencies/tests/test_load_agencies.py
```diff
@@ -2,49 +2,294 @@
...
+    )
+
+
+def _get_record_count():
+    return (
+        ToptierAgency.objects.count()
+        + SubtierAgency.objects.count()
+        + Agency.objects.count()
+        + CGAC.objects.count()
+        + FREC.objects.count()
```
emily-brents: Is there a reason to use `objects.count()` here and use `objects.all().count()` in the `test_happy_path` function below? Is there a difference?

kbard: The difference involves sleep deprivation.  They are the same.

usaspending_api/references/models/agency.py
```diff
@@ -3,22 +3,19 @@
...
-
-    toptier_agency = models.ForeignKey("references.ToptierAgency", models.DO_NOTHING, null=True, db_index=True)
-    subtier_agency = models.ForeignKey("references.SubtierAgency", models.DO_NOTHING, null=True, db_index=True)
-
-    # 1182 This flag is true if toptier agency name and subtier agency name are equal.
-    # This means the award is at the department level.
+    id = models.AutoField(primary_key=True)
+    create_date = models.DateTimeField(auto_now_add=True)
+    update_date = models.DateTimeField(auto_now=True)
+    toptier_agency = models.ForeignKey("references.ToptierAgency", models.DO_NOTHING, db_index=True)
```
sethstoudenmier: Why is toptier_agency using `models.ForeignKey` and subtier_agency is using `models.OneToOneField`?

usaspending_api/references/management/commands/load_reference_data.py
```diff
@@ -19,7 +19,7 @@ def handle(self, *args, **options):
         call_command("loaddata", "reference_fixture")
 
         self.logger.info("Loading agency list")
-        call_command("load_agencies")
+        call_command("load_agencies", "https://files.usaspending.gov/reference_data/agency_codes.csv")
```
BrianZito: If we're hardcoding the URL, why not just commit it to code?

BrianZito:
I played around with the Process Step concept, and ended up making a branch to demonstrate a more tight class-based structure: `dev-2752-agency-rework-zito`.

This is a very nice implementation of the sql-based approach, although I still have concerns with the strategy. It seems to work in my limited testing, but I'm holding back on reviewing either way until a couple questions are answered.

usaspending_api/references/models/agency.py
kbard: Because agency => toptier is many to one and subtier => agency is one to one.  Is this a trick question?

And to answer your followup question, the final design removes the agency table altogether, but we still have some data issues to sort out so it needs to stay this way for a bit longer.  Unfortunately, the currently-four-but-soon-to-be-three DoD agencies don't have subtiers so we otherwise wouldn't be able to represent them in the agency table without this structure.

And we currently need the agency table.  In the final design, the front end will need a good bit of rework because they're currently dependent on the agency surrogate id.  I want to get rid of that dependency eventually.

Hopefully somewhere in there is the answer to your question.

usaspending_api/references/management/commands/load_reference_data.py
kbard: You're preaching to the choir.  I received pushback when I suggested it and didn't feel passionate enough about the issue to press it.  Technically, the file is owned by Broker so I guess it's their decision to make.

usaspending_api/agencies/management/commands/load_agencies_sql/clean_up.sql
kbard: Good news!  It didn't dawn on me at the time (probably because one of the earliest versions didn't run in a single transaction), but because the entire process runs in a single transaction, I can use true temporary tables.  No explicit cleanup required!

usaspending_api/references/management/commands/load_reference_data.py
BrianZito: Can we put it in the readme then? This is an un-obvious place to find the latest agency list if you bork the table and just want to reload agencies.

usaspending_api/agencies/management/commands/load_agencies.py
```diff
@@ -0,0 +1,276 @@
...
+                mission=prep(agency["MISSION"]),
+                website=prep(agency["WEBSITE"]),
+                congressional_justification=prep(agency["CONGRESSIONAL JUSTIFICATION"]),
+                icon_filename=prep(agency["ICON FILENAME"]),
+                include_toptier_without_subtier=prep(agency["CGAC AGENCY CODE"]) in DOD_ARMED_FORCES_CGAC,
+            )
+            for row_number, agency in enumerate(agencies)
+        ]
+
+    def _validate_raw_agencies(self):
```
BrianZito: I saw this was an issue in codeclimate; You can make this function just validate one agency, then run that in a loop, and that will have the Cognitive Complexity

usaspending_api/references/models/agency.py
sethstoudenmier: No trick question, just still learning about agencies.

usaspending_api/references/management/commands/load_reference_data.py
tony-sappe: settings.py is a good location to store the URL. Other URLs are stored there already and they handle the prod-vs-nonprod differences

usaspending_api/agencies/README.md
```diff
@@ -0,0 +1,74 @@
...
+this writing, Treasury has 45 subtiers.  One of those subtiers, 2000, represents the entirety
+of Treasury.  It is roughly equivalent to 020 but at the subtier level.  2004, on the other
+hand, represents the Office of Inspector General within Treasury and, as such, is not equivalent
+to Treasury as a whole.  `toptier_flag` is used to identify those subtiers that are essentially
+equivalent to their toptier counterpart.  In this example, `toptier_flag` would be True for 2000
+but False for 2004.
+
+## Armed Forces
+
+There are currently four toptier agencies that get rolled up into 097 Department of Defense for
```
tony-sappe: currently *four* toptier agencies ....

* 017 Department of the Navy
* 021 Department of the Army
* 057 Department of the Air Force


is 097 included in the 4?

usaspending_api/agencies/management/commands/load_agencies.py
```diff
@@ -0,0 +1,270 @@
...
+                toptier_flag=bool(strtobool(prep(agency["TOPTIER_FLAG"]))),
+                is_frec=bool(strtobool(prep(agency["IS_FREC"]))),
+                user_selectable=bool(strtobool(prep(agency["USER SELECTABLE ON USASPENDING.GOV"]))),
+                mission=prep(agency["MISSION"]),
+                website=prep(agency["WEBSITE"]),
+                congressional_justification=prep(agency["CONGRESSIONAL JUSTIFICATION"]),
+                icon_filename=prep(agency["ICON FILENAME"]),
+                include_toptier_without_subtier=prep(agency["CGAC AGENCY CODE"]) in DOD_ARMED_FORCES_CGAC,
+            )
+            for row_number, agency in enumerate(agencies)
```
tony-sappe: Could set `start=1` in enumerate() 
https://docs.python.org/3/library/functions.html#enumerate

usaspending_api/api_contracts/contracts/v2/idvs/awards.md
```diff
@@ -117,10 +117,10 @@ List child IDVs, child awards, or grandchild awards for IDV
...
 + `award_type`: `BPA`(required, string)
 + `description`: `OTHER THAN SCHEDULE,IGF::OT::IGF` (required, string, nullable)
-+ `funding_agency`: `GENERAL SERVICES ADMINISTRATION (GSA)` (required, string)
-+ `awarding_agency`: `GENERAL SERVICES ADMINISTRATION (GSA)` (required, string)
-+ `funding_agency_id`: 634 (required, number)
-+ `awarding_agency_id`: 634 (required, number)
++ `funding_agency`: `GENERAL SERVICES ADMINISTRATION (GSA)` (required, string, nullable)
++ `awarding_agency`: `GENERAL SERVICES ADMINISTRATION (GSA)` (required, string, nullable)
++ `funding_agency_id`: 634 (required, number, nullable)
++ `awarding_agency_id`: 634 (required, number, nullable)
```
tony-sappe: Can all four of these be null?

I thought the awarding agency was a required field for transactions

usaspending_api/common/etl/primatives.py
```diff
@@ -47,7 +54,7 @@ def make_change_detector_conditional(columns: List[str], left_alias: str, right_
     return SQL(" or ").join(composed_conditionals)
 
 
-def make_column_list(columns: List[str], alias: Optional[str] = None) -> Composed:
+def make_column_list(columns: List[str], alias: str = None, overrides: ColumnOverrides = None) -> Composed:
```
tony-sappe: Shouldn't Optional[] have been left if the default values are None?
https://docs.python.org/3/library/typing.html#typing.Optional

usaspending_api/agencies/management/commands/load_agencies_sql/toptier_agency_query.sql
```diff
@@ -0,0 +1,58 @@
+select
+    cgac_code,
```
tony-sappe: Would it be better to rename this single both cgac or frec values are populating this field?

usaspending_api/agencies/README.md
kbard: I missed updating the `four` when I removed 096 from list (from another pull request).

usaspending_api/common/etl/primatives.py
kbard: I had a problem with Optional that caused me to yank it in a few places.  Sadly, I can't remember now what that problem was.  I'll add it back and see if I can jog my memory.

usaspending_api/api_contracts/contracts/v2/idvs/awards.md
kbard: Whether or not they should be, some are null.  And some transactions have invalid subtier codes so we couldn't link them to an agency even if we wanted to.

usaspending_api/agencies/management/commands/load_agencies_sql/toptier_agency_query.sql
kbard: Troll much?

Just in case you aren't: https://federal-spending-transparency.atlassian.net/browse/DEV-3593

tony-sappe: It was an honest question. Thanks for the ticket.

usaspending_api/api_contracts/contracts/v2/idvs/awards.md
tony-sappe: That makes sense, but we might need to confirm it is desired behavior.

usaspending_api/common/etl/primatives.py
```diff
@@ -129,17 +191,43 @@ def make_typed_column_list(columns: List[str], data_types: MutableMapping[str, s
...
+    composed_columns = [SQL("{} {}").format(Identifier(c), SQL(data_types[c].data_type)) for c in columns]
     return SQL(", ").join(composed_columns)
 
 
+def wrap_dblink_query(
+    dblink_name: str, sql: Union[str, Composed], alias: str, columns: List[str], data_types: DataTypes
+) -> Composed:
+    """ Wraps a query in a dblink compatible query so that it can be run on a remote server. """
+    inner_sql = convert_composable_query_to_string(sql)
+    select_columns = make_column_list(columns, "r")
```
tony-sappe: Should "r" be replaced by `alias`?

usaspending_api/agencies/management/commands/load_agencies.py
```diff
@@ -0,0 +1,270 @@
...
+
+    etl_logger_function = logger.info
+    etl_dml_sql_directory = Path(__file__).resolve().parent / "load_agencies_sql"
+
+    def add_arguments(self, parser):
+
+        parser.add_argument(
+            "agency_file",
+            metavar="AGENCY_FILE",
+            help="Path (for local files) or URI (for http(s) or S3 files) of the raw agency CSV file to be loaded.",
```
tony-sappe: No need to change the help text, just wanted to make the point that a file scheme URI works too
```file://....```

File: usaspending_api/agencies/management/commands/load_agencies.py [hunk 21]
```diff
@@ -0,0 +1,270 @@
...
+        "user_selectable",
+        "mission",
+        "website",
+        "congressional_justification",
+        "icon_filename",
+        "include_toptier_without_subtier",
+    ],
+)
+
+MAX_CHANGES = 75
```
tony-sappe: What determined this number as the threshold?

kbard: Literally just made the number up.  I wanted us to be able to make tweaks without issue, but any errors introduced by Excel or any large scale changes I wanted us to catch before they introduced hours of additional run time to the nightly pipeline.

In my mind, the work flow is:

- If you make an intentional large scale change, we should probably manually run the loader.
- If you just tweaked a description or added a couple new agencies, it should be ok to run them in the pipeline.

usaspending_api/common/etl/introspection.py
```diff
@@ -0,0 +1,95 @@
...
+_data_types = OrderedDict(
+    (
+        ("column_name", ColumnDefinition("column_name", "text", False)),
+        ("data_type", ColumnDefinition("data_type", "text", False)),
+        ("not_nullable", ColumnDefinition("not_nullable", "boolean", False)),
+    )
+)
+
+
+def _get_whatever(table_name: str, schema_name: str, dblink_name: str, sql: str, data_types: DataTypes):
```
tony-sappe: `_get_whatever`

Ummmm......

........

...

.



whatever, just keep it

usaspending_api/common/etl/etl_writable_object_base.py
```diff
@@ -0,0 +1,80 @@
...
+                raise RuntimeError("All columns listed in update_overrides must exist in database object.")
+            key_columns = [k.name for k in self.key_columns]
+            if any(c in key_columns for c in self._update_overrides):
+                raise RuntimeError("Overriding key columns is not currently supported.")
+        return self._update_overrides
+
+    @abstractmethod
+    def _get_data_types(self) -> DataTypes:
+        """ Returns a mapping of columns names to database data types. """
+        raise NotImplementedError("Must be implemented in subclasses of ETLWritableObjectBase.")
```
tony-sappe: 10/10 recommend using this for base classes

usaspending_api/agencies/management/commands/load_agencies.py
tony-sappe: I understood the workflow. Perhaps the number could be lowered to... 20?

kbard: So I kind of had in my head the number 25, but since changes could affect all three agency tables, I multipled it by 3 to get 75 to allow for double and triple counting.  Keep in mind that there are around 1,500 subtier agencies so 75 is really a drop in a bucket.

Want I should split the difference and go with 60 (3 * 20)?

usaspending_api/awards/v2/views/accounts.py
```diff
@@ -33,50 +33,26 @@
...
     select
         coalesce(sum(gfaba.transaction_obligated_amount), 0.0)         total_transaction_obligated_amount,
         taa.agency_id || '-' || taa.main_account_code                  federal_account,
         fa.account_title,
-        afmap.toptier_agency_abbreviation                              funding_agency_abbreviation,
-        afmap.toptier_agency_name                                      funding_agency_name,
-        afmap.agency_id_for_toptier                                    funding_agency_id
+        ta.abbreviation                                                funding_agency_abbreviation,
+        ta.name                                                        funding_agency_name,
+        a.id                                                           funding_agency_id
```
tony-sappe: My personal preference is to use the "base" (or most-base) table columns for values and grouping instead of joined tables, in this case replace `a.id` with `gfaba.funding_agency_id`.

I don't know of any performance impact and obviously in this case the data are the same. Doesn't need to be changed.

usaspending_api/agencies/tests/test_load_agencies.py
```diff
@@ -2,49 +2,295 @@
...
-    # Confirm nothing is empty.
-    assert CGAC.objects.all().count() > 0
-    assert FREC.objects.all().count() > 0
-    assert SubtierAgency.objects.all().count() > 0
-    assert ToptierAgency.objects.all().count() > 0
+    record_count = _get_record_count()
+
+    original_read_raw_agencies_csv = Command._read_raw_agencies_csv
+
+    def add_agency(o):
```
tony-sappe: what is `o`?

usaspending_api/agencies/management/commands/load_agencies.py
tony-sappe: Nope, 75 is great. I was only thinking of the top tier agencies

PR #2053, opened 2019-09-25
[DEV-2998] Allow specified columns on download endpoints

sethstoudenmier:
**Description:**
Using endpoints such as `/api/v2/download/awards` would previously result in a 500 error. This PR fixes it by returning all columns of a download if there no columns on a csv match the provided columns. You could argue multiple different functionalities for this, but this seemed to be the most straightforward to me.

**Technical details:**
Inside of `csv_source.py` the queryset defaults to its mapped value in `VALUES_MAPPINGS` which mean all columns for a download. Now if a user supplies a list of columns that do exist, but let's say `subaward_view` does not have any of those columns then when `row_emitter()` returns `self.queryset.values(*query_paths)` it has nothing to filter by so it returns all of the default columns. This is an issue in `apply_annotations_to_sql()` inside of `csv_generation.py` since it compares the length of the expected aliases and the columns that are selected by the queryset. Currently `columns()` inside of `csv_source.py` returns the intersection of the columns requested and the columns possible on the download. That means in a case such as `subaward_view` in the previous example that list would be empty causing the previously stated issue that results in the selected list and alias list not to match. 

This solution is to simply change the return in `columns()`. If a user supplies columns that are not found in a particular csv then all columns for that csv will be returned. 

    - [X] Backend
8. [X] Jira Ticket [DEV-2998](https://federal-spending-transparency.atlassian.net/browse/DEV-2998):

```
Updated 10/1/19: Added to, edited, and fixed the download test cases. The only exception
is one test case for checking award download status. There were issues with populating 
`vw_award_search` for the test case. Marked it as skipped so that I can move on, but 
will plan to circle back towards the end of the sprint, time permitting.
```


kbard:
Unfortunately, when I wrote the ticket I didn't research what the expected behavior SHOULD be.  I just knew we didn't want to return a 500... which this PR resolves.

Having said that, I'm pretty sure that, if a user goes to the trouble to specify columns, they probably don't want everything returned.

One suggestion would be to always return the natural key column for all returned file types regardless of whether or not it's requested.  A free/bonus/required column, if you will.

I recognize that this increases effort on the ticket.  If you think it's too much to fit in the sprint, feel free to solicit feedback from the team lead/scrum master/PO on the best way to move forward.  We could push this and create a followup ticket or push something else... or perhaps I'm just totally off base.

sethstoudenmier:
> Unfortunately, when I wrote the ticket I didn't research what the expected behavior SHOULD be. I just knew we didn't want to return a 500... which this PR resolves.
> 
> Having said that, I'm pretty sure that, if a user goes to the trouble to specify columns, they probably don't want everything returned.
> 
> One suggestion would be to always return the natural key column for all returned file types regardless of whether or not it's requested. A free/bonus/required column, if you will.
> 
> I recognize that this increases effort on the ticket. If you think it's too much to fit in the sprint, feel free to solicit feedback from the team lead/scrum master/PO on the best way to move forward. We could push this and create a followup ticket or push something else... or perhaps I'm just totally off base.

I like that idea and will go ahead and do that. Shouldn't add too much work.

usaspending_api/common/matview_manager.py
```diff
@@ -11,7 +11,10 @@
 DEPENDENCY_FILEPATH = APP_DIR / "database_scripts/matviews/functions_and_enums.sql"
 JSON_DIR = APP_DIR / "database_scripts/matview_sql_generator"
 MATVIEW_GENERATOR_FILE = APP_DIR / "database_scripts/matview_generator/matview_sql_generator.py"
-OVERLAY_VIEWS = [APP_DIR / "database_scripts/matviews/vw_award_search.sql"]
+OVERLAY_VIEWS = [
+    APP_DIR / "database_scripts/matviews/vw_award_search.sql",
+    APP_DIR / "database_scripts/etl/transaction_delta_view.sql",
```
tony-sappe: This view is for the Elasticsearch ETL script.

Was it necessary for any of the tests below?

sethstoudenmier: Was confused by this one. It wasn't needed for anything I touched, but I had a bunch of test cases fail on me when I caught up with dev on my branch. All failed because this view didn't exist.

usaspending_api/download/filestreaming/csv_source.py
```diff
@@ -61,7 +61,7 @@ def columns(self, requested):
         """Given a list of column names requested, returns the ones available in the source"""
         result = self.human_names
         if requested:
-            result = [header for header in requested if header in self.human_names]
+            result = [header for header in requested if header in self.human_names] or result[:1]
```
tony-sappe: So if the requested fields don't match any of the available csv column names, return the first key in self.human_names?

sethstoudenmier: Correct. In almost all cases that is the unique key. For some of the downloads we have multiple id columns and so I wasn't sure which to use. Elected to simply grab the first.

tony-sappe: We'll see what Ross' response is to this. I understand the justification. If it remains as-is, a comment describing "why" is a good idea

usaspending_api/common/matview_manager.py
tony-sappe: Odd

sethstoudenmier: Agreed. Here is the Travis log before I made the change from this comment: https://travis-ci.com/fedspendingtransparency/usaspending-api/jobs/241100378

martin-martin:
Thanks for working on fixing this issue!

Just to recap to correctly understand what this PR attempts to do:
* Avoid returning a `500` Error when there is no column match
* Instead return all columns (no selection)

Does it make it possible to specify columns for the Awards data download, once merged into `master`, or will that still not function as expected [and documented](https://github.com/fedspendingtransparency/usaspending-api/blob/master/usaspending_api/api_docs/api_documentation/download/advanced_search_award_download.md) regarding specifying the `columns` filter? Thank you!

sethstoudenmier:
@martin-martin this is now currently in Production. This fix should affect all download endpoints allowing for you to specify the columns, assuming the columns you specify exist in one of the CSVs that will be returned. If a CSV does not have any of the columns that you specify then it will only return the first column instead of returning the entire CSV.

PR #2059, opened 2019-09-30
DEV-3460: load_subawards refactor

kbard:
**Description:**

The original ticket was designed to address detection of Broker subaward reloads.  During testing of that enhancement, some other issues were found that were causing subawards to not get copied correctly.

This is a refactor of the subaward loader to perform full delta comparisons until such time as Broker can supply us with more granular change/deletion tracking.

**Technical details:**

In addition to fixes and adding additional tests, I also added some tools which will help with some other loaders we're working on.  I know they'll be super useful for the new agency loader.

2. [x] API is unaffected
    - [ ] Backend
4. [x] Materialized views are unaffected
5. [x] Front end is unaffected
8. [x] Jira Ticket [DEV-3460](https://federal-spending-transparency.atlassian.net/browse/DEV-3460):
    - [x] Performance is negatively impacted by full delta compares.  Runs take about 15 minutes in the development environment.  This should translate to probably 10ish minutes in production.  Previously comparisons took about 3~4 minutes.  Full reloads, on the other hand, are significantly faster.  45 minutes in development as opposed to about 2 hours previously.


BrianZito:
What's the logic behind the numbering convention, and why is it in tens?

usaspending_api/awards/management/commands/load_subawards.py
```diff
@@ -1,185 +1,155 @@
...
+            results = function(*args, **kwargs)
+            if type(results) is int and results > -1:
+                logger.info("{:,} rows affected".format(results))
+            return results
+
+    def _execute_sql_file(self, filename):
+        """ Read in a SQL file and execute it. """
+
+        filepath = (self.sql_path / filename).with_suffix(".sql")
+        return self._execute(filename, execute_update_sql, sql=filepath.read_text())
```
BrianZito: Should this still be called `execute_update_sql`? It seems like it's getting more generic

usaspending_api/common/helpers/sql_helpers.py
```diff
@@ -246,6 +217,60 @@ def fetchall_to_ordered_dictionary(cursor):
...
+        # Because django-debug-toolbar does not understand Composable queries,
+        # we need to convert the query to a string before executing it.
+        cursor.execute(convert_composable_query_to_string(sql, cursor=cursor))
+        return fetcher(cursor)
+
+
+def execute_update_sql(sql, model=Award):
+    """
+    Convenience function to execute a sql query against a database that performs
+    some sort of update (INSERT, UPDATE, DELETE, etc).  Returns the number of
```
BrianZito: This is deceptive to less experienced sql devs such as myself. A function called `execute_modification_sql` or something of that ilk would make it much clearer that this isn't just for updates.

BrianZito:
At the first pass, I think my main suggestion would be improving [inversion of control](https://stackify.com/dependency-inversion-principle/). Functions like `_execute` are a code smell to me that objects aren't being leaned on enough, will will lead to a lot of duplicate code in the long run. Take the first call:
```self._execute(
            "Copy Broker subaward table",
            operations.stage_dblink_table,
            source=remote_subaward,
            destination=broker_subaward,
            staging=temp_broker_subaward,
        )```
What information does `load_subawards` know that the subaward table doesn't? 

- The description is arbitrary, and could be defined anywhere
- the operations function is independent of the loader, and in fact needs to populate its values from the table it's working with 
- the source is dependent on the table. If the table ever changed source, you would need to update every call of execute involving it
- `destination` is the table
- the staging table is technically arbitrary, but really it would make more sense to have that tied to the table name so that two loaders never had to risk stepping on each other's toes unless they literally were loading the same thing at the same time

Suppose instead that `broker_subaward` was either an implementation of, or singleton (I don't know if Python has real singletons, so maybe a global "singleton") instance of, the ETL table class. the operation stage_dblink_table() is a function of that object, which always knows what the source table should be and where it's going to stage. In the future, we might be able to even roll up some of those one-off sql scripts into the implementations of the operation functions, since my guess is most of these will only and always be run before or after one of the CRUD operations of a specific table. 

The whole reason I made the FPDS loader the way I did is that I was afraid of redundant SQL code, and a web of dependent steps that have to be run after other steps to keep data in line. With a little more work, I think the ETLTable class is a great way to mitigate that problem.

**Comment by kbard** [comment]
> What's the logic behind the numbering convention, and why is it in tens?

The first _computer_ language I learned was BASIC.  At that time BASIC required line numbers.  By convention, line numbers were spaced by 10s (or 100s... but usually 10s) so that if you had to go back and add lines you had room to grow.

This has nothing to do with that.

It's just a way to visually keep the queries in basic runtime order... with room to grow.  If I had it to do over again, I don't know if I'd do the numbers again (a la agency loader).

File: usaspending_api/awards/management/commands/load_subawards.py (hunk 1 above)
Comment (kbard) [reply usaspending_api/awards/management/commands/load_subawards.py]: `execute_update_sql` is a convenience function intended to be used for running CRUD(-R) SQL commands.  Not sure what a better name would be.   `execute_CUD_sql` maybe?  Yes it can be used for reads, but it shouldn't be since it only acts against the master database instance instead of taking advantage of our read replica(s).

File: usaspending_api/common/helpers/sql_helpers.py (hunk 2 above)
Comment (kbard) [reply usaspending_api/common/helpers/sql_helpers.py]: Oh.  I see you addressed my previous question here.  So `execute_CUD_sql` is out then?

**Comment by kbard** [comment]
> At the first pass, I think my main suggestion would be...

Dear Professor Zito,

In a world of finite time, I often find myself painted into a corner because I took on more than I could handle in a single sprint.  I almost always have grander visions than can fit in our ridiculous two week cycles.

I have three other partial/incomplete versions of this ETL code sitting cached away for me to slowly start integrating into the app as I work on future loaders.  In fact, this code is a refinement of something I wrote for the original subaward loader but discarded because I thought it was was too messy and didn't have a clear vision on how to improve it before sprint end.

What you see here is the absolute minimum implementation I thought I could get away with to make the subaward loader work.

My point is, software development is a very iterative process.  I will file away your lecture and attempt to incorporate its lessons in future enhancements.

Thank you for taking the time to respond.

Sincerely,
Your Student

File: usaspending_api/common/helpers/sql_helpers.py (hunk 2 above)
Comment (kbard) [reply usaspending_api/common/helpers/sql_helpers.py]: After further reflection, I'm not sold on `modification`.  My first thought is that we're modifying the structure of the table.  In my head, `update` is a superset of `INSERT`, `UPDATE`, and `DELETE` since it is updating the data in the table.

Comment (BrianZito) [reply usaspending_api/common/helpers/sql_helpers.py]: I'm also at a loss for the proper term then. Still bothered by the fact that `update` shadows the SQL keyword.

File: usaspending_api/common/etl/etl_table.py [hunk 3]
```diff
@@ -0,0 +1,94 @@
...
+    def data_types(self) -> MutableMapping[str, str]:
+        """ Returns a mapping of columns names to database data types. """
+        sql = """
+            select      attname as column_name, pg_catalog.format_type(atttypid, atttypmod) as data_type
+            from        pg_attribute
+            where       attnum > 0 and attisdropped is false and attrelid = '{qualified_table_name}'::regclass
+            order by    attnum
+        """
+        sql = SQL(sql).format(qualified_table_name=self.qualified_table_name)
+        if self.dblink_name is not None:
```
tony-sappe: use `self.is_dblink()` here?

usaspending_api/common/etl/etl_table.py
```diff
@@ -0,0 +1,94 @@
...
+        return self.dblink_name is not None
+
+    @cached_property
+    def non_key_columns(self) -> List[str]:
+        """ Returns the list of columns minus the columns comprising the primary key. """
+        return [c for c in self.columns if c not in self.key_columns]
+
+    @cached_property
+    def key_columns(self) -> List[str]:
+        """ Returns the list of columns comprising the primary key. """
```
tony-sappe: Are there edge cases of tables without primary key constraints but unique constraints which are desired to be returned by this function (I can't think of any examples)?

Might be a non-issue.

usaspending_api/common/etl/operations.py
```diff
@@ -0,0 +1,179 @@
...
+from psycopg2.sql import Literal, SQL
+from typing import List
+from usaspending_api.common.etl import ETLTable, primatives
+from usaspending_api.common.helpers import sql_helpers
+
+
+def _get_shared_columns(columns: List[str], other_columns: List[str]) -> List[str]:
+    """ Return the list of columns contained in both lists. """
+
+    shared_columns = [c for c in columns if c in other_columns]
```
tony-sappe: My mind naturally thinks of `set()` for these types of operations. Nothing wrong with list comp here

usaspending_api/common/etl/etl_table.py
kbard: heh... these stupid things are the reason I created `is_dblink`.  Apparently I missed a couple.

usaspending_api/common/etl/primatives.py
```diff
@@ -0,0 +1,145 @@
...
+    Turn an alias, table, and a list of key columns into a SQL safe string containing
+    join conditionals ANDed together.
+
+        d.id1 = public.table1.id1 and d.id2 = public.table1.id2
+
+    """
+    composed_aliases = {"left_alias": Identifier(alias), "right_alias": qualified_table_name}
+    template = "{left_alias}.{column} = {right_alias}.{column}"
+    composed_conditionals = [SQL(template).format(column=Identifier(c), **composed_aliases) for c in key_columns]
+    return SQL(" and ").join(composed_conditionals)
```
tony-sappe: I would have expected `SQL(" and ".join(composed_conditionals))`

Can you confirm the code is correct performing a .join() on the SQL object instead of creating a SQL object from the string?

usaspending_api/common/etl/etl_table.py
kbard: Yes.  I have this situation in the agency loader.  It's one of the enhancements I will be making soon in that pull request.

usaspending_api/common/etl/operations.py
kbard: Mine too.  I first wrote it that way, but I think this might be more efficient than casting back and forth between sets.

usaspending_api/common/helpers/sql_helpers.py
```diff
@@ -170,68 +166,43 @@ def convert_composable_query_to_string(sql, model=Award, cursor=None):
...
-    Returns a CLOSED cursor.  Can be used for row counts and what not, but is
-    no longer operational.
-    """
-    connection = get_connection(model, False)
-    with connection.cursor() as cursor:
-        # Because django-debug-toolbar does not understand Composable queries,
-        # we need to convert the query to a string before executing it.
-        cursor.execute(convert_composable_query_to_string(sql, cursor=cursor))
-        return cursor
+def cursor_fetcher(cursor):
```
tony-sappe: What is this for?

usaspending_api/common/etl/primatives.py
kbard: This is correct.  SQL can do joins much like a string.  Pretty sure I use it this way all over the place.  Now I have to go double check everywhere that I didn't accidentally go the OTHER way anywhere...

usaspending_api/common/helpers/sql_helpers.py
kbard: Originally the update sql convenience function returned a closed cursor instead of the count.  I changed it but left the functionality around.

usaspending_api/common/etl/primatives.py
kbard: Ok.  Just confirmed.  Can't do string.join([composables]) or composable.join([strings]).   They throw exceptions.

usaspending_api/common/helpers/sql_helpers.py
```diff
@@ -170,68 +166,43 @@ def convert_composable_query_to_string(sql, model=Award, cursor=None):
...
+    """ Fetcher that simply returns the cursor. """
+    return cursor
 
 
-def execute_fetchall(sql, model=Award, fetcher=None):
+def default_fetcher(cursor):
```
tony-sappe: I preferred the original func name, I understand what "fetchall" is in context of a DB cursor. `default_fetcher()` doesn't convey much and _requires_ the docstring be read, which I argue is a code smell.

perhaps `generic_fetchall` or `default_fetchall` or `retreive_db_records` or ...

usaspending_api/awards/management/commands/load_subawards_sql/150_update_awards.sql
```diff
@@ -0,0 +1,34 @@
+-- Update counts in the awards table with new subaward totals.
```
tony-sappe: How long does it take for this to run?

I'm assuming it's going to be near constant time every run irregardless of the subaward volume loaded

usaspending_api/awards/migrations/0066_remove_brokersubaward_imported.py
```diff
@@ -0,0 +1,25 @@
...
+        migrations.RunSQL(
+            ["drop index if exists idx_broker_subaward_imported_true"],
+            ["""
+                create index if not exists idx_broker_subaward_imported_true on 
+                broker_subaward (imported) where imported is true"""],
+        ),
+        migrations.RemoveField(
+            model_name='brokersubaward',
+            name='imported',
+        ),
```
tony-sappe: Should there be a revert step in the operations?

usaspending_api/common/etl/operations.py
tony-sappe: For small lists the costs probably won't be noticeable.

Sets would change the behavior if the columns list contained duplicates. Unsure if duplicates are desired/expected/possible

usaspending_api/awards/migrations/0066_remove_brokersubaward_imported.py
kbard: I'm pretty sure the built-in migration functions revert automatically.  We don't write code to reverse field additions.

usaspending_api/common/helpers/sql_helpers.py
tony-sappe: but... the function simply returns the cursor it was provided. What workflow needs this?

usaspending_api/awards/management/commands/load_subawards_sql/150_update_awards.sql
kbard: On  `dev`:

- [150_update_awards] finished successfully after 2 m 16 s 650 ms
- [150_update_awards] finished successfully after 2 m 29 s 690 ms
- [150_update_awards] finished successfully after 2 m 43 s 681 ms

Those are with no or very few updates.  If we receive a ton of subawards all at once, I would expect this number to spike a bit.

usaspending_api/common/helpers/sql_helpers.py
kbard: It's the nature of the beast.  `execute_sql` uses dependency injection to return result.  If, for whatever reason, you want the cursor from `execute_sql` (which I did at one point but later changed my mind), we need a "fetcher" function that can do that.

If it bothers you that much I can remove it and we can add it back in later if  anyone ever needs it.

usaspending_api/common/etl/primatives.py
tony-sappe: Ah, makes sense. running .join() on a composable using an iterable of composable objects. Got it

PR #2060, opened 2019-10-01
Quick fix for fiscal year based tests

kbard:
Fixed a couple of tests that could not keep up with the times...

### DIFF
--- a/usaspending_api/search/tests/unit/test_tas_search.py
+++ b/usaspending_api/search/tests/unit/test_tas_search.py
@@ -262,7 +262,8 @@ def test_spending_over_time(client, mock_tas_data, refresh_matviews):
     data = {"group": "fiscal_year", "filters": {"tas_codes": [{"aid": "028", "main": "8006"}]}, "subawards": False}
     resp = client.post("/api/v2/search/spending_over_time", content_type="application/json", data=json.dumps(data))
     assert resp.status_code == status.HTTP_200_OK
-    assert len(resp.data["results"]) == 12
+    assert len(resp.data["results"]) > 0   # Can't use a fixed number here because this number increases every year.
+
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


usaspending_api/search/tests/unit/test_tas_search.py
```diff
@@ -262,7 +263,7 @@ def test_spending_over_time(client, mock_tas_data, refresh_matviews):
     data = {"group": "fiscal_year", "filters": {"tas_codes": [{"aid": "028", "main": "8006"}]}, "subawards": False}
     resp = client.post("/api/v2/search/spending_over_time", content_type="application/json", data=json.dumps(data))
     assert resp.status_code == status.HTTP_200_OK
-    assert len(resp.data["results"]) == 12
```
BrianZito: Can we rely on FiscalDate.today() to be accurate?

tony-sappe: What are the concerns? The question could be asked of anything:
> Can we rely on datetime.now() to be accurate?
> Can we rely on Decimal to be accurate?
> ...

kbard: > What are the concerns? The question could be asked of anything:
> 
> > Can we rely on datetime.now() to be accurate?
> > Can we rely on Decimal to be accurate?
> > Can we rely on @kbard to be accurate?
> > ...

I think I know the answer to #3...

BrianZito: Should have been for this line:
```
assert len(resp.data["results"]) == FiscalDate.today().fiscal_year - 2007
```
Didn't know if we implemented FiscalDate or not. If we did, we might be grading our own homework here.

kbard: Your updated comment is correct.  Added a better test.

tony-sappe: @BrianZito We did not implement FiscalYear, it is an external library

PR #2065, opened 2019-10-03
[DEV-3442] adding overlooked update to FK field

tony-sappe:
**Description:**
New FK column from awards to transaction_normalized (`earliest_transaction_id`) needs to be set to null along with `latest_transaction_id` when records are being purged. Value will be re-populated in a later stage of the script.

Thanks to @BrianZito for catching this in his open PR


tony-sappe:
Travis-CI checks fails because of fiscal year tests. While fixed by this PR #2060 the edits have not reached staging yet

PR #2070, opened 2019-10-04
Mod/dev 3558 new psc loader

emily-brents:
**Description:**
New PSC loader to handle the new added fields.

**Technical details:**
The file provided by Ross is "PSC_Data_June_2019_Edition_FINAL_6-20-19+DRW.xlsx", the new loader can handle excel files. The loader has to check if the rows in the file are visible before loading them, as the file contains multiple filters that restricts the data being loaded. Also includes the option to update the table to add the lengths for any codes that were previously in the table but not in the new file and therefore did not get lengths added during the load.

    - [ ] Backend
8. [x] Jira Ticket [DEV-3558](https://federal-spending-transparency.atlassian.net/browse/DEV-3558):

```
No change to frontend, universal_transaction_matview and mv_award_contract_search join to psc, but should not be affected as no codes will be removed.
```


usaspending_api/references/tests/test_load_psc.py
```diff
@@ -0,0 +1,31 @@
...
+
+@pytest.mark.django_db
+def test_load_psc():
+    """
+    Test to ensure the ingested data size is correct
+    Test to make sure previously missing information has been added
+
+    """
+
+    call_command("load_psc")
```
BrianZito: Since the command isn't a nuke and pave, does this really test anything? What if it fails to update this row?

emily-brents: the test database starts empty, so it tests if it loads the data in with the correct fields.

usaspending_api/references/management/commands/load_psc.py
```diff
@@ -1,45 +1,86 @@
...
 from usaspending_api.references.models import PSC
 import os
-import csv
 import logging
+from openpyxl import load_workbook
 
 
 class Command(BaseCommand):
-    help = "Loads program information obtained from csv file on https://www.acquisition.gov/PSC_Manual"
+    help = "Loads program information obtained from Excel file on https://www.acquisition.gov/PSC_Manual"
```
BrianZito: This kinda makes it sound like the file is being downloaded each time

usaspending_api/references/migrations/0032_auto_20190926_1308.py
```diff
@@ -0,0 +1,55 @@
...
+    dependencies = [
+        ('references', '0031_delete_cgac'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='psc',
+            name='code',
+            field=models.CharField(max_length=4),
+        ),
```
sethstoudenmier: I don't see anything wrong with this, but confused why it got added to the migration when it isn't touched on the PSC model.

emily-brents: ah yeah, i had to mess with the model a bunch when i was testing stuff, that shouldn't be there at all, good catch

PR #2076, opened 2019-10-08
[DEV-2852] Alter docker-compose and undo some changes

sethstoudenmier:
**Description:**
There was some confusion on my end about some of the AC. The confusion has been cleared up so some changes were undone and the docker-compose file was edited as was intended. 

**Technical details:**
Make the docker-compose file use `DB_SOURCE` and `DB_R1` instead of `DATABASE_URL` for the API container. 

1. [X] Unit & integration tests updated (N/A)
    - [ ] Backend
6. [X] Data validation completed (N/A)
8. [X] Jira Ticket [DEV-2852](https://federal-spending-transparency.atlassian.net/browse/DEV-2852):

```
```


usaspending_api/settings.py
```diff
@@ -169,36 +169,34 @@ def _configure_database_connection(environment_variable):
     Configure a Django database connection... configuration.  environment_variable is the name of
     the operating system environment variable that contains the database connection string or DSN
     """
-    default_options = {"options": "-c statement_timeout={0}".format(DEFAULT_DB_TIMEOUT_IN_SECONDS * 1000)}
+    default_options = {"options": "-c statement_timeout={0}".format(int(DEFAULT_DB_TIMEOUT_IN_SECONDS) * 1000)}
```
kwhickey: @sethstoudenmier I think the fix should instead be on the module variable as its set, and casting that to an int when read from the environment. 

```
DEFAULT_DB_TIMEOUT_IN_SECONDS = os.environ.get("DEFAULT_DB_TIMEOUT_IN_SECONDS") or 0
```

This solution is not tested but maybe:
```
DEFAULT_DB_TIMEOUT_IN_SECONDS = int(os.getenv("DEFAULT_DB_TIMEOUT_IN_SECONDS", "0"))
```
or
```
DEFAULT_DB_TIMEOUT_IN_SECONDS = int(os.environ.get("DEFAULT_DB_TIMEOUT_IN_SECONDS", 0))
```

usaspending_api/settings.py
```diff
@@ -169,36 +169,34 @@ def _configure_database_connection(environment_variable):
...
 
 
 # If DB_SOURCE is set, use it as our default database, otherwise use dj_database_url.DEFAULT_ENV
-# (which is "DATABASE_URL" by default).  Generally speaking, DB_SOURCE is used to support server
-# environments that support the API/website whereas DATABASE_URL is used for development and
-# operational environments (Jenkins primarily).  If DB_SOURCE is provided, then DB_R1 (read replica)
-# must also be provided.
+# (which is "DATABASE_URL" by default). Generally speaking, DB_SOURCE is used to support server
+# environments that support the API/website and docker-compose local setup whereas DATABASE_URL
+# is used for development and operational environments (Jenkins primarily). If DB_SOURCE is provided,
```
kwhickey: Not totally true if developing locally with docker now... but I think this is safe to say.

tony-sappe: None of the developers are currently developing the app within Docker

usaspending_api/settings.py
tony-sappe: This seems like the code you're looking for:
```python
DEFAULT_DB_TIMEOUT_IN_SECONDS = int(os.environ.get("DEFAULT_DB_TIMEOUT_IN_SECONDS")) or  0
```

kwhickey: @tony-sappe what does `int()` do on a `None` object?

sethstoudenmier: `TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'`

kwhickey: Tested... looks like it throws a `TypeError`

So the risk is an error if that env var is not set. It was safe the way Seth had it, because of the `None or 0` ... but when casting at the spot, it needs a default value to be provided on the `get` or `getenv`.

tony-sappe: yeah, my bad.
```python
DEFAULT_DB_TIMEOUT_IN_SECONDS = int(os.environ.get("DEFAULT_DB_TIMEOUT_IN_SECONDS", 0))
```

This will still throw an exception for strings which cannot be converted to an int. And I don't know what will happen with negative values.

PR #2094, opened 2019-10-24
COLDFIX: Update/Create routers to prevent migrations against non-USAspending databases and throw exceptions for connections pointing to AWS databases during tests

kbard:
After pulling the most recent version of `dev`, running certain tests locally increased from about a minute of run time to about 15 minutes.  This was unacceptable.

The issue was traced to [a change made as part of the new FPDS loader](https://github.com/fedspendingtransparency/usaspending-api/pull/2015/files#diff-d41742a07a26ba73b6844b153bf4b12aL195-R195) (thanks @sethstoudenmier!).

The problem was twofold:

- Part 1 was a user issue.  I had no idea my Broker connection string was actually being used for tests now (which, btw, was pointing to the PRODUCTION Broker server).  I added a check to throw an exception if it appeared any of the database connection strings are pointing to AWS during tests.  Running tests over the internet against an RDS introduces a SIGNIFICANT performance penalty.
- Part 2 was more of a Django issue.  Django apparently assumes all databases need a copy of the database so it was migrating models to all connections.  I added/updated our routers to prevent migrations to non-USAspending databases.

That's it.  The test I was running is back down to a minute.  Success!

### DIFF
--- a/usaspending_api/conftest.py
+++ b/usaspending_api/conftest.py
@@ -3,7 +3,6 @@ import os
 import pytest
 import tempfile
 
-from django.conf import settings
 from django.db import DEFAULT_DB_ALIAS
 from django.test import override_settings
 from django_mock_queries.query import MockSet
--- a/usaspending_api/conftest.py
+++ b/usaspending_api/conftest.py
@@ -192,16 +192,8 @@ def mock_matviews_qs(monkeypatch):
 
 
 def pytest_configure():
-    # This function re-added so tests would stop running against non-test databases.
-    #
-    # To make sure the test setup process doesn't try
-    # to set up another test db, remove everything but the default
-    # DATABASE_URL from the list of databases in django settings
-    test_db = settings.DATABASES.pop(DEFAULT_DB_ALIAS, None)
-    settings.DATABASES.clear()
-    settings.DATABASES[DEFAULT_DB_ALIAS] = test_db
-    # Also remove any database routers
-    settings.DATABASE_ROUTERS.clear()
+    # This function used to remove all non-default DB connections.  We undid this 10/2019 to allow testing of loaders.
+    pass
 
 
 def pytest_addoption(parser):
--- a/usaspending_api/conftest.py
+++ b/usaspending_api/conftest.py
@@ -3,7 +3,7 @@ import os
 import pytest
 import tempfile
 
-from django.db import DEFAULT_DB_ALIAS
+from django.db import connections, DEFAULT_DB_ALIAS
 from django.test import override_settings
 from django_mock_queries.query import MockSet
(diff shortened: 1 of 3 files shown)


sethstoudenmier:
Does adding this break the tests for loaders?

usaspending_api/conftest.py
```diff
@@ -191,8 +192,16 @@ def mock_matviews_qs(monkeypatch):
...
+    # This function re-added so tests would stop running against non-test databases.
+    #
+    # To make sure the test setup process doesn't try
+    # to set up another test db, remove everything but the default
+    # DATABASE_URL from the list of databases in django settings
+    test_db = settings.DATABASES.pop(DEFAULT_DB_ALIAS, None)
+    settings.DATABASES.clear()
+    settings.DATABASES[DEFAULT_DB_ALIAS] = test_db
+    # Also remove any database routers
+    settings.DATABASE_ROUTERS.clear()
```
sethstoudenmier:
```    
    # To make sure the test setup process doesn't try
    # to set up another test db, remove everything but the default
    # DATABASE_URL from the list of databases in django settings
    test_usaspending_db = settings.DATABASES.pop(DEFAULT_DB_ALIAS, None)
    test_broker_db = settings.DATABASES.pop("data_broker", None)
    settings.DATABASES.clear()
    settings.DATABASES[DEFAULT_DB_ALIAS] = test_usaspending_db
    settings.DATABASES["data_broker"] = test_broker_db
    # Also remove any database routers
    settings.DATABASE_ROUTERS.clear()```

Comment (sethstoudenmier) [reply usaspending_api/conftest.py]: This should use the `data_broker` database in `settings.DATABASES` that is defined in travis.yml

File: usaspending_api/routers/replicas.py [hunk 2]
```diff
@@ -14,22 +14,32 @@
...
     def db_for_write(self, model, **hints):
-        # write to source db only (bc read replicas)
+        """ Write to source (default) db only because replicas are read only. """
         return DEFAULT_DB_ALIAS
 
     def allow_relation(self, obj1, obj2, **hints):
-        db_list = (DEFAULT_DB_ALIAS, "db_r1")
-        if obj1._state.db in db_list and obj2._state.db in db_list:
-            return True
-        return None
```
sethstoudenmier: Since the original implementation returned `True` or `None` should the changes match that instead of returning `True` or `False`? Not sure why it returns None, but it looks intentional.

kbard: Of course you're right.

kbard: Actually, after further reading, you return None if your router has no opinion which allows subsequent routers to cast their vote.  I think in this case, we're pretty certain we do not want to allow relations to other databases.  That and we currently have no other routers.

PR #2100, opened 2019-10-30
Dev 3634 correct data descrepencies

BrianZito:
Update given discrepancy testing. location.py had a small function that would try to infer various address values. This mimics that functionality.

There is also logic for inferring address values from the ref_city_county_code table, which will come in a later PR


usaspending_api/etl/transaction_loaders/field_mappings_fpds.py
```diff
@@ -33,7 +39,6 @@
     "ultimate_parent_unique_ide": "ultimate_parent_unique_ide",
     "award_description": "award_description",
     "place_of_performance_zip4a": "place_of_performance_zip4a",
-    "place_of_performance_zip5": "place_of_performance_zip5",
```
tony-sappe: I don't think we want to change these "raw" fields in `transaction_fpds`, it's similar to the whitespace discussion.

BrianZito: Is that generally the case for all the derived fields in `transaction_fpds_functions`? What about the other tables that use these values? Are we alright with them being different?

usaspending_api/references/abbreviations.py
```diff
@@ -74,6 +74,7 @@
     "PLW": "Palau",
 }
 
+code_to_name = {k: v["name"] for (k, v) in code_to_state.items()}
```
tony-sappe: Is this necessary? The pre-existing _to_code dictionaries are changing the keys, this is simplifying the values with already present data.

Example derivation function code using the original:
```python
   code_to_state.get(broker_input["legal_entity_state_code"], {}).get("name")
```

BrianZito: Doesn't that break if the first .get() returns None? That seemed like a vulnerability in the origional.

tony-sappe: The default value for the first get is an empty dict, no vulnerability compared to the original.

usaspending_api/etl/transaction_loaders/field_mappings_fpds.py
tony-sappe: > Is that generally the case for all the derived fields in `transaction_fpds_functions`?

These are different than the existing fields being altered.

> What about the other tables that use these values? Are we alright with them being different?

It would require research on where these columns are used and how they are used.  It might be necessary to maintain the current loader behavior but it should be investigated first and  determined to be necessary.

usaspending_api/etl/transaction_loaders/field_mappings_fpds.py
```diff
@@ -508,14 +511,15 @@
...
+    "state_code": legal_entity_state_code,
+    "state_description": legal_entity_state_description,
 }
 
 # broker column name -> usaspending column name
 place_of_performance_nonboolean_columns = {
     "place_of_perform_country_c": "location_country_code",
     "place_of_perform_country_n": "country_name",
-    "place_of_performance_state": "state_code",
-    "place_of_perfor_state_desc": "state_description",
```
tony-sappe: I can't comment on line 528:
``` "place_of_performance_zip5": "zip5",```
Should that be removed from `place_of_performance_nonboolean_columns`?

BrianZito: it should

PR #2103, opened 2019-10-31
Staging -> Dev 

emily-brents:
Downmerge for the URL regex `warmfix`.


BrianZito:
Another warmfix is in the pipeline, so closing this to reopen when both are in place

PR #2118, opened 2019-11-05
FIX: [DEV-3634] Run FPDS loader with single, django-managed DB connections

kwhickey:
**Description:**
Use the Django-managed database connections in the (new) FPDS loader code, and showcase an an automated integration test of that modified loader code between broker and usaspending databases.

**Technical details:**
It was found that the use of direct `psycopg2.connect` calls used connections that were not manageable by the `pytest-django` framework. This PR refactors the data loader code to make use of Django's managed connection, rather than self-established connections. This will take advantage of general DB connection settings, but also allow for greater transaction control and testability of the code. Tests were amended after the refactor to keep them running (i.e. mock patching).

NOTE: Though a simple automated integration tests was added to assert correct behavior (in some cases) of the loader, a repeat of QA work (manual regression testing and performance testing/comparisons) needs to be done to the loader code now that the connection-handling has been changed.

    - [ ] Backend
    - [NA] Frontend <OPTIONAL>
    - [NA] Operations <OPTIONAL>
    - [ ] SME <OPTIONAL>
8. [x] Jira Ticket [DEV-3634](https://federal-spending-transparency.atlassian.net/browse/DEV-3634):

```
A scripted data load change. It will impact data and script execution, but not API or Frontend work.
```


usaspending_api/etl/transaction_loaders/fpds_loader.py
```diff
@@ -208,52 +207,52 @@ def _transform_objects(broker_objects):
...
-                if transaction_id:
-                    # Inject the Primary Key of transaction_normalized+transaction_fpds that was found, so that the
-                    # following updates can find it to update
-                    load_object["transaction_fpds"]["transaction_id"] = transaction_id
-                    _update_fpds_transaction(cursor, load_object, transaction_id)
-                else:
-                    # If there is no transaction we create a new one.
-                    transaction_id = _insert_fpds_transaction(cursor, load_object)
-
+    connection.ensure_connection()
```
BrianZito: Why is this here? Did you experience connections dropping for no apparent reason?

usaspending_api/etl/transaction_loaders/tests/integration/test_fpds_loader.py
```diff
@@ -48,36 +55,59 @@ def test_load_ids_from_broker(self):
...
+        # Award's earliest_transaction should point to transaction with earliest action_date
+        assert new_award.earliest_transaction.transaction_unique_id == "101"
+
+        # Given action_date additions, the 3rd one should push into 2011
+        assert [2010, 2010, 2011] == [_.transaction.fiscal_year for _ in usaspending_transactions]
 
 
 def _assemble_dummy_broker_data():
+    """
+    Simple way to reuse the field mappings as a def of all required fields, and then use that as teh basis for
```
BrianZito:
```suggestion
    Simple way to reuse the field mappings as a def of all required fields, and then use that as the basis for
```
You _wish_ you were deadpool

usaspending_api/etl/transaction_loaders/fpds_loader.py
kwhickey: @BrianZito drilling into that source code from Django, it has:
```
    def ensure_connection(self):
        """
        Guarantees that a connection to the database is established.
        """
        if self.connection is None:
            with self.wrap_database_errors:
                self.connect()
```

This is required when dropping down into the lower-level psycopg2 connection maintained and wrapped by Django in its `BaseDatabaseWrapper` class. 

Typically when using just `django.db.connection` or one of the connections from the `django.db.connections` dictionary, you're just getting the wrapper, and any calls to exposed methods on the wrapper will call this `ensure_connection()` under the hood to make sure the connection is actually connected. 

However, in this case, we're using the _unwrapped_ connection directly, so we have to do the work of _ensuring connection_ manually. 

The code that follows this line:
```
with connection.connection.cursor(cursor_factory=DictCursor) as cursor:
```
Notice the double `connection.connection`. The reason we need that one is to use the `DictCursor`, which you can't pass into the wrapped connection (only the unwrapped one), since Django doesn't overload/allow that method signature in the wrapped one.

usaspending_api/etl/transaction_loaders/tests/integration/test_fpds_loader.py
```diff
@@ -48,36 +55,59 @@ def test_load_ids_from_broker(self):
...
+
+        # All transactions reference that award and its id
+        tx_norm_awd = [_.transaction.award for _ in usaspending_transactions]
+        assert usaspending_awards[0] == tx_norm_awd[0] == tx_norm_awd[1] == tx_norm_awd[2]
+        tx_norm_awd_ids = [_.transaction.award_id for _ in usaspending_transactions]
+        assert new_award.id == tx_norm_awd_ids[0] == tx_norm_awd_ids[1] == tx_norm_awd_ids[2]
+
+        # And the single award created should refer to the first transaction processed that spawned the award (101)s
+        # TODO: Check if this is true??
+        assert new_award.transaction_unique_id == "101"
```
kwhickey: @tony-sappe @kbard @BrianZito - can one of you confirm this outcome and assertion is correct?

kbard: Looks right to me.  Is it failing?

In `test_fpds_loader`:
```
        broker_id_list = [101, 201, 301]
        broker_objects = _assemble_detached_award_procurement_records(broker_id_list)
```
Then in `_assemble_detached_award_procurement_records` in `test_fpds_loader`:
```
    for id in id_list:
        dummy_row = _assemble_dummy_broker_data()
        dummy_row["detached_award_procurement_id"] = id
        dummy_row["detached_award_proc_unique"] = str(id)
```
Then in the loader in `field_mappings_fpds`:
```
# broker column name -> usaspending column name
transaction_normalized_nonboolean_columns = {
    "period_of_performance_star": "period_of_performance_start_date",
    ...
    "detached_award_proc_unique": "transaction_unique_id",
}
```
Which is used by `fpds_loader` to map Broker fields to USAS fields.

tony-sappe:
@BrianZito What is the "do not merge" label for?

BrianZito:
Code freeze is until the start of the new sprint, no?

usaspending_api/etl/transaction_loaders/tests/integration/test_fpds_loader.py
kwhickey: yeah, that looks accurate. The value will be set by the loader when looking up if an award exists by the transaction's generated_unique_award_id ... and if not found, create the award referencing that spawning-transaction-id.

But does anyone know why this field is necessary? What use is it associating an award with the _first_ (and only the first) transaction that came about under this award (which may or may not be the earliest or latest transaction). Do we use this in queries in some way, and for what?

Reason I'm asking is that the logic for creation seems arbitrary/brittle, so if we have important queries using this field, we should be sure they're making the right assumptions about what this field references.

@BrianZito @tony-sappe @kbard

tony-sappe:
> Code freeze is until the start of the new sprint, no?

No, "Feature Freeze" is until the staging deploy.

"Code Freeze" is ambiguous and wasn't a successful strategy as applied by the team. Which is why it was dropped from the sprint workflow.

"Feature Freeze" is for planned sprint work (new features and planned bug tickets). It does not apply to resolving bugs/issues discovered after a PR is merged.

usaspending_api/etl/transaction_loaders/tests/integration/test_fpds_loader.py
BrianZito: Do we care? dev-3464 is to get something stable and faster for FPDS loads, and we already have an epic to remove this table structure entirely. I'd argue the viability of these fields is out of scope at this point.

PR #2121, opened 2019-11-07
[DEV-3634] Small changes to new FPDS loader

BrianZito:
**Description:**
Added a concise log that informs users of awards that failed instead of just breaking the load, as well as stylistic changes for the convenience of whomever has to clean up this mess in a year.

**Technical details:**
I used a global. I wish I hadn't, but adding it to the function return values would have been a way bigger change in a ticket that just needs to finish.

    - [x] Backend
8. [ ] Jira Ticket [DEV-3634](https://federal-spending-transparency.atlassian.net/browse/DEV-3634):

```
```

### DIFF
--- a/usaspending_api/broker/management/commands/load_fpds_transactions.py
+++ b/usaspending_api/broker/management/commands/load_fpds_transactions.py
@@ -172,7 +172,7 @@ class Command(BaseCommand):
                     ["{},".format(failure) for failure in failed_ids]
                 )
             )
-            SystemExit(1)
+            raise SystemExit(1)
 
         if options["reload_all"] or options["since_last_load"]:
             # we wait until after the load finishes to update the load date because if this crashes we'll need to load again
--- a/usaspending_api/broker/management/commands/load_fpds_transactions.py
+++ b/usaspending_api/broker/management/commands/load_fpds_transactions.py
@@ -172,6 +172,7 @@ class Command(BaseCommand):
                     ["{},".format(failure) for failure in failed_ids]
                 )
             )
+            SystemExit(1)
 
         if options["reload_all"] or options["since_last_load"]:
             # we wait until after the load finishes to update the load date because if this crashes we'll need to load again
--- a/usaspending_api/broker/management/commands/load_fpds_transactions.py
+++ b/usaspending_api/broker/management/commands/load_fpds_transactions.py
@@ -167,7 +167,7 @@ class Command(BaseCommand):
             update_c_to_d_linkages("contract")
 
         if failed_ids:
-            logger.info(
+            logger.error(
                 "The following detached_award_procurement_ids failed to load: {}".format(
                     ["{},".format(failure) for failure in failed_ids]
                 )
--- a/usaspending_api/broker/management/commands/load_fpds_transactions.py
+++ b/usaspending_api/broker/management/commands/load_fpds_transactions.py
@@ -4,7 +4,7 @@ import re
 import psycopg2
 from datetime import datetime, timezone
 
-from usaspending_api.etl.transaction_loaders.fpds_loader import load_ids, destroy_orphans, delete_stale_fpds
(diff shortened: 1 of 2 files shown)


usaspending_api/broker/management/commands/load_fpds_transactions.py
```diff
@@ -166,6 +166,13 @@ def handle(self, *args, **options):
             )
             update_c_to_d_linkages("contract")
 
+        if failed_ids:
```
tony-sappe: Good idea, we probably want the log messages to be `error` instead of `info` though. Also the script could return a non-zero exit code. using `raise SystemExit(<non-zero integer>)`

usaspending_api/broker/management/commands/load_fpds_transactions.py
```diff
@@ -166,6 +166,13 @@ def handle(self, *args, **options):
...
+            logger.info(
```
BrianZito:
```suggestion
            logger.error(
```

usaspending_api/broker/management/commands/load_fpds_transactions.py
BrianZito: If we had the program exit on failure, do we want to still update the last_load_date? I would think not since that would leave the failed awards lost forever, but not updating it risks the incremental loader falling far out of date if the warnings aren't noticed over a long weekend.

BrianZito: Just spoke to ops. Adding a non-zero code will require updates to groovy scripts, so we might want to leave this enhancement as future work with its own ticket.

tony-sappe: > If we had the program exit on failure, do we want to still update the last_load_date? I would think not since that would leave the failed awards lost forever
I agree with not updating the last load date if errors are encountered.

> but not updating it risks the incremental loader falling far out of date if the warnings aren't noticed over a long weekend.
That's where the non-zero exit code comes in, the pipeline will stop on the FPDS stage and report a failure. Makes it easier for night watch

> Just spoke to ops. Adding a non-zero code will require updates to groovy scripts, so we might want to leave this enhancement as future work with its own ticket.
OPS is talking about are when there is a non-fatal exit code which should allow the pipeline to continue. I am proposing an exit code which would stop the entire pipeline. No updates to groovy scripts are necessary since the default behavior is the desired behavior.

BrianZito: Is stopping the entire pipeline a good idea? That will add ops backlog for all the other tasks that wouldn't run that night, including not refreshing any matviews. If that's really what we want, we could just not add any of this functionality, and let the loader crash on the line with the problem. If we have to rerun the whole load any time there's an issue, we don't get much by carrying on in this stage and listing issues afterwards.

tony-sappe: > Is stopping the entire pipeline a good idea

Yes when a critical stage is unsuccessful. So much of the latter stages (and the website as a whole) are based on transaction data. If there are failures, the pipeline should die so the issues can be addressed immediately. If the pipeline reports an overall success, who is going to diligently look at the logs after each run for failures?

> That will add ops backlog for all the other tasks that wouldn't run that night, including not refreshing any matviews

It isn't adding anything, since the current loader will stop the rest of the pipeline. This falls under the night-watch responsibilities, not operations.

> If that's really what we want, we could just not add any of this functionality, and let the loader crash on the line with the problem

Up to you

> If we have to rerun the whole load any time there's an issue, we don't get much by carrying on in this stage and listing issues afterwards.

Logging the failed IDs and allowing the successful ones to be stored in the DB provides options. Currently the only remedy is to re-run the script across all of the original records since the entire transaction is rolled back. The changes were are discussing could allow someone to attempt to reload only the failed transactions to "get-back-to-good" and if the specific load is successful the last-load-date can be manually updated and the pipeline can be restarted on the next stage.

PR #2123, opened 2019-11-08
[DEV-3864] Elasticsearch ETL improvements

tony-sappe:
**Description:**
Performed refactoring to reduce the potential for human-error in the operations side of indexing data into Elasticsearch.

**Technical details:**
1. Using a new alias for incremental indexing operations (nightly)
    - Removes the need to hardcode the index name in the groovy files
1. Refactored cli arguments for es_rapidloader for more clarity
1. Additional checks during the workflow
1. Removed the redundant json mapping file so only the json template is used
1. Reduced number of disjointed hardcoded values, leveraging settings.py more
1. Refactored codebase to use [pathlib](https://docs.python.org/3.7/library/pathlib.html)'s `Path` class for `BASE_DIR` and new `APP_DIR`
1. Transferred some logic around for better organization

    - [x] Backend
    - [ ] Operations
8. [x] Jira Ticket [DEV-3864](https://federal-spending-transparency.atlassian.net/browse/DEV-3864):

```
Operations script, no impact to API behavior
```


usaspending_api/conftest_helpers.py
```diff
@@ -63,12 +64,12 @@ def _add_contents(self):
     @staticmethod
     def _refresh_materialized_views():
         """
-        This materialized view is used by transaction_delta_view.sql, so
-        we will need to refresh it in order for transaction_delta_view to see
+        This materialized view is used by the es etl view, so
+        we will need to refresh it in order for the view to see
```
BrianZito: Any special reason we wanted to be more generic here?

tony-sappe: A few hardcoded vars were consolidated and stored in settings.py. Code comments can easily become outdated from hardcoded values as well.

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -290,78 +306,91 @@ def es_data_loader(client, fetch_jobs, done_jobs, config):
...
                     "f": "ES Alias Put",
                 }
             )
-        put_alias(client, index, alias_name, award_type_codes)
+        alias_body = {"filter": {"terms": {"type": award_type_codes}}}
+        put_alias(client, index, alias_name, alias_body)
+
+    # ensure the new index is added to the alias used for incremental loads.
+    # If the alias is on multiple indexes, the loads will fail!
+    printf(
```
BrianZito: would a logger be inappropriate here?

usaspending_api/etl/es_settings.json
```diff
@@ -6,10 +6,10 @@
 		}
 	},
 	"settings": {
-		"index.max_result_window" : 500000,
-		"index": {
-			"number_of_replicas": 1,
-			"refresh_interval": "1s"
-		}
+		"index.max_result_window" : null
```
BrianZito: Does this always get overwritten? Seems dangerous...

usaspending_api/etl/es_etl_helpers.py
tony-sappe: Not inappropriate. This was written before the logger was standardized in the app. There is no need to change the current function now and it works very well for the script.

usaspending_api/etl/es_settings.json
tony-sappe: It will be overwritten by the script that parses the json.

I only know of three options (if you have others please share):

1. Duplicate hardcoded values in different files (what is currently done)
2. Move the value to the centralized Python settings file and populate the json (what the PR code does)
3. Store the value here and read the file for settings.py

Since the es_settings.json file is already being updated by the script in the current workflow, this isn't changing much.

tony-sappe: 4. Don't keep `es_settings.json`, move the JSON to be assembled by the script. Effectively storing the same info as a python dictionary populated with settings vars on creation

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -290,78 +306,91 @@ def es_data_loader(client, fetch_jobs, done_jobs, config):
...
+
+
+def swap_aliases(client, index):
     if client.indices.get_alias(index, "*"):
         printf({"msg": 'Removing old aliases for index "{}"'.format(index), "job": None, "f": "ES Alias Drop"})
         client.indices.delete_alias(index, "_all")
 
-    alias_patterns = settings.TRANSACTIONS_INDEX_ROOT + "*"
+    alias_patterns = settings.ES_TRANSACTIONS_READ_ALIAS_PREFIX + "*"
+    old_indexes = []
```
BrianZito: why is this needed?

usaspending_api/etl/es_settings.json
BrianZito: Can you point me to the script that modifies this? I don't like the sound of a mutable settings file, but I'd want to look at what's doing this before I can come up with useful suggestions.

BrianZito: Ok, it looks like just es_etl_helpers. I'm going to read through it more, but I'm confused where the duplication is coming from. Even if there is a reason it needs to be modified in more than one place, could we structure the json to show the seperation? If I see something set to `null` in a settings file, my first instinct is that someone wanted that value null, not that it's being changed behind the scenes.

tony-sappe: Would it be better if `es_settings.json` was renamed to `es_settings_template.json`?

BrianZito: Throw in a comment explaining the situation and I'd call that a good compromise to get this finished.

usaspending_api/etl/es_etl_helpers.py
tony-sappe: `old_indexes` is to store any pre-existing indexes (hopefully only one) which contained any read aliases. Currently the script attempts to close the old indexes, unfortunately since that functionality is absent the AWS ES managed service this PR alters the step to delete the old indexes.

usaspending_api/etl/es_settings.json
tony-sappe: Where?

usaspending_api/etl/es_etl_helpers.py
BrianZito: cool. I just didn't spot where it was used so I wanted to make sure it wasn't dead code

tony-sappe: Lines 379-380 and lines 389-391 in this file

usaspending_api/settings.py
```diff
@@ -8,9 +8,11 @@
 
 from django.db import DEFAULT_DB_ALIAS
 from django.utils.crypto import get_random_string
+from pathlib import Path
 
-# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+# All paths inside the project should be additive to BASE_DIR or APP_DIR
+BASE_DIR = Path(__file__).resolve().parent.parent
+APP_DIR = BASE_DIR / "usaspending_api"
```
kbard:
```
APP_DIR = Path(__file__).resolve().parent
BASE_DIR = APP_DIR.parent
```

usaspending_api/etl/es_settings.json
BrianZito: Right, this is a JSON file. I guess the rename will have to do, since any comment outside the file is less likely to be seen than the code that shows what happens directly. After that I'd be ready to approve.

usaspending_api/common/elasticsearch/elasticsearch_sql_helpers.py
```diff
@@ -0,0 +1,29 @@
...
+    The view is used to populate the Elasticsearch transaction index.
+    This function will ensure the view exists in the database.
+    """
+
+    if verify_database_view_exists(settings.ES_TRANSACTIONS_ETL_VIEW_NAME) and not force:
+        return
+
+    view_file_path = settings.APP_DIR / "database_scripts" / "etl" / "transaction_delta_view.sql"
+
+    with open(str(view_file_path)) as f:
```
kbard:
```suggestion
    view_sql = view_file_path.read_text()
```

usaspending_api/etl/es_settings.json
tony-sappe: The Python library cannot parse a JSON file with comments. Comments are not part of the JSON spec. This [SO answer](https://stackoverflow.com/a/41903882) includes a comment on why they are not in the spec

usaspending_api/settings.py
tony-sappe: I thought of that

usaspending_api/common/elasticsearch/elasticsearch_sql_helpers.py
tony-sappe: Looks good, I'm still learning how to best use pathlib

usaspending_api/etl/es_settings.json
BrianZito: That's what I was trying to say. I had my struggle with JSON and comments long ago; just renaming the file is the best hint I think we can give to future devs.

usaspending_api/settings.py
```diff
@@ -82,12 +82,15 @@
 ES_HOSTNAME = ""
 if not ES_HOSTNAME:
     ES_HOSTNAME = os.environ.get("ES_HOSTNAME")
-TRANSACTIONS_INDEX_ROOT = os.environ.get("ES_TRX_ROOT") or "future-transactions"
+ES_TRANSACTIONS_ETL_VIEW_NAME = "transaction_delta_view"
+ES_TRANSACTIONS_MAX_RESULT_WINDOW = 50000
+ES_TRANSACTIONS_NAME_SUFFIX = "transactions"
+ES_TRANSACTIONS_READ_ALIAS_PREFIX = "transaction-records"
+ES_TRANSACTIONS_WRITE_ALIAS = "transaction-load-alias"
```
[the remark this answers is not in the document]
tony-sappe: Currently there isn't a dedicated "elasticsearch module" to contain these constants. Not a bad idea, but out of scope for this PR. I wouldn't create a new module just to handle the constants, there would need to be additional benefits in the reorg.

usaspending_api/settings.py
```diff
@@ -82,12 +82,15 @@
 ES_HOSTNAME = ""
 if not ES_HOSTNAME:
     ES_HOSTNAME = os.environ.get("ES_HOSTNAME")
-TRANSACTIONS_INDEX_ROOT = os.environ.get("ES_TRX_ROOT") or "future-transactions"
+ES_TRANSACTIONS_ETL_VIEW_NAME = "transaction_delta_view"
+ES_TRANSACTIONS_MAX_RESULT_WINDOW = 50000
+ES_TRANSACTIONS_NAME_SUFFIX = "transactions"
+ES_TRANSACTIONS_READ_ALIAS_PREFIX = "transaction-records"
```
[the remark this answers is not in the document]
tony-sappe: Because ES allows wildcard searches, I thought it prudent to differentiate the index names and the alias.
`transaction-records*` won't pickup the underlying index name `*transactions` (or the sometimes previously  occurring `transactions*` index names)

Also, a search on `transactions*` won't hit the aliases.

Using `transactions-read*` could work, but there is an overlap of "transactions" which could lead to incorrect results if devs aren't careful and aware of all the indexes in the cluster.

usaspending_api/etl/management/commands/es_rapidloader.py
```diff
@@ -208,50 +166,103 @@ def controller(self):
...
+            set_final_index_config(self.elasticsearch_client, self.config["index_name"])
+            swap_aliases(self.elasticsearch_client, self.config["index_name"])
 
         if self.config["snapshot"]:
             printf({"msg": "Taking snapshot"})
-            take_snapshot(ES, self.config["index_name"], settings.ES_REPOSITORY)
+            take_snapshot(self.elasticsearch_client, self.config["index_name"], settings.ES_REPOSITORY)
+
+        if self.config["is_incremental_load"]:
+            printf({"msg": "Updating last load record in DB with {}".format(self.config["processing_start_datetime"])})
```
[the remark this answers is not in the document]
tony-sappe: `table.field` is abstracted away from this script. Would either need to import the model or hard-code the tablename and field in this log statement. Neither is better than what is currently there.

Other ETL scripts don't contain this log message. Either I scrub this one to avoid confusion, keep it as-is, or change to it like suggestion:
```Storing datetime {} for next incremental load```

File: usaspending_api/etl/management/commands/es_rapidloader.py [hunk 10]
```diff
@@ -208,50 +166,103 @@ def controller(self):
...
+
+    config["max_query_size"] = settings.ES_TRANSACTIONS_MAX_RESULT_WINDOW
+
+    config["is_incremental_load"] = not bool(config["create_new_index"]) and (
+        config["starting_date"] != default_datetime
+    )
+
+    if config["is_incremental_load"]:
+        if config["index_name"]:
+            printf({"msg": "Ignoring provided index name, using alias for incremental load"})
```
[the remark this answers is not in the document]
tony-sappe: Yes, but it's going to be the same alias stored in settings.py

PR #2126, opened 2019-11-12
Dev 1669 support older filters

BrianZito:
**Description:**
program_activity and object_class are supported in v1 awards endpoint, but not v2. More testing needs to be done, but I wanted to display this a little earlier in case I'm way off with how we architect adding things to endpoints.

**Technical details:**

    - [ ] Backend
8. [x] Jira Ticket [DEV-3869](https://federal-spending-transparency.atlassian.net/browse/DEV-3869):

```
```
**Testing Results**
```
Object Class:
{
    "filters": {
        "award_type_codes": [
            "A",
            "B",
            "C",
            "D"
        ],
        "object_class": ["257", "231", "254", "220", "233", "240", "260", "210", "251", "252", "255", "253"]
    },
    "fields": [
        "Award ID",
        "Recipient Name",
        "Start Date",
        "End Date",
        "Award Amount",
        "Awarding Agency",
        "Awarding Sub Agency",
        "Contract Award Type"
    ],
    "page": 1,
    "limit": 60,
    "sort": "Award Amount",
    "order": "desc",
    "subawards": false
}
returns in 1185 ms against DEV db


{
    "filters": {
        "tas_codes": [
            {
                "aid": "070",
                "main": "0702"
            }
        ],
        "object_class": ["940","410","430"],
        "award_type_codes": [
            "A",
            "B",
            "C",
            "D"
        ]
    },
    "subawards": false,
    "fields": [
        "Award ID",
        "Recipient Name",
        "Start Date",
        "End Date",
        "Award Amount",
        "Awarding Agency",
        "Awarding Sub Agency",
        "Contract Award Type"
    ],
    "limit": 60,
    "order": "desc",
    "page": 1,
    "sort": "Award Amount"
}
Returns in 52ms against Dev DB
```


tony-sappe:
@BrianZito I don't see the necessary FK in {{universal_transaction_matview}} or the award matviews to support these filters.

BrianZito:
> @BrianZito I don't see the necessary FK in `universal_transaction_matview` or the award matviews to support these filters.

This PR is now designed to replicate the functionality in the way that v1 did it, in order to not slow down performance with the more common advanced searches or require a backfill.

usaspending_api/common/validator/award_filter.py
```diff
@@ -26,6 +26,8 @@
     {"name": "recipient_search_text", "type": "array", "array_type": "text", "text_type": "search"},
     {"name": "recipient_type_names", "type": "array", "array_type": "text", "text_type": "search"},
     {"name": "set_aside_type_codes", "type": "array", "array_type": "text", "text_type": "search"},
+    {"name": "object_class", "type": "array", "array_type": "text", "text_type": "search"},
+    {"name": "program_activity", "type": "array", "array_type": "integer", "array_max": maxsize},
```
sethstoudenmier: Here you specify that `program_activity` is an array of integers but then in your test data you provide an array of strings. Depending on which one is correct one or the other should be updated. 

Here is the test data taken from below:
```
def legacy_filters():
    return {"award_type_codes": ["A", "B"], "object_class": ["111"], "program_activity": ["222"]}
```

BrianZito: good catch. A quirk of TinyShield is that it will work with either by trying to cast (think `==` instead of `===`), but I'd rather encourage strong typing in my examples

usaspending_api/awards/v2/filters/matview_filters.py
```diff
@@ -293,22 +293,19 @@ def keyword_parse(keyword):
 
         # Federal Account Filter
```
sethstoudenmier: With removing the `faba_flag` should this comment be removed? The same thing below.

BrianZito: No, because these are still the filters used by the federal account page. I assume the comment was denoting that, and not the the flag is about to be set.

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
```diff
@@ -172,6 +172,8 @@ List of table columns
 + `set_aside_type_codes`: `NONE` (optional, array[string])
 + `extent_competed_type_codes`: `A` (optional, array[string])
 + `tas_codes` (optional, array[TASCodeObject], fixed-type)
++ `object_class` (optional, array[string])
```
sethstoudenmier: Looking through `view_selector.py` and the usage of `get_view_queryset()` there are a lot more API Contracts that could be updated to reflect this new filter.

BrianZito: Do we want to be announcing uses of endpoints we haven't worked with or tested?

sethstoudenmier: To me, it boils down to the purpose of the API Contracts. I think of it as a way to capture the current functionality of the endpoint regardless of whether or not it is used.

BrianZito: I think of it as a document of promises we intend to keep. If we had been documenting the v1 endpoints, would you have included all the possible joins you could use, or just the ones used by the app we were comfortable supporting?

BrianZito: (this might quickly become a discussion to move off of github)

usaspending_api/awards/v2/filters/matview_filters.py
tony-sappe: What is performance like with all of the joins originating from the matviews?

The original SQL we discussed yesterday used the `faba_flag` for a separate query (unsure if it was run beforehand or as a sub-query).

usaspending_api/common/validator/award_filter.py
tony-sappe: > A quirk of TinyShield

It was an intentional design decision. Give API users the benefit of the doubt, if the string is only numeric typecast and continue.

usaspending_api/awards/v2/filters/matview_filters.py
BrianZito: Quite good, based on the testing against DEV DB shown in the PR description. The framework that was already partially set up was VERY slow (~30 seconds in local DB), likely due to the giant IN clause. This was rearranged to mimic what v1 was doing, and seems to still be performant for the use cases UI is using it for.

tony-sappe: Did you test SpendingByAwardCount? That is the current problem-child

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
```diff
@@ -172,6 +172,8 @@ List of table columns
 + `set_aside_type_codes`: `NONE` (optional, array[string])
```
tony-sappe: Same updates to usaspending_api/api_contracts/contracts/v2/search/spending_by_award_count.md?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -109,6 +110,20 @@ def validate_request_data(request_data):
         models = [
             {"name": "fields", "key": "fields", "type": "array", "array_type": "text", "text_type": "search", "min": 1},
             {"name": "subawards", "key": "subawards", "type": "boolean", "default": False},
+            {
```
tony-sappe: Not a blocker. Since these filters are identically defined in two files, they could be moved into `usaspending-api/usaspending_api/common/validator/award_filter.py` as a separate dictionary from `AWARD_FILTER` (e.g. `LEGACY_AWARD_FILTER`)

Up to you if that is done.

usaspending_api/awards/v2/filters/matview_filters.py
BrianZito: Yes, although only for accuracy, not performance. The latter relied on my linking to DEV DB, which isn't possible now that we have to put the matviews back.

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: I actually did this (with the same variable name), then thought better of it since the other "extra" fields above weren't done that way. I'm also indifferent to this one.

usaspending_api/awards/v2/filters/matview_filters.py
tony-sappe: hmmm, and staging is swamped now

tony-sappe: **very** quick checks appear OK, we'll see how they are in staging next week.

PR #2129, opened 2019-11-13
Staging -> Dev for dev 3746

BrianZito:
**Description:**
Some endpoints were missed in the original work. This is just to bring the warmfix down to DEV


BrianZito:
Doing PR in a different way to avoid conflicts

PR #2133, opened 2019-11-15
[DEV-1914] Add generated_internal_id to keyword response

tony-sappe:
**Description:**
The Keyword page needs to be able to link to award summary pages using the `generated_unique_award_id`. This PR adds the field to ES transaction index and includes it in the response (when present) as `generated_internal_id`

**Technical details:**
- Added new field to ES mapping / ETL view / and column list
- Fixed minor issue with ETL CSV path names (not broken, just not desired)
- Modified ES ETL view to use matview columns more and to be most based on transaction data instead of award data

    - [x] Backend
    - [x] Frontend
8. [x] Jira Ticket [DEV-1914](https://federal-spending-transparency.atlassian.net/browse/DEV-1914):
    - [x] Performance evaluation of affected (API | Script)

```
No edits to matviews (only 1 view)
```


kwhickey:
@tony-sappe why call it `generated_internal_id`? I think just leaving it what the name is in the DB would cut down on confusion and trouble tracing. Minimally, and especially since many of these datasets mix transaction and award-level data, I think it should mention what it's an ID _of_ (an award). If it needs to change, how about `generated_internal_award_id`. Though now that I read that, generated and internal are kind of redundant, and usually internal implies the data wouldn't make sense externally, but because it's made from a combination of business keys, it is comprehensible externally.

tony-sappe:
> @tony-sappe why call it `generated_internal_id`? I think just leaving it what the name is in the DB would cut down on confusion and trouble tracing. Minimally, and especially since many of these datasets mix transaction and award-level data, I think it should mention what it's an ID _of_ (an award). If it needs to change, how about `generated_internal_award_id`. Though now that I read that, generated and internal are kind of redundant, and usually internal implies the data wouldn't make sense externally, but because it's made from a combination of business keys, it is comprehensible externally.

@kwhickey I followed the naming convention from other API responses.

PR #2136, opened 2019-11-18
dev-3634 removed cleanup

BrianZito:
Investigation has found that any angle of trying to find orphaned values takes prohibitively long, and that the amount of orphans has little impact on runtime. Given that work is under way to remove both of these tables altogether, I think it's a better choice to let this clutter accumulate for the time being, and invest time in either removing these tables, or developing a more complete way to deal with this problem if the tables can't go for some reason.

### DIFF
--- a/usaspending_api/broker/management/commands/load_fpds_transactions.py
+++ b/usaspending_api/broker/management/commands/load_fpds_transactions.py
@@ -151,8 +151,9 @@ class Command(BaseCommand):
             self.load_fpds_incrementally(last_load)
 
         if self.modified_award_ids:
-            logger.info("cleaning orphaned metadata")
-            destroy_orphans()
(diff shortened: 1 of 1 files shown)


usaspending_api/broker/management/commands/load_fpds_transactions.py
```diff
@@ -151,8 +151,9 @@ def handle(self, *args, **options):
             self.load_fpds_incrementally(last_load)
 
         if self.modified_award_ids:
-            logger.info("cleaning orphaned metadata")
-            destroy_orphans()
+            # TODO: reactivate once this is performant. Currently the tables are too large to allow
+            # logger.info("cleaning orphaned metadata")
```
kwhickey: @BrianZito - I think this would have been a good place for `logger.warning("WARNING: Cleanup of orphaned metadata is being skipped")`

PR #2137, opened 2019-11-19
Ensure that filters work with spending_over_time

sethstoudenmier:
**Description:**
Get the spending_over_time endpoint into a more demo-able state.

**Technical details:**
Used a different bucket aggregation that works better for dates and made sure that the filters work with spending_over_time filter.


BrianZito:
*Ensure

usaspending_api/search/v2/views/spending_over_time.py
```diff
@@ -156,7 +156,7 @@ def post(self, request):
         if self.group == "fiscal_year":
             end_date = "{}-09-30".format(current_fy)
         else:
-            end_date = "{}-{}-30".format(current_fy, datetime.now(timezone.utc).month)
+            end_date = "{}-{}-28".format(current_fy, datetime.now(timezone.utc).month)
```
BrianZito: so if something is on the 31st of the month it doesn't count as being in that month?

BrianZito: It looks like calendar library can do it: https://stackoverflow.com/questions/9481136/how-to-find-number-of-days-in-the-current-month/9481305

If you don't want to import more things, you could make a datetime of day one of the next month, subtract a day, then use _that_ date to format with.

sethstoudenmier: Just saw the same thing. Doing that right now

PR #2139, opened 2019-11-19
added slash back to local download path

BrianZito:
Recent fancy string formatting changes cause local download files to save to the wrong folder, due to all slashes being deleted. This remedies the issue.

### DIFF
--- a/usaspending_api/settings.py
+++ b/usaspending_api/settings.py
@@ -244,8 +244,8 @@ USE_TZ = True
 # https://docs.djangoproject.com/en/1.11/howto/static-files/
 
 STATIC_URL = "/static/"
-STATIC_ROOT = str(APP_DIR / "static/")
-STATICFILES_DIRS = (str(APP_DIR / "static_doc_files"),)
+STATIC_ROOT = str(APP_DIR / "static/") + "/"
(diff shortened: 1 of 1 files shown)


tony-sappe:
Waiting for @jdodd2 to finish before merging and PRs

usaspending_api/settings.py
```diff
@@ -48,7 +48,7 @@
     USASPENDING_AWS_REGION = os.environ.get("USASPENDING_AWS_REGION")
 
 # AWS locations for CSV files
-CSV_LOCAL_PATH = str(BASE_DIR / "csv_downloads")
+CSV_LOCAL_PATH = str(BASE_DIR / "csv_downloads") + "/"
```
tony-sappe: Unable to comment on these lines in settings.py which also should be directories:
Line 248, 249,

usaspending_api/settings.py
```diff
@@ -48,7 +48,7 @@
...
 DOWNLOAD_ENV = ""
 BULK_DOWNLOAD_LOCAL_PATH = str(BASE_DIR / "bulk_downloads")
```
tony-sappe: and this line?

BrianZito: I don't see this setting actually used for anything. Might be wise to update it anyway

PR #2145, opened 2019-11-23
[DEV-597] TSV support in Downloads

tony-sappe:
**Description:**
Added `tsv` to the `file_format` API request parameter for all download endpoints

**Technical details:**

- Renamed a number of files, functions, and variables which contained "csv" since they are delimiter-agnostic.
- Supported far more endpoints that asked for in DEV-597 since they will most likely be future requests
- TSV values with tab characters are quoted automatically by psql

    - [ ] Backend
    - [ ] Frontend
8. [x] Jira Ticket [DEV-597](https://federal-spending-transparency.atlassian.net/browse/DEV-597):
    - [x] Performance evaluation of affected Download: no changes detected

```
No materialized views touched
No need or operations changes
```


usaspending_api/common/csv_helpers.py
```diff
@@ -5,31 +5,31 @@
 from usaspending_api.common.retrieve_file_from_uri import RetrieveFileFromUri
 
 
-def count_rows_in_csv_file(filename, has_header=True, safe=True):
+def count_rows_in_csv_file(filename, has_header=True, safe=True, delimiter=","):
```
BrianZito: Name is now incorrect.

usaspending_api/common/csv_helpers.py
```diff
@@ -5,31 +5,31 @@
...
         else:
-            row_count = sum(1 for row in csv.reader(f))
+            row_count = sum(1 for row in csv.reader(f, delimiter=delimiter))
     if has_header and row_count > 0:
         row_count -= 1
 
     return row_count
 
 
 # Function inspired by Ben Welsh: https://gist.github.com/palewire/596056
```
BrianZito: Do we really need inspiration comments?

usaspending_api/download/v2/request_validations.py
```diff
@@ -274,3 +277,12 @@ def _validate_required_parameters(request_data, required_parameters):
     for required_param in required_parameters:
         if required_param not in request_data:
             raise InvalidParameterException("Missing one or more required body parameters: {}".format(required_param))
+
+
+def _validate_file_formats(json_request: dict) -> str:
```
BrianZito: This function is a violation of Single Responsibility Principle. Is there any reason we can't move the exception logic into the existing `_validate_filters` function, and return this code to just being the default logic?

usaspending_api/download/filestreaming/download_generation.py
```diff
@@ -305,14 +305,17 @@ def wait_for_process(process, start_time, download_job):
 
     # Let the thread run until it finishes (max MAX_VISIBILITY_TIMEOUT), with a buffer of DOWNLOAD_VISIBILITY_TIMEOUT
     sleep_count = 0
-    while process.is_alive() and (time.perf_counter() - start_time) < MAX_VISIBILITY_TIMEOUT:
+    while process.is_alive():
+        if not download_job.monthly_download and (time.perf_counter() - start_time) > MAX_VISIBILITY_TIMEOUT:
+            break
```
BrianZito: What exactly was the impetus of this? Are we trying to allow the monthly download to take infinite time?

usaspending_api/download/filestreaming/download_generation.py
```diff
@@ -305,14 +305,17 @@ def wait_for_process(process, start_time, download_job):
...
         if sleep_count < 10:
             time.sleep(WAIT_FOR_PROCESS_SLEEP / 5)
         else:
             time.sleep(WAIT_FOR_PROCESS_SLEEP)
         sleep_count += 1
 
-    if (time.perf_counter() - start_time) >= MAX_VISIBILITY_TIMEOUT or process.exitcode != 0:
+    over_time = (time.perf_counter() - start_time) >= MAX_VISIBILITY_TIMEOUT
```
BrianZito: Why calculate this here when we use the same formula a few operations ago? Is accuracy that important, or are we worried about the sleeps?

usaspending_api/common/csv_helpers.py
tony-sappe: updated

usaspending_api/common/csv_helpers.py
tony-sappe: Did you look at the linked code before making this comment?

usaspending_api/download/v2/request_validations.py
tony-sappe: This follows the current workflow in this file. Ironically it is similar to `_validate_filters()`.

I'm not agreeing with the pattern, but it seems out of scope to refactor all of the validation logic for this new validation.

usaspending_api/download/filestreaming/download_generation.py
tony-sappe: If a download takes to long, kill it.

Unless it is a monthly download job, then disable the timeout check

usaspending_api/download/filestreaming/download_generation.py
tony-sappe: My changes were only to add a check for monthly download jobs. I am unsure of the need to recalculate. It could definitely be cleaner.

usaspending_api/download/v2/request_validations.py
BrianZito: As far as I can tell this would only be a few lines being moved from one function to another, and would prevent things from getting further obfuscated. I don't see any other lines using a "validate" function to generate a value.

tony-sappe: `_validate_account_level()`?

BrianZito: ...nevermind

usaspending_api/common/csv_helpers.py
tony-sappe: The code is almost a copy from the link in the comment. I believe the original dev who brought this code into the codebase did a straight-copy. It has since been improved with better variable names, and Python3 syntax.

This comment is a "weak attribution" to a snippet of code in the ether of the internet. Is it Open Source? Are we allowed to use it? Should it be removed?

At minimum, a link to the original author is provided to enlighten future devs and the public.

usaspending_api/download/tests/integration/test_download_awards.py
```diff
@@ -122,9 +122,23 @@ def test_download_awards_without_columns(client, download_test_data):
     assert ".zip" in resp.json()["url"]
 
 
+@pytest.mark.django_db
+def test_tsv_download_awards_without_columns(client, download_test_data):
+    download_generation.retrieve_db_string = Mock(return_value=generate_test_db_connection_string())
+    resp = client.post(
+        "/api/v2/download/awards/",
+        content_type="application/json",
+        data=json.dumps({"filters": {"award_type_codes": []}, "columns": []}),
```
sethstoudenmier: Going off the contract, shouldn't the tests have `file_format` in the data that is passed through the post?

```suggestion
        data=json.dumps({"filters": {"award_type_codes": []}, "columns": [], "file_format": "tsv"}),
```

tony-sappe: Isn't the field marked as optional?

sethstoudenmier: Yes, but since this test is testing that it works wouldn't you want to actually pass it along? All the other download tests are already testing that it works fine without it.

tony-sappe: AH, I understand

PR #2146, opened 2019-11-25
DEV-3995/DEV-4003: Fix for custom account download

kbard:
**Description:**

Originally there was concern that the custom account download file did not match the raw File B download for any given quarter, but once we hammered out the business logic with the Schema Team, we learned that it is possible for the files not to match.

However, in the process of researching this issue we discovered a small bug whereby results were not grouping by all of the correct fields.  Additionally, sorting was not deterministic so it was possible to receive different results for the same filtering criteria.

This pull request adds the Direct/Reimbursable flag to the uniqueness check, adds the table primary key to the sort, and adds the report period fiscal year/fiscal quarter to the download file.

    - [x] Backend
8. [x] Jira Ticket [DEV-3995](https://federal-spending-transparency.atlassian.net/browse/DEV-3995)/[DEV-4003](https://federal-spending-transparency.atlassian.net/browse/DEV-4003):


usaspending_api/download/v2/download_column_historical_lookups.py
```diff
@@ -1469,6 +1473,10 @@
         ),
         "federal_account": OrderedDict(
             [
+                (
+                    "last_reported_submission_period",
+                    "last_reported_submission_period",
```
tony-sappe: The current pattern is to set the "DB field name" value to null if it will be populated by an annotation

usaspending_api/accounts/v2/filters/account_download.py
```diff
@@ -90,10 +91,14 @@ def account_download_filter(account_type, download_table, filters, account_level
...
                 "program_activity__program_activity_code",
                 "object_class__object_class",
+                "object_class__direct_reimbursable",
             ],
         }
         distinct_cols = unique_columns_mapping[account_type]
-        order_by_cols = distinct_cols + ["-reporting_period_start"]
+        order_by_cols = distinct_cols + [
+            "-reporting_period_start",
+            "-pk",
```
tony-sappe: Does this ensure the results are more correct or is it simply to be deterministic?

If this is valuable, does the pk make it fragile to reloads like we discovered with the FPDS loader?

kbard: This is purely to make it deterministic.  I don't know that there's a "correct" order if all of the distinct columns are the same.  The non-determinism I observed was eliminated when we added Direct/Reimbursable flag so this is just a bit of defensive programming.

usaspending_api/download/v2/download_column_historical_lookups.py
kbard: Annotations work differently in the custom account download part of the site.  I spent 45 minutes trying to make them work like the other downloads and failed.  This is bigger than a breadbox and I would like to kick the can down the road so I can make progress on other tickets.

The code that I checked in works like other code for these types of downloads.

PR #2147, opened 2019-11-25
updated logs to not include duplicate award ids

BrianZito:


usaspending_api/broker/management/commands/load_fpds_transactions.py
```diff
@@ -156,7 +156,7 @@ def handle(self, *args, **options):
             # destroy_orphans()
             logger.info(
                 "updating award values ({} awards touched by transaction modifications)".format(
-                    len(self.modified_award_ids)
+                    len(tuple(self.modified_award_ids))
```
BrianZito:
```suggestion
                    len(set(self.modified_award_ids))
```

PR #2151, opened 2019-11-25
Mod/dev 3841 add url to downloads

BrianZito:
**Description:**
Added a url column, by appending website url to the generated unique key

**Technical details:**
Field was added to annotations instead of to the awards model

    - [x] Backend
8. [x] Jira Ticket [DEV-3841](https://federal-spending-transparency.atlassian.net/browse/DEV-3841):

```
```


usaspending_api/download/helpers/download_annotation_functions.py
```diff
@@ -25,6 +28,7 @@ def universal_award_matview_annotations():
         "federal_accounts_funding_this_award": StringAgg(
             "award__financial_set__treasury_account__federal_account__federal_account_code", ";", distinct=True
         ),
+        "usaspending_permalink": Concat(Value(HOST + "/#/award/"), "award__generated_unique_award_id"),
```
tony-sappe: Unfortunate question: do we need to url-encode the GUAI? (I believe yes. cc @emily-brents)
Next unfortunate question: how do we do this in SQL?

I found [this extension](https://github.com/okbob/url_encode) but it isn't supported in AWS RDS.
See [SO](https://stackoverflow.com/questions/10318014/javascript-encodeuri-like-function-in-postgresql)

emily-brents: Certain special characters have to be URL encoded for API calls, but the frontend currently results in a 404 for most special characters. There is an issue in Sprint 97 for the frontend special characters that hasn't been fixed yet -  https://federal-spending-transparency.atlassian.net/browse/DEV-3994. I would assume that the fix for that involves URL encoding the GUAI, but can't say for sure.

usaspending_api/settings.py
```diff
@@ -37,7 +37,8 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = False
 
-ALLOWED_HOSTS = ["*"]
+HOST = "localhost:3000"  # To be overwritten by Ansible
```
tony-sappe: The comment is unnecessary. Other values in this file are overwritten by the deploy process already.

usaspending_api/download/helpers/download_annotation_functions.py
tony-sappe: Choosing to defer this unfortunate "edge case" for future work.

usaspending_api/settings.py
BrianZito: That sounds like the comment is insufficient, not unnecessary. This line of code would provide a reasonable developer with the assumption that HOST will be equal to that value. Our ops code breaks that assumption by changing the value behind the scenes (which I, for one, had no idea of before working on this ticket). That's a textbook reason to have a comment. 

I'm sure other values in this file operate the same way, and should I encounter them I would like to make similar comments about those as well. Our last dev "chapter" meeting did have a broker dev say you can never have enough comments.

Even if you find it unnecessary, does it do any harm?

usaspending_api/download/helpers/download_annotation_functions.py
```diff
@@ -1,6 +1,10 @@
 from django.contrib.postgres.aggregates import StringAgg
 from usaspending_api.common.helpers.orm_helpers import FiscalYear
+from usaspending_api.settings import HOST
+from django.db.models.functions import Concat
+from django.db.models import Value
 
+award_url = HOST + "/#/award/" if "localhost" in HOST else "https://" + HOST + "/#/award/"
```
tony-sappe:
```suggestion
AWARD_URL = f"{HOST}/#/award/" if "localhost" in HOST else f"https://{HOST}/#/award/"
```

usaspending_api/settings.py
tony-sappe: #1: there is nothing questionable, odd, or confusing about this variable. It has a clear name ("host") and is populated with the default hostname from django runserver.
#2: What occurs during a deploy doesn't alter the code or behavior of the variable, it simply alters the value. This is standard for many systems.
#3: There is no reason to specifically list Ansible, if the deploy process moves to a new tool this comment will most likely be forgotten and become outdated. So yes, it can "do harm"

> Our last dev "chapter" meeting did have a broker dev say you can never have enough comments.

This is obviously not true. It is terribly easy to have too many comments. I'll admit there is room for more "good comments" but a comment isn't guaranteed to provide value.

usaspending_api/download/tests/integration/test_download_status.py
```diff
@@ -204,7 +204,7 @@ def test_download_contract_status(client, download_test_data, refresh_matviews):
 
     assert resp.status_code == status.HTTP_200_OK
     assert resp.json()["total_rows"] == 1
-    assert resp.json()["total_columns"] == len(query_paths["transaction"]["d1"])
```
tony-sappe: Why was this changed? Contracts are FPDS aka D1, IDVs are also from FPDS but they are not awards

BrianZito: That change was to have it match the slightly longer map during a period of commits where idv_transaction_history had the column while transaction still didn't. With a later change I realized it needed to be on transaction to, bringing these dicts to the same length and making this test a false negative.

PR #2153, opened 2019-11-29
Prevent matview tests from fully generating/recreating matviews every test

kbard:
**Description:**

This is a non-ticketed minor improvement to our automated tests to help with performance.  @tony-sappe noticed that we fully recreated matviews (including re-generating the matview files) every time we performed a test with matviews.

This tweak eliminates the file re-generation and defaults to refreshing the existing matviews rather than recreating them.

This knocked 40 seconds off a single test file's runtime (`test_state.py`) which should translate to a modest performance improvement across the entire test suite.

EDIT:  Looks like it shaved about 3 minutes off of overall test run time from 19 min 47 sec for a previous test to 16 min 23 sec for the test on this very pull request.


BrianZito:
If we aren't recreating the matviews, is there any point even refreshing them? The tests are now interdependent either way.

kbard:
> If we aren't recreating the matviews, is there any point even refreshing them?

We create the matviews once and refresh them as needed.

> The tests are now interdependent either way.

I have no idea what this means.

BrianZito:
> I have no idea what this means.

If I understand correctly, creating a matview will capture any changes to the model that it's pulling from, whereas refreshing just captures changes to the data. Therefore, if anything happened in a test that changed a model (very unlikely), later tests could break. In theory, changing one test could now change the results of a different one, and tests running in a different order could cause unstable results. That is test interdependence.

Personally, I don't mind that given the tests we have, but I'm curious if there's potential to optimize more for speed if we're already willing to sacrifice the "purity" of independent units. As far as I know we don't rely on refreshing the matviews to get test data into them, since we use mocked model data for our tests.

Sorry for the text wall on a semi-holiday.

kbard:
> If I understand correctly, creating a matview will capture any changes to the model that it's pulling from, whereas refreshing just captures changes to the data.

I would be pretty atypical to change a model in a test.  Models tend to be quite static.  Changing a model requires rolling out migrations.

> changing one test could now change the results of a different one

I'm not seeing it.  Tests are run in transactions.  The matview refresh would be rolled back at the end of the test.  How does this affect subsequent tests?

Back to vacationing.

usaspending_api/common/helpers/generic_helper.py
```diff
@@ -18,11 +18,11 @@
 )
 from usaspending_api.references.models import Agency
 
-logger = logging.getLogger(__name__)
 
+logger = logging.getLogger(__name__)
 TEMP_SQL_FILES = [str(DEFAULT_MATIVEW_DIR / val["sql_filename"]) for val in MATERIALIZED_VIEWS.values()]
-
 VIEW_SQL_FILES = [str(val) for val in OVERLAY_VIEWS]
+REFRESH_MATVIEWS_SQL = " ".join(f"refresh materialized view {matview_name};" for matview_name in MATERIALIZED_VIEWS)
```
tony-sappe: The SQL to refresh matviews is stored here: usaspending_api/database_scripts/matviews/refresh_matviews.sql

tony-sappe:
> > I have no idea what this means.
> 
> If I understand correctly, creating a matview will capture any changes to the model that it's pulling from, whereas refreshing just captures changes to the data. Therefore, if anything happened in a test that changed a model (very unlikely), later tests could break. In theory, changing one test could now change the results of a different one, and tests running in a different order could cause unstable results. That is test interdependence.
> 
> Personally, I don't mind that given the tests we have, but I'm curious if there's potential to optimize more for speed if we're already willing to sacrifice the "purity" of independent units. As far as I know we don't rely on refreshing the matviews to get test data into them, since we use mocked model data for our tests.
> 
> Sorry for the text wall on a semi-holiday.

-------------------------

> Therefore, if anything happened in a test that changed a model (very unlikely)

Tests should **never** alter a model past the migrations. I cannot think of a valid justification for that.

> As far as I know we don't rely on refreshing the matviews to get test data into them

Yes we do, that's the reason recreating/refreshing matviews is already a text fixture.

It's difficult to see how this PR exacerbates any negative behavior such as test interdependence.

@BrianZito

PR #2164, opened 2019-12-10
[DEV-4032] Update ES transaction index to support new ES based endpoints

sethstoudenmier:
**Description:**
Adjust the current ES transaction index, loader, and view to support future ES based endpoints such as:
* `v2/search/spending_over_time/`
* `v2/search/spending_by_geography/`
* and the breakdown of the `v2/search/spending_by_category/` endpoint

**Technical details:**
[DEV-4032_Needed_ES_Fields.txt](https://github.com/fedspendingtransparency/usaspending-api/files/3942541/DEV-4032_Needed_ES_Fields.txt)
The attached file above are the different columns used across the endpoints that we are converting to ES and the filters for Advanced Search. In addition to those, additional changes were made to Keyword search since fields were altered to be consistent. The changes were tested to make sure that the keyword search still has the same partial or exact match behavior that was expected with the previous combination of `text` and `keyword` types.

Some important parts to call out:
* The series of LPAD and CAST in universal_transaction_matview are in place to solve data issues where county and congressional codes are stored in many different ways for the same county (e.g. '01', '1', '1.0') and sometimes have letters in the beginning.
* The left join for Federal Accounts solves the issue of a transaction having multiple Federal Accounts. Since we need more than just one item from each Federal Account they needed to be stored in ES as an array of objects instead of an array of <insert data type>.
* Edits to matview_manager were added by myself because a test was failing without that a while back (from what I remember), but now everything seems to be fine. The `transaction_delta_view` does not need to stick around so took this out.
* The edits to both remove "{}.raw".format() and update TRANSACTIONS_LOOKUP were to move towards what is the ES convention to use "keyword" instead of "raw" and to have "keyword" as the multi-field instead of "text".

1. [X] Unit & integration tests updated (N/A)
    - [X] Backend
    - [ ] Operations <OPTIONAL>
8. [X] Jira Ticket [DEV-4032](https://federal-spending-transparency.atlassian.net/browse/DEV-4032):

No tests were updated, but how we handle the `transaction_delta_view` when running tests was altered.

**UPDATES:**
* Added the changes to county_code and congressional_code for both place_of_performance and recipient_location to all matviews that currently pull that data from FPDS and FABS tables.
* Updated a few transaction index fields to also include the keyword type
* Updated `es_rapidloader` to include a flag to skip the step where it deletes old indexes and swaps the aliases. This will assist with deploying changes to the ES index fields so that it does not break code. Will require manual deletion of the old index and swapping of aliases after being used.


usaspending_api/database_scripts/etl/transaction_delta_view.sql
```diff
@@ -35,9 +35,11 @@ SELECT
   UTM.award_category,
   UTM.recipient_unique_id,
   UTM.parent_recipient_unique_id,
+  CONCAT(UTM.recipient_hash, '-', case when UTM.parent_recipient_unique_id is null then 'R' else 'C' end) as recipient_hash,
   UTM.recipient_name,
 
   UTM.action_date,
+  DATE(UTM.action_date + interval '3 months') AS fiscal_action_date,
```
BrianZito: This is just to get the fiscal year, right?

sethstoudenmier: Not just fiscal year. For spending over time you need the month, day, and year as it relates to the fiscal year. Instead of splitting them up into three different components you can just add three months to the action_date and then use the date_histogram aggregation on elasticsearch.

BrianZito: Hmm. I don't have any productive ideas, but the concept of a "fiscal date" where today is December 11 in but simultaneously sometime in March in fiscal world seems like a good way to confuse a future dev. Just putting it out there.

sethstoudenmier: Oh I agree that it is weird, but that is current functionality for spending_over_time with the FiscalYear, FiscalMonth, and FiscalQuarter being the current month + 3 months since the fiscal year runs Oct 1 - Sept 30.

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -150,6 +166,15 @@ def __init__(self, *args):
 # ==============================================================================
 
 
+def convert_postgres_array_as_string_to_list(array_as_string: str) -> list:
+    """
+        Postgres arrays are stored in CSVs as strings. Elasticsearch is able to handle lists of items, but needs to
+        be passed a list instead of a string. In the case of an empty array, return null
+        For example, "{this,is,a,postgres,array}" -> ["this", "is", "a", "postgres", "array"].
+    """
+    return array_as_string.replace("{", "").replace("}", "").split(",") if len(array_as_string) > 2 else None
```
BrianZito: This has a risk of replacing things inside the array (possibly escaped). A better strategy would be to trim the first and last character.

BrianZito: This doesn't help with nested arrays, but I think we can all agree that's out of scope.

sethstoudenmier: I agree with both comments. This will only work for something that is a simple Postgres array. Will make the change suggested in the first comment.

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -276,9 +301,15 @@ def download_csv(count_sql, copy_sql, filename, job_id, skip_counts, verbose):
 
 def csv_chunk_gen(filename, chunksize, job_id):
     printf({"msg": "Opening {} (batch size = {})".format(filename, chunksize), "job": job_id, "f": "ES Ingest"})
+    # Need a specific converter to handle converting strings to correct data types (e.g. string -> array)
+    converters = {
```
BrianZito: I don't know this code well, but is there somewhere static we could put this, as a constant? The values in here aren't modified within this context, although on the other hand these aren't really "settings" are business logic so I could see the case for leaving it here.

sethstoudenmier: I originally wanted to put them at the top, but since they will not be used anywhere outside of this context I stuck with being declared and used in this function.

usaspending_api/conftest_helpers.py
```diff
@@ -43,12 +43,17 @@ def update_index(self):
...
         """
+        transaction_delta_view_sql = open(
+            str(settings.APP_DIR / "database_scripts" / "etl" / "transaction_delta_view.sql"), "r"
+        ).read()
         with connection.cursor() as cursor:
-            cursor.execute("SELECT * FROM {}".format(settings.ES_TRANSACTIONS_ETL_VIEW_NAME))
+            cursor.execute(transaction_delta_view_sql)
+            cursor.execute(f"SELECT * FROM {settings.ES_TRANSACTIONS_ETL_VIEW_NAME};")
             transactions = ordered_dictionary_fetcher(cursor)
+            cursor.execute(f"DROP VIEW IF EXISTS {settings.ES_TRANSACTIONS_ETL_VIEW_NAME};")
```
BrianZito: Why `IF EXISTS`? Do we really think this is going to drop? Harmless I suppose, but this could be a way to fail loudly...

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -11,7 +11,7 @@
 logger = logging.getLogger("console")
 
 DOWNLOAD_QUERY_SIZE = settings.MAX_DOWNLOAD_LIMIT
-KEYWORD_DATATYPE_FIELDS = ["{}.raw".format(i) for i in KEYWORD_DATATYPE_FIELDS]
+KEYWORD_DATATYPE_FIELDS = [field for field in KEYWORD_DATATYPE_FIELDS]
```
BrianZito: why not just `list(KEYWORD_DATATYPE_FIELDS`?

usaspending_api/conftest_helpers.py
sethstoudenmier: Moved it around a couple of times and didn't think to take that out. Will edit on the next set of changes.

usaspending_api/search/v2/elasticsearch_helper.py
sethstoudenmier: Good call. I will make the update.

usaspending_api/database_scripts/etl/transaction_delta_view.sql
tony-sappe: @BrianZito this is standard practice for dealing with different calendars. Month names are for calendar months, not fiscal months. It's common to only divide Fiscal years into quarters and stop there anyway.

Instead of thinking by month name, it's more productive to think of the month number.

Examples:
- October is month 10 for calendar year, but it month 1 for federal fiscal year
- April is month 4 for calendar year, month 7 for federal fiscal year, and month 1 for Booz Allen fiscal year

usaspending_api/database_scripts/etl/transaction_delta_view.sql
```diff
@@ -83,10 +96,43 @@ SELECT
...
+  UTM.business_categories
 
 FROM universal_transaction_matview UTM
-JOIN transaction_normalized TM ON (UTM.transaction_id = TM.id)
+INNER JOIN transaction_normalized TM ON (UTM.transaction_id = TM.id)
 LEFT JOIN transaction_fpds FPDS ON (UTM.transaction_id = FPDS.transaction_id)
 LEFT JOIN transaction_fabs FABS ON (UTM.transaction_id = FABS.transaction_id)
-LEFT OUTER JOIN awards AWD ON (UTM.award_id = AWD.id);
+LEFT JOIN awards AWD ON (UTM.award_id = AWD.id)
+-- Similar joins are already performed oon universal_transaction_matview, however, to avoid making the matview larger
```
tony-sappe:
```suggestion
-- Similar joins are already performed on universal_transaction_matview, however, to avoid making the matview larger
```

usaspending_api/database_scripts/matview_generator/universal_transaction_matview.json
```diff
@@ -45,18 +45,18 @@
     "  COALESCE(transaction_fpds.place_of_perform_country_n, transaction_fabs.place_of_perform_country_n) AS pop_country_name,",
     "  COALESCE(transaction_fpds.place_of_perform_country_c, transaction_fabs.place_of_perform_country_c, 'USA') AS pop_country_code,",
     "  COALESCE(transaction_fpds.place_of_performance_state, transaction_fabs.place_of_perfor_state_code) AS pop_state_code,",
-    "  COALESCE(transaction_fpds.place_of_perform_county_co, transaction_fabs.place_of_perform_county_co) AS pop_county_code,",
+    "  LPAD(CAST(CAST((REGEXP_MATCH(COALESCE(transaction_fpds.place_of_perform_county_co, transaction_fabs.place_of_perform_county_co), '^[A-Z]*(\\d+)(?:\\.\\d+)?$'))[1] AS smallint) AS text), 3, '0') AS pop_county_code,",
```
tony-sappe: Unfortunately for changes like this, they need to be propagated to all other materialized views which contain the same data field.

I believe all of these contain the same fields:
SubawardView, SummaryTransactionGeoView, SummaryTransactionMonthView, SummaryTransactionView,
ContractAwardSearchMatview, DirectPaymentAwardSearchMatview, GrantAwardSearchMatview, IDVAwardSearchMatview, LoanAwardSearchMatview, OtherAwardSearchMatview, Pre2008AwardSearchMatview

usaspending_api/etl/es_transaction_template.json
```diff
@@ -160,72 +160,137 @@
...
-          "type": "text",
-          "index": false
+          "type": "text"
         },
         "extent_competed": {
-          "type": "text",
-          "index": false
+          "type": "text"
         },
         "pulled_from": {
```
tony-sappe: Is {{pulled_from}} still useful? Originally it was used to determine if the procurement transaction was for a contract or IDV (boolean "AWARD" or "IDV" values) in conjunction with null in {{type}}. After ES was implemented the {{type}} field is populated for IDVs with "IDV*"

usaspending_api/etl/es_transaction_template.json
```diff
@@ -286,9 +371,25 @@
             }
           }
         },
-        "type": {
-          "type": "keyword",
-          "null_value": "NULL"
+        "treasury_account_identifiers": {
```
emily-brents: we talked about changing the TAS search so we didn't have to hit the database when doing that search, I don't know if you want to to include that here, but it would be like
`"treasury_accounts": {
          "type": "nested",
          "properties": {
            "aid": {
              "type": "integer"
            },
            "ata": {
              "type": "integer"
            },
            "main": {
              "type": "integer"
            },
            "sub": {
              "type": "integer"
            },
            "bpoa": {
              "type": "integer"
            },
            "epoa": {
              "type": "integer"
            },
            "a": {
              "type": "text"
            }
          }
        }`
you would also have to change the delta view to pull the data, like 
`LEFT OUTER JOIN (
  SELECT
    faba.award_id,
    JSONB_AGG(
      JSONB_BUILD_OBJECT(
        'aid', taa.agency_id,
        'ata', taa.allocation_transfer_agency_id,
        'main', taa.main_account_code,
        'sub', taa.sub_account_code,
        'bpoa', taa.beginning_period_of_availability,
        'epoa', taa.beginning_period_of_availability,
        'a', taa.availability_type_code
      )
    ) treasury_account_identifiers
  FROM
    federal_account fa
    INNER JOIN treasury_appropriation_account taa ON fa.id = taa.federal_account_id
    INNER JOIN financial_accounts_by_awards faba ON taa.treasury_account_identifier = faba.treasury_account_id
  WHERE
    faba.award_id IS NOT NULL
  GROUP BY
    faba.award_id)
 TREASURY_ACCT ON (TREASURY_ACCT.award_id = UTM.award_id);`

sethstoudenmier: Will make the change. That escaped me when working on this, but it definitely should go with this PR.

PR #2175, opened 2019-12-19
[DEV-3997] Improve File B custom account download results

kbard:
**Description:**

DEV-3997 requested that, in File B custom account downloads, we eliminate $0 rows under certain circumstances and combine rows with missing Direct/Reimbursable values in the federal account flavor of the download in other circumstances.

The business requirements are a little complicated so I recommend reading more about them in the actual Jira ticket.

**Technical details:**

We used SQL composition to solve this issue.  Rather than attempting to build a query using the ORM, we wrapped the existing query in another query that removed and combined rows as necessary to meet the requirements.

Again, a little complicated so I suggest reading the Technical section in the Jira ticket for additional details.

    - [x] Backend
5. [x] Front end unaffected
8. [x] Jira Ticket [DEV-3997](https://federal-spending-transparency.atlassian.net/browse/DEV-3997):


usaspending_api/download/filestreaming/file_b_custom_account_download_export.sql
```diff
@@ -0,0 +1,178 @@
+-- Connections are pooled which means it's possible for temporary tables to survive beyond what we
```
tony-sappe: Since it's a temp table I assumed there would be no bleed-over between download jobs.

Based on this comment, should the temporary table name be unique?

kbard: The entire SQL statement is being run in a single transaction so there (theoretically) can't be any bleed-over.  This is just an (almost) zero cost safety precaution.

kbard: I need to think on this some more.  I need to reproduce my findings from where I saw temporary tables live longer than I expected.  Now I'm wondering if that was at the cursor level or the connection level.  I may not have made that distinction when I discovered this a few months ago.  If it's at the cursor level, this is a non-issue.  If it's at the connection level, we may have a blocking issue that unique names should alleviate.

kbard: Bottom line is, in Django, temp tables live at the connection level and can live longer than a "session", however, in our implementation we're spawning a `psql` subprocess to run these queries.  Subproceses obtain their own connection and should therefore not be able to block or otherwise conflict with each other.

Having said that, I think there's value in randomizing the names a bit since I don't know how this code is going to be used in the future.  So in the spirit of forward thinking, I will add some randomness to the names just to "guarantee" (not really guaranteed) there can be no collisions.

It is trivial to implement and will cost nothing at the query/code level.

PR #2190, opened 2019-12-28
[DEV-4050] Implement spending_over_time with Elasticsearch

sethstoudenmier:
**Description:**
Implement the `/api/v2/search/spending_over_time` endpoint with Elasticsearch. This endpoint lives alongside the current Postgres implementation and is accessible only through the `X-Experimental-API` header when it is set to `elasticsearch`. The implementation attempts to match the current functionality of the Postgres implementation for all possible filters.

**Technical details:**
PR has four main components to it:
1) Implement the Advanced Search filters with Elasticsearch using `elasticsearch-dsl`.
2) Implement the `/api/v2/search/spending_over_time` endpoint with Elasticsearch and have it live alongside the current Postgres implementation. Access is controlled through the use of the `X-Experimental-API` header when it is set to `elasticsearch`.
3) Make minor changes to the ES Transaction Index that were unfortunately overlooked originally. These changes will have no effect on the current ES functionality, thankfully.
4) Write a ton of integration tests to cover both the `spending_over_time` endpoint and the new filters that use `elasticsearch-dsl`. Additionally, the tests make sure that the header that controls Postgres vs Elasticsearch is working as expected. 

    - [X] Backend
    - [X] Domain Expert <OPTIONAL>
8. [X] Jira Ticket [DEV-4050](https://federal-spending-transparency.atlassian.net/browse/DEV-4050):

```
```


usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -225,3 +227,288 @@ def concat_if_array(data):
             # This should never happen if TinyShield is functioning properly
             logger.error("Keyword submitted was not a string or array")
             return ""
+
+
+def get_base_search_with_filters(index_name: str, filters: dict) -> Search:
```
tony-sappe: The function this is based on isn't a good design. There is much room for improvement.

usaspending_api/database_scripts/etl/transaction_delta_view.sql
```diff
@@ -116,6 +124,14 @@ LEFT JOIN subtier_agency SAA ON (AA.subtier_agency_id = SAA.subtier_agency_id)
 LEFT JOIN toptier_agency TFA ON (FA.toptier_agency_id = TFA.toptier_agency_id)
 LEFT JOIN subtier_agency SFA ON (FA.subtier_agency_id = SFA.subtier_agency_id)
 LEFT JOIN references_cfda CFDA ON (FABS.cfda_number = CFDA.program_number)
+LEFT JOIN (
```
tony-sappe: Concerning recipient "hashes" + level baked into the view and index:

1. Currently the matviews do not store the level (-P, -R, -C) on the hash. I don't recall why. Perhaps it helps with aggregations.
2. This logic doesn't handle the "special case" recipients. These recipients are excluded from the Recipient Profile because their purpose is to obscure the recipients for privacy (or similar) reasons. In existing views the recipient links are often created in the business logic after data are fetched from the DB. See this subquery expression as an example for handling edge-cases: https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/common/recipient_lookups.py#L138-L149 and this example of combining the full recipient  hash + level after the records are returned by the DB: https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/search/v2/views/spending_by_category.py#L283


I don't know the best design for recipient links, someone with more time can dig into the data and make cleaner logic. Until then, it might be better to follow a similar pattern

usaspending_api/search/v2/views/spending_over_time.py
```diff
@@ -12,16 +18,30 @@
...
 from usaspending_api.common.validator.pagination import PAGINATION
 from usaspending_api.common.validator.tinyshield import TinyShield
+from usaspending_api.search.v2.elasticsearch_helper import get_base_search_with_filters
 
-
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("console")
 
 API_VERSION = settings.API_VERSION
+GROUPING_LOOKUP = {
```
tony-sappe: What prompted making this a global?

usaspending_api/search/v2/views/spending_over_time.py
```diff
@@ -89,31 +102,107 @@ def database_data_layer(self):
...
+        aggregations in a specific order to get expected results.
+        """
+        interval = "year" if self.group == "fiscal_year" else self.group
+
+        # The individual aggregations that are needed; with two different sum aggregations to handle issues with
+        # summing together floats.
+        group_by_time_period_agg = A(
+            "date_histogram", field="fiscal_action_date", interval=interval, format="yyyy-MM-dd"
+        )
+        sum_as_cents_agg = A(
```
tony-sappe: I'm not completely following the justification for counting by pennies. Doesn't the `scaled_float` data type prevent floating point rounding errors?

usaspending_api/common/helpers/generic_helper.py
```diff
@@ -104,9 +105,17 @@ def dates_are_month_bookends(start, end):
     return False
 
 
-def min_and_max_from_date_ranges(filter_time_periods):
-    min_date = min([t.get("start_date", settings.API_MAX_DATE) for t in filter_time_periods])
-    max_date = max([t.get("end_date", settings.API_SEARCH_MIN_DATE) for t in filter_time_periods])
+def min_and_max_from_date_ranges(filter_time_periods: list) -> tuple:
```
tony-sappe: These changes might be bleeding into https://federal-spending-transparency.atlassian.net/browse/DEV-4097.

If I'm reading the code correctly, a missing start_date will lead to silently using API_SEARCH_MIN_DATE (and similar logic for missing end_date) which is no longer desired behavior.

usaspending_api/search/v2/views/spending_over_time.py
```diff
@@ -89,31 +102,107 @@ def database_data_layer(self):
...
+        )
+
+        # Putting the aggregations together; in order for the aggregations to the correct structure they
+        # unfortunately need to be one after the other. This allows for nested aggregations as opposed to sibling.
+        search.aggs.bucket("group_by_time_period", group_by_time_period_agg).metric(
+            "sum_as_cents", sum_as_cents_agg
+        ).pipeline("sum_as_dollars", sum_as_dollars_agg)
+
+    def parse_elasticsearch_bucket(self, bucket):
+        key_as_date = pd.to_datetime(bucket["key_as_string"], format="%Y-%m-%d")
```
tony-sappe: `key_as_string`?

usaspending_api/search/v2/views/spending_over_time.py
```diff
@@ -89,31 +102,107 @@ def database_data_layer(self):
...
 
         # time_period is optional so we're setting a default window from API_SEARCH_MIN_DATE to end of the current FY.
         # Otherwise, users will see blank results for years
         current_fy = generate_fiscal_year(datetime.now(timezone.utc))
-        if self.groupings[self.group] == "fiscal_year":
+        if self.group == "fiscal_year":
             end_date = "{}-09-30".format(current_fy)
         else:
-            end_date = "{}-{}-30".format(current_fy, datetime.now(timezone.utc).month)
+            current_month = datetime.now(timezone.utc).month
```
tony-sappe: Any issues with the end date year being the fiscal year but the month as the calendar month?

usaspending_api/search/v2/views/spending_over_time.py
tony-sappe: Does pandas have an advantage over datetime.strptime()?

usaspending_api/search/v2/views/spending_over_time.py
sethstoudenmier: This is also used in the integration tests to make sure that the use of those groups works as expected.

usaspending_api/search/v2/views/spending_over_time.py
sethstoudenmier: There were still issues with floating point numbers because while scaled float with a factor of 100 stores its values as `203` when it is really `2.03` that value for aggregations and queries is represented as the `2.03`. So the scaled floats [are being summed together in the aggregation as regular floats](https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html#scaled-float-params). Noticed the issue during the prototype phase.

usaspending_api/search/v2/views/spending_over_time.py
sethstoudenmier: I can add a comment to explain this. But the `date_histogram` aggregation creates the `key_as_string` to represent the key since it is a date. The name is what Elasticsearch uses.

usaspending_api/search/v2/views/spending_over_time.py
sethstoudenmier: I will make a change here. To keep consistent with how we convert there searched dates over to FY this should be changed.

usaspending_api/search/v2/views/spending_over_time.py
sethstoudenmier: Pandas datetime objects have a `quarter` property unlike normal datetime objects. Could implement without pandas, but since it is already in the project I figured we should use that instead of datetime.

usaspending_api/common/helpers/generic_helper.py
sethstoudenmier: I will undo the changes since I did not notice that ticket and make a note to circle back after both are merged to make sure Postgres and ES implementations have similar functionality.

usaspending_api/search/v2/views/spending_over_time.py
tony-sappe: Makes sense. Although I would like to remove pandas from our app. It's a large package with lots of dependencies.

usaspending_api/search/v2/views/spending_over_time.py
tony-sappe: Mad world... thanks

usaspending_api/search/v2/views/spending_over_time.py
sethstoudenmier: Sad to see it go, but understandable. I will go ahead and make the change.

tony-sappe: Up to you, we aren't ready to pull the plug yet.

usaspending_api/common/experimental_api_flags.py
```diff
@@ -0,0 +1,20 @@
...
+The constants expected are defined in this file to insure consistency across different endpoints that may / will
+need to have experimental functionality alongside the expected functionality as the changes are made permanent
+in Production.
+
+EXPERIMENTAL_API_HEADER - HTTP Header expected to access experimental functionality.
+ELASTICSEARCH_HEADER_VALUE - Value expected to access Elasticsearch functionality on specific endpoints.
+"""
+from rest_framework.request import Request
+
+EXPERIMENTAL_API_HEADER = "HTTP_X_EXPERIMENTAL_API"
```
kwhickey: Cool idea to "dark release" experimental stuff. 👍

usaspending_api/common/experimental_api_flags.py
```diff
@@ -0,0 +1,20 @@
...
+ELASTICSEARCH_HEADER_VALUE = "elasticsearch"
+
+
+def is_experimental_elasticsearch_api(request: Request) -> bool:
+    """
+    Returns True or False depending on if the expected_header_value matches what is sent with the request.
+    """
+    return request.META.get(EXPERIMENTAL_API_HEADER) == ELASTICSEARCH_HEADER_VALUE
```
kwhickey: If we have multiple experimental things in play, should the value for this header instead be a `list`? And the check be whether the list includes the particular experimental feature that the user is trying to use?

usaspending_api/common/query_with_filters.py
```diff
@@ -0,0 +1,426 @@
+import logging
```
kwhickey: I would recommend moving this or parts of this into package `usaspending_api/common/elasticsearch` since it has to do with Elasticsearch querying  and the module name by itself doesn't call that out.

If truly this is the case:
> The idea is that this ABC and QueryWithFilters can be altered to work with Django as well down the road.

Then following a kind of _strategy pattern_ (e.g. [here](https://refactoring.guru/design-patterns/strategy/python/example)), the abstract (backend agnostic) parts should be here in `usaspending_api/query_with_filters` then the elasticsearch-specific implementation would go in `usaspending_api/common/elasticsearch/query_with_filters.py` and django-specific implementation would go in `usaspending_api/common/django/query_with_filters.py`

usaspending_api/search/v2/views/spending_over_time.py
sethstoudenmier: I wanted to add that I also found this: https://github.com/elastic/elasticsearch/pull/37169. Once we upgrade to a new version of Elasticsearch the issue might actually go away. Apparently there were some precision errors found with `scaled_float` and it was fixed as part of `6.7`.

usaspending_api/common/query_with_filters.py
```diff
@@ -0,0 +1,426 @@
...
+
+    @classmethod
+    def generate_elasticsearch_query(cls, filters: dict) -> ES_Q:
+        must_queries = []
+
+        for filter_type, filter_values in filters.items():
+            # Validate the filters
+            if filter_type in cls.unsupported_filters:
+                msg = "API request included '{}' key. No filtering will occur with provided value '{}'"
+                logger.info(msg.format(filter_type, filter_values))
```
kwhickey: Best to log this at the `warning` level.

usaspending_api/search/tests/integration/test_spending_over_time.py
```diff
@@ -0,0 +1,1073 @@
...
+    )
+    assert resp.status_code == status.HTTP_400_BAD_REQUEST
+
+
+"""
+As of 12/26/2019 these are intended for the experimental Elasticsearch functionality that lives alongside the Postgres
+implementation. These tests verify that ES performs as expected, but that it also respects the header put in place
+to trigger the experimental functionality. When ES for spending_over_time is used as the primary implementation for
+the endpoint these tests should be updated to reflect the change.
+"""
```
kwhickey: Really nice and thorough set of integration tests following here. Well done!

usaspending_api/search/v2/views/spending_over_time.py
kwhickey: If it's _only_ ever used outside this module by tests, and is otherwise considered only to be used by this module, I consider it okay to make it a "protected" global, and directly import that protected global into the tests (i.e. allow tests to violate that "weak private/protected" contract).

e.g. 
`_GROUPING_LOOKUP = { ... }`

usaspending_api/database_scripts/etl/transaction_delta_view.sql
sethstoudenmier: Made some changes to the `recipient_id` filter and the view SQL that make it function the same way only with ES instead of Django. 

Regarding the special cases, from my understanding, it should allow for the same functionality as the current `spending_by_category` implementation now.

usaspending_api/common/experimental_api_flags.py
sethstoudenmier: I like the idea, but it would require passing the headers as some sort of formatted string such as "header1/header2/header3" because I don't believe HTTP Headers will take a list as a valid value. It seems to convert the values to strings.

usaspending_api/common/query_with_filters.py
kwhickey: Worth further discussion though _**IF**_ we think there is value in being able to conjoin the "interface" to backend querying -- between Django and Elasticsearch. 

That is,  doing some queries or filters would use the same object as an abstraction/interface to a backend, and the developer wouldn't really need to know whether the backend was django or elasticsearch. 

My thought is that since Django is still going to be used heavily for various other endpoints (and likely in different ways than how it is/was used for Adv. Search), and Django will eventually be phased-out of Adv. Search ... that it is probably not worth the effort to make the two use a single backend querying interface. 

That shouldn't prevent you from good programming and good abstraction in building the Elasticsearch querying interface/layer -- just means that we can defer or choose not to build a django-specific implementation for it, which would likely require some compromises in the interface+impls to make it compatible for both.

usaspending_api/common/elasticsearch/client.py
```diff
@@ -48,7 +53,9 @@ def create_es_client():
         logger.error("Error creating the elasticsearch client: {}".format(e))
 
 
-def es_client_query(index, body, timeout="1m", retries=5):
+def es_client_query(
+    index: str = None, body: dict = None, timeout: str = "1m", retries: int = 5, search: Search = None
+) -> Union[Elasticsearch, Response]:
```
tony-sappe: Confused by the `Union` typehint. Seems like it would be something like `Optional[Union[dict, Response]]` (unsure if that is proper) since None is a possible return value

If desired, the response type of _es_search() can be defined as a new type, which could be utilized in other functions' I/O.

usaspending_api/common/query_with_filters.py
tony-sappe: > and Django will eventually be phased-out of Adv. Search 

The plan is still to use Postgres for downloads. Advanced Search functionality contains downloads. As previously discussed there is a huge concern with creating dissonance between the future ES API results and download results. The idea of a shared filter object which can be reused for the various data stores and _hopefully_ ease the burden of keeping results consistent appears to be a clean design.

usaspending_api/common/helpers/generic_helper.py
```diff
@@ -59,9 +60,15 @@ def generate_fiscal_month(date):
     """ Generate fiscal period based on the date provided """
     validate_date(date)
 
-    if date.month in [10, 11, 12, "10", "11", "12"]:
-        return int(date.month) - 9
-    return int(date.month) + 3
+    if date.month in [10, 11, 12]:
```
tony-sappe: _If this recommendation cascades into problems.... ignore._


A new "base class" for obtaining the fiscal date which is used for the other functions:

```python
def convert_date_to_fiscal_cal(date: datetime) -> FiscalDate:
    validate_date(date)
    return FiscalDate(date.year, date.month, date.day)

def generate_fiscal_quarter(date: datetime) -> int:
    return convert_date_to_fiscal_cal(date).quarter

# repeat for year and month....
```

It might encourage devs to migrate to using `convert_date_to_fiscal_cal()` instead of calling several of the other functions.

Could even use datetime.now(timezone.utc) as the default date since that is a common value passed into the functions

usaspending_api/common/query_with_filters.py
```diff
@@ -0,0 +1,426 @@
...
+        if filter_values is None:
+            raise InvalidParameterException(f"Invalid filter: {cls.underscore_name} has null as its value.")
+
+        return cls._generate_elasticsearch_query(filter_values)
+
+    @classmethod
+    @abstractmethod
+    def _generate_elasticsearch_query(cls, filter_values: Union[str, list]) -> Union[ES_Q, List[ES_Q]]:
+        """ Returns a Q object used to query Elasticsearch. """
+        raise NotImplementedError("Must be implemented in subclasses of _Filter.")
```
tony-sappe: Is this exception redundant when using `@abstractmethod`?

sethstoudenmier: Replaced with `pass` because it is redundant.

usaspending_api/common/helpers/generic_helper.py
sethstoudenmier: Like the idea, but I feel that this will result in more changes then what I would like to make right now if not needed.

usaspending_api/common/elasticsearch/client.py
sethstoudenmier: Created a unique type to use instead. Also, helps to keep the line from being really long.

usaspending_api/common/query_with_filters.py
kwhickey: But the downloads (from Adv Search) won't be running Django filters, will they? Isn't the approach to execute the filters on ES, get all the postgres IDs from the ES docs, and then run the download as a query to postgres with a giant kind of `IN` clause (no filtering aside from that)?

If you're thinking about the general Downloads section of the site (custom account/award download) we could adapt this interface to abstract those django queries. I'm not clear on how much similarity there is between the backend data structures of those (matviews?) and these ES indexes. If very close, then a common filtering interface could be good. But if they differ in what is filterable or the semantics of what some filterable field represents, forcing them to conform to the same interface could cause confusion (e.g. trying to add a filter that's not supported by one backed or another).

tony-sappe: You're correct on the current process of using the list of IDs from ES to drive downloads. While it is functional it plays into weaknesses of both Elasticsearch and Postgres, there is concern it will be the process long-term

Currently Custom Award Downloads use the same filters and SQL for downloads as Advanced Search.

PR #2192, opened 2019-12-31
[DEV-4120] Adding place_of_perform_scope to Assistance data

tony-sappe:
**Description:**
Adding new assistance field from Broker for downloads: `place_of_perform_scope`

**Technical details:**
Adding new columns, includes backfill script. FABS ETL automatically picks it up since the column name matches.

    - [x] Backend
8. [x] Jira Ticket [DEV-4120](https://federal-spending-transparency.atlassian.net/browse/DEV-4120):

```
No changes to matviews or API request/responses
```


usaspending_api/database_scripts/job_archive/backfill_primary_place_of_performance_scope.py
```diff
@@ -0,0 +1,174 @@
...
+import math
+import psycopg2
+import time
+
+from os import environ
+
+
+# DEFINE THESE ENVIRONMENT VARIABLES BEFORE RUNNING!
+SPENDING_CONNECTION_STRING = environ["DATABASE_URL"]
+BROKER_CONNECTION_STRING = environ["DATA_BROKER_DATABASE_URL"]
```
BrianZito: Is there a reason we aren't using `connections["broker"]` for this?

tony-sappe: This a standalone script in job_archive so it doesn't import the django app. (just like all the other scripts in this dir)

BrianZito:
Not merging due to incomplete checklist

usaspending_api/awards/management/commands/load_subawards_sql/050_enhance_with_transaction_data.sql
```diff
@@ -20,3 +20,17 @@ from
...
+
+set
+    place_of_perform_scope = fabs.place_of_performance_scope
+
+from
+    transaction_fabs fabs
+
+where
+    fabs.transaction_id = temp_load_subawards_subaward.latest_transaction_id and
+    temp_load_subawards_subaward.latest_transaction_id is not null;
```
kbard: I would suggest putting this query in its own file.  If memory serves, Postgres only returns the row count of the last query executed.  If we want the row count from the query above logged, these two queries will need to be in their own files.

Also, `and temp_load_subawards_subaward.latest_transaction_id is not null` is not strictly necessary.  In SQL NULL != NULL so the first part of the _where_ clause should never match if either side of the `=` is NULL.  I don't think affects anything to leave it so not a showstopper.

usaspending_api/database_scripts/job_archive/backfill_primary_place_of_performance_scope.py
```diff
@@ -0,0 +1,174 @@
...
+        logging.info("[{}] {:,} rows updated in {}".format(transaction_type, row_count, t.elapsed_as_string))
+        return row_count
+
+
+if __name__ == "__main__":
+    log_format = "[%(asctime)s] [%(levelname)s] - %(message)s"
+    logging.getLogger()
+    logging.basicConfig(level=logging.INFO, format=log_format, datefmt="%Y/%m/%d %H:%M:%S (%Z)")
+
+    fabs_row_count, fabs_row_count = 0, 0
```
kbard: Is `fabs_row_count` supposed to be declared twice here or was one of them supposed to be something else?

usaspending_api/database_scripts/job_archive/backfill_primary_place_of_performance_scope.py
```diff
@@ -0,0 +1,174 @@
...
+    (VALUES {}) AS broker(
+        published_award_financial_assistance_id,
+        place_of_performance_scope
+    )
+WHERE
+    fabs.published_award_financial_assistance_id = broker.published_award_financial_assistance_id
+    AND (
+        fabs.place_of_performance_scope IS DISTINCT FROM broker.place_of_performance_scope
+    )
+RETURNING fabs.transaction_id
```
kbard: I don't see us using this return value anywhere.  Think there's any efficiency to be gained if Postgres doesn't have to keep track of ids to return?

usaspending_api/awards/management/commands/load_subawards_sql/050_enhance_with_transaction_data.sql
tony-sappe: Should I remove the extraneous check from the original SQL as well?

kbard: Might as well.  We'll have to track down the original developer and flog them for such nonsense.

PR #2170, opened 2019-12-16
DEV-2895 Port sqs_work_dispatcher.py framework from Broker

kwhickey:
**Description:**
Bring over from Broker the queueing framework that sits between background app jobs and the AWS SQS queue, and handles error, cleanup, retries, and worker EC2 node terminations.

**Technical details:**
This port is just to get the framework into USAspending. It is not integrated into any USAspending background jobs yet. It is being broken up for easier review, and merging this could be deferred til that integration happens. 

Of note, changes from broker implementation are:

1. Allowing callers to override the `daemon` value of `multiprocessing.Process`. This is needed by USAspending to allow its workers to spawn their own child processes.
2. No reliance on a backing DB table for a fake queue that can be used locally or for unit tests. This allows the unit tests to be more like true unit tests and less like integration tests.

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [NA] API documentation updated
3. [x] Necessary PR reviewers:
    - [x] Backend
    - [NA] Frontend <OPTIONAL>
    - [NA] Operations <OPTIONAL>
    - [NA] Domain Expert <OPTIONAL>
4. [NA] Matview impact assessment completed
5. [NA] Frontend impact assessment completed
6. [NA] Data validation completed
7. [NA] Appropriate Operations ticket(s) created
8. [ ] Jira Ticket [DEV-2895](https://federal-spending-transparency.atlassian.net/browse/DEV-2895):
    - [x] Link to this Pull-Request
    - [NA] Performance evaluation of affected (API | Script | Download)
    - [NA] Before / After data comparison

**Area for explaining above N/A when needed:**
```
Just frameworking so far. No alteration or integration with running app code yet.
```

PR #2194, opened 2019-12-31
DEV-2896 bulk downloader using sqs work dispatcher

kwhickey:
**Description:**
Put our "resiliency and logging" wrapper around the bulk downloader worker processes that handle downloads.

After porting the `sqs_work_dispatcher.py` and accompanying files over from Broker (https://github.com/fedspendingtransparency/usaspending-api/pull/2170), this PR incorporates utilizing that framework to run, monitor, error-handle, log, and retry python subprocesses that handle SQS messages to run a bulk download. 

**Technical details:**
- Updated logging configuration pattern in `settings.py` to allow distinct logging levels per module
- Altered the `download_sqs_worker` to be wrapped by the sqs work dispatcher, and to handle errors coming from the worker or the parent process
- Made changes to the fake queue implementation so that transient queues are used for unit testing and a persistent local queue is used for local development. Both backed by a file in the `/bulk_downloads/local_queue` directory
- Altered docker setup to run a bulk-downloader service separately from the API and added a `settings.py` setting influenced by presence of an environment variable which determines whether download requests will be run eagerly, in-process of the API, or queued in the local file-backed queue for pickup and processing from the bulk-downloader service. 
    - by default when running as docker containers, it is queued and processed by the bulk-downloader container
- Made changes to the `sqs_work_dispatcher.py` framework for better logging/diagnostics and alignment of download logs to work dispatcher logs
- Standardizing where and how the boto queue resource is retrieved for use
- Doing better cleanup of file handles using context managers when data is written to csv or zip
- Altering the way that zips are written to with `'w'` mode rather than `'a'` in case there is a queue-retry and a previous zip file with the same name encountered. 

    - [x] Backend
    - [NA] Frontend <OPTIONAL>
    - [NA] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-2896](https://federal-spending-transparency.atlassian.net/browse/DEV-2896):

```
Not affecting domain/business logic, frontend, or API
```


usaspending_api/settings.py
```diff
@@ -43,6 +43,16 @@
 # Define local flag to affect location of downloads
 IS_LOCAL = True
 
+# How to handle downloads locally
+# True: process it right away by the API;
+# False: leave the message in the local file-backed queue to be picked up and processed by the bulk-download container
+RUN_LOCAL_DOWNLOAD_IN_PROCESS = (
```
tony-sappe: Simplify the logic, and make it a one-liner.
Also the conditional was switched to make false the default.

```suggestion
RUN_LOCAL_DOWNLOAD_IN_PROCESS = os.environ.get("RUN_LOCAL_DOWNLOAD_IN_PROCESS", "").lower() in ["true", "1", "yes"]
```

usaspending_api/common/sqs/sqs_work_dispatcher.py
```diff
@@ -953,19 +954,47 @@ def _set_message_visibility(self, new_visibility):
...
             )
-            log_job_message(logger=self._logger, message=message, job_type=self.worker_process_name, is_exception=True)
-            raise QueueWorkDispatcherError(message) from exc
+            log_dispatcher_message(self, message=message, is_exception=True)
+            raise QueueWorkDispatcherError(
+                message, worker_process_name=self.worker_process_name, queue_message=self._current_sqs_message
+            ) from exc
+
+
+class AbstractQueueError(Exception, metaclass=ABCMeta):
```
tony-sappe: nice ABC. I recommend the exception classes be stored in a dedicated file.

usaspending_api/settings.py
kwhickey: @tony-sappe it looks cleaner but I don't think `False` as the default will go over well with the team, and will break unit tests. 

Unit Tests: (I had it this way originally, and found this out) are performing checks on the response to the download endpoint and therefore require that the download get processed right-away, in-process of the test session. Otherwise the download request gets enqueued and not processed at all, and the test fails (e.g. "total_rows" = None).

Devs: Most people probably are not currently running docker-based local dev envs, and those that are may not always run _both_ the api and bulk downloader containers (maybe just the API and/or DB). So where they are expecting a download to fire off the download code immediately, they'll sit there and wait (since it will just put it in the queue and wait for _something_ (that they're not running) to pick it up.

usaspending_api/common/sqs/sqs_work_dispatcher.py
kwhickey: I'm not opposed but is there a general good rule for this? I was thinking that they'll likely only be used (instantiated) by the `SQSWorkDispatcher`, so stay in the same module alongside it. 

Granted though, they will and can be _caught_  in lots of code throughout apps that use the dispatcher.

kwhickey:
Adding `do not merge` label after some discoveries during testing. 

It appears that by making this worker process non-daemonic, to support creation of psql and zip child processes, it does not now clean up orphaned processes. Those processes can become orphaned if there's an issue in our bulk download code... which would leave for instance a long-running and significant-to-the-db `psql` process running, even if there is no download worker process that could pick up the result of the psql output and send it back to the user.

usaspending_api/settings.py
tony-sappe: good points

```python
RUN_LOCAL_DOWNLOAD_IN_PROCESS = os.environ.get("RUN_LOCAL_DOWNLOAD_IN_PROCESS", "").lower() not in ["false", "0", "no"]
```

usaspending_api/common/sqs/sqs_work_dispatcher.py
tony-sappe: Rule in this codebase or general practice?

This codebase does have a few files which exclusively contain custom exception classes. It seems straight-forward to me since they are self-contained objects and I want to encourage there usage by raising awareness of their existence. If nothing else it cleanly reduces the line count in larger files

kwhickey:
> Adding `do not merge` label after some discoveries during testing.
> 
> It appears that by making this worker process non-daemonic, to support creation of psql and zip child processes, it does not now clean up orphaned processes. Those processes can become orphaned if there's an issue in our bulk download code... which would leave for instance a long-running and significant-to-the-db `psql` process running, even if there is no download worker process that could pick up the result of the psql output and send it back to the user.
> 
> Need to find a way to detect process exiting (that part should already exist, at least for the parent dispatcher process) and then walk a tree of descendant processes to kill them all.

Fixed as good as can be with latest commits. 
Still slightly vulnerable to orphan processes continuing to run in the rare case that the worker receives a termination signal prior to the parent dispatcher. 

Removing `do not merge` label.

usaspending_api/settings.py
kwhickey: Done.

usaspending_api/common/sqs/sqs_work_dispatcher.py
kwhickey: Done.

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,185 +1,154 @@
...
-        return
-    try:
-        message.change_visibility(VisibilityTimeout=new_visibility)
-    except botocore.exceptions.ClientError:
-        write_to_log(message="Unable to set VisibilityTimeout. Message might have previously been released or removed")
+def _update_download_job_status(download_job_id, status, error_message=None, overwrite_error_message=False):
+    """Handles status updates on the DownloadJob records
 
+    Args:
+        download_job_id (int): id of the DownloadJob record being processed
```
tony-sappe: No reason to change now. If the parameters types are useful to document then adding type hints are the recommended documentation

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,185 +1,154 @@
...
+            status = "failed"
+            _update_download_job_status(download_job_id, status, stack_trace)
+            log_job_message(
+                logger=logger,
+                message=f"Marked DownloadJob with id = {download_job_id} as {status}",
+                job_type=JOB_TYPE,
+                job_id=download_job_id,
+                is_exception=True,
+            )
+    except:  # noqa
```
tony-sappe: `except Exception:` is the recommended syntax, seems cleaner than a #noqa comment

usaspending_api/download/filestreaming/zip_file.py
```diff
@@ -3,9 +3,26 @@
 
 
 def append_files_to_zip_file(file_paths, zip_file_path):
+    """
+    Create zip archive at the specified zip_file_path if it does not exist, and add all the files at provided
+    file_paths to it.
 
-    # Append all of the provided files to the specified zip file.
+    NOTE: If a zip file already exists at zip_file_path, the given files will be added in addition to the ones
```
tony-sappe: No edits necessary. Writing thoughts


The function name `append_files_to_zip_file()`  leads me to believe appending to an existing file is the intended workflow. Recent changes to include timestamps in the data filenames should prevent duplicate file names even if a failure occurred during a previous attempt (unlike zip files which are stored in the DB and should remove any pre-existing file on collision).

usaspending_api/download/filestreaming/download_generation.py
```diff
@@ -120,6 +122,7 @@ def generate_download(download_job):
         # Remove generated file
         if not settings.IS_LOCAL and os.path.exists(zip_file_path):
             os.remove(zip_file_path)
+        _kill_spawned_processes(download_job)
```
tony-sappe: Could there be spawned processes still running here?

usaspending_api/common/sqs/sqs_handler.py
```diff
@@ -152,16 +192,11 @@ def receive_messages(
...
-# Pseudo-singleton instances of the Fake Queue and Fake Dead Letter Queue
-FAKE_QUEUE = _FakeFileBackedSQSQueue()
-FAKE_DEAD_LETTER_QUEUE = _FakeStatelessLoggingSQSDeadLetterQueue()
-
-
 class FakeSQSMessage:
     """Fakes portions of a boto3 SQS resource ``Message`` object.
     See: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#message"""
 
-    def __init__(self, queue_url: str = FAKE_QUEUE.UNITTEST_FAKE_QUEUE):
```
tony-sappe: could have left the type hint

usaspending_api/download/filestreaming/download_generation.py
kwhickey: It'd be rare but I think possible. 

What I realized is that the queue framework should stop trying to be responsible for taking care of spawned child processes in the "processing-app code" that it is dispatching to. That should be a cleanup responsibility of the processing-app. In this case, since the download code spawns sub-processes, it should handle cleanup of them in case of an error. 

Now, it's likely that the error is going to either be in psql, in which case it will kill itself, or in zip, which won't run til after psql, and will similarly kill itself. So this would only occur if the python code around psql or the zip (e.g. in the wait_for_process) somehow encountered an error. So that's why this was added to the finally as a just-in-case step.

PR #2205, opened 2020-01-07
FIX: [DEV-2896] Fixing download zip file overwriting

kwhickey:
**Description:**
Revert to using "append" mode when creating the download zip to fix the problem of each additional file added re-starting the zip from scratch. 

**Technical details:**
- Write mode `'w'` was introduced when creating the download zip as a precaution against re-using a stale, half-built zip file on a retry attempt of the bulk downloader (queue retry)
- Turns out our code does not create the zip in one-go, but in several calls to add files. 
- So put append mode `'a'` back into place instead
- And added a precautionary removal of a pre-established zip with the same file name on new download generations

    - [ ] Backend
    - [NA] Frontend <OPTIONAL>
    - [NA] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-2896](https://federal-spending-transparency.atlassian.net/browse/DEV-2896):
    - [NA] Performance evaluation of affected (API | Script | Download)
    - [NA] Before / After data comparison

```
Small fix to a prior approved PR
```

### DIFF
--- a/usaspending_api/download/filestreaming/download_generation.py
+++ b/usaspending_api/download/filestreaming/download_generation.py
@@ -25,7 +25,7 @@ from usaspending_api.download.filestreaming import NAMING_CONFLICT_DISCRIMINATOR
 from usaspending_api.download.filestreaming.download_source import DownloadSource
 from usaspending_api.download.filestreaming.generate_export_query import generate_default_export_query
 from usaspending_api.download.filestreaming.file_description import build_file_description, save_file_description
-from usaspending_api.download.filestreaming.zip_file import append_files_to_zip_file, write_files_to_zip_file
+from usaspending_api.download.filestreaming.zip_file import append_files_to_zip_file
 from usaspending_api.download.helpers import (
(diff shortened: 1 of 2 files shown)


usaspending_api/download/filestreaming/download_generation.py
```diff
@@ -96,7 +100,7 @@ def generate_download(download_job):
         raise Exception(download_job.error_message) from e
     finally:
         # Remove working directory
-        if os.path.exists(working_dir):
+        if working_dir and os.path.exists(working_dir):
```
BrianZito: Why is this check needed? It seems like working_dir is still always being set.

sethstoudenmier: Possible that the try block could fail before `working_dir` is set and then it would have a value of None in the Finally block.

kwhickey: Yep, that's it @sethstoudenmier

PR #2207, opened 2020-01-09
[DEV-3495] Tweak federal account name derivation

kbard:
**Description:**

The business rules for how we derive federal account names have changed.  This pull request reflects those changes and fixes a couple bugs in the TAS loader.

2. [x] API should be unaffected
    - [ ] Backend
8. [x] Jira Ticket [DEV-3495](https://federal-spending-transparency.atlassian.net/browse/DEV-3495):


usaspending_api/references/account_helpers.py
```diff
@@ -0,0 +1,89 @@
...
+        cursor.execute(
+            """
+                update  treasury_appropriation_account as tu
+                set     federal_account_id = fa.id
+                from    treasury_appropriation_account as t
+                        left outer join federal_account as fa on
+                            t.agency_id = fa.agency_identifier and
+                            t.main_account_code = fa.main_account_code
+                where   tu.treasury_account_identifier = t.treasury_account_identifier and
+                        tu.federal_account_id is distinct from fa.id;
```
BrianZito: Will this last check save any time?

kbard: Probably not a significant amount for tables this size, but why update rows that don't need updating?

BrianZito: Only time. Often assignment takes less time than a boolean check.

kbard: Updates in a database tend to be very expensive since they will copy the row.

kbard: But again, these tables are small, so it would probably be unnoticeable.  Regardless, as a general habit, I don't want to update rows I don't have to.

usaspending_api/references/account_helpers.py
```diff
@@ -0,0 +1,89 @@
...
+        distinct on (agency_id, main_account_code)
+        agency_id as agency_identifier,
+        main_account_code,
+        concat(agency_id, '-', main_account_code) as federal_account_code,
+        account_title
+    from
+        treasury_appropriation_account
+    order by
+        agency_id,
+        main_account_code,
```
BrianZito: Why are these first in ordering? Doesn't this defeat the point of the rest of the ticket?

kbard: It's a requirement of `distinct on`.

https://www.postgresql.org/docs/10/sql-select.html#SQL-DISTINCT

> The DISTINCT ON expression(s) must match the leftmost ORDER BY expression(s).

How does that defeat the point of the rest of the ticket?

BrianZito: Ok, that clears that up.

PR #2208, opened 2020-01-10
Mod/dev 4115 filter out transaction obligated amount

sethstoudenmier:
**Description:**
Filter out the File C rows that have `transaction_obligated_amount` set to either 0 or null. 

**Technical details:**
Noticed that the `financial_accounts_by_awards` PK is not referenced on any other table and instead other tables reference `financial_accounts_by_awards` by the different FKs it has. My understanding is that we should be able to filter out `WHERE transaction_obligated_amou IS NOT NULL and transaction_obligated_amou != 0` when loading File C from Broker and then run `DELETE FROM financial_accounts_by_awards WHERE transaction_obligated_amount = 0 OR transaction_obligated_amount IS NULL;` on the different environments to remove old records.

This PR is mostly moving around test cases for loading File C to use an instance of a test Broker DB instead of a mocked DB cursor.

Also this means that the Matviews will need to be refreshed so that they no longer have the old `financial_accounts_by_awards` records.

    - [x] Backend
8. [X] Jira Ticket [DEV-4115](https://federal-spending-transparency.atlassian.net/browse/DEV-4115):

```
```


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -111,13 +111,13 @@ def signal_handler(signal, frame):
...
-            submission_id
+        award_financial_query = (
+            "SELECT * FROM certified_award_financial"
+            f" WHERE submission_id = {submission_id}"
+            " AND transaction_obligated_amou IS NOT NULL AND transaction_obligated_amou != 0"
         )
-        if isinstance(db_cursor, PhonyCursor):  # spoofed data for test
-            award_financial_frame = pd.DataFrame(db_cursor.db_responses[award_financial_query])
-        else:  # real data
-            award_financial_frame = pd.read_sql(award_financial_query, connections["data_broker"])
```
BrianZito: I'm all for getting rid of test hooks when we can, but are we sure this wasn't used anywhere?

sethstoudenmier: Was only be used by the `load_submission` test I edited from what I noticed. With Broker integration tests we should be able to move away from mocking the DB Cursor.

usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -674,8 +674,6 @@ def load_file_c(submission_attributes, db_cursor, award_financial_frame):
     awards_touched = []
 
     # format award_financial_frame
-    float_cols = ["transaction_obligated_amou"]
-    award_financial_frame[float_cols] = award_financial_frame[float_cols].fillna(0)
```
BrianZito: Why isn't this needed anymore?

sethstoudenmier: Previously this was used to backfill the `transaction_obligated_amou` values that were Null to be 0. Now that we won't be taking in any Null or 0 values there is no need to run this.

PR #2211, opened 2020-01-10
[DEV-4121] Change download endpoint response

sethstoudenmier:
**Description:**
Break apart the `/download/<type>/` and `/download/status?file_name=` endpoints so that the download endpoint has more meaningful information and links to the status endpoint.

**Technical details:**
Since we cache our downloads majority of the response for any of the `/download/<type>` endpoints is useless and requires that someone use the `/download/status?file_name=` endpoint if they want to find out anything other than the file_name. This change breaks apart the two responses and adds a link to the download status endpoint for all of the download endpoints.

Sample Old Response:
```{
    "status": "ready",
    "url": "https://files-nonprod.usaspending.gov/generated_downloads/dev/IDV_UNKNOWN_2020-01-10_H18M46S31395245.zip",
    "message": null,
    "file_name": "IDV_UNKNOWN_2020-01-10_H18M46S31395245.zip",
    "total_size": null,
    "total_columns": null,
    "total_rows": null,
    "seconds_elapsed": null
}
```

Sample New Response:
```
{
    "status_url": "http://localhost:8000/api/v2/download/status?file_name=IDV_UNKNOWN_2020-01-10_H17M40S19533231.zip",
    "file_name": "IDV_UNKNOWN_2020-01-10_H17M40S19533231.zip",
    "file_url": "/Users/sethstoudenmier/Documents/DATA_Act/projects/usaspending-api/csv_downloads/IDV_UNKNOWN_2020-01-10_H17M40S19533231.zip",
    "download_request": {
        "account_level": "treasury_account",
        "columns": [],
        "download_types": [
            "idv_federal_account_funding",
            "idv_orders",
            "idv_transaction_history"
        ],
        "file_format": "csv",
        "filters": {
            "award_type_codes": [
                "IDV_B_A",
                "B",
                "A",
                "D",
                "IDV_E",
                "IDV_A",
                "IDV_B_C",
                "IDV_B_B",
                "IDV_B",
                "IDV_C",
                "IDV_D",
                "C"
            ],
            "idv_award_id": 41813500
        },
        "include_data_dictionary": true,
        "include_file_description": {
            "destination": "readme.txt",
            "source": "/Users/sethstoudenmier/Documents/DATA_Act/projects/usaspending-api/usaspending_api/data/idv_download_readme.txt"
        },
        "is_for_idv": true,
        "limit": 500000,
        "piid": null,
        "request_type": "idv"
    }
}
```

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
8. [X] Jira Ticket [DEV-4121](https://federal-spending-transparency.atlassian.net/browse/DEV-4121):

```
```


usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -98,17 +100,36 @@ def get_download_response(self, file_name):
...
+                protocol = "http"
+                host = "localhost:8000"
+            elif settings.DOWNLOAD_ENV == "production":
+                protocol = "https"
+                host = "api.usaspending.gov"
+            else:
+                protocol = "https"
+                host = f"{settings.DOWNLOAD_ENV}-api.usaspending.gov"
+
+            status_url = f"{protocol}://{host}/api/v2/download/status?file_name={file_name}"
```
BrianZito: Could we generalize this logic somewhere? This is pretty brittle logic...

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -98,17 +100,36 @@ def get_download_response(self, file_name):
...
-            "message": download_job.error_message,
-            "file_name": file_name,
-            # converting size from bytes to kilobytes if file_size isn't None
-            "total_size": download_job.file_size / 1000 if download_job.file_size else None,
-            "total_columns": download_job.number_of_columns,
-            "total_rows": download_job.number_of_rows,
-            "seconds_elapsed": download_job.seconds_elapsed(),
-        }
+        base_response = {"file_name": file_name, "file_url": file_path}
+        if is_download_status:
```
BrianZito: What was the impetus for this boolean? Just so that the two endpoints can share the functionality of `base_response` and the boto code? It seems to me like this is two distinct views smashed into one. Splitting them up and keeping them in the same class would let you use all the common constants and functions you want and lower Cognitive Complexity.

sethstoudenmier: Good call, will make the change

usaspending_api/download/v2/base_download_viewset.py
sethstoudenmier: I agree I am not the biggest fan but did not see a much better way. This should only be used in the download endpoint so there is no need to break this out into a different function or create another item in settings.py similar to the DATA_DICTIONARY_DOWNLOAD_URL

BrianZito: reuse isn't the only reason to use helper functions. Just having a private function in this class that returns status_url will drop cognitive complexity by 3 and shorten each function.

PR #2216, opened 2020-01-13
[DEV-3767] Spending by Awards Elasticsearch Endpoint

emily-brents:
**Description:**
Adds functionality to spending_by_award endpoint to allow searching using Elasticsearch. Spending_by_award uses the awards index in elasticsearch, which will need to be set up before the code can be used.

**Technical details:**
When passing the experimental API header (`X-Experimental-API/elasticsearch`) and hitting the `spending_by_award` endpoint, you can switch to using Elasticsearch for advanced search. Currently the elasticsearch cluster can only return 50,000 results, so if paginating past that the query will fail. To get around this, you can paginate using `search_after`, which will require passing the `last_id` and `last_value` parameters for each request. The API response for the previous page will include these values. However, the sort order will be different when using this method, so if planning on paginating past 50,000 results, the user should default to the search_after method in order to ensure that they don't miss any records.

    - [ ] Backend
8. [x] Jira Ticket [DEV-3767](https://federal-spending-transparency.atlassian.net/browse/DEV-3767):

```
No impact to frontend, feature is currently hidden, access by using API headers
```


usaspending_api/awards/v2/lookups/elasticsearch_lookups.py
```diff
@@ -25,6 +25,72 @@
     "Last Date to Order": "ordering_period_end_date",
 }
 
+default_mapping = {
```
BrianZito: is this a "default" that ever gets used directly, or a "base" that is only ever appended to?

emily-brents: It only gets appended to

BrianZito: Probably better to name it `base_mapping` for clarity. In general in this area, this could be shortened a lot by defining rather than modifying dependent dictionaries:
```
base_mapping = {"blah":"blah"}
contracts_mapping = {**base_mapping, **{"yada":"yada"}}
```

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -55,6 +71,12 @@
             **{award_type: loan_award_mapping for award_type in loan_type_mapping},
             **{award_type: non_loan_assistance_award_mapping for award_type in non_loan_assistance_type_mapping},
         },
+        "elasticsearch_type_code_to_field_map": {
+            **{award_type: contracts_mapping for award_type in contract_type_mapping},
+            **{award_type: idv_mapping for award_type in idv_type_mapping},
+            **{award_type: loan_mapping for award_type in loan_type_mapping},
+            **{award_type: non_loan_assist_mapping for award_type in non_loan_assistance_type_mapping},
+        },
```
BrianZito: 💯

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -98,12 +120,20 @@ def post(self, request):
             "sort_order": json_request["order"],
             "upper_bound": json_request["page"] * json_request["limit"] + 1,
         }
+        self.elasticsearch = is_experimental_elasticsearch_api(request)
+        self.last_id = json_request.get("last_id")
+        self.last_value = json_request.get("last_value")
```
BrianZito: Won't these only be provided if the user is sending the experimental elasticsearch flag?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -197,6 +236,21 @@ def get_sort_by_fields(self):
...
+            sort_by_fields = ["display_award_id"]
+        else:
+            if set(self.filters["award_type_codes"]) <= set(contract_type_mapping):
+                sort_by_fields = [contracts_mapping[self.pagination["sort_key"]]]
+            elif set(self.filters["award_type_codes"]) <= set(loan_type_mapping):
+                sort_by_fields = [loan_mapping[self.pagination["sort_key"]]]
+            elif set(self.filters["award_type_codes"]) <= set(idv_type_mapping):
+                sort_by_fields = [idv_mapping[self.pagination["sort_key"]]]
+            elif set(self.filters["award_type_codes"]) <= set(non_loan_assistance_type_mapping):
+                sort_by_fields = [non_loan_assist_mapping[self.pagination["sort_key"]]]
```
BrianZito:
```suggestion
                sort_by_fields = [non_loan_assist_mapping[self.pagination["sort_key"]]]
            else
                raise [some exception]
```
Since `self.filters` is provided by the user, it's quite possible that a request will qualify for none of the above cases, resulting in sort_by_fields being None and a 500 error later down the line. It would be better to explicitly raise a 400 error for an incorrect set of filters.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -226,3 +280,122 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
+            "last_value": last_value,
+            "messages": [get_time_period_message()],
+        }
+
+    def query_elasticsearch(self) -> list:
+        filter_query = QueryWithFilters.generate_elasticsearch_query(self.filters, query_type="awards")
+        sort_field = self.get_elastic_sort_by_fields()
+        sorts = [{field: self.pagination["sort_order"]} for field in sort_field]
+        if self.last_id:
+            sorts.append({"generated_unique_award_id.keyword": "desc"})
```
BrianZito: so if the user doesn't supply a last_id, we don't sort the results?

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: The endpoint already checks for both empty filters and for incorrect award type codes and throws a 422, so it would be redundant.

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: No, we add sorting by `generated_unique_award_id` because there could be duplicate values for other sort fields and that could result in missing records. The normal sort fields are generated two lines above this in `sort_field = self.get_elastic_sort_by_fields()`

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: yeah. that's why they use `get`

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: Would it be possible to add this functionality to get_elastic_sort_by_fields? As it stands this violates the Single Responsibility Principle and risks logic getting lost.

usaspending_api/search/v2/views/spending_by_award.py
BrianZito: Not super dangerous (particularly since this is an intermediary state), but I would rather see these in the conditional below line 132 to ease the distinction between workflows.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -226,3 +280,122 @@ def populate_response(self, results: list, has_next: bool) -> dict:
             "page_metadata": {"page": self.pagination["page"], "hasNext": has_next},
             "messages": [get_time_period_message()],
         }
+
+    def populate_elastic_response(
```
BrianZito: What's the difference between the roles of `populate_elastic_response` and `construct_es_response`?

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: sure

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: well they do totally different things, but `populate_elastic_response` is pretty pointless so i will remove it

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: sure

usaspending_api/common/query_with_filters.py
```diff
@@ -132,7 +132,7 @@ def _generate_elasticsearch_query(cls, filter_values: List[str], query_type) ->
         fields = ["recipient_name"]
 
         for v in filter_values:
-            upper_recipient_string = v.upper()
+            upper_recipient_string = v.upper() + "*"
```
sethstoudenmier: What is the expected functionality? Currently in Prod if you search for "Boo" you only get two hits, but if you search for "Booz" you will get hits on Booz Allen Hamilton. Won't adding a wildcard make it so that "Boo" will also search for "Booz Allen Hamilton"?

emily-brents: Yeah, that's the idea. This was discussed with the PO - we want searches to return results if they partially match the search text - i.e., `boo` to `Booz Allen` and `lab` to `labs`, `laboratory`, etc.

sethstoudenmier: Few small things:

- Should we be able to search for "ooz" as well? Would need a wildcard at the front of the string.
- The wildcard needs to be added in the conditional (or a different variable needs to be created) since this will have unexpected results with the `if` of the conditional below. 
- The AC for your story and mine (spending_over_time) needs to be udpated to reflect this decision since they currently say they should mimic current functionality.

emily-brents: 1. The PO did not specify, I've asked for follow up, but generally we want to avoid having leading wildcards, because they cause performance hits compared to trailing wildcards
2. Already dealt with
3. Sure

usaspending_api/common/query_with_filters.py
```diff
@@ -20,16 +20,16 @@ class _Filter(metaclass=ABCMeta):
     underscore_name = None
 
     @classmethod
-    def generate_query(cls, filter_values: Union[str, list]) -> dict:
+    def generate_query(cls, filter_values: Union[str, list], query_type) -> dict:
```
tony-sappe: Since the type hints are already present, best to include for the new parameter

usaspending_api/awards/v2/lookups/elasticsearch_lookups.py
tony-sappe: Good thoughts. The proposed syntax was unavailable when the original code this is based on was written.

usaspending_api/search/tests/integration/test_spending_over_time.py
```diff
@@ -382,7 +382,7 @@ def _test_correct_response_for_keywords(client):
             {
                 "group": "fiscal_year",
                 "filters": {
-                    "keywords": ["test", "recipient_name_", "recipient_name_1"],
```
tony-sappe: was there an issue with the trailing underscores?

sethstoudenmier: I put that in the original test case to make sure that it wasn't being counted. Now that we are using the wildcard it would trigger every single recipient in the test.

usaspending_api/common/query_with_filters.py
tony-sappe: Leading wildcards are discouraged due to performance concerns. I doubt there is a strong-enough desire for that functionality to incur the performance penalty

usaspending_api/search/tests/integration/test_spending_over_time.py
tony-sappe: ah, thank you

usaspending_api/common/query_with_filters.py
emily-brents: fair, fixed

usaspending_api/awards/v2/lookups/elasticsearch_lookups.py
emily-brents: sure

PR #2213, opened 2020-01-10
[DEV-4109] Part 2: Improve Award Update Functions

tony-sappe:
**Description:**
Update functions and SQL in `award_helpers.py` for better award data quality

**Technical details:**
1. `update_date` is now being updated when the functions update the records (!!!)
1. SQL is stored outside of functions so other scripts can import them
1. SQL and functions are more consistent
1. Update SQL joins using `unique_award_key` instead of surrogate keys

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated (N/A)
3. [x] Necessary PR reviewers:
    - [x] Backend
4. [x] Matview impact assessment completed (N/A)
5. [x] Frontend impact assessment completed (N/A)
6. [x] Data validation completed
7. [x] Appropriate Operations ticket(s) created (N/A)
8. [ ] Jira Ticket [DEV-4109](https://federal-spending-transparency.atlassian.net/browse/DEV-4109):
    - [x] Link to this Pull-Request
    - [ ] Performance evaluation of affected API
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
No edits to materialized views or API behavior
```

PR #2212, opened 2020-01-10
[DEV-4109] Part 1: Minor FPDS loader edits

tony-sappe:
**Description:**
During the investigation as part of DEV-4109, minor edits were made to the loader for logging and clarity. This is the first PR for DEV-4109 and doesn't significantly change the loader. See below for details

**Technical details:**

1. Awards are updated immediately after transaction deletes are processed
1. All awards which had deleted transactions are re-calculated. Not just the awards with nulled FK
1. More logging
1. 3x Chunk size for batch processing which was faster locally (still smaller than the legacy script)
1. Removed "recipient_location" objects and mapping since they were not for the data load after #2199
1. Fixed batch sizes when reading IDs from a file

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated (N/A)
3. [x] Necessary PR reviewers:
    - [x] Backend
4. [x] Matview impact assessment completed (N/A)
5. [x] Frontend impact assessment completed (N/A)
6. [x] Data validation completed
7. [x] Appropriate Operations ticket(s) created (N/A)
8. [x] Jira Ticket [DEV-4109](https://federal-spending-transparency.atlassian.net/browse/DEV-4109):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected API
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
No edits to matviews or API
```

PR #2219, opened 2020-01-13
[DEV-4109] Part 3: Update Scripts

tony-sappe:
**Description:**
Building on #2212 and #2213, this PR updates scripts used to correct existing data problems 

**Technical details:**
- Leverage SQL in `award_helpers.py` instead of duplication
- New script to correct data. Also includes `--inspect` which will be included in automated ETL jobs

    - [x] Backend
8. [x] Jira Ticket [DEV-4109](https://federal-spending-transparency.atlassian.net/browse/DEV-4109):

```
No matviews were harmed in the making of this PR
No changes to API
```


usaspending_api/database_scripts/job_archive/management/commands/dev_4109_repair_awards.py
```diff
@@ -0,0 +1,122 @@
...
+
+logger = logging.getLogger("console")
+
+CHECK_MISSING_AWARDS = """
+    SELECT tf.{field}
+    FROM {table} AS tf
+    LEFT OUTER JOIN awards a on a.generated_unique_award_id = tf.unique_award_key
+    WHERE a.id IS NULL;
+"""
+
```
BrianZito: No comment explaining where it comes from and when it's obsolete?

tony-sappe: hectic day

usaspending_api/database_scripts/job_archive/management/commands/dev_4109_repair_awards.py
```diff
@@ -0,0 +1,169 @@
...
+        with Timer("Running Part 2: Fixing Transaction FKs"):
+            with connection.cursor() as cursor:
+                cursor.execute(fix_txn_fk_to_awards)
+                transactions = [str(row[0]) for row in cursor.fetchall()]
+            if transactions:
+                logger.info(f"Updated transactions:\n{', '.join(transactions)}")
+
+    def part_three(self):
+        logger.info("Part 3: Re-Computing Awards")
+        logger.warning("Run usaspending-api/usaspending_api/scripts/archive/recompute_all_awards.py")
```
BrianZito: It doesn't actually run the command?

tony-sappe: I hope that is a rhetorical question.

tony-sappe:
Adding do not merge. Will need to make changes to this script to facilitate new functionality in the daily ETL pipeline. This will most likely be shifted to a warmfix PR later this week

usaspending_api/database_scripts/job_archive/management/commands/dev_4109_repair_awards.py
BrianZito: I was asking if it was run somewhere else. It seems strange that part three of the script is to do absolutely nothing.

tony-sappe: Log statements are "absolutely nothing" now?

The script it is directing someone to run is pre-existing and the run last time took over 24hrs. While it could be called by this script the benefit seems minimal at best. The context for these scripts are to correct data and are manually run on an ad-hoc basis. These log statements are code documentation in case the operations ticket is lost or not easily found.

BrianZito: I misinterpreted that message as a statement of what the script was currently doing. Technically the future tense hints at that, but you might want to say "please run" or something to be more clear. Since it's a short-lived script I'm less worried about this since future generations of devs will (hopefully) never see this.

usaspending_api/database_scripts/job_archive/management/commands/dev_4109_repair_awards.py
```diff
@@ -0,0 +1,231 @@
...
+Jira Ticket Number(s): DEV-4109
+
+Expected CLI:
+
+    $ python3 manage.py dev_4109_repair_awards
+
+Purpose:
+
+    Correct Faulty FKs and missing award records
+"""
```
BrianZito: Is this script general use, or is there a time where we expect it to become obsolete?

tony-sappe: This script is to fix faulty data in the database. Optimally it would only need to be run once. However since I wasn't able to conclusively prove what the root cause is/was there is no guarantee the problem won't occur again. So the --inspect flag was added so the SQL can run on a daily basis to alert if the issue crops up again.

Between general use, and planned obsolescence it is most definitely the latter.

BrianZito: I've seen "can be safely deleted after " comments on these scripts before. I don't know if we consider that general practice.

tony-sappe: Currently no.
Here is the mandatory header in the README: https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/database_scripts/job_archive/README.md

PR #2231, opened 2020-01-17
[DEV-4206] Implement Awards and Transactions Downloads with Elasticsearch

sethstoudenmier:
**Description:**
Implement the `/download/awards` and `/download/transactions` endpoints with Elasticsearch and have it live alongside the current Postgres implementation, accessible via an HTTP Header.

**Technical details:**
Two parts:

1. The download endpoints were implemented so that we can condense the `where` clause in the Postgres queries that we use for downloads. Similar to current Keyword search downloads, the filters are used to search for all Award or Transaction IDs on Elasticsearch. This list of IDs is then put into the `where` clause of a Postgres query where it searches for all IDs that are in the Array.
2. Refactored the use of the `elasticsearch-dsl` Search object to having unique objects for both of our indexes. Hope was to also clean up how `client.py` was being used to generate an Elasticsearch object, handle errors, and handle retrying the ES query. Left `client.py` in place for now while we have Elasticsearch being used with and without the DSL.

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [X] Jira Ticket [DEV-4206](https://federal-spending-transparency.atlassian.net/browse/DEV-4206):

```
```


usaspending_api/common/query_with_filters.py
```diff
@@ -12,6 +13,11 @@
 logger = logging.getLogger(__name__)
 
 
+class _QueryType(Enum):
```
emily-brents: nice 👍

usaspending_api/download/helpers/elasticsearch_download_functions.py
```diff
@@ -0,0 +1,95 @@
...
+        return flat_ids
+
+    @classmethod
+    @abstractmethod
+    def query(cls, filters: dict) -> QuerySet:
+        pass
+
+
+class AwardsElasticsearchDownload(_ElasticsearchDownload):
+    _source_field = "award_id"
```
emily-brents: I know we switched to using `generated_unique_award_id` to identify awards on the website, do we want to use those here as well? I don't think it'll actually make any difference, but it would be more consistent.

sethstoudenmier: Was torn on this. I am not against changing it, but feel more comfortable using `award_id` in the API since it is the PK for the `vw_award_search` view.

PR #2233, opened 2020-01-21
[DEV-4109] Part 3.1: Follow-up tweaks to script after test runs

tony-sappe:
**Description:**
Using `unique_award_key` for the award update SQL caused two issues:
1. Degraded performance (known)
1. Need to properly handle `unique_award_key` values with quotes

Also there was an issue with the fix script which didn't appropriately return a non-0 exit code for Jenkins to handle.

**Technical details:**
Since one script will check and correct bad FK values in `transaction_normalized`, using the FKs will not introduce data quality issues. The script will be run on a daily basis with a flag to test the FKs and ensure any drift is caught immediately and flagged.

    - [ ] Backend
8. [x] Jira Ticket [DEV-4109](https://federal-spending-transparency.atlassian.net/browse/DEV-4109):

```
No changes to API behavior
No matviews were altered
```

### DIFF
--- a/usaspending_api/database_scripts/job_archive/recompute_all_awards.py
+++ b/usaspending_api/database_scripts/job_archive/recompute_all_awards.py
@@ -8,13 +8,12 @@ Expected CLI:
 
 Purpose:
 
-    Adding new earliest_transaction to awards
-    Improved logic for determining the initial transaction of an award which is
-    used as the "earliest transaction" FK. Also make a mirror alteration to
-    the logic determining "latest transaction" for an award.
-    Re-calculate awards using the unique_award_key
+    Re-calculate all award fields to ensure the awards are accurate and up-to-date
+    Can choose a range of award.id values to process, can also limit to FABS or FPDS
+    for subset update runs.
 
-    All of these changes requires updating 100% of award records
+    Same SQL used in usaspending_api/etl/award_helpers.py
+
 
 """
 import argparse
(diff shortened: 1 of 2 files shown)


usaspending_api/database_scripts/job_archive/recompute_all_awards.py
```diff
@@ -44,7 +43,7 @@
     CONNECTION_STRING = environ["DATABASE_URL"]
 except Exception:
     print("SET env var DATABASE_URL!!!\nTerminating script")
-    raise SystemExit(1)
+    raise SystemExit(100)
```
BrianZito: Why status 100? Is this meaningful to Jenkins?

tony-sappe: More like "why not?"
No significance.

BrianZito: Giving non-standard feedback to something well established like program exit codes can risk confusion. "1" is a standard "The program broke, and this code has no special information why" response, while "100" seems like something specific and intentional. No harm besides making some poor guy do some pointless googling though.

tony-sappe: It is useful to standardize on scripts to use the same exit codes for similar errors. That way someone knowledgeable with the script can use the exit code to reference where and why it failed.

We don't do that at the moment. 100 isn't known to be tied to a specific meaning, so it won't be confused with existing patterns. It would be trivial to set them all to 1 if you believe it is too confusing

PR #2241, opened 2020-01-23
[DEV-1666] New source transaction models

tony-sappe:
**Description:**
Groundbreaking PR for the [database re-architecture initiative](https://federal-spending-transparency.atlassian.net/browse/DEV-3752).

- Two new tables: `source_assistance_transaction` & `source_procurement_transaction`
- 1-to-1 mapping to the data source tables in Broker
- Rudimentary ETL integration tests in-place

**Technical details:**

- No data transformation is occurring, these tables are designed to hold perfect copies of Broker data
- Leveraging project ETL objects for abstracting the copy SQL
- The Insert From operations are fast, at least 1650/s (~6 million / hr)

    - [ ] Backend
8. [x] Jira Ticket [DEV-1666](https://federal-spending-transparency.atlassian.net/browse/DEV-1666):

```
No impact to the API
No changes to materialized views
```


usaspending_api/transactions/tests/integration/test_assistance_etl.py
```diff
@@ -0,0 +1,268 @@
...
+(E'2018-11-28 20:12:41.143701',E'2018-11-28 20:12:41.143706',89126418,E'20181003',E'A',E'10',E'AGRICULTURAL RISK COVERAGE PROG - COUNTY',E'Redacted due to PII',NULL,E'012',E'12346T',E'12D2',NULL,E'NON',E'P',E'10.112',NULL,NULL,E'12FA00PY57991082',1468,NULL,E'012',E'12346T',E'12D2',NULL,NULL,NULL,E'USA',NULL,NULL,NULL,E'57366',NULL,0,NULL,E'20181003',E'20181003',E'SD48460',E'00',E'USA',NULL,E'57366',3,E'SAI EXEMPT',NULL,E'00',E'1468.00',E'Price Loss Coverage',E'Department of Agriculture (USDA)',E'Farm Service Agency',E'Department of Agriculture (USDA)',E'Farm Service Agency',NULL,E'Hutchinson',E'South Dakota',E'PARKSTON',E'PARKSTON',E'067',E'Hutchinson',E'SD',E'South Dakota',E'2018-11-28 20:12:41.14262',E'12D2_12
```
sethstoudenmier: Don't think we need this test since it already exists as part of `test_broker_integration.py`.

tony-sappe: Correct. And I can remove it from both of the new test files.

It takes a fraction of a second to run and is useful for troubleshooting.

usaspending_api/common/custom_django_fields.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class NaiveTimestampField(models.Field):
+    """timestamp column type without timezone"""
+
+    def db_type(self, connection):
+        return "timestamp without time zone"
+
+
+class BooleanFieldWithDefault(models.Field):
```
sethstoudenmier: I wonder if the issue you had with default value being ignored was fixed between 1.11 and 2.2. Tried to find something, but had no luck.

usaspending_api/transactions/loader_functions.py
```diff
@@ -0,0 +1,52 @@
...
+def store_ids_in_file(id_iter: List, file_name: str = "temp_file") -> Tuple[Path, int]:
+    """Store the provided values in a file
+        returning filepath and count of ids
+    """
+    total_ids = 0
+    file_path = Path(file_name)
+
+    with open(str(file_path), "w") as f:
+        for id_string in id_iter:
+            if not id_string:
```
sethstoudenmier: Couldn't this be simplified to `if id_string is not None` and then the contents of the "for"? Avoiding the use of `continue` when it isn't needed.

usaspending_api/transactions/tests/integration/test_assistance_etl.py
```diff
@@ -0,0 +1,247 @@
...
```
sethstoudenmier: You don't actually use monkeypatch in either of your tests.

usaspending_api/common/custom_django_fields.py
tony-sappe: Possibly. Haven't tested anything in Django 2.2 yet

usaspending_api/transactions/loader_functions.py
tony-sappe: I try to think in terms of guard clauses https://refactoring.guru/replace-nested-conditional-with-guard-clauses and place the least-amount of code nested under the conditional

usaspending_api/transactions/tests/integration/test_assistance_etl.py
tony-sappe: thanks

usaspending_api/common/custom_django_fields.py
kbard: What was the problem?

Remember that Django defaults are only at the ORM level.  They aren't mirrored in the database as this is doing.  If you want defaults in the database, better to use an unmanaged model.  When we synced the database with the ORM a year or so ago, I had to add some SQL to the migrations to mimic defaults for some database columns that had them.

Doing it this way allows Django to support callable defaults (functions).

usaspending_api/common/custom_django_fields.py
```diff
@@ -0,0 +1,22 @@
+from django.db import models
+
+
+class NumericField(models.Field):
```
kbard: In Postgres, `numeric` is equivalent to `decimal`.  Could you not just use the built in `decimal` data type here or was the point of this specifically not to have to supply precision and scale values?

usaspending_api/common/custom_django_fields.py
tony-sappe: The "problem" was the Django migration didn't set the default value to `false` which is how the Broker DB field is defined. Using the custom field was an easy approach to define the field in Djnago and allow migrations to occur with the desired end state. There is no discrepancy between the model and the table.  With the exception of the materialized views, all table schemas should be managed by django.

This is the resultant column in the table when using the custom field:
```is_active boolean NOT NULL DEFAULT false,```

usaspending_api/common/custom_django_fields.py
tony-sappe: My goal was the duplicate Broker's schema 100%. Their columns are defined as `numeric` without precision and scale. It isn't what I would choose, but that's the source data

usaspending_api/common/custom_django_fields.py
kbard: If we're copying the data from Broker, is having a default value even necessary?  Missing values would have already been converted to their defaults in their data.

tony-sappe: Probably not necessary.
Very true.

Rigorously matching the Broker's schema might cause issues down the road. We can weigh the pros and cons later if others devs have good ideas.

PR #2243, opened 2020-01-23
[DEV-4224] ES Pagination Past 50,000 records

emily-brents:
**Description:**
Adds pagination past 50,000 records to the `/api/v2/spending_by_award/` endpoint when using elasticsearch. There are some issues with the implementation that limit the solutions we can use. 
- The `scroll` api or the partitioning method we use for downloads do not sort the results, which means we can't use them for advanced search
- The `search_after` method does allow sorting ( it requires sorting), but requires a value to mark the location in the results to search after (hence the name). This means that we cannot support jumping to random pages - so we have to enforce the infinite scroll / page by page approach.
- The only way to allow jumping to a random page, as far as I can tell, is to do a postgres query for the last record before the requested page, then doing a search after query on elasticsearch using that value from postgres. This defeats the entire point of switching to elasticsearch, so I don't recommended actually doing this.

    - [x] Backend
    - [ ] Frontend
    - [x] Operations 
    - [x] Domain Expert 
7. [X] Appropriate Operations ticket(s) created - https://federal-spending-transparency.atlassian.net/browse/OPS-1092
8. [X] Jira Ticket [DEV-4224](https://federal-spending-transparency.atlassian.net/browse/DEV-4224):

```
No changes to matviews
```
Request - 
```
{
    "filters": {
        "time_period": [{"start_date": "2007-10-01", end_date": "2020-09-30"}],
        "award_type_codes": ["A","B", "C","D"]
    },
    "fields": ["Award ID","Recipient Name","Start Date","End Date","Award Amount", "Awarding Agency", "Awarding Sub Agency", "Contract Award Type","recipient_id", "prime_award_recipient_id"],
    "page": 500,
    "limit": 100,
    "sort": "Award Amount",
    "order": "desc",
    "subawards": false
    "last_record_unique_id": 17680500,
    "last_record_sort_value": "349.18"
}
```

Response (Before) - 1986ms
```
{
    "limit": 100,
    "results": [
        {
            "internal_id": 3820500,
            "Award ID": "0427",
            "Recipient Name": "GARY JET CENTER, INC.",
            "Start Date": "2016-01-26",
            "End Date": "2016-01-26",
            "Award Amount": 349.17,
            "Awarding Agency": "Department of Defense",
            "Awarding Sub Agency": "Defense Logistics Agency",
            "Contract Award Type": "DELIVERY ORDER",
            "recipient_id": "57260d32-949a-398a-a674-b280da6920a8-C",
            "prime_award_recipient_id": null,
            "generated_internal_id": "CONT_AWD_0427_9700_SP060012D0037_9700"
        },
        {
            "internal_id": 21132500,
            "Award ID": "GSNPNE4W2270",
            "Recipient Name": "F.P. WOLL & COMPANY",
            "Start Date": "2010-06-21",
            "End Date": "2010-07-24",
            "Award Amount": 349.1,
            "Awarding Agency": "General Services Administration",
            "Awarding Sub Agency": "Federal Acquisition Service",
            "Contract Award Type": "DO",
            "recipient_id": "694ba8a1-feb9-c577-a31e-cd5158856b66-C",
            "prime_award_recipient_id": null,
            "generated_internal_id": "CONT_AWD_GSNPNE4W2270_4730_GS02FS0005_4730"
        },
        {
            "internal_id": 21142500,
            "Award ID": "GSNPNE6R5001",
            "Recipient Name": "F.P. WOLL & COMPANY",
            "Start Date": "2010-10-28",
            "End Date": "2010-11-30",
            "Award Amount": 349.1,
            "Awarding Agency": "General Services Administration",
            "Awarding Sub Agency": "Federal Acquisition Service",
            "Contract Award Type": "DO",
            "recipient_id": "694ba8a1-feb9-c577-a31e-cd5158856b66-C",
            "prime_award_recipient_id": null,
            "generated_internal_id": "CONT_AWD_GSNPNE6R5001_4732_GS02FS0005_4730"
        },
...
]
,
    "page_metadata": {
        "page": 500,
        "hasNext": true
    },
    "messages": [
        "For searches, time period start and end dates are currently limited to an earliest date of 2007-10-01.  For data going back to 2000-10-01, use either the Custom Award Download feature on the website or one of our download or bulk_download API endpoints as listed on https://api.usaspending.gov/docs/endpoints."
    ]
```

After -  355ms
```
{
    "limit": 100,
    "results": [
        {
            "internal_id": 3820500,
            "Award ID": "0427",
            "Recipient Name": "GARY JET CENTER, INC.",
            "Start Date": "2016-01-26",
            "End Date": "2016-01-26",
            "Award Amount": 349.17,
            "Awarding Agency": "Department of Defense",
            "Awarding Sub Agency": "Defense Logistics Agency",
            "Award Type": null,
            "recipient_id": "57260d32-949a-398a-a674-b280da6920a8-C",
            "prime_award_recipient_id": null,
            "generated_internal_id": "CONT_AWD_0427_9700_SP060012D0037_9700"
        },
        {
            "internal_id": 21142500,
            "Award ID": "GSNPNE6R5001",
            "Recipient Name": "F.P. WOLL & COMPANY",
            "Start Date": "2010-10-28",
            "End Date": "2010-11-30",
            "Award Amount": 349.1,
            "Awarding Agency": "General Services Administration",
            "Awarding Sub Agency": "Federal Acquisition Service",
            "Award Type": null,
            "recipient_id": "694ba8a1-feb9-c577-a31e-cd5158856b66-C",
            "prime_award_recipient_id": null,
            "generated_internal_id": "CONT_AWD_GSNPNE6R5001_4732_GS02FS0005_4730"
        },
        {
            "internal_id": 21132500,
            "Award ID": "GSNPNE4W2270",
            "Recipient Name": "F.P. WOLL & COMPANY",
            "Start Date": "2010-06-21",
            "End Date": "2010-07-24",
            "Award Amount": 349.1,
            "Awarding Agency": "General Services Administration",
            "Awarding Sub Agency": "Federal Acquisition Service",
            "Award Type": null,
            "recipient_id": "694ba8a1-feb9-c577-a31e-cd5158856b66-C",
            "prime_award_recipient_id": null,
            "generated_internal_id": "CONT_AWD_GSNPNE4W2270_4730_GS02FS0005_4730"
        },
     ...
    ],
    "page_metadata": {
        "page": 500,
        "hasNext": true,
        "last_record_unique_id": 12176000,
        "last_record_sort_value": "345.36"
    },
    "messages": [
        "For searches, time period start and end dates are currently limited to an earliest date of 2007-10-01.  For data going back to 2000-10-01, use either the Custom Award Download feature on the website or one of our download or bulk_download API endpoints as listed on https://api.usaspending.gov/docs/endpoints."
    ]
}
```


usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -287,6 +278,30 @@ def query_elasticsearch(self) -> list:
         filter_query = QueryWithFilters.generate_awards_elasticsearch_query(self.filters)
         sort_field = self.get_elastic_sort_by_fields()
         sorts = [{field: self.pagination["sort_order"]} for field in sort_field]
+        record_num = (self.pagination["page"] - 1) * self.pagination["limit"]
+        if record_num >= settings.ES_AWARDS_MAX_RESULT_WINDOW and not self.last_id and not self.last_value:
```
kwhickey: Are `self.last_id` and `self.last_value` taken as integers or strings? And can one be populated and not the other? 

Questions pointing towards some ambiguity. It may be best to use `None` as a comparison rather than falsey in case `total_obligation` is valid as `0`. 

Also adding parentheses: 

If both have to be there to to be used in manually specifying what page to land on:
`if record_num >= settings.ES_AWARDS_MAX_RESULT_WINDOW and (self.last_id is None and self.last_value is None):`

If only one needs to be there to be used in manually specifying what page to land on:
`if record_num >= settings.ES_AWARDS_MAX_RESULT_WINDOW and not (self.last_id is None and self.last_value is None):`

... though I think I may have messed up that last one, but hopefully you get the gist.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -287,6 +278,30 @@ def query_elasticsearch(self) -> list:
...
+            )
+            self.pagination["upper_bound"] = record_num
+            self.pagination["lower_bound"] = record_num - 1
+            queryset = self.construct_queryset()
+            if len(queryset) != 1:
+                return {}
+            results = {}
+            for result in queryset:
+                results["award_id"] = result.get("award_id")
+                results["total_obligation"] = float(result.get("total_obligation"))
```
kwhickey: Since you already forcibly  checked it to have 1 result, maybe no loop and just get the one value, and not results dict?
```
last_record_award_id = queryset[0].get("award_id")
last_record_total_obligation = float(queryset[0].get("total_obligation"))
# ...
.extra(search_after=[last_record_total_obligation, last_record_award_id])[0 : self.pagination["limit"]]
```
?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -287,6 +278,30 @@ def query_elasticsearch(self) -> list:
...
         sorts = [{field: self.pagination["sort_order"]} for field in sort_field]
+        record_num = (self.pagination["page"] - 1) * self.pagination["limit"]
+        if record_num >= settings.ES_AWARDS_MAX_RESULT_WINDOW and not self.last_id and not self.last_value:
+            sorts.append({"award_id": self.pagination["sort_order"]})
+            logger.warning(
+                "Jumping to random page past 50,000 records will have a performance hit when using Elasticsearch"
+            )
+            self.pagination["upper_bound"] = record_num
+            self.pagination["lower_bound"] = record_num - 1
+            queryset = self.construct_queryset()
```
kwhickey: I see now. So this is where it runs a database query for 1 supposed record, in order to use that db record to jump the search_after ahead. 

It's not elegant, and does chain us back to the db. But it's probably more efficient than what I was thinking - which was to "scroll" through the ES pages one page at a time on the server-side until you reached the desired page.

This does run the risk of not accurately paging to the right page in ES if the source DB data (matview) and ES index are not completely in sync.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -287,6 +278,30 @@ def query_elasticsearch(self) -> list:
...
+                return {}
+            results = {}
+            for result in queryset:
+                results["award_id"] = result.get("award_id")
+                results["total_obligation"] = float(result.get("total_obligation"))
+            search = (
+                AwardSearch()
+                .filter(filter_query)
+                .sort(*sorts)
+                .extra(search_after=[results["total_obligation"], results["award_id"]])[0 : self.pagination["limit"]]
```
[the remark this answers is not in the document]
emily-brents: No, `award_id` is the award table pk from the database, so if there are duplicates there then we have bigger problems.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -287,6 +278,30 @@ def query_elasticsearch(self) -> list:
         filter_query = QueryWithFilters.generate_awards_elasticsearch_query(self.filters)
         sort_field = self.get_elastic_sort_by_fields()
         sorts = [{field: self.pagination["sort_order"]} for field in sort_field]
+        record_num = (self.pagination["page"] - 1) * self.pagination["limit"]
+        if record_num >= settings.ES_AWARDS_MAX_RESULT_WINDOW and not self.last_id and not self.last_value:
+            sorts.append({"award_id": self.pagination["sort_order"]})
+            logger.warning(
+                "Jumping to random page past 50,000 records will have a performance hit when using Elasticsearch"
+            )
```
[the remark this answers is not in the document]
emily-brents: yeah, good point, will fix

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: last_id should be an integer, last_value can be any of the sort fields, so it should be a string. The none comparison is a good point, will change

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: yeah that's fair, will fix

usaspending_api/search/v2/views/spending_by_award.py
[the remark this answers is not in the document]
kwhickey: Ah ... Sorry, i had it backwards. I thought we were sorting by award_id and using `total_obligation` as the tie breaker. It's the other way around.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -338,24 +395,22 @@ def construct_es_reponse(self, response) -> dict:
...
             "page_metadata": {
                 "page": self.pagination["page"],
-                "hasNext": response.hits.total - (self.pagination["page"] - 1) * self.pagination["limit"]
-                > self.pagination["limit"],
-                "last_id": last_id,
-                "last_value": last_value,
+                "hasNext": total - (self.pagination["page"] - 1) * self.pagination["limit"] > self.pagination["limit"],
+                "last_record_unique_id": last_record_unique_id,
+                "last_record_tie_break_value": last_record_sort_value,
+                "last_record_sort_value": last_record_tie_break_value,
```
kwhickey: Did these get mixed up? `name: value`. The lefts don't match the rights. 
Or am I not seeing correctly how the the pagination logic is applied?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -155,14 +156,27 @@ def validate_request_data(request_data):
...
                 "type": "text",
                 "text_type": "search",
                 "required": False,
                 "allow_nulls": True,
             },
-            {"name": "last_value", "key": "last_value", "type": "float", "required": False, "allow_nulls": True},
+            {
+                "name": "last_record_tie_break_value",
+                "key": "last_record_tie_break_value",
+                "type": "float",
```
kwhickey: If this is going to be `award_id`, should its type be `integer`? Either way, double check the types for all these fields.

I may have confused things when suggesting the names for these too. 

Effectively you're looking for a field too sort by, and a field that is unique, if the sort field is duplicated. The 2nd one (unique) is _also_ deemed the tie-breaker. Right? 

So it's possibly you only need fields like
1. `last_record_sort_value`
2. `last_record_unique_id` 

... and not mention anything about a tie-breaker field.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,34 +294,80 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
+                    record=self.pagination["lower_bound"],
+                    es_limit=settings.ES_AWARDS_MAX_RESULT_WINDOW,
+                )
+            )
+            self.pagination["upper_bound"] = record_num
+            self.pagination["lower_bound"] = record_num - 1
+            queryset = self.construct_queryset()
+            if len(queryset) != 1:
+                return {}
+            results = []
```
kwhickey: Redundant here. Set right below.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,34 +294,80 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
                 .filter(filter_query)
-                .sort(*sorts)[
-                    ((self.pagination["page"] - 1) * self.pagination["limit"]) : (
-                        ((self.pagination["page"] - 1) * self.pagination["limit"]) + self.pagination["limit"]
-                    )
-                ]
+                .sort(*sorts)
+                .extra(
+                    search_after=[
+                        self.date_to_epoch_millis(self.last_record_sort_value),
```
kwhickey: So in this case, the api user has specified the `last_record_sort_value` directly -- effectively asking to jump to a _that_ page. If this is a new field in the API response (especially if we're going to just get rid of it soon), could you just enforce that they have to provide its value in epoch millis?

I'm also find leaving it as is. Seems to be a nice convenience to make the api call easier. But if the date conversion stuff gets messy, maybe just fallback to expecting it in millis.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,34 +294,80 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
         )
         response = search.handle_execute()
 
         return response
 
-    def construct_es_reponse(self, response) -> dict:
+    def construct_es_response(self, response) -> dict:
         results = []
+        if response is None:
+            return {"There was an error connecting to the Elasticsearch cluster"}
```
kwhickey: In general we should not return lower level details about backend problems through the API. 
- If this is a hint to the developer, it should be logged, so the developer can inspect logs
- For the API caller, it should probably throw an exception with a friendly non-detailed error message that something went wrong (or however we would return a `500` code with generic error message)

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: YUP that's just a mistake

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: I ended up adding a third sort field, since postgres has a default sort on the matviews (`total_obligation`), and in order to compare results I wanted to mimic that as close as possible. That's why this is a `float` not an `int`. I don't know if we want to keep that behavior, I don't know if that's something that we deliberately did or if it was done to try and speed up advanced search since the total_obligation sort is the default sort on the advanced search page.

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -68,6 +68,7 @@ SELECT
   vw_award_search.pop_zip5,
   vw_award_search.pop_congressional_code,
   vw_award_search.pop_city_name,
+  vw_award_search.pop_city_code,
```
tony-sappe: What about the addition of `vw_award_search.recipient_location_city_code` ?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -338,24 +385,26 @@ def construct_es_reponse(self, response) -> dict:
...
+            has_next = True
+
+        if len(response) > 0 and has_next:
+            last_record_unique_id = response[len(response) - 2].meta.sort[1]
+            last_record_sort_value = response[len(response) - 2].meta.sort[0]
+
         return {
             "limit": self.pagination["limit"],
-            "results": results,
+            "results": results[0 : self.pagination["limit"]],
```
tony-sappe:
```suggestion
            "results": results[:self.pagination["limit"]],
```

usaspending_api/database_scripts/etl/award_delta_view.sql
emily-brents: I was basing the fields on the lookups in `matview_lookups.py`, which only included the `pop_city_code`. We can include the `recipient_location_city_code` as well, it's just not currently included.

tony-sappe: Gotcha. The lookups are for the table columns, but not the search/filter operations. It is possible `recipient_location_city_code` isn't necessary like `pop_city_code`.

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: Removed

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: Fixed

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: I ended up removing the third field, so disregard the previous comment

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,43 +288,82 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
                 row[field] = hit.get(
-                    self.constants["elasticsearch_type_code_to_field_map"][hit[self.constants["award_semaphore"]]].get(
-                        field
-                    )
+                    str(
+                        self.constants["elasticsearch_type_code_to_field_map"][
+                            hit[self.constants["award_semaphore"]]
+                        ].get(field)
+                    ).replace(".keyword", "")
                 )
```
kwhickey: ew. Sorry you had to do this ... dictionary-maths. Looks like a holdover from maybe a not-so good kind of polymorphism. 

(It doesn't help that I don't think `semaphore` is being used correctly here. Probably should just be `award_type_code` or `award_discriminator` or `award_partition_key`...)

But, since it's a non-obvious piece of code, please put a comment at the top. Like
`# Parsing API response values from ES query result JSON`

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,43 +288,82 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
             for field in self.fields:
                 row[field] = hit.get(
-                    self.constants["elasticsearch_type_code_to_field_map"][hit[self.constants["award_semaphore"]]].get(
-                        field
-                    )
+                    str(
+                        self.constants["elasticsearch_type_code_to_field_map"][
+                            hit[self.constants["award_semaphore"]]
+                        ].get(field)
+                    ).replace(".keyword", "")
```
kwhickey: And why is the `.keyword` dropped?

Even though we query with `.keyword` on the field name, does ES truncate that in the results JSON?

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -338,29 +382,38 @@ def construct_es_reponse(self, response) -> dict:
...
-                "last_id": last_id,
-                "last_value": last_value,
+                "hasNext": has_next,
+                "last_record_unique_id": last_record_unique_id,
+                "last_record_sort_value": last_record_sort_value,
             },
             "messages": [get_time_period_message()],
         }
 
     def append_recipient_hash_level(self, result) -> dict:
```
kwhickey: I'm fine with the change here... but I read on to the impl of this function. 
Has this always been there, or is it new for the ES impl?
Issuing an additional SQL query for each result to be returned in the response? 

Seems like a major bottleneck risk for performance. 
Maybe not so bad when paginating only 60 at a time, but if accessing this not through the website, and getting many more results ... not so good.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,43 +288,82 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
                 AwardSearch()
                 .filter(filter_query)
-                .sort(*sorts)[
-                    ((self.pagination["page"] - 1) * self.pagination["limit"]) : (
-                        ((self.pagination["page"] - 1) * self.pagination["limit"]) + self.pagination["limit"]
-                    )
-                ]
+                .sort(*sorts)
+                .extra(
+                    search_after=[self.date_to_epoch_millis(self.last_record_sort_value), self.last_record_unique_id]
```
kwhickey: So here, in the scenario where the API caller has _provided_ `last_record_sort_value` in the API request POST json...

Are we expecting this value to be provided as a normal date? Or as epoch-millis?

Because, we _return_ (in the HTTP Response) it from this endpoint as epoch-millis ... so for consistency, I would expect them to provide it back to us in the same format (in which case it does not need to be converted).

usaspending_api/etl/es_award_template.json
```diff
@@ -29,11 +29,9 @@
   },
   "sort": {
     "fields": [
-      "total_obligation",
-      "action_date"
+      "total_obligation"
```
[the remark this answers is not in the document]
kwhickey: Nvm. Looks like this sorting will need to be removed entirely once the two branches are merged together. (due to incompatibility of sorting indexes and nested docs)

usaspending_api/awards/v2/lookups/elasticsearch_lookups.py
```diff
@@ -27,22 +27,22 @@
...
-    "Awarding Agency": "awarding_toptier_agency_name",
-    "Awarding Agency Code": "awarding_toptier_agency_code",
-    "Awarding Sub Agency": "awarding_subtier_agency_name",
-    "Awarding Sub Agency Code": "awarding_subtier_agency_code",
-    "Funding Agency": "funding_toptier_agency_name",
-    "Funding Agency Code": "funding_toptier_agency_code",
-    "Funding Sub Agency": "funding_subtier_agency_name",
-    "Funding Sub Agency Code": "funding_subtier_agency_code",
-    "Place of Performance City Code": "pop_city_code",
+    "Recipient Name": "recipient_name.keyword",
```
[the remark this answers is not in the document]
kwhickey: Talked: 
- should have had this `.keyword` on for all these. 
- would have just not matched queries against these fields' values (now it will)
- would need to do a deeper analysis on which of these are used for full-text search and which aren't

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: `award_semaphore` is just a fancy name for ``award_type_flag`.
If the variable name accuracy is in question, it could be something like `award_type_code_fieldname`

Why could't the dict fields be modified cleanly beforehand like this?
https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/awards/v2/lookups/elasticsearch_lookups.py#L91
```python
TRANSACTIONS_SOURCE_LOOKUP = {key: value.replace(".keyword", "") for key, value in TRANSACTIONS_LOOKUP.items()}
```

usaspending_api/search/v2/views/spending_by_award.py
tony-sappe: Max number of results per API request is 100. While I agree this isn't optimal, the minor performance hit was documented when it was added.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,43 +288,82 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
+                )[: self.pagination["limit"] + 1]
             )
+            if self.last_record_sort_value is not None and self.last_record_unique_id is not None
+            else (AwardSearch().filter(filter_query).sort(*sorts)[record_num : record_num + self.pagination["limit"]])
```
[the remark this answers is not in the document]
emily-brents: changed to if-else block

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,43 +288,82 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
-                    ((self.pagination["page"] - 1) * self.pagination["limit"]) : (
-                        ((self.pagination["page"] - 1) * self.pagination["limit"]) + self.pagination["limit"]
-                    )
-                ]
+                .sort(*sorts)
+                .extra(
+                    search_after=[self.date_to_epoch_millis(self.last_record_sort_value), self.last_record_unique_id]
+                )[: self.pagination["limit"] + 1]
             )
+            if self.last_record_sort_value is not None and self.last_record_unique_id is not None
```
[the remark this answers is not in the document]
emily-brents: added exception for if one parameter is missing

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -283,43 +288,82 @@ def populate_response(self, results: list, has_next: bool) -> dict:
...
+
     def query_elasticsearch(self) -> list:
         filter_query = QueryWithFilters.generate_awards_elasticsearch_query(self.filters)
         sort_field = self.get_elastic_sort_by_fields()
         sorts = [{field: self.pagination["sort_order"]} for field in sort_field]
-        search = (
-            (
+        record_num = (self.pagination["page"] - 1) * self.pagination["limit"]
+
+        if record_num >= settings.ES_AWARDS_MAX_RESULT_WINDOW and (
```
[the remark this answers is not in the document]
emily-brents: added

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: Used Tony's suggestion of using a source lookup, removed the `.replace(".keyword","")`

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: Remove the conversion on the API-provided value, kept it for the postgres lookup, which does need to be converted from date to epoch-millis

PR #2258, opened 2020-01-31
Preemptive increase of max API date to FY2024

tony-sappe:
**Description:**
Changing from end of FY2020 (7 months away) to the end of FY 2024

**Technical details:**
N/A

    - [ ] Backend
8. [ ] Jira Ticket (N/A)

```
Minor config change
```


BrianZito:
2024 seems arbitrary. Is there any reason to have a max API date? If so, is there some reason why four years from now is the right amount to push it to?

tony-sappe:
> 2024 seems arbitrary. Is there any reason to have a max API date? If so, is there some reason why four years from now is the right amount to push it to?

Yes it is arbitrary. The upper date was set in-place when the start and end dates were optional for each time period. Setting to 2024 means it doesn't need to be updated for 4.5 years

BrianZito:
> > 2024 seems arbitrary. Is there any reason to have a max API date? If so, is there some reason why four years from now is the right amount to push it to?
> 
> Yes it is arbitrary. The upper date was set in-place when the start and end dates were optional for each time period. Setting to 2024 means it doesn't need to be updated for 4.5 years

Do we have concerns with users requesting date ranges in the future, and getting "inaccurate" results? If we don't, why not set the max to 3024 and prevent this from ever becoming an issue again? Since we don't have any data from the future I don't see how it could be a performance concern.

tony-sappe:
> > > 2024 seems arbitrary. Is there any reason to have a max API date? If so, is there some reason why four years from now is the right amount to push it to?
> > 
> > 
> > Yes it is arbitrary. The upper date was set in-place when the start and end dates were optional for each time period. Setting to 2024 means it doesn't need to be updated for 4.5 years
> 
> Do we have concerns with users requesting date ranges in the future, and getting "inaccurate" results? If we don't, why not set the max to 3024 and prevent this from ever becoming an issue again? Since we don't have any data from the future I don't see how it could be a performance concern.

Hard to get inaccurate results for data which don't exist yet :-)
Potentially the max date can be refactored out in the near future after recent changes to the API. If not, I would recommend the max date be dynamically set to the end of the next FY. But that is for future developers to figure out. This PR was just to complete a reoccurring reminder since November.

PR #2264, opened 2020-02-05
[DEV-4063][DEV-4064] Implement Spending by Awarding Agency and Subagency

sethstoudenmier:
**Description:**
Implemented Spending by Category for Awarding Agency and Awarding Subagency with Elasticsearch and refactored the current Postgres implementation.

**Technical details:**
Created a base class to be used by the different `spending_by_category` endpoints. This PR implements both the Awarding Agency and Awarding Subagency endpoints since they are both similar and to ensure that the base class would work as intended. Also, noticed that I did not fully understand the needs for Agency IDs when the `transaction_delta_view` was updated a few sprints ago. As a result, some changes were needed to allow for the agency endpoints to match current Postgres implementation

    - [x] Backend
8. [X] Jira Tickets [DEV-4063](https://federal-spending-transparency.atlassian.net/browse/DEV-4063) [DEV-4064](https://federal-spending-transparency.atlassian.net/browse/DEV-4064):

```
```


usaspending_api/search/helpers/spending_by_category_helpers.py
```diff
@@ -0,0 +1,66 @@
+import logging
+from typing import Tuple
+
+from usaspending_api.recipient.models import StateData
+from usaspending_api.references.models import Agency, Cfda, PSC, RefCountryCode, NAICS
+
+logger = logging.getLogger(__name__)
+
+
+def fetch_agency_tier_id_by_agency(agency_name: str, is_subtier: bool = False) -> int:
```
tony-sappe: Any function which includes
```python
return None
```

needs wrap the return type with `Optional[]` 
https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type

usaspending_api/search/helpers/spending_by_category_helpers.py
```diff
@@ -0,0 +1,66 @@
+import logging
+from typing import Tuple
+
+from usaspending_api.recipient.models import StateData
+from usaspending_api.references.models import Agency, Cfda, PSC, RefCountryCode, NAICS
+
+logger = logging.getLogger(__name__)
```
tony-sappe: 👍

usaspending_api/database_scripts/etl/transaction_delta_view.sql
```diff
@@ -120,6 +124,26 @@ LEFT JOIN toptier_agency TAA ON (AA.toptier_agency_id = TAA.toptier_agency_id)
 LEFT JOIN subtier_agency SAA ON (AA.subtier_agency_id = SAA.subtier_agency_id)
 LEFT JOIN toptier_agency TFA ON (FA.toptier_agency_id = TFA.toptier_agency_id)
 LEFT JOIN subtier_agency SFA ON (FA.subtier_agency_id = SFA.subtier_agency_id)
+-- The next four joins are needed to get the Agency ID for all combinations of Awarding / Funding
```
tony-sappe: Not tracking why this is necessary

usaspending_api/download/helpers/elasticsearch_download_functions.py
```diff
@@ -52,7 +52,7 @@ def _get_download_ids_generator(cls, search: Union[AwardSearch, TransactionSearc
             yield results
 
     @classmethod
-    def _get_download_ids(cls, filters: dict, size: int = 100000) -> QuerySet:
+    def _get_download_ids(cls, filters: dict, size: int = 10000) -> QuerySet:
```
tony-sappe: Smaller chunks for downloads? Was an issue encountered or perhaps this is a performance improvement?

usaspending_api/search/v2/urls.py
```diff
@@ -4,13 +4,17 @@
...
+from usaspending_api.search.v2.views.spending_by_category_views.awarding_agency import AwardingAgencyViewSet
+from usaspending_api.search.v2.views.spending_by_category_views.awarding_subagency import AwardingSubagencyViewSet
 from usaspending_api.search.v2.views.spending_by_geography import SpendingByGeographyVisualizationViewSet
 from usaspending_api.search.v2.views.spending_over_time import SpendingOverTimeVisualizationViewSet
 
 urlpatterns = [
     url(r"^new_awards_over_time", NewAwardsOverTimeVisualizationViewSet.as_view()),
     url(r"^spending_by_award_count", SpendingByAwardCountVisualizationViewSet.as_view()),
     url(r"^spending_by_award", SpendingByAwardVisualizationViewSet.as_view()),
+    url(r"^spending_by_category/awarding_agency", AwardingAgencyViewSet.as_view()),
```
tony-sappe: Should  the new spending_by_category endpoints be in their own url file? There will be over a dozen by the time all of the endpoints are created

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,153 @@
...
+from usaspending_api.common.elasticsearch.search_wrappers import TransactionSearch
+from usaspending_api.common.experimental_api_flags import is_experimental_elasticsearch_api
+from usaspending_api.common.helpers.generic_helper import get_simple_pagination_metadata, get_time_period_message
+from usaspending_api.common.query_with_filters import QueryWithFilters
+from usaspending_api.common.validator.award_filter import AWARD_FILTER
+from usaspending_api.common.validator.pagination import PAGINATION
+from usaspending_api.common.validator.tinyshield import TinyShield
+from usaspending_api.search.v2.elasticsearch_helper import get_number_of_unique_terms
+
+logger = logging.getLogger("console")
```
tony-sappe:
```suggestion
logger = logging.getLogger(__name__)
```

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,153 @@
...
+        self.elasticsearch = validated_payload.get("elasticsearch")
+        self.subawards = validated_payload["subawards"]
+        self.pagination = {
+            "page": validated_payload["page"],
+            "limit": validated_payload["limit"],
+            "lower_limit": (validated_payload["page"] - 1) * validated_payload["limit"],
+            "upper_limit": validated_payload["page"] * validated_payload["limit"] + 1,
+        }
+
+        if self.elasticsearch and not self.subawards:
```
tony-sappe: Could this conditional be simplified?

if self.subawards:
....
elif self.elasticsearch:
....
else:
....

usaspending_api/download/helpers/elasticsearch_download_functions.py
sethstoudenmier: Intended this to be 10,000 but I added an additional 0 apparently and so this is fixing the mistake. It is also a performance boost.

usaspending_api/database_scripts/etl/transaction_delta_view.sql
sethstoudenmier: When hitting the `awarding_agency` endpoint the `awarding_agency_id` is the ID from Agency table, but it is not the correct ID for the response since the `toptier_flag` might be False for that particular Agency. Needed to take the `toptier_flag` into account when pulling the ID that matches the Agency record. For example, that are dozens of records with the `toptier_agency_id = 68` in the Agency table, but only one of them has the `toptier_flag = True`.

tony-sappe: Oh right, that's a new concern. Let me shoot you something outside Github to brainstorm something

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,153 @@
...
+
+        self.filters = validated_payload.get("filters", {})
+        self.elasticsearch = validated_payload.get("elasticsearch")
+        self.subawards = validated_payload["subawards"]
+        self.pagination = {
+            "page": validated_payload["page"],
+            "limit": validated_payload["limit"],
+            "lower_limit": (validated_payload["page"] - 1) * validated_payload["limit"],
+            "upper_limit": validated_payload["page"] * validated_payload["limit"] + 1,
+        }
```
BrianZito: I would compress this into a helper function within the class, which would leave this function with just two big chucks of code to parse.

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,153 @@
...
+        models.extend(copy.deepcopy(AWARD_FILTER))
+        models.extend(copy.deepcopy(PAGINATION))
+
+        validated_payload = TinyShield(models).block(request.data)
+
+        validated_payload["elasticsearch"] = is_experimental_elasticsearch_api(request)
+
+        return Response(self.handle_business_logic(validated_payload))
+
+    def handle_business_logic(self, validated_payload: dict) -> dict:
```
BrianZito:
```suggestion
    def perform_search(self, validated_payload: dict) -> dict:
```
happy to back off on this one since "handle_business_logic" is the norm among our endpoints, but I think there's an opportunity for self-documenting code here.

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,153 @@
...
+    @cache_response()
+    def post(self, request: Request) -> Response:
+        models = [
+            {"name": "subawards", "key": "subawards", "type": "boolean", "default": False, "optional": True},
+        ]
+        models.extend(copy.deepcopy(AWARD_FILTER))
+        models.extend(copy.deepcopy(PAGINATION))
+
+        validated_payload = TinyShield(models).block(request.data)
+
```
BrianZito:
```suggestion
```
related logic

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,153 @@
...
+    def handle_elasticsearch_logic(self, search: TransactionSearch) -> list:
+        results = self.get_elasticsearch_results_generator(search)
+        flat_results = list(itertools.chain.from_iterable(results))
+        return flat_results
+
+    @abstractmethod
+    def apply_elasticsearch_aggregations(
+        self, search: TransactionSearch, curr_partition: int, num_partitions: int, size: int
+    ) -> None:
```
BrianZito:
```suggestion
    ) -> TransactionSearch:
```
Side effects can be dangerous, particularly when working with something quirky like Elasticsearch where a non referentially-transparent function like this could be doing things a developer wouldn't expect. Having this work as a builder might make it clearer that only the TransactionSearch is being modified.

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,153 @@
...
+class BaseSpendingByCategoryViewSet(APIView, metaclass=ABCMeta):
+    """
+    Base class inherited by the different category endpoints.
+    """
+
+    category: Category
+    elasticsearch: bool
+    filters: dict
+    obligation_column: int
+    pagination: dict
```
BrianZito: Since this is used by implementers, it might be wise to make this a dataclass

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
BrianZito: I'm not sure if subawards should be in this design at all, since that's very use case specific. If this chunk of logic is going to be different for each kind of category, this is probably a place to have another abstract method for the implementers.

tony-sappe: You don't expect the ability to surface sub-award data for each/most categories?

The abstract methods are called here depending on the request JSON and headers (subawards vs prime awards, elasticsearch vs postgres), I'm not tracking the benefit for making this re-defined in each child class.

BrianZito: Right, I forget that subawards is a whole tab sometimes. Given the logical seperation here, I would argue in favor of keeping two levels and simplifying the first conditional:
```
if self.elasticsearch:
...
else:
    if self.subawards:
    ...
    else:
```
to isolate the es logic more easily.

tony-sappe: Funny, I thought my suggestion was following the logical separation better. Essentially, sub-awards will always be separate from prime-awards no matter where the data are stored. The Postgres/Elasticsearch separated business logic is temporary and is still searching and surfacing the same data.

If possible, it is always a good idea to "flatten" the code. It wasn't obvious what the nested conditionals added

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def common_db_query(self, queryset: QuerySet, django_filters: dict, django_values: list) -> QuerySet:
+        return (
+            queryset.filter(**django_filters)
+            .values(*django_values)
+            .annotate(amount=Sum(self.obligation_column))
+            .order_by("-amount")
+        )
+
+    def get_elasticsearch_results_generator(self, search: TransactionSearch, size: int = 1000) -> list:
```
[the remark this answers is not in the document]
datbossg310: It's a great idea I believe I was trying to I believe I was trying to do a pull request

datbossg310:
It's great I believe I was trying to pull do a pull request did I do
something wrong

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -0,0 +1,165 @@
...
+        return Response(self.perform_search(validated_payload))
+
+    def perform_search(self, validated_payload: dict) -> dict:
+
+        self.filters = validated_payload.get("filters", {})
+        self.elasticsearch = validated_payload.get("elasticsearch")
+        self.subawards = validated_payload["subawards"]
+        self.pagination = self._get_pagination(validated_payload)
+
+        if self.subawards:
```
BrianZito: So if it's subawards AND elasticsearch, it still queries django?

usaspending_api/search/v2/views/spending_by_category_views/awarding_agency.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+        # Have to create a separate dictionary for the bucket_sort values since "from" is a reserved word
+        bucket_sort_values = {
+            "sort": {"sum_as_dollars": {"order": "desc"}},
+            "from": (self.pagination.page - 1) * self.pagination.limit,
+            "size": self.pagination.limit + 1,
+        }
+        sum_bucket_sort = A("bucket_sort", **bucket_sort_values)
+
```
BrianZito: Is this going to change much between implementations? While we have two helper classes, this might be a good opportunity for code reuse.

usaspending_api/search/v2/views/spending_by_category_views/awarding_agency.py
```diff
@@ -0,0 +1,89 @@
...
+            field="awarding_toptier_agency_name.keyword",
+            include={"partition": curr_partition, "num_partitions": num_partitions},
+            size=size,
+        )
+        group_by_agency_abbreviation = A("terms", field="awarding_toptier_agency_abbreviation.keyword")
+        group_by_agency_id = A("terms", field="awarding_toptier_agency_id")
+        sum_as_cents = A("sum", field="generated_pragmatic_obligation", script={"source": "_value * 100"})
+        sum_as_dollars = A(
+            "bucket_script", buckets_path={"sum_as_cents": "sum_as_cents"}, script="params.sum_as_cents / 100"
+        )
```
BrianZito: A lot of this logic is duplicated with awarding subagency. If these two are the only two implementations to use this functionality, I would suggest another abstract class capturing all the things they have in common. If this is more of a quirk and you would expect future implementations to use some but not all of these, I would advise we move these functions to a library. If EVERY implementation will be using all these, we might want to make the base class function not abstract, and have the implementations do the equivilent of a super() call.

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
sethstoudenmier: Correct. Currently, we do not use Elasticsearch for Subawards.

usaspending_api/search/v2/elasticsearch_helper.py
```diff
@@ -225,3 +229,29 @@ def concat_if_array(data):
...
+def get_number_of_unique_terms(filter_query: ES_Q, field: str) -> int:
+    search = TransactionSearch().filter(filter_query)
+    cardinality_aggregation = A("cardinality", field=field)
+    search.aggs.metric("field_count", cardinality_aggregation)
+    response = search.handle_execute()
+    response_dict = response.aggs.to_dict()
+    return response_dict.get("field_count", {"value": 0})["value"]
+
+
+def get_sum_aggregations(field_to_sum: str, pagination: Pagination) -> Dict[str, A]:
```
BrianZito: Could this title be more specific to its reusable use case? Not a rhetorical question, I honestly don't know.

usaspending_api/api_contracts/contracts/v2/search/spending_by_category/awarding_agency.md
```diff
@@ -0,0 +1,166 @@
...
+        
+            {
+                "filters": { 
+                    "keywords": ["Filter is required"] 
+                }
+            }
+
++ Response 200 (application/json)
+    + Attributes (object)
+        + `category`: `awarding_agency` (required, string)
```
tony-sappe: Typically the Attributes don't have example values and we include an example JSON body

tony-sappe: Followup-thought:

This value should always be `awarding_agency ` which means it is unnecessary.
Although, I assume it is present for interoperability between the original SpendingByCategory response.

Open to discussion if it is necessary

usaspending_api/search/v2/elasticsearch_helper.py
sethstoudenmier: Also couldn't think of a better name. Ideally this could be used to generate the sum aggregations for any Elasticsearch query that is using the DSL.

usaspending_api/api_contracts/contracts/v2/search/spending_by_category/awarding_agency.md
sethstoudenmier: Your assumption is correct. It doesn't need to be there, but not having there would make it differ from the current SpendingByCategory response. Removing it means adding logic that will eventually be redundant when we move away from a single SpendingByCategory response.

usaspending_api/search/v2/views/spending_by_category_views/awarding_agency.py
datbossg310: Ok

datbossg310:
I am  yes up every time I I think I believe is merged already everytime I do hit the link I believe I'm getting an error message a404

PR #2259, opened 2020-01-31
[DEV-4155] Upgrade ES to 7.1

emily-brents:
**Description:**
Code changes to support moving to Elasticsearch 7.1

**Technical details:**
upgrades elasticsearch-py and elastichsearch-dsl packages, changes code to support removal of `doc_types`, makes adjustments to any code that calls `response["hits"]["total"]`

**Depends On:**
- https://github.com/fedspendingtransparency/usaspending-config/pull/611

**Requirements for PR merge:**

1. [X] Unit & integration tests updated
2. [X] API documentation updated
3. [ ] Necessary PR reviewers:
    - [x] Backend
    - [ ] Operations 
    - [ ] Domain Expert 
4. [N/A] Matview impact assessment completed
5. [N/A] Frontend impact assessment completed
6. [X] Data validation completed
7. [X] Appropriate Operations ticket(s) created - https://federal-spending-transparency.atlassian.net/browse/OPS-1012
8. [ ] Jira Ticket [DEV-4155](https://federal-spending-transparency.atlassian.net/browse/DEV-4155):
    - [X] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #2273, opened 2020-02-07
[DEV-4460] Remove nested data types elasticsearch

emily-brents:
**Description:**
Removes all 1nested` data types for both the award and transaction indices in elasticsearch.

**Technical details:**
No changes to the actual delta views, but we change the way the data is loaded into ES so that they are stored as arrays of strings instead of nested fields

    - [x] Backend
7. [X] Appropriate Operations ticket(s) created - https://federal-spending-transparency.atlassian.net/browse/OPS-1092
8. [x] Jira Ticket [DEV-4460](https://federal-spending-transparency.atlassian.net/browse/DEV-4460):

```
```


usaspending_api/database_scripts/etl/transaction_delta_view.sql
```diff
@@ -137,8 +137,7 @@ LEFT JOIN (
        )
      ) treasury_accounts
  FROM
-   federal_account fa
```
tony-sappe: Was this table helping to reduce duplicate/invalid values?

usaspending_api/database_scripts/etl/transaction_delta_view.sql
```diff
@@ -163,4 +162,4 @@ LEFT JOIN (
     faba.award_id IS NOT NULL
   GROUP BY
     faba.award_id
-) FEDERAL_ACCT ON (FEDERAL_ACCT.award_id = UTM.award_id);
+) FEDERAL_ACCT ON (FEDERAL_ACCT.award_id = UTM.award_id);
```
tony-sappe: Missing trailing newline

usaspending_api/etl/es_award_template.json
```diff
@@ -408,30 +408,7 @@
...
-              "type": "integer"
-            },
-            "epoa": {
-              "type": "integer"
-            },
-            "a": {
-              "type": "text"
-            }
-          }
+          "type": "keyword"
```
tony-sappe: I'm surprised we can run regex on a keyword field based on [this help doc](https://www.elastic.co/guide/en/elasticsearch/reference/7.1/keyword.html)

image

sethstoudenmier: If my understanding is correct that means that the `regexp` query will need to provide a regex that matches the entire string representation of the treasury_account instead of only a partial match. Struggling to find documentation that mentions exactly that.

usaspending_api/database_scripts/etl/transaction_delta_view.sql
sethstoudenmier: Speaking to only the transaction_delta_view, I believe that was a result of a bad copy and paste on my part between Federal Account and TAS.

usaspending_api/etl/es_transaction_template.json
```diff
@@ -395,24 +395,10 @@
...
+          "type": "keyword"
         },
         "federal_accounts": {
-          "type": "nested",
-          "properties": {
-            "id": {"type": "integer"},
-            "account_title": {"type": "keyword"},
-            "federal_account_code": {"type": "keyword"}
-          }
+          "type": "keyword"
```
kwhickey: Are each/either of `treasury_accounts` or `federal_accounts` expected to have their text value of their codes searchable by  the Keyword Search string?

If so, a `keyword` data type is not going to allow text analysis that would allow these to be mapped. In that case we'd need to add a 2nd field that has the compact/standard TAS / FAS symbol format stored and make that `text` data type, and make that one of the fields targeted by Keyword Search.

usaspending_api/etl/es_award_template.json
kwhickey: I would think this falls in the same boat as a wildcard query (e.g. `"wildcard": "uni*"`).
... Taking the `exact value` to mean the value provided in the source doc, and not any of the analyzed derivatives made. 
i.e. When not `keyword`, and using `text` with a "stemming" analyzer combined with "stop words" analyzer would take `"plays on his collegiate football team"` and then perform text searching on any of the terms `["play", "college", "football", "team"]`. 

Whereas use of `wildcard` or `regexp` on a `keyword` data type would be running matches on the exact value of the text as provided: `"plays on his collegiate football team"`

So I think it'll work. But it'd be good to have tests to verify.

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -245,13 +247,24 @@ def convert_postgres_array_as_string_to_list(array_as_string: str) -> list:
...
-        In the case of an empty array, return null.
+        Postgres JSON arrays (jsonb) are stored in CSVs as strings. Since we want to avoid nested types
+        in Elasticsearch the JSON arrays are converted to dictionaries to make parsing easier and then
+        converted back into a formatted string.
     """
-    return json.loads(json_array_as_string) if json_array_as_string else None
+    if json_array_as_string is None or len(json_array_as_string) == 0:
+        return None
+    result = []
+    json_array = json.loads(json_array_as_string)
```
kwhickey: Is this guaranteed to create a list preserving order? Such that the components will always be pulled from the DB and pushed to ES always in the same order to satisfy  the search?

usaspending_api/etl/es_award_template.json
```diff
@@ -416,30 +416,7 @@
           "type": "text"
         },
         "treasury_accounts": {
-          "type": "nested",
```
kwhickey: One of the major benefits we were trying to get out of removing nested fields was allowing the sort to be put back in place, and the sort to default to the same sort that @emily-brents was putting into her default Advanced Search queries. 

@emily-brents took the sort out of the mapping [in her PR because the nested fields prohibited it](https://github.com/fedspendingtransparency/usaspending-api/pull/2259/files#r374893886). Here its left as-was which is:
```
  "sort": {
    "fields": [
      "total_obligation",
      "action_date"
    ],
    "order": [
      "desc",
      "desc"
    ]
  },
```
That's not the ideal sort because her PR uses `total_obligation` and `award_id`. 

We need to coordinate who will be getting the sort in, and using the right fields.

usaspending_api/etl/es_transaction_template.json
```diff
@@ -395,24 +395,10 @@
           }
         },
         "treasury_accounts": {
-          "type": "nested",
```
kwhickey: Similar to the comment in the award mapping re: sorting. 
```
  "sort": {
    "fields": [
      "transaction_amount",
      "action_date"
    ],
    "order": [
      "desc",
      "desc"
    ]
  },
```

It's defaulted to this here, but we need to determine where it will get changed to use `award_id`, (or in this case, whatever the default sort should be for transaction data)

Looks like its set, but not sure if `action_date` is best.
https://github.com/fedspendingtransparency/usaspending-api/pull/2259/files#diff-08acb4d2c9ca3adffa8189ccbc80cb25R9

usaspending_api/etl/es_award_template.json
kwhickey: Looks like in her 7.1 PR it's being set, but not to the fields that she needs I don't think:
https://github.com/fedspendingtransparency/usaspending-api/pull/2259/files#diff-570f89ea9b600b880909d51919150cdbR9

(I see `action_date` and not `award_id`)

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -80,36 +80,36 @@ SELECT
...
-  LEFT OUTER JOIN (
+  FROM
+    recipient_lookup AS rlv
+) recipient_lookup ON (recipient_lookup.duns = vw_award_search.recipient_unique_id AND vw_award_search.recipient_unique_id IS NOT NULL)
+LEFT JOIN (
   SELECT
     faba.award_id,
     JSONB_AGG(
-      JSONB_BUILD_OBJECT(
+      DISTINCT JSONB_BUILD_OBJECT(
```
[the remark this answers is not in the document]
sethstoudenmier: If you pull it over as an array of strings from Postgres then you still have to do some special handling since the Postgres values are stored in a CSV as strings. Now for TAS that isn't too much of a problem since you only have numbers that will make up a concatenated string. But for Federal Accounts where the `account_title` could have a wide variety of different characters, you would run into some difficulties making sure that you pick a unique character to concat on with Postgres. By using an array of JSONs we can leverage json.loads() to handle the parsing for us.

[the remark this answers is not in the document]
datbossg310: it looks so perfect I like that so what I would do is fix it I will put a count title looking so perfect though and I will pick a unique Handler

usaspending_api/etl/es_transaction_template.json
datbossg310: Glad y'all are try fixing it making it perfect I mean y'all really are the best

datbossg310:
Thanks alot guys

datbossg310:
Now do I revert over or will that be yes I should be revert over right

datbossg310:
@emily-brents @kwhickey @tony-sappe    need info. for broker

lisaam:
@datbossg310  Please see my comment at https://github.com/fedspendingtransparency/usaspending-api/commit/5bdbfafbddd9877b1352ec09ac60a8b78e6e3643#commitcomment-37178282

PR #2288, opened 2020-02-11
reduced duplicate code for spending_by_category_tests

BrianZito:
Further cleans things up in preparation for making a lot more of these endpoints.


BrianZito:
Better improvements made on a different branch.

PR #2289, opened 2020-02-12
Modular fixture data

BrianZito:
**Description:**
Allows mock data to be generated dynamically per test, reducing effort needed to produce more specialized unit tests while improving readability of test errors by tying mock data to a logical structure by default. If needed for a special test, these patterns can still be broken without making other tests fragile.


usaspending_api/search/tests/integration/spending_by_category/test_spending_by_awarding_agency.py
```diff
@@ -79,11 +60,36 @@ def test_correct_response(client, monkeypatch, elasticsearch_transaction_index,
         "limit": 10,
         "page_metadata": {"page": 1, "next": None, "previous": None, "hasNext": False, "hasPrevious": False},
         "results": [
-            {"amount": 10.0, "name": "Awarding Toptier Agency 3", "code": "TA3", "id": 1005},
```
sethstoudenmier: The purpose of this test is to make sure that we are correctly grabbing the Toptier Agency ID by using the `is_toptier` flag. If the new use of fixtures resulted in this change then it is no longer fulfilling its original intent. The fixture should be updated to match the original relations so that this test has the same purpose as before.

usaspending_api/search/tests/integration/spending_by_category/test_spending_by_funding_agency.py
```diff
@@ -79,8 +60,8 @@ def test_correct_response(client, monkeypatch, elasticsearch_transaction_index,
         "limit": 10,
         "page_metadata": {"page": 1, "next": None, "previous": None, "hasNext": False, "hasPrevious": False},
         "results": [
-            {"amount": 10.0, "name": "Funding Toptier Agency 4", "code": "TA4", "id": 1004},
```
sethstoudenmier: Similar as the above comment. Need to double check the fixture changes, but the difference in the TA2 agency having an ID of 1006 meant that it successfully grabbed the ID of a different agency that had the toptier flag.

usaspending_api/search/tests/integration/spending_by_category/spending_test_fixtures.py
```diff
@@ -0,0 +1,71 @@
...
+        id=2,
+        award_id=2,
+        awarding_agency_id=1003,
+        funding_agency_id=1006,
+        federal_action_obligation=10,
+        action_date="2020-01-02",
+    )
+
+
+def _setup_agency(id, subteirs, special_name):
```
sethstoudenmier: subtiers not subteirs

usaspending_api/search/tests/integration/spending_by_category/test_spending_by_awarding_agency.py
BrianZito: The ID changed because the agencies now come from a helper function that picks ids in a more standard way, but agency 1003 (defined in line 24 of the fixture class) is the same structure as the old agency 1005 (defined in line 23 of the old fixture class). The new helper method just standardizes that the `agency` table entry for the toptier ends with the same number as the `toptier_agency` table entry, whereas the old code set it up so that the agency row for the toptier always ended with 5, while subtiers ended with the same number as the agency row.

usaspending_api/search/tests/integration/spending_by_category/test_spending_by_funding_agency.py
BrianZito: The flipping of order was a quirk (I assume of splitting the fixtures), but these similarly were defined in lines 7 and 26 of the fixtures, and have slightly different IDs to better avoid collisions. Unless there's an error in the helper method of that file, anything with "Toptier" in it should ALWAYS be toptier, since nothing is named by hand anymore.

usaspending_api/search/tests/integration/spending_by_category/test_spending_by_awarding_agency.py
sethstoudenmier: The original structure had transaction_normalzied record with an awarding_agency_id of 1003 (which is the same as the new structure). However, the difference is that in the old the `is_toptier` flag is set to False and in the new it is set to True. That distinction is important since it checks that the code is looking up the correct awarding_toptier_agency_id of 1005 (where the `is_toptier` flag is set to True). The new structure doesn't perform that check since 1003 is set to True and 1005 is set to False, meaning no need to perform that check.

BrianZito: Doesn't that still make the code perform the check, just with 1003 returning true and 1005 returning false?

PR #2290, opened 2020-02-12
DEV-4208 Fork existing requests to load test Elasticsearch backend

kwhickey:
**Description:**
Forking a mirrored request that is sent back to the API with the Experimental Elasticsearch header on it to leverage existing production traffic for load testing ES

**Technical details:**
Pick up the request at the point where we check for the ES header, and when _not_ there, spin up a new request using the `requests` library, with the existing as a template, and add this header. Eagerly timeout on that request and ignore its response/errors.

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-4208](https://federal-spending-transparency.atlassian.net/browse/DEV-4208):

```
Should have no impact on front, no impact on matviews. No ops tickets necessary
```


usaspending_api/common/experimental_api_flags.py
```diff
@@ -17,4 +25,46 @@ def is_experimental_elasticsearch_api(request: Request) -> bool:
     """
     Returns True or False depending on if the expected_header_value matches what is sent with the request.
     """
-    return request.META.get(EXPERIMENTAL_API_HEADER) == ELASTICSEARCH_HEADER_VALUE
+    use_es = request.META.get(EXPERIMENTAL_API_HEADER) == ELASTICSEARCH_HEADER_VALUE
+    if not use_es:
+        mirror_request_to_elasticsearch(request)
```
BrianZito: This is an `is` function, which implies it can be called as many times as desired without side-effects. There might be implementations where code checks for being elasticsearch in two places, making two forked requests.

usaspending_api/common/experimental_api_flags.py
```diff
@@ -17,4 +25,46 @@ def is_experimental_elasticsearch_api(request: Request) -> bool:
...
+    data = json.dumps(request.data)
+    headers = {
+        **request.headers,
+        HttpHeaders.parse_header_name(EXPERIMENTAL_API_HEADER): ELASTICSEARCH_HEADER_VALUE,
+    }
+
+    logger.warning(
+        f"Mirroring inbound request with elasticsearch experimental header.\n"
+        f"\tOriginial Request details: {request._request}\n"
+        f"\tMirrored Request details: \n\t\turl = {url}, \n\t\tdata = {data}, \n\t\theaders = {headers}"
```
BrianZito: do any of our requests have large response bodies? We don't have infinite log space.

usaspending_api/common/experimental_api_flags.py
```diff
@@ -17,4 +25,46 @@ def is_experimental_elasticsearch_api(request: Request) -> bool:
...
+    )
+
+    # NOTE: Purposely desiring an immediate timeout, with ignoring of that timeout error,
+    # since this is a fire-and-forget way to siphon off duplicate load-testing traffic to the server,
+    # without disrupting the primary request
+    try:
+        if request.method == "GET":
+            requests.get(url, data, headers=headers, timeout=0.01)
+        else:
```
BrianZito: What about "OPTIONS" requests? Do we do an "UPDATE" or "INSERT" anywhere? I doubt it, but it might be safer to ignore anything other than GET and POST in order to avoid giving a bunch of 405s everywhere.

usaspending_api/common/experimental_api_flags.py
kwhickey: Oh, excellent catch @BrianZito. I added that only so I personally could see the data when developing with this. That verbose logging should definitely be gotten rid of.

I think that whole log statement should go. Presence of the experimental header should be enough for use to delineate what requests were generated by the app vs. those that were natural.

usaspending_api/common/experimental_api_flags.py
kwhickey: I concur. Safer to explicitly check `POST` with `elif` and then just do nothing in all other cases.

usaspending_api/common/experimental_api_flags.py
kwhickey: Fair point. Can you find any?

BrianZito: I can't see into the future

usaspending_api/common/experimental_api_flags.py
BrianZito: Follow question then: with no logging statements, what does the test assert?

emily-brents: I'd rather keep a shorted version of the log statement, just to be able to differentiate between what is being duplicated by the site and what is dev and/or testers checking something through postman etc..

kwhickey:
@BrianZito @emily-brents , I realized a bit too late, there was a Tech Task (which should have led to an AC item) in the Jira issue: https://federal-spending-transparency.atlassian.net/browse/DEV-4208

Around making sure that this "test-traffic" does not have its response data PUT into the Redis cache. Reason: this path of code is still "under-test" and may provide inaccurate response data that we don't want surfacing from the cache when people navigate the postgres-backed path of that endpoint. 

Was that accounted for in this impl? And if not, can you please create a high-priority follow-on improvement to warm-fix it?

emily-brents:
No, that was not specifically addressed with this implementation, here is the follow up ticket: https://federal-spending-transparency.atlassian.net/browse/DEV-4693

PR #2291, opened 2020-02-14
[OPS-1096] HOTFIX Added chunking to submission loader

kbard:
**DO NOT MERGE:** Still testing.

**Description:**

Our largest File C submission ever has managed to break the nightly pipeline.  This pull requests chunks the submission to reduce memory footprint.

1. [x] Unit tests unaffected
    - [ ] Backend
5. [x] Front end unaffected
7. [x] Operations Ticket [OPS-1096](https://federal-spending-transparency.atlassian.net/browse/OPS-1096)

### DIFF
--- a/usaspending_api/etl/management/helpers/load_submission.py
+++ b/usaspending_api/etl/management/helpers/load_submission.py
@@ -1,0 +1,166 @@
+import numpy as np
+import pandas as pd
+
+from collections import deque
+from django.db import connections
+from django.utils.functional import cached_property
+from usaspending_api.references.models import ObjectClass, RefProgramActivity
+
+
+class Bunch:
+    """Generic class to hold a group of attributes."""
+
+    def __init__(self, **kwds):
+        self.__dict__.update(kwds)
+
+
+def get_or_create_program_activity(row, submission_attributes):
+    # We do it this way rather than .get_or_create because we do not want to duplicate existing pk's with null values
+    filters = {
+        "program_activity_code": row["program_activity_code"],
+        "program_activity_name": row["program_activity_name"].upper()
+        if row["program_activity_name"]
+        else row["program_activity_name"],
+        "budget_year": submission_attributes.reporting_fiscal_year,
+        "responsible_agency_id": row["agency_identifier"],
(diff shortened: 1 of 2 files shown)


usaspending_api/etl/management/helpers/load_submission.py
```diff
@@ -0,0 +1,166 @@
...
+        self.__dict__.update(kwds)
+
+
+def get_or_create_program_activity(row, submission_attributes):
+    # We do it this way rather than .get_or_create because we do not want to duplicate existing pk's with null values
+    filters = {
+        "program_activity_code": row["program_activity_code"],
+        "program_activity_name": row["program_activity_name"].upper()
+        if row["program_activity_name"]
+        else row["program_activity_name"],
```
BrianZito: I assume this is just to not try and uppercase an empty string? Just checking cause it looks like there might be a mistake here...

kbard: Literally a straight copy.  None of these functions were touched.  In this file, the only things I'm responsible for are `CertifiedAwardFinancial` and `CertifiedAwardFinancialIterator`.

If you think it's a bug, let's make a note and loop back around after the dust settles... unless we're 100% sure it's a bug, it's super trivial to fix, and doesn't require any backfilling.

kbard: Just as a side note, that code looks to me like it's trying to avoid uppercasing nulls (Nones).

usaspending_api/etl/management/helpers/load_submission.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __init__(self, submission_attributes):
+        self.submission_attributes = submission_attributes
+
+    @cached_property
+    def count(self):
+        sql = f"""
+            select  count(*)
+            from    certified_award_financial
+            where   submission_id = {self.submission_attributes.broker_submission_id} and
```
sethstoudenmier: It doesn't really matter but could define the where clause as a string and pass it to the iterator. Seems they are both the exact same except for the conditional `_last_id` check.

PR #2294, opened 2020-02-16
[FIX] Fixing ETL issues

tony-sappe:
## Issue 1
**Elasticsearch ETL for awards is broken**
Highly confident this has the fix, but unable to test locally due to odd error on missing index name......
Please review and test **before** merging!!!

## Issue 2
**New Assistance transaction ETL didn't do very well without inactive FABS**


tony-sappe:
Tested the scripts against production

PR #2301, opened 2020-02-19
Filter recipient_id from spending_by_award and add unused filter messages

BrianZito:
**Description:**
This PR does two things:
1) Add a message to inform users when filters are being ignored
2) Ignore `recipient_id` filter for spending_by_award

**Technical details:**
Actual message text is made up by developer, and likely just a placeholder.

    - [x] Backend
    - [x] Frontend 
8. [x] Jira Ticket [DEV-4272](https://federal-spending-transparency.atlassian.net/browse/DEV-4272):

```
```


usaspending_api/download/v2/download_transaction_count.py
```diff
@@ -47,7 +48,11 @@ def post(self, request):
             "calculated_transaction_count": total_count,
             "maximum_transaction_limit": settings.MAX_DOWNLOAD_LIMIT,
             "transaction_rows_gt_limit": total_count > settings.MAX_DOWNLOAD_LIMIT,
-            "messages": [get_time_period_message()],
+            "messages": [
+                get_generic_filters_message(
+                    set(list(self.original_filters.keys())), [elem["name"] for elem in AWARD_FILTER]
```
BrianZito: This got tagged by Code Climate as being duplicate code with the other places it's used. I'm reluctant to put this in a helper function for helper's sake, since this line's logic is very dependent on how the data is stored in this class. The helper would be tightly coupled against 4 view classes, which IMO is a worse issue than just having some boiler plate. Any thoughts?

usaspending_api/search/v2/views/spending_by_category.py
```diff
@@ -100,9 +100,11 @@ def post(self, request: Request) -> Response:
         models.extend(copy.deepcopy(PAGINATION))
 
         # Apply/enforce POST body schema and data validation in request
+        original_filters = request.data.get("filters")
         validated_payload = TinyShield(models).block(request.data)
 
         validated_payload["elasticsearch"] = is_experimental_elasticsearch_api(request)
+        validated_payload["original_filters"] = original_filters
```
tony-sappe: Why is the original JSON being stored in the validated payload?
Doesn't that undermine the purpose of "validated request JSON"

BrianZito: It was a tough call for me, but I thought that linking the history onto the payload object was a little cleaner than two tightly coupled parameters that are derived from each other (not to mention that would introduce conflicts with everyone's ES tickets). The validated payload is still guarenteed to be "valid", in that it no longer has filters that will cause issues; the original_filters variable is more of a tag or description. It is no longer a reflection of the original JSON now that there's a new key, but I feel like this is comparable to something like adding a timestamp to the request.

usaspending_api/download/v2/download_transaction_count.py
tony-sappe: > since this line's logic is very dependent on how the data is stored in this class.

This leads to the question: why? Why couldn't the attributes be standardized in the classes? We could re-implement the parent class `APIView` to maintain the attributes. `get_generic_filters_message()` could then be run for all API responses and contingent on booleans in the child class (or some other flag/toggle/list)

BrianZito: There is a lot of functionality that can be DRYed up in this classes. I'm concerned about going on TWO unrequested refactoring rampages this sprint though...

usaspending_api/search/v2/views/spending_by_category.py
tony-sappe: Part of  "validating" the request data is to provide some protection/control of the data in the business logic. This would negate some of that. However, the conversation on the other comment might make this design decision moot.

usaspending_api/download/v2/download_transaction_count.py
tony-sappe: You're the one who started the conversation 😄 I'm simply providing thoughts.

BrianZito: Looks like this convo got a little sidetracked. Do you feel like this is a situation where it's safe to overrule CodeClimate for this PR, or should I be working on refactoring this?

tony-sappe: > Looks like this convo got a little sidetracked
How so? There is room for cleaning this up so CC is happy

> Do you feel like this is a situation where it's safe to overrule CodeClimate for this PR, or should I be working on refactoring this?
The repeated set(list(...)) and list comprehension isn't a nice line of code to duplicate. Is there an easy way to simplify the function call code? I'd be willing to override CC if the code isn't as messy

usaspending_api/search/v2/views/spending_by_category.py
tony-sappe: Appears like `validated_payload["original_filters"] = original_filters` is no longer necessary here.

It should also be removed from BaseSpendingByCategoryViewSet and replaced by making original_filters a class attribute

PR #2304, opened 2020-02-20
seperated awards from ref data

BrianZito:
**Description:**
Since we all have to add different aspects to these transactions and awards, splitting that off from the ref data will avoid a lot of conflicts


BrianZito:
In your example, I was hoping the solution would be to make a new fixture building off the first, starting with the basic award and adding more transactions as that test needed. Would there be an issue with doing that?

PR #2307, opened 2020-02-21
[DEV-1671] Modify existing transaction ETL to use new source tables

tony-sappe:
**Description:**
Instead of fetching the same data from Broker twice, leverage the new source_*_transaction tables for the existing transaction ETLs populating `transaction_fabs`, `transaction_fpds`, `transaction_normalized`, and `awards`.

**Technical details:**

- The deleted transaction IDs in S3 have always been a pain point since they were implemented. The code has been refactored and the FPDS and FABS code has been unified. Still WIP for writing  to S3.
- Needed to remove the FABS submission parameter and query from FABS ETL since that table does not exist in `data_store_api`. It shouldn't cause any issues since that is not used during routine operations and the initial correction to no longer overlook records was implemented in the new `transfer_assistance_records` management command

    - [x] Backend
    - [x] Operations
    - [x] Domain Expert
8. [x] Jira Ticket [DEV-1671](https://federal-spending-transparency.atlassian.net/browse/DEV-1671):

```
No changes to API or materialized views
```


usaspending_api/common/helpers/s3_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+    s3 = boto3.resource("s3", region_name=region_name)
+    return s3.Bucket(bucket_name)
+
+
+def access_s3_object(bucket_name: str, obj: "boto3.resources.factory.s3.ObjectSummary") -> io.BytesIO:
+    """Return the Bytes of an S3 object"""
+    bucket = get_s3_bucket(bucket_name=bucket_name)
+    data = io.BytesIO()
+    bucket.download_fileobj(obj.key, data)
+    data.seek(0)  # Like rewinding a VCR cassette
```
BrianZito: If I understand this functionality correctly, this wouldn't be thread safe. Is that a concern?

tony-sappe: Perhaps you can elaborate more on "this" (entire function? `data.seek(0)`?)

Are any of the current management commands which would use this multi-threaded?

BrianZito: specifically the use of `data.seek(0)`, if this means the S3 object has some kind of global state that needs to be reset, makes `access_s3_object` not thread safe. I don't know if this is a problem in practice, but maybe might be an issue if multiple management commands are being run in Jenkins, and this is connecting to some state outside the app. The [base documentation](https://docs.python.org/3/library/io.html) of io seems mostly about file reading, so I wasn't able to find a quick answer. 
If this isn't a concern, and we know that io is just storing a pointer to the last line locally (which is what I would expect this kind of library to be doing), is there any need to rewind a cassette that we're just going to throw away?

tony-sappe: The s3 object data is being downloaded using `. download_fileobj()` into `data` (a byteIO object) and the seek on the python variable holding the data is immediately after, the code isn't manupulating anything in AWS nor is it expecting external state other than an existing s3 bucket name and object name. I don't see a risk of something occurring between those two lines even if called multiple times by a multithreaded script.

tony-sappe:
Local tests were run using a script outlined below. A diff was run on the CSVs ignoring the `transaction_id` field. No data discrepancies were found in either FABS or FPDS.
```
1. git checkout mod/dev-1671-modify-existing-etl
2. Delete transactions (updated_at >= '${start_date}')
3. python3 manage.py transfer_assistance_records --date="${start_date}"
4. python3 manage.py fabs_nightly_loader --start-datetime="${start_date}"
5. psql dump records to csv (new)
6. Delete transactions (updated_at >= '${start_date}')
7. git checkout dev
8. python3 manage.py fabs_nightly_loader --start-datetime="${start_date}"
9. psql dump records to csv (original)
```

usaspending_api/common/helpers/s3_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    logger.info("Gathering all deleted transactions from S3")
+    try:
+        bucket = get_s3_bucket(bucket_name=bucket_name)
+        bucket_objects = list(bucket.objects.all())
+    except Exception:
+        msg = (
+            "Verify settings.USASPENDING_AWS_REGION and settings.DELETED_TRANSACTION_JOURNAL_FILES are correct "
+            "or env variables: USASPENDING_AWS_REGION and DELETED_TRANSACTION_JOURNAL_FILES are set"
```
BrianZito: This message isn't very reusable. Would it be possible to use the bucket name to construct this string?

usaspending_api/transactions/management/commands/delete_procurement_records.py
```diff
@@ -20,51 +18,31 @@ class Command(AgnosticDeletes, BaseCommand):
...
-            data.seek(0)
-            reader = csv.reader(data.read().decode("utf-8").splitlines())
-            next(reader)  # skip the header
-
-            full_key_list = [rows[0] for rows in reader]
-            numeric_keys = list([int(rows) for rows in full_key_list if rows.isnumeric()])
-            odd_ids = set(full_key_list).symmetric_difference([str(id) for id in numeric_keys])
+        for date, transaction_ids in full_key_list.items():
+            numeric_keys = list([int(row) for row in transaction_ids if row.isnumeric()])
+            odd_ids = set(transaction_ids).symmetric_difference([str(id) for id in numeric_keys])
```
BrianZito: Is the objective of this just to get the non-numeric keys? If not, this code's purpose is unclear.

tony-sappe:
```python
    if odd_ids:
                logger.info(f"Unexpected non-numeric IDs in file: {list(odd_ids)}")
```

usaspending_api/common/helpers/s3_helpers.py
tony-sappe: good catch

usaspending_api/transactions/management/commands/delete_procurement_records.py
BrianZito: the couldn't odd_ids just be `list([int(row) for row in transaction_ids if not row.isnumeric()])`? That would be clearer IMO.

tony-sappe: I don't recall the original reasoning for the method. It is possible the original justification no longer holds. This PR is just altering the variable name in those lines.

usaspending_api/common/helpers/s3_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+        logger.exception(msg)
+        bucket_objects = None
```
kbard: Should this be re-raising the exception?  Currently it is swallowing it.  When I run locally it throws up all over my screen but keeps running as if nothing happened.

It sort of looks like it's intentional but I'm not sure I understand why.  If it can't access the S3 isn't that a problem?

usaspending_api/common/helpers/s3_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+import re
+
+from typing import Optional, List
+
+from usaspending_api import settings
+
+logger = logging.getLogger("script")
+
+
+def access_s3_object_list(
```
kbard: Are we doing away with the ability to be able to run loaders locally?  I could have sworn the old FPDS loader used to look in my `csv_downloads` directory for delete files using the `IS_LOCAL` setting.

usaspending_api/etl/transaction_loaders/fpds_loader.py
```diff
@@ -39,7 +37,7 @@ def delete_stale_fpds(date):
     Returns list of awards touched
     """
 
-    detached_award_procurement_ids = get_deleted_fpds_data_from_s3(date)
+    detached_award_procurement_ids = retrieve_deleted_fpds_transactions(start_datetime=date)
```
kbard: This is a problem.  It is returning a dictionary of `{date: [transaction_id, ..], ...}` which doesn't work with the remainder of the code in this function.

tony-sappe: darn. I must have lost those edits. Good find

usaspending_api/common/helpers/s3_helpers.py
tony-sappe: It is intentional. I'm open to a discussion about it

usaspending_api/common/helpers/s3_helpers.py
tony-sappe: Yes it did. 
1. The local file functionality was poorly designed and implemented. If has a benefit should  remain it needs to be refactored.
2. I am trying to make the FABS and FPDS loaders as shared as possible. If there is a benefit to having local files for deletes it should be added to FABS too.

usaspending_api/common/helpers/s3_helpers.py
tony-sappe: "...throws up all over my screen" == stack trace in log? If the stack trace is unwelcome it can easily be removed :-)

kbard: haha... that wasn't the point.  I was just concerned about continuing after not being able to access the S3.  If we can't access the S3 we can't pick up delete files.  If we can't pick up delete files then there's a pretty good chance some will be skipped on the next run since it's driven by date.

usaspending_api/common/helpers/s3_helpers.py
kbard: I just wanted to make sure it was a conscious decision.  It did make it a little easier to test but I agree that that was clunky code.  And FABS is a whole other thing since it writes.  That makes it extra scary to test against S3.

usaspending_api/common/helpers/s3_helpers.py
tony-sappe: I understood your concerns. So either:
A: Cannot run the loader without a working connection to an S3 bucket
B: Re-implement the local mode

tony-sappe: Added the snippet below to `usaspending_api/transactions/transaction_delete_journal_helpers.py` so the inability to connect to S3 stops the script
```python
    ...
        if objects is None:
            raise Exception("Problem accessing S3 for deleted records")
    ...
```

usaspending_api/etl/transaction_loaders/fpds_loader.py
tony-sappe: See Line 45 after recent commit

03-03-2020_delete_records_award_0000000000.csv
```diff
@@ -0,0 +1,2 @@
+sausage
+rootbeer
```
kbard: I'm guessing someone was a little hungry and/or thirsty when they created this file...

usaspending_api/common/helpers/s3_helpers.py
kbard: This will fail if a delete file is not created that night for whatever reason.  Looking through the S3 bucket, I see a couple gaps in the delete file dates, specifically 1/18/20 and 2/22/20.

tony-sappe: 1. The behavior should be different between an exception and no files returned (`None` vs empty list `[]`)
2. The current FABS only uploads a file when there are delete records while the new scripts will always create a file. It will be empty in the event no records were deleted in the timeframe which tracks with FPDS delete files

kbard: > The behavior should be different between an exception and no files returned

Yup.  You're right.  My bad.

03-03-2020_delete_records_award_0000000000.csv
tony-sappe: Needed non-numeric values. Who doesn't like food?

PR #2326, opened 2020-03-03
Ftr/dev 4264 tas filter tree

BrianZito:
**Description:**
Adds a series of new endpoints to support the TAS filter on website. 

**Technical details:**
Because the TAS filter is not fractal (i.e. the "parent" of a TAS is not itself a TAS), I decided to not follow the model of NAICS. Instead, these endpoints are named and supported by the concept of a "filter tree", with standardized nodes that draw data from different tables in different contexts. This allows one restful endpoint to support the frontend, rather than several endpoints, most of which wouldn't be retrieving "TAS" objects at all. This structure should support the logic of PSC as well, and should we ever have ample free time I suggest we refactor NAICS into the same structure. 

A "depth" parameter is provided allowing users to specify how many layers of children are populated. This was introduced based on early conversations with UI developers who thought this functionality would be needed. That doesn't look as likely now, but the functionality proved to not be that difficult to support so I kept it in. This also allows someone to view the entire TAS forrest at a glance (albeit a ~30 second glance) if desired.

The trickiest business logic is the association between federal accounts and agencies, which I have attempted to build out in a reusable and clear way [here](https://github.com/fedspendingtransparency/usaspending-api/compare/ftr/dev-4624-tas-filter-tree?expand=1#diff-1280147fe0074e832ed61172c7a6046a). Because of how murky our logic gets, however, this area could always use extra scrutiny.

    - [ ] Backend
    - [ ] Frontend 
8. [x] Jira Ticket [DEV-4275](https://federal-spending-transparency.atlassian.net/browse/DEV-4275):
    - [x] Performance evaluation of affected (API | Script | Download) (N/A)

```
```


BrianZito:
I've spent a little time addressing codeclimate's concerns, and I think I want talk it out with peers before implementing the recommended solutions. The drop in test coverage is because Code Climate can't track what lines are covered by integration tests. I started building unit tests to cover the same ground, but I've stopped that effort since they aren't testing anything new. 

The style errors are simple: Code Climate doesn't like early return statements, and presumably wants me to use flags instead. I happen to be a big fan of early return, so I would rather talk it out with you guys than simply change the code blindly.

usaspending_api/references/v2/views/filter_trees/filter_tree.py
```diff
@@ -0,0 +1,63 @@
...
+
+@dataclass
+class Node:
+    id: str
+    ancestors: list
+    description: str
+    count: int
+    children: list
+
+    def toJSON(self):
```
emily-brents: camel case? in python code?

usaspending_api/common/helpers/agency_logic_helpers.py
```diff
@@ -0,0 +1,39 @@
...
+    :return: string, which is the toptier_code for the correct agency to count this combination under
+    """
+    if (cgac, frec) in DOD_ARMED_FORCES_TAS_CGAC_FREC:
+        return DOD_CGAC
+
+    if cgac in DOD_SUBSUMED_CGAC:
+        return DOD_CGAC
+
+    if cgac in DHS_SUBSUMED_CGAC:
+        return DHS_CGAC
```
kbard: Is this how we're handling FEMA as a toptier now?  Previously, it was treated as a toptier but given lower visibility.  Is the goal to not show FEMA as a toptier in the top tier of the TAS control or is this a new site-wide thing?

usaspending_api/references/v2/views/filter_trees/filter_tree.py
kbard: seconded

usaspending_api/references/v2/views/filter_trees/filter_tree.py
```diff
@@ -0,0 +1,63 @@
+from abc import ABCMeta, abstractmethod
+from dataclasses import dataclass
+
+DEFAULT_CHILDREN = 0
+
+
+@dataclass
+class Node:
+    id: str
+    ancestors: list
```
emily-brents: I mentioned this in person, but I find it very odd to have a tree where each node has a complete list of its ancestors. I feel like it would make more sense to either have just the parent, and traverse the tree from child to parent; or to not have the parent/ancestors at all and traverse from parent to child. If having the ancestor list works that's fine, it just seems more complicated than necessary.

usaspending_api/references/v2/views/filter_trees/tas_filter_tree.py
```diff
@@ -0,0 +1,94 @@
+from usaspending_api.common.helpers.agency_logic_helpers import cfo_presentation_order
+from usaspending_api.accounts.models import TreasuryAppropriationAccount, FederalAccount
+from usaspending_api.common.helpers.agency_logic_helpers import agency_from_identifiers
+from usaspending_api.references.v2.views.filter_trees.filter_tree import DEFAULT_CHILDREN, Node, FilterTree
+from usaspending_api.references.models import ToptierAgency
+
+
+class TASFilterTree(FilterTree):
+    def toptier_search(self):
```
kbard: Are we supposed to be limiting these to only those agencies that have File C data?  Usually we don't want to show an agency if there's no way it can produce results.

usaspending_api/references/v2/views/filter_trees/tas_filter_tree.py
```diff
@@ -0,0 +1,94 @@
...
+        }
+        agency_dictionaries = [self._dictionary_from_agency(agency) for agency in agency_set]
+        cfo_sort_results = cfo_presentation_order(agency_dictionaries)
+        return cfo_sort_results["cfo_agencies"] + cfo_sort_results["other_agencies"]
+
+    def _dictionary_from_agency(self, agency):
+        if agency:
+            return {"toptier_code": agency.toptier_code, "name": agency.name}
+        else:
+            return {"toptier_code": "-1", "name": "failed to find agency"}
```
kbard: Usually we don't show agencies that are not in our toptier table.  I'm pretty sure we don't want -1 "failed to find agency" in our response (which I'm seeing).

usaspending_api/references/v2/views/filter_trees/tas_filter_tree.py
```diff
@@ -0,0 +1,94 @@
...
+    def toptier_search(self):
+        agency_id_sets = TreasuryAppropriationAccount.objects.values("fr_entity_code", "agency_id").distinct(
+            "agency_id"
+        )
+        agency_set = {
+            ToptierAgency.objects.filter(
+                toptier_code=agency_from_identifiers(elem["agency_id"], elem["fr_entity_code"])
+            ).first()
+            for elem in agency_id_sets
+        }
```
kbard: The way this is written, it is pinging the database about 320 times to generate the top level of the tree which results in a 30 second response time when I run it against `dev`.  I believe this can be written in a single query which should help.

usaspending_api/references/v2/views/filter_trees/filter_tree.py
tony-sappe: This method isn't returning JSON

usaspending_api/references/v2/views/filter_trees/tas_filter_tree.py
BrianZito: on line 10 the list of agency ids to search for is calculated from the TAS objects in order to ensure this.

usaspending_api/references/v2/views/filter_trees/filter_tree.py
BrianZito: Would you consider "to_dictionary" more accurate?

usaspending_api/common/helpers/agency_logic_helpers.py
BrianZito: Ross said, "any 058 federal account would be listed under 070 for the TAS filter" in conversations for clarification. Combined with the requirement to not show agencies without TAS under them, I took the interpretation that FEMA should not be shown at all.

usaspending_api/references/v2/views/filter_trees/filter_tree.py
BrianZito: It might end up being removable, depending on how the many refactors of the front end pan out, but if we end up doing TAS similarly to how NAICS is being done as of time of writing, we need a reliable way to confirm if one node is ancestor of another, even if the nodes aren't provided in the same tree. Given that I'm not sure it will be removed and it was relatively easy to build, I was intending to start with the higher amount of requirements and pull it off if it's not necessary.

BrianZito:
Added "Do Not Merge" since performance is still prohibative on the need to find possible FREC matches.

usaspending_api/references/v2/views/filter_trees/tas_filter_tree.py
```diff
@@ -0,0 +1,95 @@
...
+from usaspending_api.references.v2.views.filter_trees.filter_tree import DEFAULT_CHILDREN, Node, FilterTree
+from usaspending_api.references.models import ToptierAgency
+
+
+class TASFilterTree(FilterTree):
+    def toptier_search(self):
+        agency_id_sets = TreasuryAppropriationAccount.objects.values("fr_entity_code", "agency_id").distinct(
+            "agency_id"
+        )
+        agency_set = set(
```
tony-sappe: Is `set()` being used to remove duplicates?  If yes, it likely would be optimal to leverage distinct on the ORM query. If not, perhaps use `list()` to avoid confusion such as mine from future version of ourselves.

usaspending_api/references/v2/views/filter_trees/filter_tree.py
tony-sappe: Yes. Would you?

BrianZito: Technically, yes, but the way people use the term colloquially it's kind of like the difference between url and uri. The Response class we use has a "json()" method that returns a dictionary. Not a hill to die on though.

tony-sappe: I was amused people jumped on the camel case when the name was inaccurate. The example is different since the Request object contains JSON and provides that data in a native datatype. Keep `to_json()` if it works for the devs actually reviewing.

usaspending_api/common/helpers/agency_logic_helpers.py
kbard: I wish we had a place to record such decisions.  I've implemented many off-the-cuff decisions like this.  Wonder if this needs to apply anywhere else or just in the tree.  Thanks for the clarification.

usaspending_api/references/v2/views/filter_trees/tas_filter_tree.py
kbard: Having a TAS doesn't guarantee there's File C data for that TAS.

But that's sort of irrelevant.  I meant File D and edited my original comment accordingly.  My apologies.

Since this is an Advanced Search filter, it likely needs to only include those agencies, federal accounts, and treasury accounts that can be linked to File D data.  Maybe we don't needs this for minimum viability and can spin that off into its own ticket.  Might want to check with the PO.

usaspending_api/references/v2/views/filter_trees/filter_tree.py
BrianZito: language-specific casing is like unused import statements; we all know it's not important, but it's really obvious, and I feels good to make a comment knowing you can't be wrong.

kbard:
I feel compelled to make some progress on my own tickets for a while.  I will pick this review back up later this afternoon.

BrianZito:
Just found new logic for mapping federal accounts to agencies in https://federal-spending-transparency.atlassian.net/browse/DEV-3796. Will hold off any merging until this logic is reflected here.

usaspending_api/common/helpers/agency_logic_helpers.py
kwhickey: These types of decisions heard from the PO, whether coming to _any_ of the tester, developer, or scrum master, **should be recorded as _acceptance criteria_** in the ticket, and re-affirmed by the PO and acknowledged by the tester by way of Approving the AC (-> _Ready for Sprint_)

BrianZito: I've more than once had a chat with the PO that ended with, "Talk to Kirk, he knows this stuff". We're moving into retro territory, but the reality is that this business logic is fluid. The code above, in fact, I know to be wrong now.

kbard: We're in deep trouble if I'm the source of business rules.  I wish I knew a good way to record these things.  I've been on a couple of projects where we've attempted to capture business rules, but they've never really worked very well (in my opinion).  The closest we came was a Lotus Notes repository where we organized business rules.  Is Lotus Notes still a thing?  Man I'm old.

usaspending_api/references/v2/views/filter_trees/filter_tree.py
kwhickey: > we need a reliable way to confirm if one node is ancestor of another

Can we just provide a `parent` or `parent_node` or `parent_tas` attribute back for the requested node that allows the answer to the question: _"Do you have a parent? Who is your parent"_?

BrianZito: To be honest, I forget if that would suffice or not. I'll check in with Max now that we have one working filter tree and see what logic he's using at this point (remember that full ancestry like this is an inherent feature of NAICS)

usaspending_api/api_contracts/contracts/v2/references/filter_tree/tas.md
```diff
@@ -0,0 +1,67 @@
...
+                    + `2` 
+        With this tree structure, only TAS will be returned, and the tree depth will always be one, reguardless of provided depth.
+
++ Response 200 (application/json)
+    + Attributes (object)
+        + `results` (required, array[TASFilterTreeNode], fixed-type)
+
+## Data Structures
+
+### TASFilterTreeNode (object)
```
MaxwellKendall: There's not a name attribute or anything...?

BrianZito: Forgot to add  this to the documentation. Should be corrected now.

usaspending_api/references/v2/views/filter_trees/filter_tree.py
BrianZito: Will this still give you enough information to use the rest  of the API? In  order  to be as RESTful as possible and avoid possible ambiguity, the other endpoints introduced in this PR require the full path down  to the level you're searching for  children of.

usaspending_api/api_contracts/contracts/v2/references/filter_tree/tas.md
```diff
@@ -0,0 +1,68 @@
...
+
++ `id` (required, string)
++ `description` (required, string)
++ `ancestors` (required, array[string])
++ `count` (required, number)
++ `children` (required, array[TASFilterTreeNode], nullable)
```
MaxwellKendall: We need a way to know what depth we're at...

If we do decide to only provide the immediate ancestor, can I infer this anyway based on the nature of the data or is it necessary for you to tell me explicitly in the response object?

BrianZito: inference is fine

BrianZito: It would be possible to infer, but that will involve the front end doing  business logic, which I would prefer to avoid.

usaspending_api/references/constants.py
```diff
@@ -27,3 +27,6 @@
     ("011", "8238"),
     ("011", "8242"),
 ]  # Federal Account (AID, MAIN)s that are to be reported under DoD.
+
+# Agencies which should be excluded from dropdowns.
+EXCLUDE_CGAC = ["000", "067"]
```
kbard, line 32: Why was this added back?  It's obsolete now.

BrianZito: Must have been an artifact of reverting stuff when I was adding constants in the old implementation.

usaspending_api/references/tests/integration/filter_tree/conftest.py
```diff
@@ -0,0 +1,22 @@
+from usaspending_api.references.tests.integration.filter_tree.tas.tas_data_fixtures import (
+    basic_agency,
+    cfo_agencies,
+    non_cfo_agencies,
+    unsupported_agencies,
+    multiple_federal_accounts,
+    agency_with_unsupported_fa,
+    multiple_tas,
+    fa_with_unsupported_tas,
+)
```
kbard, line 10: Is there a value in doing it this way over just putting all the fixtures in conftest.py?  It seems like unnecessary work to put all the fixtures in one file and then import them into another.  I suppose if we ever have multiple files, but we can burn that bridge when we get to it.

BrianZito: I'd love to hear your alternative. I hold no love for these conftest files in the first place; I just put up with them as a nessicary step for the otherwise quite nice pytest fixture patter.

usaspending_api/references/v2/views/filter_tree/tas_filter_tree.py
```diff
@@ -0,0 +1,100 @@
...
+from django.db.models import Exists, OuterRef
+
+
+class TASFilterTree(FilterTree):
+    def toptier_search(self):
+        agency_set = (
+            ToptierAgency.objects.annotate(
+                has_faba=Exists(
+                    FinancialAccountsByAwards.objects.filter(
+                        treasury_account__federal_account__parent_toptier_agency=OuterRef("pk")
```
kbard, line 15: This is my fault.  Since we're applying this to advanced search, we really want to check if there's File D data.  This is checking for File C only.  We need to take it one step further.

To this filter function (and the others like it), can we add  `award_id__isnull=False'?

Sorry about that.

usaspending_api/references/tests/integration/filter_tree/conftest.py
kbard: Oh, I was just suggesting copying the contents of tas_data_fixtures.py into this file and deleting tas_data_fixtures.py.  You don't have to, though.  I was just wondering if there was some value I wasn't seeing doing it this-a-way.

usaspending_api/references/v2/views/filter_tree/tas_filter_tree.py
BrianZito: I've trusted you this far...

kbard: aaaand look where that's gotten you...

tony-sappe: walked right into that one

usaspending_api/api_contracts/contracts/v2/references/filter_tree/tas.md
```diff
@@ -0,0 +1,1035 @@
+FORMAT: 1A
```
tony-sappe, line 1: Please remove `usaspending_api/api_contracts/contracts/v2/references/filter_tree/__init__.py`

usaspending_api/api_contracts/contracts/v2/references/filter_tree/tas.md
```diff
@@ -0,0 +1,1035 @@
...
+
++ Response 200 (application/json)
+    + Attributes (object)
+        + `results` (required, array[TASFilterTreeNode], fixed-type)
+    + Body
+
+            {
+            "results": [
+            {
+            "id": "012-8226",
```
tony-sappe, line 73: Is there a benefit having 800+ lines of a response? Truncating to a few result objects should be sufficient to document the response and for a mock server

usaspending_api/api_contracts/contracts/v2/references/filter_tree/tas.md
```diff
@@ -0,0 +1,1035 @@
...
+            "ancestors": [],
+            "description": "Department of Agriculture",
+            "count": 139,
+            "children": null
+            }
+            ]
+            }
+
+
+## Search by Agency [GET /api/v2/references/filter_tree/tas/{agency}/{?depth}]
```
tony-sappe, line 46: Other contract files contain a single endpoint per file. Please split these into different files with appropriate directory structures and names to match the URL?

usaspending_api/references/tests/integration/filter_tree/conftest.py
tony-sappe: It does seem unnecessary, curious to learn the benefit

usaspending_api/api_contracts/contracts/v2/references/filter_tree/tas.md
BrianZito: As a heads up, our NAICS reference endpoint (https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/api_contracts/contracts/v2/references/naics.md) doesn't work that way.

If we plan to split both files, do you have a recommendation on how to name them? I see that we have an `awards/id.md` file, but that strikes me as very confusing to the unenlightened browser.

@MaxwellKendall @ebdabbs since these policies effect you most directly.

usaspending_api/api_contracts/contracts/v2/references/filter_tree/tas.md
tony-sappe: @BrianZito Wrong init file was removed

image

usaspending_api/api_contracts/contracts/v2/references/filter_tree/tas.md
tony-sappe: I believe the existing pattern would make the files:

- /v2/references/filter_tree/tas.md
- v2/references/filter_tree/tas/agency.md
- v2/references/filter_tree/tas/agency/federal_account.md


We can't devote enough time to always accommodate the random "unenlightened browser."
There are currently two user-friendly methods to find the correct contract if the directory structure is confusing to someone:
1. Endpoints included in https://api.usaspending.gov/docs/endpoints contain links to the correct contract 
1. The search functionality in Github's U/I

PR #2331, opened 2020-03-05
dev-4638 Allow searching on parent and grandparent NAICS

BrianZito:
**Description:**
Previously NAICS search only looked for exact matches. This change allows shorter NAICS to be searched, returning all results matching children.

**Technical details:**

    - [ ] Backend
8. [x] Jira Ticket [DEV-4638](https://federal-spending-transparency.atlassian.net/browse/DEV-4638):

```
```


BrianZito:
Testing results:

calling spending_by_award_count with body:
```
{
    "filters": {
        "time_period": [
            {
                "start_date": "2007-10-01",
                "end_date": "2019-09-30"
            }
        ],
        "naics_codes": [
            "111110"
        ],
        "award_type_codes": [
            "B",
            "A",
            "D",
            "C"
        ]     
    },
    "fields": [
        "Award ID",
        "Recipient Name",
        "Start Date",
        "End Date",
        "Award Amount",
        "Awarding Agency",
        "Awarding Sub Agency",
        "Contract Award Type"
    ],
    "page": 1,
    "limit": 60,
    "sort": "Award Amount",
    "order": "desc",
    "subawards": false
}
```
was used against local (in branch) and dev for various 6-digit naics to check for consistancy. Results were consistant for 111110, 237110, 237990, and 541330

BrianZito:
Testing results part 2: search with 
```
"naics_codes": [
            "5413"
        ],
```
returns 383858 results in dev, the same as searching the mv_contract_award_search matview in postico with filters:
naics_code LIKE 5413%
action_date >= 2007-10-01

usaspending_api/awards/v2/filters/matview_filters.py
```diff
@@ -249,9 +249,9 @@ def keyword_parse(keyword):
                 queryset = queryset.filter(cfda_number__in=in_query)
 
         elif key == "naics_codes":
-            in_query = [v for v in value]
-            if len(in_query) != 0:
-                queryset = queryset.filter(naics_code__in=in_query)
+            if value:
```
kbard: Can we add a validation to this?  It is very easy to submit requests that crash the database query which has me worried.  Turns out, you can do it in the TinyShield filter.  Search for `{"name": "naics_codes", "type": "array", "array_type": "text", "text_type": "search"}` and replace it with `{"name": "naics_codes", "type": "array", "array_type": "integer", "text_type": "search"}`.

You'll also need to update this bit here to convert all the ints to strings probably using a list comprehension `regex = f"^({'|'.join([str(v) for v in value])}).*"`.

sethstoudenmier:
Can you make the necessary change to https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/common/query_with_filters.py#L307 as well? Want to keep the ES filters for Advanced Search page up to date with the Postgres filters.

usaspending_api/awards/v2/filters/matview_filters.py
BrianZito: That looks like a good solution to this problem, but later we will need to add this functionality for TAS and PSC, which iirc will not be strictly numeric. I'm trusting Django to sanitize strings here, so can you provide an example of what might crash the DB?

kbard: Django sanitizes the strings for SQL not for regex.  Any regex construct could potentially be used to crash the query.  Try searching for a close paren... or brace... or bracket... etc.

BrianZito: Does TinyShield specify a regex for allowed inputs? That's the best way I can think of to deal with this, if our library supports it...

kbard: I don't believe so.

BrianZito:
> Can you make the necessary change to https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/common/query_with_filters.py#L307 as well? Want to keep the ES filters for Advanced Search page up to date with the Postgres filters.

Do we have any existing examples of regex in elasticsearch?

emily-brents:
LGTM, tested locally, didn't see any issues.

usaspending_api/common/query_with_filters.py
```diff
@@ -312,7 +312,7 @@ def _generate_elasticsearch_query(cls, filter_values: List[str], query_type: _Qu
         naics_codes_query = []
 
         for v in filter_values:
-            naics_codes_query.append(ES_Q("match", naics_code__keyword=v))
+            naics_codes_query.append(ES_Q("wildcard", naics_code__keyword=f"{v}*"))
```
kwhickey: +1 for changing the `match` and `match_phrase_prefix` to `wildcard`, and being able to take advantage of the `naics_code.keyword` subfield. It should make the ES impl faster.

usaspending_api/awards/v2/filters/matview_filters.py
```diff
@@ -249,9 +249,9 @@ def keyword_parse(keyword):
...
+                regex = f"^({'|'.join([str(elem) for elem in value])}).*"
```
kwhickey: Doing some tests with Brian against DEV with larger datasets, this regexp query with Postgres seemed to lag a whole bunch. 
We need a broader discussion about how best to implement these types of postgres searches/matches of hierarchical selections with e.g. @kbard and @tony-sappe and @emily-brents before this code can move forward. 

We may never get it where we want it with postgres, hence providing an Elasticsearch backend impl, but we should at least be consistent with using the best-possible postgres impl.

kbard:
> Postgres-backed impl has a very high latency when there's a high number of matches.

This was true of the original version of the endpoint as well.  I did some performance comparisons  and saw little to no difference in performance.

BrianZito:
> @BrianZito -
> Looks like we're at status quo performance with the regex impl.
> 
> Can you tell me if you tried any of these and if the performance gets much faster when doing so?
> 
> See here: https://stackoverflow.com/questions/2245536/sql-combining-multiple-like-queries
> 
> 1. `WHERE naics_code SIMILAR TO '11%|22%|3344%'`
> 2. `WHERE naics_code LIKE ANY(ARRAY['11%', '22%', '3344%'])`
> 3. Converting given codes into numeric ranges:
>    
>    * for each prefix given, derive the lowerbound and upper bound of the range
>    * lower: zero-pad to 6 digits
>    * upper: add one, then zero-pad to 6 digits, then subtract 1
>    * include any results that match `AND naics_code BETWEEN {lower} AND {upper}`
>    * Example: Code = `14`. lower = 140000; upper = 14 + 1 -> 150000 - 1 = 149999:  `AND naics_code BETWEEN 140000 AND 149999`
>    * There could be some pre-processing on the API python side to collapse ranges.
>    * If there are a lot of disjoint ranges, this may not really work out performance wise
> 
> Firstly I'm curious if these are indeed faster when running as SQL than the regex style.
> Secondly, can we make Django ORM do these things?

I have not, and unless we plan to bring the PSQL functionality back into scope, I suggest we don't. I only left the Postgres functionality in there for ease of testing, and to avoid making exception logic to ensure the endpoints won't use those filters unless elasticsearch is activated.

kwhickey:
@BrianZito I suggest you try at least the first 2 styles. Should take less than an hour to connect to a DB with significant data and compare regex style with these two. 

The reason is: We're still considering a _partial_ release of the Elasticsearch converted Advanced Search endpoints, to deliver on some value already achieved. This would leave the unconverted endpoints _still using 100% postgres-backed_ search and filter implementation. And, if we release the new NAICS filter with the partial release of ES, we'll need the postgres impl to be as fast as it can be (balancing complexity of the impl). 

So if either of those other 2 approaches prove much faster, they give us viable options to pursue a partial release of the ES conversion without tanking certain endpoints that use the NAICS filter.

kbard:
So, of these suggestions, the only one that caught my eye was converting the NAICS to an integer.  I cannot remember whether or not I attempted that.  I just attempted it (again?).  It didn't help.  It still resulted in a minute and a half run time for my sample query.

I'm messing around with statistics now.  I'll keep tinkering during quite moments today.

PR #2336, opened 2020-03-12
Mod/dev 4640 add exclusion logic

BrianZito:
**Description:**
Adds logic to filter on NAICS which should NOT appear.

**Technical details:**

    - [x] Backend
    - [x] Frontend 
8. [x] Jira Ticket [DEV-4640](https://federal-spending-transparency.atlassian.net/browse/DEV-4640):

```
```


BrianZito:
Simple arithmetic tests to determine logic:

```
TEST 1 (no exclusion)
{
  "filters": {
    "time_period": [
      {
        "start_date": "2018-10-01",
        "end_date": "2019-09-30"
      }
    ],
    "award_type_codes": [
      "A",
      "B",
      "C",
      "D"
    ],
    "naics_codes": [
    "23"	
    ]
  },
  "page": 1,
  "limit": 60,
  "sort": "Award Amount",
  "order": "desc",
  "subawards": false
}
"results": {
        "contracts": 129,
        "direct_payments": 0,
        "grants": 0,
        "idvs": 0,
        "loans": 0,
        "other": 0
    },


TEST 2 (with exclusion)
{
  "filters": {
    "time_period": [
      {
        "start_date": "2018-10-01",
        "end_date": "2019-09-30"
      }
    ],
    "award_type_codes": [
      "A",
      "B",
      "C",
      "D"
    ],
    "naics_codes": [
    "23"	
    ],
    "exclude_naics_codes": [
    "2383"	
    ]
  },
  "page": 1,
  "limit": 60,
  "sort": "Award Amount",
  "order": "desc",
  "subawards": false
}
"results": {
        "contracts": 122,
        "direct_payments": 0,
        "grants": 0,
        "idvs": 0,
        "loans": 0,
        "other": 0
    },

TEST 3 (only excluded class):
{
  "filters": {
    "time_period": [
      {
        "start_date": "2018-10-01",
        "end_date": "2019-09-30"
      }
    ],
    "award_type_codes": [
      "A",
      "B",
      "C",
      "D"
    ],
    "naics_codes": [
    "2383"	
    ]
  },
  "page": 1,
  "limit": 60,
  "sort": "Award Amount",
  "order": "desc",
  "subawards": false
}
"results": {
        "contracts": 7,
        "direct_payments": 0,
        "grants": 0,
        "idvs": 0,
        "loans": 0,
        "other": 0
    },

```

usaspending_api/api_contracts/contracts/v2/download/awards.md
```diff
@@ -164,3 +164,7 @@ This route sends a request to the backend to begin generating a zipfile of award
 + `city` (optional, string)
 + `district` (optional, string)
 + `zip` (optional, string)
+
+### NAICSCodeObject (object)
++ `require`: [`33`]
```
tony-sappe: These fields need the types to be defined instead of showing an example (`optional, array[string]`)

usaspending_api/awards/v2/filters/matview_filters.py
```diff
@@ -255,10 +255,23 @@ def keyword_parse(keyword):
...
-            if value:
-                regex = f"^({'|'.join([str(elem) for elem in value])}).*"
+            if isinstance(value, list):
+                requires = value
+                exclude = []
+            elif isinstance(value, dict):
+                requires = value.get("require") or []
+                exclude = value.get("exclude") or []
+            else:
+                raise InvalidParameterException(f"naics_codes must be an array or object")
```
sethstoudenmier: The `f` here is not necessary since it is just a string.

usaspending_api/awards/v2/filters/matview_filters.py
```diff
@@ -255,10 +255,23 @@ def keyword_parse(keyword):
                 queryset = queryset.filter(cfda_number__in=in_query)
 
         elif key == "naics_codes":
-            if value:
-                regex = f"^({'|'.join([str(elem) for elem in value])}).*"
+            if isinstance(value, list):
```
sethstoudenmier: Should the validation for the `naics_code` be in the filter code?

BrianZito: The exception can be removed since TinyShield is already covering this. I was just trying to fail loudly in  the event both of these checks fail, mostly to not have the code look strange here.

sethstoudenmier: Personally would like to see the error removed. If we have the same error thrown in multiple places it turns into an easter egg hunt to find where it came from.

usaspending_api/search/elasticsearch/filters/naics.py
```diff
@@ -0,0 +1,97 @@
...
+    def generate_elasticsearch_query(cls, filter_values, query_type: _QueryType) -> ES_Q:
+        # legacy functionality permits sending a single list of naics codes, which is treated as the required list
+        if isinstance(filter_values, list):
+            requires = filter_values
+            exclude = []
+        elif isinstance(filter_values, dict):
+            requires = filter_values.get("require") or []
+            exclude = filter_values.get("exclude") or []
+        else:
+            raise InvalidParameterException(f"naics_codes must be an array or object")
```
sethstoudenmier: Same as for the matview_filter here if you go to remove the exceptions.

usaspending_api/search/tests/unit/filters/naics.py
```diff
@@ -0,0 +1,37 @@
+from usaspending_api.search.elasticsearch.filters.naics import NaicsCodes
+
+
+def test_primative_naics_filter():
+    assert NaicsCodes._query_string(["11"], []) == "(11*)"
```
sethstoudenmier: Small details but might be good to change the test cases to call out the parameter since it is just two lists next to each other. For example, `assert NaicsCodes._query_string(requires=["11"], exclude=[]) == "(11*)"`. Will help anyone who goes behind you on these test cases.

usaspending_api/search/elasticsearch/filters/naics.py
```diff
@@ -0,0 +1,97 @@
...
+
+        requires = [str(code) for code in require]
+        exclude = [str(code) for code in exclude]
+
+        return ES_Q("query_string", query=cls._query_string(requires, exclude), default_field="naics_code")
+
+    @classmethod
+    def _query_string(cls, require, exclude):
```
kwhickey: Doc strings always helpful when things aren't totally obvious. Saves time groking the code, even just to summarize the gist.

usaspending_api/search/elasticsearch/filters/naics.py
```diff
@@ -0,0 +1,97 @@
...
+        positive_query = " OR ".join([node.get_query() for node in positive_nodes])
+        negative_query = " AND ".join([node.get_query() for node in negative_nodes])
+
+        if positive_query and negative_query:
+            return f"{positive_query} AND {negative_query}"
+        else:
+            return positive_query + negative_query  # We know that exactly one is blank thanks to TinyShield
+
+    @staticmethod
+    def _order_naics_codes(requires, exclude, all_codes):
```
kwhickey: Doc string as well.

usaspending_api/search/elasticsearch/filters/naics.py
```diff
@@ -0,0 +1,97 @@
...
+        negative_codes = {
+            "top": [code for code in exclude if len([root for root in all_codes if code.startswith(root)]) == 1]
+        }
+        postive_codes["sub"] = [code for code in requires if code not in postive_codes["top"] + negative_codes["top"]]
+        negative_codes["sub"] = [code for code in exclude if code not in postive_codes["top"] + negative_codes["top"]]
+
+        return postive_codes, negative_codes
+
+
+class _NaicsNode:
```
kwhickey: Doc string...

usaspending_api/search/tests/unit/filters/naics.py
```diff
@@ -0,0 +1,37 @@
+from usaspending_api.search.elasticsearch.filters.naics import NaicsCodes
```
kwhickey: 👍 this is a good start. i was really hoping to see this, because that code is fraught with subtle assumptions about what is and is not possible, with zero documentation. 

However, I think you ought to add more of these test cases. 
e.g. specifically, add some tests with _many_ elements in the arrays, that have overlapping prefixes, and just many in general, and assert that the result is what was expected. 
Also if order has any bearing, try in the default order, then reverse the arrays and try again.

kwhickey: I'd also actually like to see an integration test that hits ES to be sure that: not only does it generate the expected `query_string` syntax, but that your assumptions about how ES will execute that query string on an indexed data set are correct.

Please add that, since we all make assumptions about ES that we are discovering new things about every day.

BrianZito: Integration test was modified at https://github.com/fedspendingtransparency/usaspending-api/pull/2336/files#diff-53a443dd0745aa53a400aa5732f26ebd

Tons more tests could be made of course, but I was trying to respect the simple structure that existed, in order to avoid having excessive tests checking the same thing.

kwhickey: Good to see we have at least one real hit of ES data. 
However, can you piggy back on that test and run a few more data scenarios / queries that force execution of some of the more convoluted query strings and assert the results are as expected. 
Since the testbed is already there, should just by a bit of copy-paste-edit-rename.

BrianZito: I'd want to be a little cleaner than a copy/paste strategy, particularly with fixtures, since what we have for spending_by_award is a bit of a mess and it will only get harder to work with (and ensure that failed tests are failing for the reason we think they are) as the fixture gets built out to support more complex cases.
While cleaning up the fixture data and restructuring this test class is certainly something I'd like to do, there's a question of scope in doing so, since the way I'd want to add more integration tests is more effort than a copy/paste job.

BrianZito: Can you give an example for "overlapping prefixes"?

kwhickey: @BrianZito Take a look at this branch as a starter `mod/dev-4640-add-exclusion-logic-integ-tests`, that can be merged into yours. 

Should just be a few things that need to change if you build off of the fixture data that already exists.

https://github.com/fedspendingtransparency/usaspending-api/compare/mod/dev-4640-add-exclusion-logic...mod/dev-4640-add-exclusion-logic-integ-tests

BrianZito: Your example shows cake pattern, which I was asked not to do before. Are we okay with adding fixture data outside python fixtures now?

kwhickey: There is overlap among these 3
```
"naics_codes": {
    "require": ["11", "112233"],
    "exclude": ["1122"]
}
```

And these three
```
"naics_codes": {
    "require": ["1122"],
    "exclude": ["11", "112233"]
}
```

And these three
```
"naics_codes": {
    "require": ["112233"],
    "exclude": ["11", "1122"]
}
```

And ...
```
"naics_codes": {
    "require": ["11", "112233"],
    "exclude": ["112233"]
}
```
And
```
"naics_codes": {
    "require": ["112233", "112233"],
    "exclude": ["112233", "112233"]
}
```
And
```
"naics_codes": {
    "require": ["11", "112233"],
    "exclude": ["1122"]
}
```
And
```
"naics_codes": {
    "require": ["112233", "112233"],
    "exclude": ["112233", "1122", "112233"]
}
```

While you're at it, it would also be good to add a unit test to verify logic when an empty string, 1-digit string, 3-digit string, or 5-digit string or 7+ digit string is passed.

kwhickey: The fixture is already shared among tests. And the data created in each test will be torn down after that test since it uses a _`function`_ scope for the fixture, yes? That includes the data from the fixture, plus any modifications or additions or deletions to that data needed for your specific test.

If we wanted to isolate test data to a per-test basis, then we should not create that data as a shareable fixture, but just create it within the individual test, so there is not a ripple effect when shared test data changes. 

I'm not sure where the rule stands. If you know, then choose one of two approaches:
1) Use the fixture as a shared fixture, add what you need onto it in your test
2) Copy the fixture code into your test, add what you need onto it in your test

Either way you have a path forward.

usaspending_api/search/elasticsearch/filters/naics.py
```diff
@@ -0,0 +1,103 @@
...
+        elif isinstance(filter_values, dict):
+            require = filter_values.get("require") or []
+            exclude = filter_values.get("exclude") or []
+        else:
+            raise InvalidParameterException(f"naics_codes must be an array or object")
+
+        requires = [str(code) for code in require]
+        exclude = [str(code) for code in exclude]
+
+        return ES_Q("query_string", query=cls._query_string(requires, exclude), default_field="naics_code")
```
kwhickey: ❗️ This needs to be changed to `naics_code.keyword` yes? 

Please use the keyword sub-fields or fields with keyword types as much as possible in Advanced Search filters.

usaspending_api/awards/v2/filters/matview_filters.py
```diff
@@ -255,9 +255,20 @@ def keyword_parse(keyword):
...
+                if value.get("exclude"):
+                    raise NotImplementedException(
+                        "NOT IMPLEMENTED: postgres endpoint does not currently support excluded naics!"
+                    )
+
+            else:
+                raise InvalidParameterException("naics_codes must be an array or object")
+
+            regex = f"^({'|'.join([str(elem) for elem in require])}).*"
+            queryset = queryset.exclude(naics_code__regex=regex)
```
dpb-bah: If I'm understanding this right. It's excluding the elements we're requiring? Is this a typo or am I missing something?

BrianZito: that looks like a huge typo

MaxwellKendall:
@BrianZito ✅ for contracts.

BrianZito:
Doing one more update with dev. Keep in mind that because of codeclimate I can't merge this PR.

usaspending_api/awards/v2/filters/matview_filters.py
kwhickey: Did this code get changed per the comment @BrianZito?
@sethstoudenmier if you had code that leveraged the 501 response, can you cite it here. Not sure if you can use it Brian, but would be good to distinguish "something unknown when wrong" (`500`), "an error condition we know about is being reported (`501`)"

BrianZito: We  discussed this when we went over the code  in person. Did  we come to any  conclusions, Seth?

usaspending_api/common/validator/award_filter.py
```diff
@@ -24,7 +24,22 @@
...
+        "name": "naics_codes",
+        "type": "any",
+        "models": [
+            {
+                "name": "naics_codes",
+                "type": "object",
+                "min": 0,
+                "object_keys": {
+                    "require": {"type": "array", "array_type": "integer", "text_type": "search", "min": 0},
+                    "exclude": {"type": "array", "array_type": "integer", "text_type": "search", "min": 0},
```
kwhickey: Is this for tiny shield validations? 
I see `"array_type": "integer"`, but aren't the naics_code arrays `string` arrays per the API docs?

BrianZito: It turns out that TinyShield's integer type allows strings so long as they parse into integers.

usaspending_api/awards/v2/filters/matview_filters.py
sethstoudenmier: There is a `NotImplementedException` which throws a 501, but in this case, I think an InvalidParamaterException is a correct response. My comment was around removing the thrown error all-together if we are confident that it is going to be caught by something such as TinyShield. Otherwise, I am fine with it here.

usaspending_api/search/tests/unit/filters/naics.py
kwhickey: I was thinking to limit setup time on each test, you could have multiple asserts in one test, modifying the data for each part of the test. That way you could pack a bunch of different `require`/`exclude` combos into 1 test. But these 3 are fine.

kwhickey: I was thinking to limit setup time on each test, you could have multiple asserts in one test, modifying the data for each part of the test. That way you could pack a bunch of different `require`/`exclude` combos into 1 test. But these 3 are fine.

BrianZito: Keep in mind these are unit tests, whose setup takes milliseconds.

BrianZito: Keep in mind that this would still require reindexing the elasticsearch each time no matter what, which I presume is where most of the time comes from.

usaspending_api/api_contracts/contracts/v2/download/awards.md
kwhickey: This looks resolved, but the [`33`] should be in quotes as a string, yes?

BrianZito: I don't think so. MSON isn't a very strict language https://apiblueprint.org/documentation/mson/specification.html#232-value-member-types

usaspending_api/common/validator/award_filter.py
kwhickey: Closing loop: this is being kept this way because it allows tinyshield to verify the array items are in fact integers (even though they need to be passed in quotes).

PR #2345, opened 2020-03-17
[DEV-4649] Remove ES page jumping

emily-brents:
**Description:**
Removes the "jumping" feature on spending_by_award when using Elasticsearch due to performance issues.

**Technical details:**
The "jumping" worked by determining the correct record based on the page number and the limit, then doing a postgres search to get the value of the records preceding the desired first record, then querying elasticsearch using the `search_after`. The performance for this was bad. 

Note: Removing this code will require using a different API request for spending_by_award, but only after query through 50,000 records. I.E. - for pages 1-833 of the advanced search page (using limit: 60), the API will return fine. After that, you will have to use the last_record_unique_id" and "last_record_sort_value" to get the correct results.

    - [x] Backend
8. [X] Jira Ticket [DEV-4649](https://federal-spending-transparency.atlassian.net/browse/DEV-4649):

```
No changes to matview, no ops tickets needed
```

### DIFF
--- a/usaspending_api/search/v2/views/search_elasticsearch.py
+++ b/usaspending_api/search/v2/views/search_elasticsearch.py
@@ -56,7 +56,7 @@ class SpendingByTransactionVisualizationViewSet(APIView):
         record_num = (validated_payload["page"] - 1) * validated_payload["limit"]
         if record_num >= settings.ES_TRANSACTIONS_MAX_RESULT_WINDOW:
             raise UnprocessableEntityException(
-                "Accessing page {page} with limit {limit} exceeds the Elasticsearch limit of {es_limit}. Please use downloads to access the full set of results.".format(
+                "Page #{page} of size {limit} is over the maximum result limit ({es_limit}). Consider using custom data downloads to obtain large data sets.".format(
                     page=validated_payload["page"],
                     limit=validated_payload["limit"],
                     es_limit=settings.ES_TRANSACTIONS_MAX_RESULT_WINDOW,
--- a/usaspending_api/search/v2/views/search_elasticsearch.py
+++ b/usaspending_api/search/v2/views/search_elasticsearch.py
@@ -7,7 +7,11 @@ from rest_framework.views import APIView
 
 from usaspending_api.common.api_versioning import api_transformations, API_TRANSFORM_FUNCTIONS
 from usaspending_api.common.cache_decorator import cache_response
-from usaspending_api.common.exceptions import ElasticsearchConnectionException, InvalidParameterException
+from usaspending_api.common.exceptions import (
+    ElasticsearchConnectionException,
+    InvalidParameterException,
+    UnprocessableEntityException,
+)
 from usaspending_api.common.helpers.generic_helper import get_simple_pagination_metadata
 from usaspending_api.common.validator.award_filter import AWARD_FILTER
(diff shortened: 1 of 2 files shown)


usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -313,50 +313,14 @@ def query_elasticsearch(self) -> list:
...
         if (self.last_record_sort_value is None and self.last_record_unique_id is not None) or (
             self.last_record_sort_value is not None and self.last_record_unique_id is None
         ):
             # malformed request
             raise Exception(
                 "Using search_after functionality in Elasticsearch requires both last_record_sort_value and last_record_unique_id."
             )
-
-        # API request is asking to jump to a random, non-sequential page of results
-        if record_num >= settings.ES_AWARDS_MAX_RESULT_WINDOW and (
```
tony-sappe: What is the behavior when an API request is attempting to fetch a random page from ES? Is it handled gracefully?

emily-brents: No, it'll just return a 500, which is the same as keyword search currently. I think adding a specific error message would be good, so should I add it to both advanced and keyword?

tony-sappe: Yes, a 422 with a custom message would be much better. We shouldn't tolerate 500 responses

emily-brents: Alright, added 422 to both spending_by_award and spending_by_transaction. Let me know if you have any thoughts on the wording

usaspending_api/search/v2/views/search_elasticsearch.py
```diff
@@ -49,6 +53,16 @@ def post(self, request):
                 m["optional"] = False
         validated_payload = TinyShield(models).block(request.data)
 
+        record_num = (validated_payload["page"] - 1) * validated_payload["limit"]
+        if record_num >= settings.ES_TRANSACTIONS_MAX_RESULT_WINDOW:
+            raise UnprocessableEntityException(
+                "Accessing page {page} with limit {limit} exceeds the Elasticsearch limit of {es_limit}. Please use downloads to access the full set of results.".format(
```
tony-sappe: It's a good practice to hide backend architecture from the API. Something along these lines?

`Page #{page} of size {limit} is over the maximum result limit ({es_limit}). Consider using custom data downloads to obtain large data sets.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -313,50 +314,25 @@ def query_elasticsearch(self) -> list:
...
-            search = (
-                AwardSearch()
-                .filter(filter_query)
-                .sort(*sorts)
-                .extra(search_after=[*results])[: self.pagination["limit"] + 1]
-            )
-            response = search.handle_execute()
-            return response
-
+        # es_limit=settings.ES_AWARDS_MAX_RESULT_WINDOW,
```
tony-sappe: intentional comment?

emily-brents: fun fact: no.

usaspending_api/search/v2/views/search_elasticsearch.py
emily-brents: changed.

PR #2349, opened 2020-03-17
[DEV-4069] Implement Spending by CFDA with ES

sethstoudenmier:
**Description:**
Implement Spending by CFDA as a new endpoint. Additionally, allow the use of Elasticsearch for both the old and new endpoint with CFDA.

**Technical details:**
While implementing Spending by CFDA using a new approach the Spending by Agency work needed to be refactored. The new implementation pre-aggregates the fields needed to determine both uniqueness and the response into a JSON formatted string. For example:
```
"{'number':'<cfda_number>','description':'<cfda_program_title>','id':'<cfda_id_in_usas_db>'}"
```

In a case where `cfda_number` does not exist for a record (e.g. FPDS transactions) then a value of `null` is stored in Elasticsearch. Storing a value of `null` allows for the document to be ignored when created aggregations. To avoid issues with data quality the `cfda_program_title` and `cfda_id` are both retrieved from a reference table, `references_cfda`. Since the JSON format is utilized parsing the information for each unique bucket is as easy as `json.loads(<bucket_key>)`.

Should note that part of the `transaction_delta_view.sql` needs to be left to avoid breaking current functionality with the Spending by Agency endpoints. This can be removed as a warmfix into Staging once the Sprint 103 deploy is completed. 

**Performance**
To test performance the nonprod performance cluster was reindexed using the changes in this PR and a series of test filters were used to get an accurate gauge. The worst performance was found with Recipient Type and Award Amount (two filters that are known for not having great performance) at about 15 seconds and 9 seconds respectively. In addition to these, slow times of 16 seconds were also noted when retrieving all possible records by setting minimum and maximum time periods. If these times are taken out then the average response time is 2.5 seconds where there are results and 1 second when there are no results found. The cardinality check at the beginning of the search helps to short circuit requests without performing a full query.

    - [x] Backend
8. [X] Jira Ticket [DEV-4069](https://federal-spending-transparency.atlassian.net/browse/DEV-4069):

```
```


usaspending_api/api_contracts/contracts/v2/search/spending_by_category/cfda.md
```diff
@@ -0,0 +1,166 @@
...
+            The page of results to return based on the limit
+        + `subawards` (optional, boolean)
+            Determines whether Prime Awards or Sub Awards are searched
+    + Body
+        
+            {
+                "filters": { 
+                    "keywords": ["Filter is required"] 
+                }
+            }
```
kbard: Now that we're being more targeted with endpoints, should our sample bodies reflect that?  I feel like this one leaves quite a bit to the imagination.

tony-sappe: I don't recommend a sample body section for the request, only for the response

kbard: May I ask why?  I find both to be quite useful when reading documentation.

tony-sappe: The response JSON is much more useful for tools such as mock servers.

The requests, as you mentioned, can leave much to the imagination since they are very complex. To capture all possible filters  they would be very detailed or perhaps they would be random combinations of filters with varying levels of benefit. Additionally, the docs would need to be maintained which was a problem pre-API Contract. Typically the responses are more static than the global search filters and tied to a specific endpoint so it is easier for a dev to update.

If someone wants to spend the time writing an example request I wouldn't force them to remove it. But AFAIK it was never discussed as an important section to include.

kbard: Just thought this would be a good opportunity to (at least in my opinion) improve our contracts/documentation a little.  If the only other opinion is a dissenting one, then I don't care enough to press it.  Consider my original musing withdrawn.

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_industry_codes.py
```diff
@@ -0,0 +1,75 @@
...
+)
+
+
+class IndustryCodeType(Enum):
+    CFDA = "cfda_number"
+    PSC = "product_or_service_code"
+    NAICS = "naics_code"
+
+
+class BaseIndustryCodeViewSet(BaseSpendingByCategoryViewSet, metaclass=ABCMeta):
```
kwhickey: Stylistic thing, but one I'm fond of: when creating an abstract class, name it with `Abstract`, e.g. `AbstractIndustryCodeViewSet`. It implies that it's base, and also alerts people that it's not instantiable. If that's too verbose, you could use `AIndustryCodeViewSet` or `ACFDAViewSet` (of course we use uppercase for acronyms on this project 😐 ... so that one looks bad)

usaspending_api/search/v2/views/spending_by_category_views/spending_by_industry_codes.py
```diff
@@ -0,0 +1,16 @@
+from usaspending_api.search.v2.views.spending_by_category_views.base_spending_by_category import Category
+from usaspending_api.search.v2.views.spending_by_category_views.base_spending_by_industry_codes import (
+    IndustryCodeType,
+    BaseIndustryCodeViewSet,
+)
+
+
+class CfdaViewSet(BaseIndustryCodeViewSet):
```
kwhickey: Oh wait. A dissident. _To the revolution!_

usaspending_api/etl/es_transaction_template.json
```diff
@@ -286,20 +274,41 @@
             }
           }
         },
-        "awarding_toptier_agency_code": {
-          "type": "keyword"
-        },
-        "funding_toptier_agency_code": {
-          "type": "keyword"
+        "awarding_toptier_agency_agg_field": {
```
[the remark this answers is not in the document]
kwhickey: Nevermind. I see this naming convention is used in _lots_ of places.

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_category.py
```diff
@@ -124,39 +123,42 @@ def common_db_query(self, queryset: QuerySet, django_filters: dict, django_value
...
 
         The top-most aggregation is expected to use partitions to make sure that all buckets are retrieved.
         """
-        pass
+        # Create the filtered Search Object
+        search = TransactionSearch().filter(filter_query)
+
+        # Get count of unique buckets; terminate early if there are no buckets matching criteria
+        bucket_count = get_number_of_unique_terms(filter_query, f"{self.category.agg_field}.hash")
+        if bucket_count == 0:
```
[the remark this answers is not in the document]
sethstoudenmier: Logging a warning and throwing an error that notifies a user to use more filters should this case come up. As of now this case should not occur, more of a just in case.

usaspending_api/search/v2/views/spending_by_category_views/base_spending_by_industry_codes.py
```diff
@@ -0,0 +1,75 @@
...
+                    "name": industry_code_info.get("description") or None,
+                }
+            )
+
+        return results
+
+    def query_django(self, base_queryset: QuerySet) -> List[dict]:
+        if self.subawards:
+            if self.industry_code_type == IndustryCodeType.PSC or self.industry_code_type == IndustryCodeType.NAICS:
+                self.raise_not_implemented()
```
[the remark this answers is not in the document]
sethstoudenmier: Now throws a 501.

usaspending_api/search/tests/integration/spending_by_category/test_elasticsearch_header.py
```diff
@@ -18,7 +18,7 @@ def test_elasticsearch_headers(client, monkeypatch, elasticsearch_transaction_in
 def elasticsearch_http_header_helper(client, monkeypatch, elasticsearch_transaction_index, endpoint_name):
     logging_statements = []
     monkeypatch.setattr(
-        "usaspending_api.search.v2.views.spending_by_category_views.base_spending_by_category.logger.info",
+        "usaspending_api.search.v2.views.spending_by_category_views.abstract_spending_by_category.logger.info",
```
kwhickey: @sethstoudenmier now that I see this, I'm thinking it would be a more concise, more pythonic packaging structure to just put the `abstract` class and any of its concrete implementations into the same module.
That is,
- rename `abstract_spending_by_category.py` to `spending_by_category.py`
- and the others too, e.g. `abstract_spending_by_industry_codes.py` -> `spending_by_industry_codes.py`, `abstract_spending_by_agency.py` -> `spending_by_agency.py`
- and then put the definition for e.g. `class CfdaViewSet(AbstractIndustryCodeViewSet):` into `spending_by_industry_codes.py`

This current structure of having a module (file) per abstract class and concrete impl feels more Java-esque where it requires a file per Class, and I think python code would lean towards bundling them in the same module. 

Is there a reason to separate them? e.g. did code climate complain on length of the abstract file?

sethstoudenmier: No real reason to have them broken apart. Codeclimate probably will complain though.

PR #2347, opened 2020-03-17
[DEV-4693] Preventing caching when using elasticsearch header

emily-brents:
**Description:**
Prevents requests sent with the `X-Experimental-API` header from being stored in the redis cache.

**Technical details:**
Alters the caching decorator to check for the `X-Experimental-Api` header and prevents the request from being pickled in the cache if it is present and matches `elasticsearch`. This will need to be removed in the future if we move to a elasticsearch implementation.d
**Requirements for PR merge:**

1. [X] Unit & integration tests updated
2. [X] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
4. [N/A] Matview impact assessment completed
5. [N/A] Frontend impact assessment completed
6. [X] Data validation completed
7. [N/A] Appropriate Operations ticket(s) created
8. [ ] Jira Ticket [DEV-4693](https://federal-spending-transparency.atlassian.net/browse/DEV-4693):
    - [x] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
No ops tickets, no changes to matviews
```

PR #2358, opened 2020-03-23
[DEV-4693] Prevent ES caching

emily-brents:
Prevents Elasticsearch requests from populating the cache. Same as https://github.com/fedspendingtransparency/usaspending-api/pull/2347.

### DIFF
--- a/usaspending_api/common/cache_decorator.py
+++ b/usaspending_api/common/cache_decorator.py
@@ -2,11 +2,18 @@
 import logging
 from rest_framework_extensions.cache.decorators import CacheResponse
 
+from usaspending_api.common.experimental_api_flags import is_experimental_elasticsearch_api
+
 logger = logging.getLogger("console")
 
 
 class CustomCacheResponse(CacheResponse):
     def process_cache_response(self, view_instance, view_method, request, args, kwargs):
+        if is_experimental_elasticsearch_api(request):
+            # bypass cache altogether
+            response = view_method(view_instance, request, *args, **kwargs)
+            response = view_instance.finalize_response(request, response, *args, **kwargs)
+            return response
         key = self.calculate_key(
             view_instance=view_instance, view_method=view_method, request=request, args=args, kwargs=kwargs
         )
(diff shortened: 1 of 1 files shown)


usaspending_api/common/cache_decorator.py
```diff
@@ -2,11 +2,18 @@
...
+
 logger = logging.getLogger("console")
 
 
 class CustomCacheResponse(CacheResponse):
     def process_cache_response(self, view_instance, view_method, request, args, kwargs):
+        if is_experimental_elasticsearch_api(request):
+            # bypass cache altogether
+            response = view_method(view_instance, request, *args, **kwargs)
+            response = view_instance.finalize_response(request, response, *args, **kwargs)
```
BrianZito: Are we sure we don't need to set `response["Cache-Trace"]` to something here? It's set, but not used, in the other code, so maybe it's being requested elsewhere in the program?

emily-brents: What would you suggest we change it to? My understanding was that the "Cache-Trace" header is only set if the endpoint is actually interacting with the cache, which in this case it isn't.

BrianZito: "no-cache" seems reasonable to me...

BrianZito: That's what it does below when it checks but doesn't find anything

kwhickey: @BrianZito But in the case where there is no cache decorator on an endpoint, this code isn't even encountered, and that header will never be set. 

That's the scenario this is intended to imitate:
- No even bothering with the cache

Rather than...

- i looked in the cache and didn't find anything

BrianZito: I'm happy to see the code merged either way as  long as our  code  is never looking for the "header" (it's not a header, is it? It looks like a part of the response body).

tony-sappe: Cache-Trace is a header

Added to help devs and testers troubleshoot caching issues

PR #2366, opened 2020-03-26
[DEV-4736] Use correct column for FPDS pop_country_name

sethstoudenmier:
**Note:**
Instead of crowding our open PRs I will open up a PR into QAT once this is approved and before merging.

**Description:**
Change the column used when referencing `pop_country_name` in matviews to use the correct column. The old column is not populated across any of FPDS records. 

Verified that after recreating matviews locally performing a Custom Award Download searching for `All Foreign Countries` now returns the same result count as searching for `All` countries and then filtering down to non-USA transactions.

**Technical details:**
Fix was three parts:
* Replaced all references in matviews of `transaction_fpds.place_of_perform_country_n` to the correct column of `transaction_fpds.place_of_perf_country_desc`.
* Updated test cases that referenced `place_of_perform_country_n` as part of TransactionFPDS model to now use `place_of_perf_country_desc`.
* Updates to load_base and threaded_data_loader to use a different set of mappings since the key of `primary_place_of_performance_country_name` has a different value for FABS and FPDS.

    - [ ] Backend
8. [X] Jira Ticket [DEV-4736](https://federal-spending-transparency.atlassian.net/browse/DEV-4736):

```
```


usaspending_api/common/threaded_data_loader.py
```diff
@@ -252,7 +253,10 @@ def load_data_into_model(self, model_instance, fields, field_map, value_map, row
             # If our field is the 'long form' field, we need to get what it maps to
             # in the data so we can map the data properly
             elif model_field in LONG_TO_TERSE_LABELS:
-                source_field = LONG_TO_TERSE_LABELS[model_field]
+                if type(mod) == TransactionFPDS:
```
tony-sappe: I don't understand why this mapping is necessary. Which table has `primary_place_of_performance_country_name`?

sethstoudenmier: I couldn't find where this would be hit but put it in place in case this is ever reached where the model is TransactionFPDS just in case.

tony-sappe: If I understand it correctly, these mappers are for data loads (specifically from Broker -> USAspending) and I don't see that field name in either FPDS DB  table.

sethstoudenmier: Yeh, I couldn't find it anywhere either. Will remove changes around this.

PR #2405, opened 2020-04-10
[DEV-4506] Obtain recipient_id when record doesn't contain DUNS

tony-sappe:
**Description:**
Original SQL only worked when both the award's latest transaction and the recipient lookup record contained a DUNS. Now, in the event of no DUNS in the transaction/award the lookup will be performed using the name and null DUNS

**Technical details:**
N/A

    - [x] Backend
8. [x] Jira Ticket [DEV-4506](https://federal-spending-transparency.atlassian.net/browse/DEV-4506):

```
Changes simply increase the results which contain a populated `recipient_id` value
```

### DIFF
--- a/usaspending_api/common/recipient_lookups.py
+++ b/usaspending_api/common/recipient_lookups.py
@@ -151,12 +151,21 @@ def annotate_recipient_id(field_name, queryset):
                 recipient_profile rp
                 inner join recipient_lookup rl on rl.recipient_hash = rp.recipient_hash
             where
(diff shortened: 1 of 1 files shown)


tony-sappe:
> This should work (and appears to when I test it locally), but I have some reservations matching on name alone when we have names like "", "1", ".", "X", etc in the database. There are many such low-quality names.

I don't disagree with the sentiment. However, if a recipient profile page exists for that "recipient" then the hyperlink should be provided and functional.

PR #2421, opened 2020-04-20
[DEV-4375] Historical CARS File A loader

kbard:
**Description:**

Quick and dirty loader (and supporting plumbing) for historical CARS File A data.

1. [x] Unit & integration tests currently not implemented
    - [x] Backend
5. [x] Front end unaffected
8. [x] Jira Ticket [DEV-4375](https://federal-spending-transparency.atlassian.net/browse/DEV-4375):


usaspending_api/common/retrieve_file_from_uri.py
```diff
@@ -57,6 +57,13 @@ def copy(self, dest_file_path):
         else:
             raise NotImplementedError("No handler for scheme: {}!".format(self.parsed_url_obj.scheme))
 
+    def copy_to_temporary_file(self):
+        """Sometimes it is super helpful to just have a nice, concrete, local file to work with."""
+        with tempfile.NamedTemporaryFile() as tf:
```
tony-sappe, line 62: Does this reliably work?

My understanding of the documentation is after the file is closed (which should occur when exiting the context manager) the file isn't guaranteed to persist. It is up to the garbage collection.

One theory is that `copy_to_temporary_file()` effectively only generates a temp file name (since the temp file is removed immediately), then `self.copy()` re-creates the file when it copies the data into it. If this is the case, does the file always persist? I see os.remove below, so I'm assuming this is what is occurring.

Is that desirable behavior to implement here?

usaspending_api/references/management/commands/load_historical_appropriation_account_balances.py
```diff
@@ -0,0 +1,151 @@
...
+                        headers = self._get_headers(file_path)
+                    with Timer(f"Create temporary table {table_name}"):
+                        self._create_temporary_table(table_name, headers)
+                    with Timer(f"Import {file_name}"):
+                        self._import_file(file_path, table_name)
+                    os.remove(file_path)
+
+                destination_table_name = HistoricalAppropriationAccountBalances._meta.db_table
+                with Timer(f"Empty {destination_table_name}"):
+                    cursor.execute(f"delete from {destination_table_name}")
```
tony-sappe, line 130: since this isn't wrapped in a transaction block and it always performs a full reload, I do prefer `TRUNCATE TABLE <table> RESTART IDENTITY`

usaspending_api/common/retrieve_file_from_uri.py
kbard: This is a very good question and I will have to get back to you... probably not today.

usaspending_api/references/management/commands/load_historical_appropriation_account_balances.py
kbard: It is wrapped in a transaction.  It's using psycopg2 which transactions by default.  Django doesn't transaction by default.

tony-sappe: Interesting.

Even more interesting when I recall Django's connections are on top of psycopg2

kbard: /shrug

It's been Django's position for a while.

https://docs.djangoproject.com/en/2.2/topics/db/transactions/#autocommit-details

usaspending_api/common/retrieve_file_from_uri.py
kbard: There's certainly no shortage of ways to create temporary files in Python.  The goal of this snippet is to get me a file path that is guaranteed to be unique and creatable.  I thought actually creating the file would be a great way to do that, but then I want the file gone because I intend to use it.

A quick walk through `NamedTemporaryFile` seems to indicate that it's a hard delete as soon as the context manager `__exit__` is called.  I also created a little test app that checked for the file immediately after the context manager ends and it was gone.  I don't think it's garbage collected.  Where did you see that?

I'm not necessarily married to this solution.  It was just an example of one way to do it that I had handy.  If you have a more preferred method, I'm all ears.

tony-sappe: > I don't think it's garbage collected. Where did you see that?

Something in the stdlib documentation page. Might have misread it.

PR #2425, opened 2020-04-22
[DEV-4873] Fix S3 file datetime filtering for FPDS loader

kbard:
**Description:**

Previously, the FPDS loader was obtaining the delete file date from the date contained in the delete file's name.  This was only a date which meant the datetime was treated as though the delete occurred at midnight.  The last run timestamp is a full timestamp which means it captures the time as well as the date.  Since the timestamp is usually a few minutes after midnight, this means files created on the same day as the last run were not being picked up.

That's confusing.  Perhaps a real life example will help:
```
2020-04-21 00:05:46  Last run datetime
2020-04-21 14:57:30  Actual file create date  (04-21-2020_delete_records_IDV_1587481048.csv)                                        
2020-04-21 00:00:00  The date the FPDS loader THINKS the delete file was created since
                     time is being truncated
```
Since `2020-04-21 00:00:00` < `2020-04-21 00:05:46`, the file is ignored.

The fix we went with for this HOTFIX was to use the actual `last_modified` date returned from S3.  We may want to switch to using the epoch in the file name for the permanent fix since that won't change if the files are copied around or otherwise touched such that their `last_modifed` date is altered.

### DIFF
--- a/usaspending_api/transactions/transaction_delete_journal_helpers.py
+++ b/usaspending_api/transactions/transaction_delete_journal_helpers.py
@@ -1,15 +1,14 @@
 import csv
 import logging
 
-
 from collections import defaultdict
 from datetime import datetime
 from typing import Optional, List
-
 from usaspending_api import settings
-from usaspending_api.common.helpers.date_helper import datetime_is_ge, datetime_is_lt
+from usaspending_api.common.helpers.date_helper import cast_datetime_to_naive, datetime_is_ge, datetime_is_lt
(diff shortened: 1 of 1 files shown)


BrianZito:
This code is also run by the FABS loader. Was that smoke-tested?

kbard:
> This code is also run by the FABS loader. Was that smoke-tested?

Yes, but know that this code is exceedingly difficult to test properly.  We should keep very close tabs on both FABS and FPDS for a few days.

PR #2428, opened 2020-04-23
New TAS Filter

BrianZito:
**Description:**
While legacy functionality of `tas_codes` is still supported, this splits the functionality of TAS into `tas_codes`, which follows a similar hierarchical model to NAICS, and `treasury_account_components` serving the tab of that name.

**Technical details:**
The old `treasury_accounts` column was  split  into `tas_paths` and `tas_components` in both the award  and transaction index. Both columns are arrays of strings instead of arrays of JSON objects, in order to better support Elasticsearch's querystring syntax. `tas_paths` now also contains the codes of the toptier agency and Federal Account actually linked to in the database, which due to bucketing logic is not always the one implied in the TAS itself.
Additional details available upon request.

    - [x] Backend
    - [x] Frontend
 * Pre deploy: (https://federal-spending-transparency.atlassian.net/browse/DEV-4971)
 * Post deploy: (https://federal-spending-transparency.atlassian.net/browse/DEV-4984)
8. [x] Jira Ticket [DEV-4268](https://federal-spending-transparency.atlassian.net/browse/DEV-4268):

```
```


usaspending_api/api_contracts/search_filters.md
```diff
@@ -379,6 +379,35 @@ Request parameter description:
...
+
+**Description:** There are two filters for TAS: tas_codes and treasury_account_components. Unlike other filters, these two work on OR logic with each other (but AND logic with all other filters)
+
+**Example Request:**
+```
+{
+    "tas_codes": {
+        "require": [["091"]],
+        "exclude": [["091","091-0800"]]
+    }
```
BrianZito, line 392:
```suggestion
    },
```

usaspending_api/search/elasticsearch/filters/tas.py
```diff
@@ -0,0 +1,96 @@
...
+
+class TASNode(Node):
+    def _basic_search_unit(self):
+        return search_regex_of(self.code)
+
+    def clone(self, code, positive, positive_naics, negative_naics):
+        return TASNode(code, positive, positive_naics, negative_naics)
+
+
+class _TreasuryAccounts(_Filter):
```
BrianZito, line 74: not private anymore

usaspending_api/search/helpers/matview_filter_helpers.py
```diff
@@ -36,6 +37,8 @@ def build_tas_codes_filter(queryset, tas_filters):
         - Grabbing the list of all treasury_account_identifiers that match the filters provided.
         - Assembling an integer array overlap (&&) query to search for those integers.
     """
+    if isinstance(tas_filters, dict):
+        raise UnprocessableEntityException("Heirarchical TAS filters not supported in Postgres implementation")
```
BrianZito, line 41: add test for this

usaspending_api/api_contracts/contracts/v2/search/spending_by_award_count.md
```diff
@@ -120,6 +120,10 @@ This endpoint takes award filters, and returns the number of awards in each awar
 + `exclude`: `3333` (optional, array[string])
 
 ### TASCodeObject (object)
++ `require`: [[`091`]] (optional, array[array[string]])
++ `exclude`: [[`091`, `091-0800`]] (optional, array[array[string]])
+
+### TreasuryAccountComponentsObject (object)
```
kbard, line 126: Does the `AdvancedFilterObject` object need to include these changes?

usaspending_api/api_contracts/contracts/v2/search/spending_over_time.md
```diff
@@ -138,6 +138,10 @@ This endpoint returns a list of aggregated award amounts grouped by time period
 + `exclude`: `3333` (optional, array[string])
 
 ### TASCodeObject (object)
++ `require`: [[`091`]] (optional, array[array[string]])
++ `exclude`: [[`091`, `091-0800`]] (optional, array[array[string]])
+
+### TreasuryAccountComponentsObject (object)
```
kbard, line 144: Does the `AdvancedFilterObject` object need to include these changes?

usaspending_api/api_contracts/contracts/v2/search/spending_by_geography.md
```diff
@@ -169,6 +169,10 @@ This endpoint takes award filters, and returns aggregated obligation amounts in
 + `exclude` (optional, array[string])
 
 ### TASCodeObject (object)
++ `require`: [[`091`]] (optional, array[array[string]])
++ `exclude`: [[`091`, `091-0800`]] (optional, array[array[string]])
+
+### TreasuryAccountComponentsObject (object)
```
kbard, line 175: Does the `AdvancedFilterObject` object need to include these changes?

usaspending_api/api_contracts/contracts/v2/search/spending_by_category.md
```diff
@@ -151,6 +151,10 @@ This endpoint returns a list of the top results of specific categories sorted by
 + `exclude` (optional, array[string])
 
 ### TASCodeObject (object)
++ `require`: [[`091`]] (optional, array[array[string]])
++ `exclude`: [[`091`, `091-0800`]] (optional, array[array[string]])
+
+### TreasuryAccountComponentsObject (object)
```
kbard, line 157: Does the `AdvancedFilterObject` object need to include these changes?

usaspending_api/search/elasticsearch/filters/tas.py
```diff
@@ -0,0 +1,103 @@
...
+        + code_lookup["ata"]
+        + code_lookup["sub"]
+        + code_lookup["bpoa"]
+        + code_lookup["epoa"]
+        + code_lookup["a"]
+    )
+
+    # TODO: move this to a Tinyshield filter
+    if not re.match(r"^(\d|\w|-|\*|=)+$", search_regex):
+        print(search_regex)
```
tony-sappe, line 67: Print statement

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
```diff
@@ -232,6 +233,10 @@ List of table columns
 + `exclude`: `3333` (optional, array[string])
 
 ### TASCodeObject (object)
++ `require`: [[`091`]] (optional, array[array[string]])
```
tony-sappe, line 236: `array[array[string]]` ?

BrianZito: Is that inappropriate syntax? The endpoint DOES use a 2D array.

usaspending_api/api_contracts/contracts/v2/search/spending_by_award_count.md
tony-sappe: Yes. @BrianZito please add `TreasuryAccountComponentsObject` to the object

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
tony-sappe: Oh interesting. Can you share a request JSON example?

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -94,15 +94,15 @@ LEFT JOIN (
 LEFT JOIN (
   SELECT
     faba.award_id,
-    JSONB_AGG(
-      DISTINCT JSONB_BUILD_OBJECT(
```
tony-sappe, line 98: The distinct was important before. Can you confirm the new SQL doesn't contain duplicate TAS in the array?

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
BrianZito:
```
"tas_codes": {
        	"require": [["091"]],
        	"exclude": [["091","091-0800"]]
        },
```

BrianZito: Because the relationship between Agency, Federal Account, and TAS are arbitrary (at least to the degree that you can't know for sure what agency a federal account belongs to from the federal account code), the API needs the full ancestor array in order to link search nodes for the hierarchical logic.

usaspending_api/database_scripts/etl/award_delta_view.sql
BrianZito: The new SQL did produce duplicates, fixed with latest commit

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
tony-sappe: Can you show a useful example of leveraging the 2D array? The copy-paste from the API Contract isn't showing why a 2d array is necessary.

BrianZito:
```
"tas_codes": {
        	"require": [["011"]],
        	"exclude": [["011","011-1082","097-011-2007/2012-1082-000"]]
        },
```
would be an example of where it is nessicary. For whatever reason, the tas loader's rules decided that this TAS was under agency 011, NOT agency 097 like the ata in  the TAS code might imply. Without the path provided (which  is always provided by the filter endpoints themselves, so there's no issue of the javascript  code not having the path), the search logic would have no way to know if  "097-011-2007/2012-1082-000" is a descendent of "011". Even if there was a way,  you are now putting context-specific logic in either the frontend or the search filter, introducing risk of desync with the sql performing the data load.

BrianZito:
Please keep reviewing the rest of the PR, but I'm pretty sure I've found a bug in the logic. Stand by.

usaspending_api/common/validator/award_filter.py
```diff
@@ -12,6 +13,16 @@
     "listed on https://api.usaspending.gov/docs/endpoints."
 )
 
+OLD_TAS_FILTER = {
```
tony-sappe, line 16: Usually using "old" means it is around for the interim and will be removed once the old functionality reaches production (or similar scenario)

Is this in the process of being replaced and will be removed later?

usaspending_api/common/validator/award_filter.py
```diff
@@ -4,6 +4,7 @@
 
 from usaspending_api.awards.v2.lookups.lookups import award_type_mapping
 from usaspending_api.common.validator.helpers import TINY_SHIELD_SEPARATOR
+from usaspending_api.search.elasticsearch.filters.tas import TasCodes, TreasuryAccounts
```
tony-sappe, line 7: I understand why these imports into the global TinyShield rule models were done. Still, it seems to be a slippery slope on the past of cluttering this file with imports when the goal of the file is simply defining a few dictionaries. Thoughts?

BrianZito: My first thought would be that the danger of magic strings dwarfs any concern from having too many import statements at the top. Obviously I can't speak to whatever happened in the past, but if something is in two places I think another import is a small price to pay to keep code DRY.

tony-sappe: Oops, ~past~ => `path`.

My typo misdirected the reader to look back when I'm thinking forward. To be logically consistent, all of the new Elasticsearch filter objects would be imported into this file to reuse the underscore_name.

usaspending_api/search/tests/integration/hierarchical_filters/test_tas_filter_irregular_formats.py
```diff
@@ -0,0 +1,127 @@
...
+        client,
+        [
+            {
+                "a": "R"  # "R" doesn't represent anything special here, it just makes sure the code is ok with any capital letter
+            }
+        ],
+        None,
+    )
+
+    print(resp.json())
```
tony-sappe, line 116: print()

usaspending_api/common/validator/award_filter.py
BrianZito: That sounds like a good thing.

usaspending_api/api_contracts/search_filters.md
```diff
@@ -379,6 +379,35 @@ Request parameter description:
...
+    "treasury_account_components": [{"aid":"005","bpoa":"2015","epoa":"2015","main":"0107","sub":"000"}]
+}
+```
+
+`tas_codes` (TASFilterObject) : Two nullable arrays of arrays of strings: `require` and `exclude`. 
+* When `require` is provided, search will only return results that have a TAS code that is a descendant of one of the paths from the require list. 
+* When `exclude` is provided,  search will only return results that do NOT have a TAS code that is a descendant of one of the paths from the exclude list. 
+* If an element matches both lists, the more specific rule (longer array) supercedes.
+
+`treasury_account_components`: List of objects. Each object can have any of the following keys, and will filter results down to those that match the value for each key provided:
```
tony-sappe, line 402: `tas_codes` references the ObjectName, while `treasury_account_components` does not. Consistency is nice

usaspending_api/tests/test_award_index_elasticsearch.py
```diff
@@ -102,11 +102,6 @@ def test_date_range(award_data_fixture, elasticsearch_award_index):
 
 def test_tas(award_data_fixture, elasticsearch_award_index):
     elasticsearch_award_index.update_index()
-    search_regex = (
```
tony-sappe, line 105: Is this regex in line with the actual code implementation?

usaspending_api/search/elasticsearch/filters/tas.py
```diff
@@ -0,0 +1,110 @@
...
+                "main": v.get("main", ".*"),
+                "ata": v.get("ata", ".*"),
+                "sub": v.get("sub", ".*"),
+                "bpoa": v.get("bpoa", ".*"),
+                "epoa": v.get("epoa", ".*"),
+                "a": v.get("a", ".*"),
+            }
+
+            search_regex = f".*aid={code_lookup['aid']}main={code_lookup['main']}ata={code_lookup['ata']}sub={code_lookup['sub']}bpoa={code_lookup['bpoa']}epoa={code_lookup['epoa']}a={code_lookup['a']}"
+            print(search_regex)
```
tony-sappe, line 106: print

usaspending_api/search/elasticsearch/filters/tas.py
```diff
@@ -0,0 +1,110 @@
...
+                "aid": v.get("aid", ".*"),
+                "main": v.get("main", ".*"),
+                "ata": v.get("ata", ".*"),
+                "sub": v.get("sub", ".*"),
+                "bpoa": v.get("bpoa", ".*"),
+                "epoa": v.get("epoa", ".*"),
+                "a": v.get("a", ".*"),
+            }
+
+            search_regex = f".*aid={code_lookup['aid']}main={code_lookup['main']}ata={code_lookup['ata']}sub={code_lookup['sub']}bpoa={code_lookup['bpoa']}epoa={code_lookup['epoa']}a={code_lookup['a']}"
```
tony-sappe, line 105: I'm curious about performance after adding a leading wildcard to the regex. @sethstoudenmier @emily-brents is this OK from a performance perspective?

usaspending_api/search/elasticsearch/filters/tas.py
```diff
@@ -0,0 +1,110 @@
...
+        + code_lookup["bpoa"]
+        + code_lookup["epoa"]
+        + code_lookup["a"]
+    )
+
+    # TODO: move this to a Tinyshield filter
+    if not re.match(r"^(\d|\w|-|\*|=)+$", search_regex):
+        raise UnprocessableEntityException(f"Unable to parse TAS filter")
+
+    print(search_regex)
```
tony-sappe, line 75: print

usaspending_api/search/tests/integration/hierarchical_filters/test_tas_filter_basic_cases.py
```diff
@@ -0,0 +1,165 @@
...
+    _setup_es(client, monkeypatch, elasticsearch_award_index)
+    resp = query_by_tas(client, {"require": [_agency_path(BASIC_TAS) + [TAS_STRINGS[ATA_TAS]]]})
+
+    assert resp.json()["results"] == []
+
+
+@pytest.mark.django_db
+def test_match_from_ata_tas(client, monkeypatch, elasticsearch_award_index, award_with_ata_tas):
+    _setup_es(client, monkeypatch, elasticsearch_award_index)
+    print(_tas_path(3))
```
tony-sappe, line 60: print

usaspending_api/tests/test_award_index_elasticsearch.py
BrianZito: It is not, which is why it was deleted. The new regex is on line 115, which still doesn't match the code, but to my understanding this is testing the index, not the filter. If we don't think it's important to  test the index by itself, I would rather just delete this test suite and trust the filter tests to test filters.

usaspending_api/search/elasticsearch/filters/tas.py
emily-brents: Generally no, we want to avoid leading wildcards specifically because they cause performance hits - from the Elasticsearch docs: 

> Avoid beginning patterns with * or ?. This can increase the iterations needed to find matching terms and slow search performance.

BrianZito: Would it be alright to just add another column to the index? There is no way to order the tas codes to never start with a wildcard, but we could always keep adding different strings to regex on.

tony-sappe: If I understand the searches, FA and TAS should be two separate fields in the index.

BrianZito: So there are two searches: the component search (which this line of code is doing), and the search by path (in the class above in this file). In the component search, Federal Account does not matter; only TAS components are searched on. In the search by path, the agency, federal account, and TAS all matter. Right now, this index column `treasury_accounts` holds data for all three, in order  to support both searches with one index. Right now I'm experimenting locally with a second index with just TAS data to support the component search without needing to start with a wildcard.

BrianZito: Keep in mind that Federal Account is a property of a TAS, not an Award or a Transaction.

tony-sappe: > Keep in mind that Federal Account is a property of a TAS, not an Award or a Transaction.

But the data are pulled from different tables/columns because they don't match perfectly, right? Otherwise the new "FA" fields wouldn't be needed since they would be duplicating pre-existing TAS fields.

BrianZito: Yes, they are a join from the tas table to fa and agency. What I was trying to say is that they aren't one-to-one with awards or transactions, so even if it was a new column in the index, it would still need to be associated with tas.

usaspending_api/search/tests/integration/hierarchical_filters/test_tas_filter_irregular_formats.py
```diff
@@ -0,0 +1,101 @@
...
+
+    assert resp.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY, "Failed to return 422 Response"
+
+
+@pytest.mark.django_db
+def test_treasury_account_component_filter_appropriate_characters(
+    client, monkeypatch, elasticsearch_award_index, award_with_tas
+):
+    _setup_es(client, monkeypatch, elasticsearch_award_index)
+    resp = query_by_treasury_account_components(
```
tony-sappe, line 81:
```suggestion
     # "R" doesn't represent anything special here, it just makes sure the code is ok with any capital letter
     resp = query_by_treasury_account_components(client, [{"a": "R" }], None)
```

usaspending_api/search/tests/integration/hierarchical_filters/es_search_test_helpers.py
```diff
@@ -0,0 +1,60 @@
+import json
+from django.conf import settings
+
+from usaspending_api.search.elasticsearch.filters.tas import TasCodes, TreasuryAccounts
+from usaspending_api.common.experimental_api_flags import EXPERIMENTAL_API_HEADER, ELASTICSEARCH_HEADER_VALUE
+
+
+def _setup_es(client, monkeypatch, elasticsearch_award_index):
```
sethstoudenmier, line 8: If you have time would be nice to replace this and its usage with the `setup_elasticsearch_test()` function defined under `usaspending_api/search/tests/data/utilities.py`. It will setup elasticsearch according to the type of index you pass it.

BrianZito: Just implemented a simple refactor. When I circle back to write more tests for the transaction index, I'll imrpove this further.

PR #2431, opened 2020-04-23
[DEV-4943] Use file epoch instead of S3 last modified for filtering delete files

kbard:
**Description:**

Last sprint we introduced a hotfix to use S3 last modified date when filtering FPDS/FABS delete files.  This will work right up until we move the files somewhere else.  Thankfully, the original developers added the file creation time epoch to the file name so we are able to persistently determine when those files were created.

This pull request converts the S3 code to use that epoch instead of the last modified date.  It also cleans up some of the logging we added when trying to track down the problem.  Finally it attempts to simplify a couple of the lower level functions.

1. [x] Unit & integration tests unaffected
    - [ ] Backend
5. [x] Front end unaffected
8. [x] Jira Ticket [DEV-4943](https://federal-spending-transparency.atlassian.net/browse/DEV-4943):


kbard:
> Looks good. Did you confirm both epochs are created from UTC?

USAspending uses time.time() which claims to be UTC.  Broker does their own math for some reason, but is also using naive UTC dates/times.

PR #2434, opened 2020-04-24
Ftr/dev 4526 psc filter tree

BrianZito:
**Description:**
Sets up a filter tree for PSC following the same model as TAS

**Technical details:**

    - [x] Backend
    - [ ] Frontend 
4. [ ] Matview impact assessment completed (N/A)
7. [ ] Appropriate Operations ticket(s) created (N/A)
8. [x] Jira Ticket [DEV-4904](https://federal-spending-transparency.atlassian.net/browse/DEV-4904):
    - [x] Performance evaluation of affected (API | Script | Download) (N/A)

```
```


usaspending_api/api_contracts/contracts/v2/references/filter_tree/psc.md
```diff
@@ -0,0 +1,71 @@
...
+       
+## Data Structures
+
+### FilterTreeNode (object)
+
++ `id` (required, string)
++ `description` (required, string)
++ `ancestors` (required, array[string])
++ `count` (required, number)
++ `children` (required, array[FilterTreeNode], nullable)
```
tony-sappe, line 71: Please don't put __init__.py files in the contract directories.

usaspending_api/api_contracts/contracts/v2/references/filter_tree/psc.md
```diff
@@ -0,0 +1,71 @@
...
+                "$schema": "http://json-schema.org/draft-04/schema#",
+                "type": "string"
+            }
+
++ Response 200 (application/json)
+    + Attributes (object)
+        + `results` (required, array[FilterTreeNode], fixed-type)
+    + Body
+    
+            {
```
tony-sappe, line 37: Please use 4-space indentation levels for all JSON in the contracts (pretty-print)

usaspending_api/api_contracts/contracts/v2/references/filter_tree/psc.md
```diff
@@ -0,0 +1,71 @@
...
+These endpoints are used to power USAspending.gov's PSC search component on the advanced search page.
+The response is a forest of filter search nodes, which despite having a unified structure represent different
+database fields based on depth in the tree.
+
+## Toptier Search [GET /api/v2/references/filter_tree/psc/{?depth}{?filter}]
+
+Returns the basic groupings of Product Service Codes
++ Request (application/json)
+    + Parameters
+        + `depth` (optional, enum[number])
```
tony-sappe, line 15: Add the default value under this parameter. Perhaps this snippet will work?

```
+ Default
    0
```

usaspending_api/references/v2/views/filter_tree/psc_filter_tree.py
```diff
@@ -0,0 +1,42 @@
...
+        elif len(tiered_keys) == 1:
+            return self._psc_from_group(tiered_keys[0])
+        else:
+            return self._psc_from_parent(tiered_keys[-1])
+
+    def _toptier_search(self):
+        return PSC_GROUPS.keys()
+
+    def _psc_from_group(self, group):
+        if group in PSC_GROUPS:
```
tony-sappe, line 20: Does the code allow non-keys to be provided to this method?

usaspending_api/api_contracts/contracts/v2/references/filter_tree/psc/group/group.md
```diff
@@ -0,0 +1,70 @@
...
+These endpoints are used to power USAspending.gov's PSC search component on the advanced search page.
+The response is a forest of filter search nodes, which despite having a unified structure represent different
+database fields based on depth in the tree.
+
+## Toptier Search [GET /api/v2/references/filter_tree/psc/{group}/{?depth}{?filter}]
+
+Returns PSCs under the provided group
++ Request (application/json)
+    + Parameters
+        + `group`: `Product` (required, string)
```
tony-sappe, line 15: Does `group` only support a set list of values? If yes please turn into an enum[string]

usaspending_api/references/v2/views/filter_tree/psc_filter_tree.py
BrianZito: Yes, just like it allows a user to search for children of a non existant PSC, NAICS or Federal Account. The most Restful thing might be to do a search a level above and return 404 on that condition, but I think the case could be made either way here, and that strategy seems to double computation time without too much benefit.

usaspending_api/api_docs/markdown/endpoints.md
```diff
@@ -93,7 +93,11 @@ The currently available endpoints are listed in the following table.
 |[/api/v2/references/data_dictionary/](/api/v2/references/data_dictionary/)|GET| Returns a JSON structure of the Schema team's Rosetta Crosswalk Data Dictionary|
 |[/api/v2/references/filter_tree/tas/](/api/v2/references/filter_tree/tas/)|GET| Returns a list of toptier agencies that have at least one TAS affiliated with them|
 |[/api/v2/references/filter_tree/tas/<AGENCY\>/](/api/v2/references/filter_tree/tas/020/)|GET| Returns a list of federal accounts associated with the specified agency|
-|[/api/v2/references/filter_tree/tas/<AGENCY\>/<FEDERAL_ACCOUNT\>](/api/v2/references/filter_tree/tas/020/020-0550/)|GET| Returns a list of Treasury Account Symbols associated with the specified federal account|
+|[/api/v2/references/filter_tree/tas/<AGENCY\>/<FEDERAL_ACCOUNT\>/](/api/v2/references/filter_tree/tas/020/020-0550/)|GET| Returns a list of Treasury Account Symbols associated with the specified federal account|
+|[/api/v2/references/filter_tree/psc/](/api/v2/references/filter_tree/psc/)|GET| Returns a list of PSC groupings|
```
tony-sappe, line 97: isn't "P" before "T" ?

usaspending_api/references/v2/views/filter_tree/psc_filter_tree.py
tony-sappe: Perhaps I didn't ask clearly. It appears the code is checking the parameter `group` is in the hard coded dict keys above:

```python
if group in PSC_GROUPS:
```
=>
```python
if group in("Research and Development", "Service", "Product")
```

I haven't dived into the code, but that seems odd to need this check this far down in the business logic. Shouldn't the group already be vetted to ensure it is an acceptable value?

BrianZito: For a GET request there is no such thing as an unacceptable value, only an unacceptable format. Just like you can search for the snapshot of a non-existant FA (https://api.usaspending.gov/api/v2/federal_accounts/8899/fiscal_year_snapshot/2017/) or the Federal Accounts under a non-existant agency (https://dev-api.usaspending.gov/api/v2/references/filter_tree/tas/998/), there's nothing wrong with looking for PSC codes under a category that doesn't exist. There's an argument to be made that all of these should return 404 instead of an empty list, but that would be a change to make  for the filter trees as a whole, not special behavior that only happens at one level of one tree.
  The check happens here because it's the first area of code dedicated to this specific level of logic for this tree. And even when the check happens, it's to avoid a null pointer on the table, not to validate the request; the endpoint still returns an empty list if the group isn't found, there's just no appropriate regex to pass to the next function. Now that I think about it, I might just do a get_or_default call and smooth out this code a little.

usaspending_api/references/tests/integration/filter_tree/psc/psc_data_fixtures.py
```diff
@@ -0,0 +1,138 @@
+from model_mommy import mommy
+import pytest
+
+
+@pytest.fixture
+def no_data(db):
```
tony-sappe, line 6: How is this different than simply providing `db` as a test function parameter?

BrianZito: This ultimately comes down to "it made sense at the time". I have no opinions one way or the other on how to define this.

usaspending_api/api_contracts/contracts/v2/references/filter_tree/psc/group.md
```diff
@@ -0,0 +1,73 @@
...
+The response is a forest of search filter nodes, which despite having a unified structure represent different
+database fields based on depth in the tree.
+
+## Toptier Search [GET /api/v2/references/filter_tree/psc/{group}/{?depth}{?filter}]
+
+Returns PSCs under the provided group
++ Request (application/json)
+    + Parameters
+        + `group`: `Product` (required, string)
+        Parent group of PSCs to return
```
tony-sappe, line 16: Since this parameter will allow any string and simply produce no results, can you list the three group strings which have the potential to return data in the contracts?

usaspending_api/api_contracts/contracts/v2/references/filter_tree/psc/group.md
```diff
@@ -0,0 +1,73 @@
...
+        + `depth` (optional, number)
+            Defines how many levels of descendants to return under each node. For example, depth=0 will 
+            return a flat array, while depth=2 will populate the children array of each top level node 
+            with that node's children and grandchildren. The actual depth of each tree may be less than 
+            the value of depth if returned nodes have no children. Negative values are treated as 
+            infinite, returning all descendants.  
+            + Default: 0
+        
+        + `filter` (optional, string) 
+            Restricts results to nodes with a `id` or `description` matching the filter string. If depth is
```
tony-sappe, line 26: What is the `id` field, is it the PSC?

usaspending_api/api_contracts/contracts/v2/references/filter_tree/psc/group/psc/psc.md
```diff
@@ -0,0 +1,77 @@
...
+       
+## Data Structures
+
+### FilterTreeNode (object)
+
++ `id` (required, string)
++ `description` (required, string)
++ `ancestors` (required, array[string])
++ `count` (required, number)
++ `children` (required, array[FilterTreeNode], nullable)
```
tony-sappe, line 77: probably should list `fixed-type`

usaspending_api/api_contracts/contracts/v2/references/filter_tree/psc/group.md
BrianZito: Unless it's the group name. Like in the TAS filter tree, the meaning of `id` and `description` are context-dependent, hence their more generic naming.

usaspending_api/references/v2/views/filter_tree/filter_tree.py
```diff
@@ -80,7 +81,7 @@ def raw_search(self, tiered_keys: list, filter_string: str) -> list:
         pass
 
     @abstractmethod
-    def unlinked_node_from_data(self, ancestors: list, data) -> UnlinkedNode:
+    def unlinked_node_from_data(self, ancestors: list, data: any) -> UnlinkedNode:
```
tony-sappe, line 84: This should be `Any` imported from the typing lib. `any` is a builtin function

PR #2436, opened 2020-04-27
WARMFIX [DEV-4963] Sanitize elasticsearch text input

emily-brents:
**Description:**
Adds sanitization for elasticsearch text input to advanced search filters.

**Technical details:**
Strips all lucene special characters from the text entered into the advanced search filters. This is slightly different from the keyword search method, which strips only specific characters as the keyword search functionality allows using certain characters.

Additionally, for whatever reason importing the sanitization function from elasticsearch_helpers.py would not work in query_with_filters.py so we had to redefine it there.

    - [x] Backend
8. [X] Jira Ticket [DEV-4963](https://federal-spending-transparency.atlassian.net/browse/DEV-4963):

```
No changes to matviews, no ops tickets needed
```

### DIFF
--- a/usaspending_api/search/v2/es_sanitization.py
+++ b/usaspending_api/search/v2/es_sanitization.py
@@ -1,9 +1,20 @@
 import logging
 import re
 
-from usaspending_api.search.v2.elasticsearch_helper import concat_if_array
-
 logger = logging.getLogger("console")
+
+
+def concat_if_array(data):
+    if isinstance(data, str):
+        return data
+    else:
+        if isinstance(data, list):
+            str_from_array = " ".join(data)
+            return str_from_array
+        else:
+            # This should never happen if TinyShield is functioning properly
+            logger.error("Keyword submitted was not a string or array")
+            return ""
 
 
 def es_sanitize(input_string):
--- a/usaspending_api/search/v2/es_sanitization.py
+++ b/usaspending_api/search/v2/es_sanitization.py
@@ -1,6 +1,8 @@
 import logging
 import re
 
+from usaspending_api.search.v2.elasticsearch_helper import concat_if_array
+
 logger = logging.getLogger("console")
 
 
@@ -12,3 +14,14 @@ def es_sanitize(input_string):
         msg = "Stripped characters from input string New: '{}' Original: '{}'"
         logger.info(msg.format(processed_string, input_string))
     return processed_string
+
+
+def es_minimal_sanitize(keyword):
+    keyword = concat_if_array(keyword)
+    """Remove Lucene special characters instead of escaping for now"""
+    processed_string = re.sub(r"[%{}/:!^\[\]]", "", keyword)
+    if len(processed_string) != len(keyword):
+        msg = "Stripped characters from ES keyword search string New: '{}' Original: '{}'"
+        logger.info(msg.format(processed_string, keyword))
+        keyword = processed_string
+    return keyword
--- a/usaspending_api/search/v2/es_sanitization.py
+++ b/usaspending_api/search/v2/es_sanitization.py
@@ -1,0 +1,14 @@
+import logging
+import re
+
+logger = logging.getLogger("console")
+
+
+def es_sanitize(input_string):
+    """ Escapes reserved elasticsearch characters and removes when necessary """
+
+    processed_string = re.sub(r'([-&!|{}()^~*?:\\/"+\[\]<>])', "", input_string)
+    if len(processed_string) != len(input_string):
+        msg = "Stripped characters from input string New: '{}' Original: '{}'"
+        logger.info(msg.format(processed_string, input_string))
+    return processed_string
--- a/usaspending_api/common/query_with_filters.py
+++ b/usaspending_api/common/query_with_filters.py
@@ -9,9 +9,18 @@ from elasticsearch_dsl import Q as ES_Q
 from usaspending_api.search.elasticsearch.filters.filter import _Filter, _QueryType
 from usaspending_api.search.elasticsearch.filters.naics import NaicsCodes
 from usaspending_api.common.exceptions import InvalidParameterException
-from usaspending_api.search.v2.elasticsearch_helper import es_sanitize
 
 logger = logging.getLogger(__name__)
+
+
+def es_sanitize(input_string):
+    """ Escapes reserved elasticsearch characters and removes when necessary """
+
+    processed_string = re.sub(r'([-&!|{}()^~*?:\\/"+\[\]<>])', "", input_string)
+    if len(processed_string) != len(input_string):
+        msg = "Stripped characters from input string New: '{}' Original: '{}'"
+        logger.info(msg.format(processed_string, input_string))
+    return processed_string
 
 
 class _Keywords(_Filter):
--- a/usaspending_api/common/query_with_filters.py
+++ b/usaspending_api/common/query_with_filters.py
@@ -1,5 +1,4 @@
 import logging
-import re
 
 from django.conf import settings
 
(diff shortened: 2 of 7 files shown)


emily-brents:
> I still see several filters that don't have sanitization. As for those that follow the standard list of strings format for filter_values, I would rather see the sanitization done in advance, so that the onus isn't on the implementer of each filter to remember this requirement. I wrote the `pre-sanitize` branch to demonstrate this. Didn't test it though.

Most of the advanced search filters do not allow text input, so it's redundant to sanitize them. Additionally, not all the filters use text values, so we can't just sanitize them all the same way without it causing errors

emily-brents:
> I still see several filters that don't have sanitization. As for those that follow the standard list of strings format for filter_values, I would rather see the sanitization done in advance, so that the onus isn't on the implementer of each filter to remember this requirement. I wrote the `pre-sanitize` branch to demonstrate this. Didn't test it though.

emily-brents:
> I still see several filters that don't have sanitization. As for those that follow the standard list of strings format for filter_values, I would rather see the sanitization done in advance, so that the onus isn't on the implementer of each filter to remember this requirement. I wrote the `pre-sanitize` branch to demonstrate this. Didn't test it though.

So sanitizing all the fields the same won't work, since not all the fields are plain text fields some are dicts, some are. Even if we did want to go through all the dictionaries to sanitize them, we wouldn't actually want to do that, since cleaning the text input requires removing characters that work in elasticsearch for non-text fields, so it would break the time period filter entirely. I can go through an add the sanitization to fields that won't break, but it would still be mostly redundant for any field that isn't direct text input, which is only Keyword, recipient, and award id.

tony-sappe:
> > I still see several filters that don't have sanitization. As for those that follow the standard list of strings format for filter_values, I would rather see the sanitization done in advance, so that the onus isn't on the implementer of each filter to remember this requirement. I wrote the `pre-sanitize` branch to demonstrate this. Didn't test it though.
> 
> So sanitizing all the fields the same won't work, since not all the fields are plain text fields some are dicts, some are. Even if we did want to go through all the dictionaries to sanitize them, we wouldn't actually want to do that, since cleaning the text input requires removing characters that work in elasticsearch for non-text fields, so it would break the time period filter entirely. I can go through an add the sanitization to fields that won't break, but it would still be mostly redundant for any field that isn't direct text input, which is only Keyword, recipient, and award id.

Put another way: the intention for this ticket is to sanitize user-input text fields on the website's Advanced Search page.

usaspending_api/search/v2/es_sanitization.py
```diff
@@ -0,0 +1,14 @@
+import logging
```
tony-sappe, line 1: If `es_sanitize` is being segregated, would it be agreeable to move both it and `es_minimal_sanitize` together in the `usaspending-api/usaspending_api/common/elasticsearch/` directory?

BrianZito:
> > > I still see several filters that don't have sanitization. As for those that follow the standard list of strings format for filter_values, I would rather see the sanitization done in advance, so that the onus isn't on the implementer of each filter to remember this requirement. I wrote the `pre-sanitize` branch to demonstrate this. Didn't test it though.
> > 
> > 
> > So sanitizing all the fields the same won't work, since not all the fields are plain text fields some are dicts, some are. Even if we did want to go through all the dictionaries to sanitize them, we wouldn't actually want to do that, since cleaning the text input requires removing characters that work in elasticsearch for non-text fields, so it would break the time period filter entirely. I can go through an add the sanitization to fields that won't break, but it would still be mostly redundant for any field that isn't direct text input, which is only Keyword, recipient, and award id.
> 
> Put another way: the intention for this ticket is to sanitize user-input text fields on the website's Advanced Search page.

Aren't all of these fields user-input by an API user? As long as non-text fields don't need to be sanitized this still shouldn't be a problem though.

usaspending_api/search/v2/es_sanitization.py
emily-brents: that's very agreeable

tony-sappe:
> > > > I still see several filters that don't have sanitization. As for those that follow the standard list of strings format for filter_values, I would rather see the sanitization done in advance, so that the onus isn't on the implementer of each filter to remember this requirement. I wrote the `pre-sanitize` branch to demonstrate this. Didn't test it though.
> > > 
> > > 
> > > So sanitizing all the fields the same won't work, since not all the fields are plain text fields some are dicts, some are. Even if we did want to go through all the dictionaries to sanitize them, we wouldn't actually want to do that, since cleaning the text input requires removing characters that work in elasticsearch for non-text fields, so it would break the time period filter entirely. I can go through an add the sanitization to fields that won't break, but it would still be mostly redundant for any field that isn't direct text input, which is only Keyword, recipient, and award id.
> > 
> > 
> > Put another way: the intention for this ticket is to sanitize user-input text fields on the website's Advanced Search page.
> 
> Aren't all of these fields user-input by an API user? As long as non-text fields don't need to be sanitized this still shouldn't be a problem though.

Technically yes, all values provided to the API are "user-input", which is why the previous comment was clear on the website user-provided text fields. The ticket was written to address specific errors noticed in the production log statements. It is a warm-fix to ensure it reaches production ahead of "flipping the switch" to Elasticsearch.

BrianZito:
> > > > > I still see several filters that don't have sanitization. As for those that follow the standard list of strings format for filter_values, I would rather see the sanitization done in advance, so that the onus isn't on the implementer of each filter to remember this requirement. I wrote the `pre-sanitize` branch to demonstrate this. Didn't test it though.
> > > > 
> > > > 
> > > > So sanitizing all the fields the same won't work, since not all the fields are plain text fields some are dicts, some are. Even if we did want to go through all the dictionaries to sanitize them, we wouldn't actually want to do that, since cleaning the text input requires removing characters that work in elasticsearch for non-text fields, so it would break the time period filter entirely. I can go through an add the sanitization to fields that won't break, but it would still be mostly redundant for any field that isn't direct text input, which is only Keyword, recipient, and award id.
> > > 
> > > 
> > > Put another way: the intention for this ticket is to sanitize user-input text fields on the website's Advanced Search page.
> > 
> > 
> > Aren't all of these fields user-input by an API user? As long as non-text fields don't need to be sanitized this still shouldn't be a problem though.
> 
> Technically yes, all values provided to the API are "user-input", which is why the previous comment was clear on the website user-provided text fields. The ticket was written to address specific errors noticed in the production log statements. It is a warm-fix to ensure it reaches production ahead of "flipping the switch" to Elasticsearch.

Is there a ticket for the API user-provided text fields that's going to be a warmfix? If we're looking at a vulnerability we need to patch it at both points.

tony-sappe:
> > > > > > I still see several filters that don't have sanitization. As for those that follow the standard list of strings format for filter_values, I would rather see the sanitization done in advance, so that the onus isn't on the implementer of each filter to remember this requirement. I wrote the `pre-sanitize` branch to demonstrate this. Didn't test it though.
> > > > > 
> > > > > 
> > > > > So sanitizing all the fields the same won't work, since not all the fields are plain text fields some are dicts, some are. Even if we did want to go through all the dictionaries to sanitize them, we wouldn't actually want to do that, since cleaning the text input requires removing characters that work in elasticsearch for non-text fields, so it would break the time period filter entirely. I can go through an add the sanitization to fields that won't break, but it would still be mostly redundant for any field that isn't direct text input, which is only Keyword, recipient, and award id.
> > > > 
> > > > 
> > > > Put another way: the intention for this ticket is to sanitize user-input text fields on the website's Advanced Search page.
> > > 
> > > 
> > > Aren't all of these fields user-input by an API user? As long as non-text fields don't need to be sanitized this still shouldn't be a problem though.
> > 
> > 
> > Technically yes, all values provided to the API are "user-input", which is why the previous comment was clear on the website user-provided text fields. The ticket was written to address specific errors noticed in the production log statements. It is a warm-fix to ensure it reaches production ahead of "flipping the switch" to Elasticsearch.
> 
> Is there a ticket for the API user-provided text fields that's going to be a warmfix? If we're looking at a vulnerability we need to patch it at both points.

No. This PR is the warmfix under [DEV-4963](https://federal-spending-transparency.atlassian.net/browse/DEV-4963). For several weeks 100% of Advanced Search visualization API requests were forked and sent to Elasticsearch. A high number of Elasticsearch query errors were directly due to the keyword filter text values.

Sanitizing all inputs is a good idea, that's where TinyShield comes into play. But this PR and the ticket is specifically to address the errors occurring now.

PR #2441, opened 2020-04-29
[DEV-4875] Convert old Elasticsearch code

emily-brents:
**Description:**
Moves the city search elasticsearch code from the basic elasticsearch package to the elasticsearch-dsl package to make it more consistent with rest of the site's elasticsearch implementation

**Technical details:**
Removes the es_client_query function, and swaps all usage of that function to use the `TransactionSearch` method which uses the elasticsearch-dsl package.

    - [x] Backend
8. [X] Jira Ticket [DEV-4875](https://federal-spending-transparency.atlassian.net/browse/DEV-4875):

```
```


usaspending_api/common/elasticsearch/client.py
```diff
@@ -54,29 +53,6 @@ def create_es_client() -> Elasticsearch:
...
-            logger.info(f"Failure using these: Index='{index}', Body={json.dumps(body)}")
-        if response is None:
-            logger.info(f"Failure using these: Body={json.dumps(search.to_dict())}")
-        else:
-            return response
-    logger.error(f"Unable to reach elasticsearch cluster. {retries} attempt(s) made")
-    return None
-
-
 def _es_search(
```
sethstoudenmier, line 56: I believe you can also delete this since it should only be used by the `es_client_query` method above that is getting deleted.

usaspending_api/references/v2/views/city.py
```diff
@@ -69,23 +69,18 @@ def prepare_search_terms(request_data):
...
-        "size": 0,
-        "query": {"bool": {"filter": {"bool": query_string}}},
-        "aggs": {
-            "cities": {"terms": {"field": "{}.keyword".format(return_fields[0]), "size": city_buckets}, "aggs": aggs}
-        },
-    }
+        aggs = {"field": return_fields[1], "size": 100}
+    agg_values = {"field": "{}.keyword".format(return_fields[0]), "size": city_buckets}
+    agg_key = A("terms", **agg_values)
+    query.aggs.bucket("cities", agg_key).pipeline("states", A("terms", **aggs))
```
sethstoudenmier, line 83: Can you change the second terms aggregation to be a `.bucket()` instead of `.pipeline()` since that is the type of aggregation that it is. They have the same effect in this case since the only difference in the DSL is what they return, but I can see this possibly causing confusion.

emily-brents: sure thing

usaspending_api/common/elasticsearch/client.py
emily-brents: that is correct

PR #2459, opened 2020-05-05
Updating Dockerfile: Referenced public rpm no longer exists.

bminahankcfrb:
rpm downloaded from download.postgresql.org no longer is available. referencing latest to increase our chances it will be available in the future.

**Description:**
When building the docker container, I could not build it because the rpm file on postgres' website was no longer available. This PR will update the referenced rpm file to the latest for the desired architecture.

**Technical details:**
can't build the docker container if the publicly hosted rpm is not available.


bminahankcfrb:
opening up one from branch created in a repo instead of from a fork.

PR #2460, opened 2020-05-05
Updating Dockerfile: Referenced public rpm no longer exists.

bminahan-kc:
rpm downloaded from download.postgresql.org no longer is available. referencing latest to increase our chances it will be available in the future.

**Description:**
When building the docker container, I could not build it because the rpm file on postgres' website was no longer available. This PR will update the referenced rpm file to the latest for the desired architecture.

**Technical details:**
can't build the docker container if the publicly hosted rpm is not available.

**Description:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN yum -y install wget gcc openssl-devel bzip2-devel libffi libffi-devel zlib-d
 RUN yum -y groupinstall "Development Tools"
 
 ##### Install PostgreSQL 10 client (psql)
-RUN yum -y install https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/pgdg-redhat10-10-2.noarch.rpm
+RUN yum -y install https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
 RUN yum -y install postgresql10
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN yum -y install wget gcc openssl-devel bzip2-devel libffi libffi-devel zlib-d
 RUN yum -y groupinstall "Development Tools"
 
 ##### Install PostgreSQL 10 client (psql)
-RUN yum -y install https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/pgdg-redhat10-10-2.noarch.rpm
+RUN yum -y install https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
```
tony-sappe, line 22: Unfortunately this did not work for my docker-compose container build locally. Even though I can download the rpm file on my host. Thoughts?

image

bminahankcfrb: I just tried `docker build --no-cache .` again and it built for me just fine. Maybe a network fluke? have you tried multiple times to build that layer or just the once?

bminahankcfrb: are you using any cached layers?

tony-sappe: must have been a VPN issue. Worked this afternoon

PR #2463, opened 2020-05-11
[DEV-4925] New program activity endpoint for agency profile page

sethstoudenmier:
**Description:**
Implement a new endpoint that returns a list of program activities for a specific toptier agency and fiscal year. 

**Technical details:**
Implemented the endpoint using a similar approach as the count endpoint. Added onto the agency_base class to take into account the different requirements for a GET vs POST. A minor adjustment was made to the count so that it will not count records that have a value of 0 for both the obligated amount and gross overlay.

    - [x] Backend
    - [x] Frontend <OPTIONAL>
8. [X] Jira Ticket [DEV-4925](https://federal-spending-transparency.atlassian.net/browse/DEV-4925):

```
```


usaspending_api/api_contracts/contracts/v2/agency/toptier_code/program_activity.md
```diff
@@ -0,0 +1,104 @@
...
+                "$schema": "http://json-schema.org/draft-04/schema#",
+                "type": "number"
+            }
+    + Parameters
+        + `toptier_code`: 086 (required, number)
+            The toptier code of an agency (could be a CGAG or FREC) so only numeric character strings of length 3-4 are accepted.
+    + Attributes
+        + `fiscal_year` (optional, number)
+            The desired appropriations fiscal year. Defaults to the current FY.
+        + `filter` (optional, string)
```
BrianZito, line 23: I wasn't helping with the `filter` path param on the filter tree endpoints, but would a `filter` body parameter  get confused with the `filters` param used in advanced search?

tony-sappe: `filter` is a common parameter or key for the app's API requests. There would be a benefit to reducing the usage of it when the term could be more specific for the request. However I don't believe it adds confusion.

usaspending_api/agency/v2/views/program_activity_list.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Obtain the list of program activity categories for a specific agency in a
+    single fiscal year based on whether or not that program activity has ever
+    been submitted in File B.
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/agency/toptier_code/program_activity.md"
+
+    @cache_response()
+    def post(self, request: Request, *args: Any, **kwargs: Any) -> Response:
```
kbard, line 21: Do we need to support `post` here?  With the exception of the advanced search which has a massive filter list that I'm told exceeds maximum query string of older browsers when fully populated, I had envisioned us moving toward more `get` based requests which is typical for REST-like APIs.  `post` is usually saved for operations that add data.

kbard: I realize there's an argument to be made for being consistent across our endpoints (I've even made that argument), but I feel like this is our opportunity for a clean break.  This is a whole new app.  We can start a new tradition.  One we can call our own.  One that tomorrow's developers can complain about.  "Why'd they do it this way?!?!?!"

usaspending_api/agency/v2/views/agency_base.py
```diff
@@ -44,3 +51,26 @@ def toptier_agency(self):
     @property
     def standard_response_messages(self):
         return [get_account_data_time_period_message()] if self.fiscal_year < 2017 else []
+
+    @cached_property
+    def pagination(self):
+        if self.request.method == "GET":
```
kbard, line 57: This seems unnecessary.  The only way you can get to this method is if you actually call it.  I assume you wouldn't call it if your endpoint didn't support it.  Plus, I could see this being useful regardless of whether or not we end up with `POST` or `GET` requests.  (See another comment I'm about to make concerning combining `request.data` with `request.query_params`.)

usaspending_api/agency/v2/views/agency_base.py
```diff
@@ -44,3 +51,26 @@ def toptier_agency(self):
...
+            page=request_data["page"],
+            limit=request_data["limit"],
+            lower_limit=(request_data["page"] - 1) * request_data["limit"],
+            upper_limit=(request_data["page"] * request_data["limit"]) + 1,
+            sort_key=request_data.get("sort", "obligated_amount"),
+            sort_order=request_data["order"],
+        )
+
+    @property
+    def filter(self):
```
kbard, line 73: I'm a little on the fence about adding this (and `pagination`) to the base class since it is not ubiquitous to most of the endpoints.  I know both of these are dirty words, but we could subclass this specifically for endpoints that provide pagination and filtering or we could create a mixin that supports that functionality.

A third option would be to support this compositionally (which is a lot like subclassing and/or mixins except your class uses the base class rather than actually inheriting from it).  Composition is so IN right now.  You want to be cool, don't you?

usaspending_api/agency/v2/views/program_activity_list.py
```diff
@@ -0,0 +1,66 @@
...
+            financialaccountsbyprogramactivityobjectclass__final_of_fy=True,
+            financialaccountsbyprogramactivityobjectclass__treasury_account__funding_toptier_agency=self.toptier_agency,
+            financialaccountsbyprogramactivityobjectclass__submission__reporting_fiscal_year=self.fiscal_year,
+        ).exclude(
+            financialaccountsbyprogramactivityobjectclass__obligations_incurred_by_program_object_class_cpe=0,
+            financialaccountsbyprogramactivityobjectclass__gross_outlay_amount_by_program_object_class_cpe=0,
+        )
+
+        if self.filter:
+            queryset_results = queryset_results.filter(program_activity_name__contains=self.filter)
```
kbard, line 46: `icontains`?

usaspending_api/agency/v2/views/program_activity_list.py
sethstoudenmier: If there is no pushback I am always in favor of more RESTful endpoints. As you stated, this is purely done as an attempt to be consistent.

usaspending_api/agency/v2/views/program_activity_list.py
```diff
@@ -1,14 +1,14 @@
...
 from typing import Any, List
-from usaspending_api.agency.v2.views.agency_base import AgencyBase
+from usaspending_api.agency.v2.views.agency_base import AgencyBase, ListMixin
 from usaspending_api.common.cache_decorator import cache_response
 from usaspending_api.common.helpers.generic_helper import get_simple_pagination_metadata
 from usaspending_api.references.models import RefProgramActivity
 
 
-class ProgramActivityList(AgencyBase):
+class ProgramActivityList(AgencyBase, ListMixin):
```
kbard, line 11: Mixins are typically listed to the left of base classes in python because method overrides are set up based on a right to left reading of the superclasses (i.e. if you want mixin methods to override base class methods, they need to be listed to the left of the base class).  It's not currently an issue in this case since there are no overrides, but if someone decides to override a base method/property this will become a problem.

usaspending_api/agency/v2/views/program_activity_count.py
```diff
@@ -37,7 +37,12 @@ def get_program_activity_count(self):
                         final_of_fy=True,
                         treasury_account__funding_toptier_agency=self.toptier_agency,
                         submission__reporting_fiscal_year=self.fiscal_year,
-                    ).values("pk")
+                    )
+                    .exclude(
```
kbard, line 41: Note to self, @kbard check this SQL out in the morning to ensure it doesn't express same odd behavior as the other query we already replaced.

kbard: Does not exhibit odd behavior we saw in other query, likely because it's not part of an aggregation.  👍

usaspending_api/api_contracts/contracts/v2/agency/toptier_code/program_activity.md
```diff
@@ -0,0 +1,94 @@
+FORMAT: 1A
+HOST: https://api.usaspending.gov
+
+# List Program Activity [/api/v2/agency/{toptier_code}/program_activity/{?fiscal_year,filter,order,sort,page,limit}]
+
+Returns a list of Program Activity in the Agency's appropriations for a single fiscal year
```
ebdabbs, line 6: I think this would read better as `Program Activities` plural

usaspending_api/api_contracts/contracts/v2/agency/toptier_code/program_activity.md
```diff
@@ -0,0 +1,94 @@
...
+                + `desc`
+                + `asc`
+        + `sort` (optional, enum[string])
+            Optional parameter indicating what value results should be sorted by.
+            + Default: `obligated_amount`
+            + Members
+                + `name`
+                + `obligated_amount`
+                + `gross_outlay_amount`
+        + `page` (optional, number)
```
ebdabbs, line 37: Does `page` default to 1?

usaspending_api/api_contracts/contracts/v2/agency/toptier_code/program_activity.md
```diff
@@ -0,0 +1,94 @@
...
+        + `limit` (optional, number)
+            How many results are returned.
+            + Default: 10
+
++ Response 200 (application/json)
+    + Attributes
+        + `toptier_code` (required, string)
+        + `fiscal_year` (required, number)
+        + `limit` (required, number)
+        + `page_metadata` (required, PageMetadata, fixed-type)
```
ebdabbs, line 48: Since this table will use a pager element (rather than infinite scroll or next / prev buttons) we'll need a `count` of total results. We don't really need `next`, `previous`, `hasNext`, or `hasPrevious` for USAS frontend purposes, but no harm in including them.

usaspending_api/api_contracts/contracts/v2/agency/toptier_code/program_activity.md
sethstoudenmier: Good point, will make the change.

usaspending_api/api_contracts/contracts/v2/agency/toptier_code/program_activity.md
sethstoudenmier: Yes it does, will add that.

usaspending_api/api_contracts/contracts/v2/agency/toptier_code/program_activity.md
sethstoudenmier: Since we have the count endpoints to go along with these list endpoints should the total still be added into the page_metadata here?

ebdabbs: True, we could pass down the totals from the tab component!

PR #2465, opened 2020-05-12
dev-4268 fixed logic bug with sibling codes

BrianZito:
**Description:**
Fixes subtle bug with complex TAS filters

**Technical details:**
Neither the code nor the tests covered the case of multiple require codes on the same level as one more more negative codes. 

    - [x] Backend
8. [x] Jira Ticket [DEV-4268](https://federal-spending-transparency.atlassian.net/browse/DEV-4268):

```
```


usaspending_api/search/filters/elasticsearch/HierarchicalFilter.py
```diff
@@ -17,7 +17,7 @@ def _query_string(cls, require, exclude) -> str:
         negative_query = " AND ".join([node.get_query() for node in negative_nodes])
 
         if positive_query and negative_query:
-            return f"{positive_query} AND {negative_query}"
+            return f"({positive_query}) AND {negative_query}"
```
tony-sappe, line 20: Curious why the negative query doesn't need surrounding parenthesis.

BrianZito: The simple answer is that the query is joined by ANDs instead of the ORs of the positive query, and A and B and C is equivelent to A and (B and C), but there's no reason not to add a set there just in case. And I smell room for more tests!

tony-sappe: Recall that Elasticsearch doesn't "honor the usual precedence rules, so parentheses should be used whenever multiple operators are used together"

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_boolean_operators

BrianZito: I have become all too familiar with that rule during this section of work. I've already gone ahead and put parens around any groupings of negative queries (and added a couple new tests to double check)

PR #2488, opened 2020-05-18
Reconfigure uwsgi.ini for a smaller server node

tony-sappe:
**Description:**
Reducing the deployed API server node from 32GB of RAM to 8GB requires some changes to the uwsgi config

**Technical details:**
Comments in the file explain the details

**OPS link**
fedspendingtransparency/usaspending-config#686

### DIFF
--- a/config/uwsgi.ini
+++ b/config/uwsgi.ini
@@ -26,7 +26,7 @@ cheaper-overload = 1
 cheaper = 50
 
 # Max Workers
-workers = 100
+workers = 200
 
 # Memory Limits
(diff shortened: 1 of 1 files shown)


config/uwsgi.ini
```diff
@@ -26,14 +26,16 @@ cheaper-overload = 1
 cheaper = 50
 
 # Max Workers
-workers = 200
+workers = 100
```
kwhickey, line 29: The plan was to keep this at 200 and monitor closely. With the limits in place too, uwsgi hopefully keeps the instance from falling over due to memory, but still allows near 200 workers if memory is okay.

PR #2493, opened 2020-05-21
[DEV-5091] Remove old code from move to Elasticsearch

sethstoudenmier:
**Description:**
Following up on the move to Elasticsearch for Advanced Search where code was commented out, this PR removes the old Postgres code and some additional code that was being used by what was commented out and is no longer needed.

**Technical details:**
Fairly straightforward since a lot of the deleted code was commented out. The bulk of non-commented out code that was deleted was used by the Postgres code to determine which matview it would utilize. It was left in place in case we had to re-instate the old Postgres code.

    - [x] Backend
8. [X] Jira Ticket [DEV-5091](https://federal-spending-transparency.atlassian.net/browse/DEV-5091):

```
```


usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -57,7 +57,7 @@
 from usaspending_api.common.exceptions import UnprocessableEntityException
 from usaspending_api.submissions.models import SubmissionAttributes
 
-logger = logging.getLogger("console")
+logger = logging.getLogger(__name__)
```
tony-sappe, line 60: I expected a lot more could be removed from this file. If this endpoint is completely done using materialized views what is `matview_search_filter_determine_award_matview_model()` used for? Same with all of the non-elasticsearch/non-subaward mappings

usaspending_api/awards/v2/filters/view_selector.py
```diff
@@ -206,88 +206,6 @@ def can_use_view(filters, view_name):
     return True
```
tony-sappe, line 207: I imagine all of the dictionaries defined in `MATVIEW_SELECTOR` can be removed except for the three listed in `recipient_totals()`.

usaspending_api/search/v2/views/spending_by_award.py
sethstoudenmier: Will need to step through the imports some more.

usaspending_api/awards/v2/filters/view_selector.py
sethstoudenmier: Hesitant to remove them from this PR since those matviews are still around as of right now. Was thinking they might be removed as part of removing the old matviews.

tony-sappe: Using that logic, Emily could say "I can't remove these matviews, they are still used by the code!" 😆 

Time-permitting, it would be good to be as aggressive with code removal as possible. Shifting priorities might not leave much more time for this

sethstoudenmier: Touché. Will remove that in my next commit.

emily-brents: I'd agree, seems more in line with removing the matviews, but it should be fine either way

PR #2501, opened 2020-05-28
Warmfix/update rosetta loader

BrianZito:
**Description:**
Modifications to the load_rosetta command to support a wider array of files, more specifically the new column of DAIMS 2.0

**Technical details:**
This change removed DB_TO_XLSX_MAPPING, and instead trusts column names provided by the file. The actual rosetta table stores the entire file as a JSON object anyway, so there's no concern with not having the right DB column names. This does leave the possibility of having a file with the wrong column names, but DAIMS has shown they can and do add or remove columns, so I don't think we have a choice but to trust our inputs here.

Structure of the raw names was changed to be based on the display name and index of the column. This will ensure that it is always unique. Related frontend PR (https://github.com/fedspendingtransparency/usaspending-website/pull/1698) has a change to correspond with the initial sorted column, but otherwise front doesn't reference any of these raw names directly. Because of this PR, front-end will need to merge before backend.

There will be a related frontend change to make the colors work right, but there is no need to coordinate the merges. 

    - [x] Backend
8. [ ] Jira Ticket [DEV-4176](https://federal-spending-transparency.atlassian.net/browse/DEV-4176):

```
```


usaspending_api/references/tests/integration/test_load_rosetta.py
```diff
@@ -40,18 +40,18 @@ def test_rosetta_fresh_load():
...
-            {"raw": "award_file", "display": "Award File"},
-            {"raw": "award_element", "display": "Award Element"},
-            {"raw": "subaward_file", "display": "Subaward File"},
-            {"raw": "subaward_element", "display": "Subaward Element"},
-            {"raw": "account_file", "display": "Account File"},
-            {"raw": "account_element", "display": "Account Element"},
-            {"raw": "legacy_award_file", "display": "Award File"},
-            {"raw": "legacy_award_element", "display": "Award Element"},
-            {"raw": "legacy_subaward_element", "display": "Subaward Element"},
+            {"raw": "A:element", "display": "Element"},
```
tony-sappe, line 43: What does the column name help with?

BrianZito: uniqueness of the `raw` value. Frontend uses these to find the column to sort on.

BrianZito: Previously the hard-coded list implicitly guarenteed this property, this PR changes it to be explicit.

BrianZito:
> Because this script reads the XLSX file and creates a JSON document which is provided by the API there needs to be some exceptions set for the file data. There are a lot of assumptions provided in the PR description justification.

I'm unsure what expectations can be held for the file. We could verify the shape of the data (equal columns on each row), but the file can have nulls. We could verify the list or order of columns, or their association under sections, but those will change every iteration. Since our input is also the authoritative source, it seems like any validation that actually threw an exception would already be obsolete.

tony-sappe:
> > Because this script reads the XLSX file and creates a JSON document which is provided by the API there needs to be some exceptions set for the file data. There are a lot of assumptions provided in the PR description justification.
> 
> I'm unsure what expectations can be held for the file. We could verify the shape of the data (equal columns on each row), but the file can have nulls. We could verify the list or order of columns, or their association under sections, but those will change every iteration. Since our input is also the authoritative source, it seems like any validation that actually threw an exception would already be obsolete.

Leaving `DB_TO_XLSX_MAPPING` in place to maintain strict columns would be a good first step since they have a direct impact to the API response shape. As discovered today, having a flexible container isn't the only concern to the frontend. Wouldn't it be preferable to fail fast and hard rather than store a JSON doc which can cause undesirable behavior on the frontend?

BrianZito:
Loud failures are always nice, but I have concerns to the accuracy of those failures by using DB_TO_XLSX_MAPPING. The last two updates to DAIMS (2.0 and 1.4) both added columns, and Ross is talking to scema to do another update to 2.0 to change the column order again. In both cases, the mapping wouldn't have been matched, causing an error. In both cases, that error would be wrong and the correct course of action would be to simply update the list directly from the document. In fact, unless the document was tampered with for some reason, I believe that this would ALWAYS be the case for a major change (as defined by the document on changing the data dictionary), and the mapping would never catch an issue for a minor change where no columns were changed. 

At the end of the day, I think there's two key points here. The first is that the structure of our data is arbitrary, since Schema could provide a legitimate file with any set of columns, as our input is the source of truth for its own validity. Second, the purpose of the data dictionary page that this endpoint powers is to display an excel document (which has already been uploaded by the time this command is run) in-browser. If a "bad" file was uploaded and this loader ran, I think that the correct behavior would be to display it accurately, according to how the file would look if downloaded. 

That said, throwing an error would give the DEV or OPS staff a chance to see the mistake and re-upload the file, but only if there was some validation rule that could prove that the file they were looking at wasn't the file they were supposed to have. Personally, I don't think a hard-coded column list accomplishes that objective. If we still want to go ahead with it, I would suggest something more minimalistic: instead of building this mapping and making it an integral part of the data load, just have a list of the last accepted columns, and if they don't match directly tell the user "This file has a different set of columns than before. Was this intended?"

BrianZito:
https://github.com/fedspendingtransparency/usaspending-website/pull/1698 is merged.

PR #2502, opened 2020-05-29
[DEV-5252] Elasticsearch sanitization fix

emily-brents:
**Description:**
Removes several characters from the list of characters that we remove from text inputs that are used by the elasticsearch cluster.

**Technical details:**
Removes `^`, `~`, `/`, `!`, and `-` from the list of characters being stripped from the elasticsearch text inputs. `^`, `~`, `/`, and `-` all require escaping to be used in the queries, so they are now being escaped in the 

    - [x] Backend
8. [X] Jira Ticket [DEV-5252](https://federal-spending-transparency.atlassian.net/browse/DEV-5252):

```
No changes to matviews, no ops tickets needed
```


usaspending_api/common/query_with_filters.py
```diff
@@ -166,12 +166,11 @@ class _RecipientSearchText(_Filter):
     @classmethod
     def generate_elasticsearch_query(cls, filter_values: List[str], query_type: _QueryType) -> ES_Q:
         recipient_search_query = []
-        fields = ["recipient_name"]
 
         for v in filter_values:
             upper_recipient_string = es_sanitize(v.upper())
             recipient_name_query = ES_Q(
-                "query_string", query=upper_recipient_string + "*", default_operator="AND", fields=fields
+                "match", recipient_name={"query": upper_recipient_string + "*", "operator": "AND"}
```
sethstoudenmier, line 173: I don't think this is working as expected. The `match` query is going to look for a literal `*` since it doesn't support the same query_string syntax. I haven't tried the code locally, but trying a `match` against my local ES cluster and it did not support wildcard character.

sethstoudenmier: Saw you changed it, but just to add to this the tests were still passing because apparently the wildcard character is simply ignored. Query for `Recipient_name_1*` yields transactions with `Recipient_name_1` and not `Recipient_name_10`.

emily-brents: We should probably have tests that check the wildcard is working - I believe we have some on the keyword filter for advanced search, so it's odd that the recipient filter doesn't check that as well.

usaspending_api/search/tests/unit/test_elasticsearch_helpers.py
```diff
@@ -109,15 +109,15 @@ def test_get_download_ids(monkeypatch, transaction_type_data, elasticsearch_tran
 
 
 def test_es_sanitize():
```
tony-sappe, line 111: Duplicated test? I know it is pre-existing but this diff highlights the duplication with usaspending_api/search/tests/unit/test_es_sanitize.py

emily-brents: Hm yeah, that is a duplicate. I think it got missed when we moved where the sanitization functions were being kept

PR #2514, opened 2020-06-09
[DEV-5132] Add DEFC to award response

kbard:
**Description:**

Add DEFC to response for an award.  Pretty trivial, really.

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [ ] Frontend @MaxwellKendall 
4. [x] Materialized views unaffected
5. [x] Frontend impact assessment completed
6. [x] Data validation completed
7. [x] No Operations ticket required
8. [x] Jira Ticket [DEV-5132](https://federal-spending-transparency.atlassian.net/browse/DEV-5132):
    - [x] Link to this Pull-Request

PR #2520, opened 2020-06-09
[DEV-5136] Add account data/file C data to award endpoint

emily-brents:
**Description:**
Adds both the sum of obligations and outlays to the award endpoint response and a breakdown of the outlays and obligations by the disaster emergency fund code

    - [x] Backend
    - [x] Frontend 
8. [X] Jira Ticket [DEV-5136](https://federal-spending-transparency.atlassian.net/browse/DEV-5136):

```
```


kbard:
I can't help but notice that these changes could have been used to answer the question answered by #2514.  Should we back out #2514?  cc: @tony-sappe

usaspending_api/api_contracts/contracts/v2/awards/award_id.md
```diff
@@ -406,6 +428,28 @@ This endpoint returns a list of data that is associated with the award profile p
...
+                    {
+                        "code": "P",
+                        "amount": 2.0
+                    },
+                    {
+                        "code": "L",
+                        "amount": 1.0
+                    }
+                },
+                "account_obligations": {
```
kbard, line 443: Ohhhh... so close... these are returning an array of objects... not an object of objects  😛

usaspending_api/api_contracts/contracts/v2/awards/award_id.md
```diff
@@ -769,6 +843,10 @@ This endpoint returns a list of data that is associated with the award profile p
 + `place_of_performance` (required, Location, fixed-type)
 + `executive_details` (required, Executive, fixed-type)
 + `cfda_info` (required, array[CFDAInfo], fixed-type)
++ `total_account_outlay` (required, number)
++ `total_account_obligation` (required, number)
++ `account_obligations` (required, array[object], fixed-type)
++ `account_outlays` (required, array[object], fixed-type)
```
kbard, line 849: Usually we define the actual object and reference that definition rather than use generic objects.  Also, should  these mention that they're obligations/outlays broken down by defc?

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +586,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+
+def fetch_account_details_idv(award_id: int) -> dict:
+
+    child_award_sql = """
+    select
+        ac.id as award_id
+    from
+        parent_award pap
+        inner join awards ap on ap.id = pap.award_id
+        inner join awards ac on ac.fpds_parent_agency_id = ap.fpds_agency_id and ac.parent_award_piid = ap.piid
```
kbard, line 628: I _think_ we need to exclude child IDVs here too like we do for grandchild awards (`ac.type not like 'IDV%'`).  Most of the time we don't seem to care about values in the IDV as much as we do their non-IDV children.

emily-brents: Do we know why we would exclude them? It seems to me that if a child IDV had a specific DEF code or obligated amount unique to it and not its child awards we would want to include it.

kbard: We need to confirm this with someone smarter than me, but I believe IDVs don't usually have these values or, if they do, they're not supposed to.  Either way, I think we're supposed to ignore most dollar figures on IDVs.  Kristine or Ross would probably be our best bet.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -128,6 +134,9 @@ def construct_idv_response(requested_award_dict: dict) -> OrderedDict:
         return None
     response.update(award)
 
+    account_data = fetch_account_details_award(award["id"])
```
kbard, line 137: Is this supposed to be `fetch_account_details_idv`?

emily-brents: :o yeah

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +586,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+
+    where
+        pap.award_id = {award_id}
+    """
+    children = execute_sql_to_ordered_dictionary(child_award_sql.format(award_id=award_id))
+    grandchildren = execute_sql_to_ordered_dictionary(grandchild_award_sql.format(award_id=award_id))
+    award_ids = []
+    award_ids.extend([x["award_id"] for x in children])
+    award_ids.extend([x["award_id"] for x in grandchildren])
+    award_ids.append(award_id)
```
kbard, line 650: _IF_ we're not supposed to be summing IDV values, then we also need to not append `award_id`.

usaspending_api/awards/v2/data_layer/orm.py
tony-sappe: The api/v2/awards/ response should only be data on that specific IDV record.

There is another endpoint (perhaps it's api/v2/awards/idvs/amounts/) to provide the aggregated amounts for child and grandchild IDV orders

emily-brents: yup, tony is correct, i completely misread the ticket, the IDV amounts should be under a different endpoint

kbard: The question still applies to the other endpoint, I suppose.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +586,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
     results = {"toptier_code": toptier_code, "midtier_code": midtier_code, "base_code": base_code}
     return results
+
+
+def fetch_account_details_award(award_id: int) -> dict:
+    queryset = (
+        FinancialAccountsByAwards.objects.filter(award_id=award_id)
+        .values("disaster_emergency_fund")
+        .annotate(
+            obligated_amount=Sum("obligations_delivered_orders_unpaid_total_cpe"),
```
kwhickey, line 596: This is changing to `total_obligation_amount`, yes?

usaspending_api/api_contracts/contracts/v2/awards/award_id.md
```diff
@@ -680,6 +746,10 @@ This endpoint returns a list of data that is associated with the award profile p
     `null` if the contract has no parent
 + `naics_hierarchy` (required, NAICSHierarchy, fixed-type)
 + `psc_hierarchy` (required, PSCHierarchy, fixed-type)
++ `total_account_outlay` (required, number)
++ `total_account_obligation` (required, number)
++ `account_obligations` (required, array[DEFCAmount], fixed-type)
++ `account_outlays` (required, array[DEFCAmount], fixed-type)
```
kwhickey, line 752: Here and below - 
If this naming is not finalized, it appears too generic of a name for something that will be very specific to DEF codes. `account_outlays_by_defc` is good (same for obligations). Unless we want to make this an arbitrary breakdown where code could be for anything (Program Activity, object class, TAS, etc.), in which case `account_outlay_breakdown` or just `account_outlays` or `account_outlay_details` is fine.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +586,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+    grandchildren = execute_sql_to_ordered_dictionary(grandchild_award_sql.format(award_id=award_id))
+    award_ids = []
+    award_ids.extend([x["award_id"] for x in children])
+    award_ids.extend([x["award_id"] for x in grandchildren])
+    award_ids.append(award_id)
+    queryset = (
+        FinancialAccountsByAwards.objects.filter(award_id__in=award_ids)
+        .values("disaster_emergency_fund")
+        .annotate(
+            obligated_amount=Sum("obligations_delivered_orders_unpaid_total_cpe"),
```
kwhickey, line 655: Also changing this to `total_obligated_amount`, yes?

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +586,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
     return results
+
+
+def fetch_account_details_award(award_id: int) -> dict:
+    queryset = (
+        FinancialAccountsByAwards.objects.filter(award_id=award_id)
+        .values("disaster_emergency_fund")
+        .annotate(
+            obligated_amount=Sum("obligations_delivered_orders_unpaid_total_cpe"),
+            gross_outlay_amount=Sum("gross_outlay_amount_by_award_cpe"),
```
kwhickey, line 597: This is a tricky field that needs special handling beyond FY 2020. See slack thread: https://data-transparency-bfs.slack.com/archives/C0122HUCD17/p1591885771087400

and **CARES Award Obligation and Awards Outlay … by Budget Category (File C):**
section in https://federal-spending-transparency.atlassian.net/browse/DEV-5360 which addresses this.

IF we absolutely **MUST** defer the work that makes this work post-2020, put a comment in with `ENABLE_CARES_ACT_FEATURES` in the comment that states that we need to further accommodate this problem post-launch.

From thread:
> If you want to get ALL-TIME (life of the award) value, it’s trickier, and you need to look at all FYs that had this award, and get the latest reported period’s value in any of those FYs. e.g. Let’s say we’re in FY21, Q2, period 6 (period 5 has closed):
> 1. If we only need to look back at 2020, get FY2020 P12 value for this award. 
> 2. Add that to FY2021 P05 value for this award
> 3. That gives the lifetime outlay value for this award over the 2 FYs it existed.

There's a rough attempt at what sql would look like that grabs that cross-year values.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +586,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+    award_ids = []
+    award_ids.extend([x["award_id"] for x in children])
+    award_ids.extend([x["award_id"] for x in grandchildren])
+    award_ids.append(award_id)
+    queryset = (
+        FinancialAccountsByAwards.objects.filter(award_id__in=award_ids)
+        .values("disaster_emergency_fund")
+        .annotate(
+            obligated_amount=Sum("obligations_delivered_orders_unpaid_total_cpe"),
+            gross_outlay_amount=Sum("gross_outlay_amount_by_award_cpe"),
```
kwhickey, line 656: same as prior comment.

usaspending_api/api_contracts/contracts/v2/awards/award_id.md
MaxwellKendall: `by` IMO indicates the value is a map. `defc_account_outlays`. I really don't care what it's named TBH.

usaspending_api/awards/v2/data_layer/orm.py
tony-sappe: I believe the code is correct after the changes since the comments. Is that correct?

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +583,89 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+
+
+def fetch_account_details_award(award_id: int) -> dict:
+    queryset = (
+        FinancialAccountsByAwards.objects.filter(award_id=award_id)
+        .values("disaster_emergency_fund")
+        .annotate(
+            obligated_amount=Sum("obligations_delivered_orders_unpaid_total_cpe"),
+            gross_outlay_amount=Sum("gross_outlay_amount_by_award_cpe"),
+        )
```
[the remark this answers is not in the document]
tony-sappe: @emily-brents can you point out where `null` are being filtered out? Running a slightly modified version of the SQL produces many null DEFC

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +585,93 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+    where faba.award_id = '{award_id}'
+    group by faba.disaster_emergency_fund_code, total_outlay;
+    """
+    results = execute_sql_to_ordered_dictionary(defc_sql.format(award_id=award_id))
+    outlay_by_code = []
+    obligation_by_code = []
+    total_outlay = 0
+    total_obligations = 0
+    for row in results:
+        total_outlay += row["total_outlay"] if row["total_outlay"] is not None else 0
```
tony-sappe, line 667: Wouldn't the coalesce for `total_outlay` in the SQL prevent null/None?
Additionally, it would be cleaner to add coalesce into the SQL for `total_obligations`

usaspending_api/idvs/v2/views/amounts.py
```diff
@@ -9,12 +10,167 @@
...
 from usaspending_api.awards.models import ParentAward
 from usaspending_api.common.cache_decorator import cache_response
+from usaspending_api.common.helpers.sql_helpers import execute_sql_to_ordered_dictionary
 from usaspending_api.common.validator.award import get_internal_or_generated_award_id_model
 from usaspending_api.common.validator.tinyshield import TinyShield
 
 
 logger = logging.getLogger("console")
 
+defc_sql = """
```
tony-sappe, line 20: How similar is this SQL to the new SQL in usaspending_api/awards/v2/data_layer/orm.py? Could they be merged into a helper function?

usaspending_api/awards/v2/data_layer/orm.py
emily-brents: yes

usaspending_api/awards/v2/data_layer/orm.py
emily-brents: yeah

usaspending_api/idvs/v2/views/amounts.py
emily-brents: they're essentially the same, the only real difference is that the idv sql queries against a list of award_ids, which is easy enough to change

usaspending_api/awards/v2/data_layer/orm.py
emily-brents: yes

usaspending_api/api_contracts/contracts/v2/idvs/amounts/award_id.md
```diff
@@ -22,11 +22,18 @@ Get aggregated award counts and funding amounts for IDV contracts
         * `child_award_total_obligation`: 363410.59 (required, number)
         * `child_award_base_and_all_options_value`: 297285.59 (required, number)
         * `child_award_base_exercised_options_val`: 297285.59 (required, number)
+        * `child_total_account_outlay`: 0 (required, number)
+        * `child_total_account_obligation`: 0 (required, number)
+        * `child_account_outlays_by_defc`: [] (required, array[DEFCAmount])
+        * `child_account_obligations_by_defc`: [] (required, array[DEFCAmount])
```
kbard, line 28: These arrays (and the ones below) need `fixed-type` else the type isn't enforced.

kbard: Now you're just messin' with me aren't you...

kbard, line 28:
```suggestion
        * `child_account_obligations_by_defc`: [] (required, array[DEFCAmount], fixed-type)
```

usaspending_api/api_contracts/contracts/v2/idvs/amounts/award_id.md
```diff
@@ -22,11 +22,18 @@ Get aggregated award counts and funding amounts for IDV contracts
         * `child_award_total_obligation`: 363410.59 (required, number)
         * `child_award_base_and_all_options_value`: 297285.59 (required, number)
         * `child_award_base_exercised_options_val`: 297285.59 (required, number)
+        * `child_total_account_outlay`: 0 (required, number)
+        * `child_total_account_obligation`: 0 (required, number)
+        * `child_account_outlays_by_defc`: [] (required, array[DEFCAmount])
```
kbard, line 27:
```suggestion
        * `child_account_outlays_by_defc`: [] (required, array[DEFCAmount], fixed-type)
```

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +585,98 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+        -- This is NOT right, since period dates are not the same as submission window close dates,
+        -- but an approximation for illustration purposes until we have the reporting dates table up
+        select
+            distinct concat(p.reporting_fiscal_year::text,
+            lpad(p.reporting_fiscal_period::text, 2, '0')) as fyp,
+            p.reporting_fiscal_year, p.reporting_fiscal_period
+        from submission_attributes p
+        order by p.reporting_fiscal_year desc, p.reporting_fiscal_period desc
+    ),
+    eligible_submissions as (
```
kbard, line 614: `eligible_submissions` isn't being used.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +585,98 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+            faba.transaction_obligated_amount,
+            faba.disaster_emergency_fund_code,
+            s.reporting_fiscal_year,
+            s.reporting_fiscal_period,
+            s.submission_id
+        from financial_accounts_by_awards faba
+        inner join submission_attributes s on faba.submission_id=s.submission_id
+        where {award_id_sql} and faba.disaster_emergency_fund_code is not null
+    ),
+    closed_periods as (
```
kbard, line 604: Is `closed_periods` doing anything?  It seems to just be returning all periods.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +585,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+        faba.disaster_emergency_fund_code,
+        coalesce(ffy.prior_fys_outlay, 0) + coalesce(cfy.current_fy_outlay, 0) as total_outlay,
+        coalesce(sum(faba.transaction_obligated_amount), 0) as obligated_amount
+    from eligible_file_c_records faba
+    left join fy_final_balances ffy on ffy.award_id = faba.award_id and ffy.disaster_emergency_fund_code = faba.disaster_emergency_fund_code
+    left join current_fy_balance cfy
+        on cfy.reporting_fiscal_period != 12 -- don't duplicate the year-end period's value if in unclosed period 01
+        and cfy.award_id = faba.award_id
+        and cfy.disaster_emergency_fund_code = faba.disaster_emergency_fund_code
+    group by faba.disaster_emergency_fund_code, total_outlay;
```
kwhickey, line 653: I was having a hard time following this all the way through, but tested and it verifies the group by to be okay. But to confirm, you can group-by `total_outlay` here, because effectively this query is going to _repeat_ the same outlay+defc values for each distinct obligation amount present for the award in the `faba` table? True?

This is a nice way to get outlay and obligation sums in 1 query (if a bit hard to follow) in order to not have to run 2 similar queries for each of outlay and obligation.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +585,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
...
+def fetch_account_details_award(award_id: int) -> dict:
+    award_id_sql = "faba.award_id = {award_id}".format(award_id=award_id)
+    results = execute_sql_to_ordered_dictionary(defc_sql.format(award_id_sql=award_id_sql))
+    outlay_by_code = []
+    obligation_by_code = []
+    total_outlay = 0
+    total_obligations = 0
+    for row in results:
+        total_outlay += row["total_outlay"]
+        total_obligations += row["obligated_amount"]
```
kwhickey, line 666: Here and below, this is going to give bad sums. 

Looks like you removed this part from the sql:
```
inner join disaster_emergency_fund_code defc on defc.group_name = 'covid_19' and defc.code = faba.disaster_emergency_fund_code
```

Which was necessary to get the breakdowns by ALL DEFCs for the award details. However, the total covid obligation and total covid outlay should be just based on DEFC rows that have `group_name = 'covid_19'`. 

So probably run a query on the defc reference table before running this query to get the covid_19 defcs, and then in this row only do the sum if the row's defc is one of those in the group.

This is based on the assumption that we want ALL DEFC funding breakdowns eventually, and not just the covid ones. If we're only working with covid ones ... add that inner join back to the CTE for eligible file c records.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -577,3 +585,91 @@ def fetch_naics_hierarchy(naics: str) -> dict:
         pass
     results = {"toptier_code": toptier_code, "midtier_code": midtier_code, "base_code": base_code}
     return results
+
+
+defc_sql = """
```
kwhickey, line 590: I recommend putting this and any functions in this .py file that use this or build upon this into a different module named `usaspending_api/awards/v2/data_layer/sql.py`. Being in `orm.py` implies that I could find data access code that uses the Django ORM. This goes straight to SQL so isn't really orm-code.

usaspending_api/awards/v2/data_layer/orm.py
kwhickey: It would be great to also see another unit test that creates file c rows with DEFCs that are and are _not_ COVID-19 defcs, and assert that the totals are only inclusive of the ones that are covid DEFCs. 

Also once the closed periods stuff is in, a similar test would be good for excluding outlay balances that are either from non-period-12 in prior FYs or from not-the-latest-closed-period-in-this-FY.

kbard: So as part of my testing I wrote up a bunch of additional tests.  Once I'm done helping Melissa, I'll pass them along to Emily.  They added non COVID DEF codes and a bunch of other things.

usaspending_api/awards/v2/data_layer/orm.py
emily-brents: Yes, the outlays are calculated in the CTEs (current fy balance and fy final balances), so they are the same for every entry in the faba CTE. They get grouped by because they aren't aggregated like obligations.

usaspending_api/awards/v2/data_layer/orm.py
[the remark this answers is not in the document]
emily-brents: No, that was removed from here after the sql was updated to coalesce the outlays and obligations, i just forgot to remove it there.

usaspending_api/awards/v2/data_layer/sql.py
```diff
@@ -0,0 +1,67 @@
...
+        where {award_id_sql} and faba.disaster_emergency_fund_code is not null
+    ),
+    closed_periods as (
+    select
+        w.submission_fiscal_year,
+        w.submission_fiscal_month,
+        w.is_quarter
+    from dabs_submission_window_schedule w
+    where
+        w.submission_reveal_date <= now() -- change "period end" with "window close"
```
kwhickey, line 22: Minor thing: you should remove that comment on this line

usaspending_api/awards/v2/data_layer/orm.py
kwhickey: @emily-brents - I think this is still an issue with the new SQL. The new SQL doesn't limit faba records to just those with `group_name = 'covid_19'` so this looping and summing will sum across all/any DEFCs submitted (including Q and 9, etc. that are not COVID-related). 

Whereas the AC for this effort is to show total outlay and total obligation for just COVID-related def codes. 

You need some condition in the loop that says something like:
```
covid_defcs = DisasterEmergencyFundCodes.objects().filter("group_name = 'covid_19'").values_list('code', flat=True)
# ...
for row in results:
    if row["disaster_emergency_fund_code"] in covid_defcs:
        total_outlay += row["total_outlay"]
        total_obligations += row["obligated_amount"]
# ...
```

usaspending_api/idvs/v2/views/amounts.py
```diff
@@ -9,12 +10,98 @@
...
+        child_obligation_by_code.append(
+            {"code": row["disaster_emergency_fund_code"], "amount": row["obligated_amount"]}
+        )
+    grandchild_outlay_by_code = []
+    grandchild_obligation_by_code = []
+    grandchild_total_outlay = 0
+    grandchild_total_obligations = 0
+    for row in grandchild_results:
+        grandchild_total_outlay += row["total_outlay"]
+        grandchild_total_obligations += row["obligated_amount"]
```
kwhickey, line 87: Same problem as above. Totals need to be for covid-only DEFCs

usaspending_api/idvs/v2/views/amounts.py
```diff
@@ -9,12 +10,98 @@
...
+        if grandchild_award_ids != []
+        else {}
+    )
+    child_outlay_by_code = []
+    child_obligation_by_code = []
+    child_total_outlay = 0
+    child_total_obligations = 0
+    for row in child_results:
+        child_total_outlay += row["total_outlay"]
+        child_total_obligations += row["obligated_amount"]
```
kwhickey, line 76: Same problem as above. Totals need to be for covid-only DEFCs

usaspending_api/idvs/v2/views/amounts.py
emily-brents: I was under the impression that the totals where just the total spending we had for all DEFC, and the frontend would be filtering out which DEFC they wanted to use on their end.
But actually thinking about it it doesn't make much sense to do it that way, because how would they know which DEFC are covid-related, so I will in fact change this

tony-sappe: The website is using the reference endpoint which does give the necessary info on which DEFC are COVID-19
https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/api_contracts/contracts/v2/references/def_codes.md

PR #2538, opened 2020-06-19
[DEV-5230] Add COVID data to ES / spending_by_award

emily-brents:
**Description:**
Adds COVID-19 related outlays and obligations to the elasticsearch indexes. Also modifies the `spending_by_award` endpoint to allow that information to be returned.

**Technical details:**
The data being added to each index is as follows:
Award Index:
- Adds all DEF codes associated with the award in an array
- Adds COVID-related (based on DEF code) outlays associated with the award. Based on the `gross_outlay_amount_by_award_cpe` field. This field is aggregated by summing the value for the final period per fiscal year. If there is no entry for the final period of the fiscal year it is assumed to be 0.
- Adds COVID-related (based on DEF code) obligations associated with the award. Based on the  sum of the `transaction_obligated_amount` field.

Transactions:
- Adds all DEF codes associated with the parent award of the transaction.

    - [ ] Backend
7. [X] Appropriate Operations ticket(s) created [DEV-5232](https://federal-spending-transparency.atlassian.net/browse/DEV-5232)
8. [ X Jira Ticket [DEV-5230](https://federal-spending-transparency.atlassian.net/browse/DEV-5230):

```
```


usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -81,7 +81,10 @@ SELECT
   vw_award_search.naics_code,
   vw_award_search.naics_description,
   TREASURY_ACCT.tas_paths,
-  TREASURY_ACCT.tas_components
+  TREASURY_ACCT.tas_components,
+  DEFC.defc as disaster_emergency_fund_codes,
+  DEFC.total_covid_obligations,
```
tony-sappe, line 86: Is it possible to store the obligation and outlay per DEFC in the index? That will make it useful for the COVID-19 page.

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -92,6 +95,79 @@ LEFT JOIN (
...
+        FROM submission_attributes s
+        WHERE concat(s.reporting_fiscal_year::text, lpad(s.reporting_fiscal_period::text, 2, '0')) IN (
+            SELECT fyp FROM closed_periods
+        )
+    ),
+    eligible_file_c_records AS (
+        SELECT
+            faba.award_id,
+            CASE WHEN faba.disaster_emergency_fund_code IN ('L','M','N','O','P') THEN faba.gross_outlay_amount_by_award_cpe else 0 END AS gross_outlay_amount_by_award_cpe,
+            CASE WHEN faba.disaster_emergency_fund_code IN ('L','M','N','O','P') THEN faba.transaction_obligated_amount else 0 END AS transaction_obligated_amount,
```
tony-sappe, line 123: There is a strong desire to prohibit hardcoding the COVID-19 DEFC into code. the lookup table provides a `covid_19` group which can be used to obtain the list.

usaspending_api/etl/es_award_template.json
```diff
@@ -418,6 +418,27 @@
...
           "type": "keyword"
+        },
+        "disaster_emergency_fund_codes": {
+          "type": "keyword"
+        },
+        "total_covid_obligations": {
+          "type": "scaled_float",
+          "scaling_factor": 100,
+          "fields": {
+            "keyword": {
```
tony-sappe, line 429: What is a keyword field adding for a scaled_float?

usaspending_api/database_scripts/etl/award_delta_view.sql
kwhickey: @tony-sappe 
I think the effort to do this would be a lot. It would be akin to what we tried to do with the TAS records at one point in time -- making them _nested objects_ (sub json docs to the main award doc).

An alternative would be to bake those into an array of json strings like we do with TAS now (we had to get away from nested objects for indexing performance reasons). But that's usually only good for _search_ I think. And the DEFC amounts are numeric values we'd want to use for aggregation. 

There also is nothing yet in the _current_ mocks for Jul 30th that require further subdivision of award data by DEFC. Eventually if a DEFC filter comes along, it would drive something like this. 
a) We'd either need to get it into awards ES index (could be dirty and compromise the index for other uses)
b) Switch those APIs from using the awards ES index to using postgres tables instead
c) make a new account index that brings over award-specific metadata at index-time by joining `financial_accounts_by_awards` to `awards`

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -92,6 +95,80 @@ LEFT JOIN (
...
+        SELECT
+            faba.award_id,
+            CASE WHEN defc.group_name = ('covid_19') THEN faba.gross_outlay_amount_by_award_cpe else 0 END AS gross_outlay_amount_by_award_cpe,
+            CASE WHEN defc.group_name = ('covid_19') THEN faba.transaction_obligated_amount else 0 END AS transaction_obligated_amount,
+            faba.disaster_emergency_fund_code,
+            s.reporting_fiscal_year,
+            s.reporting_fiscal_period
+        FROM financial_accounts_by_awards faba
+        INNER JOIN eligible_submissions s ON s.submission_id = faba.submission_id
+        INNER JOIN disaster_emergency_fund_code defc on defc.code = faba.disaster_emergency_fund_code
```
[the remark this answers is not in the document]
kwhickey: I think I see why you had to do the case statement here. This string of CTEs is attaching ALL DEFCs to the award record as an array, not just COVID-ones. Hence the `ARRAY_AGG(DISTINCT faba.disaster_emergency_fund_code) AS defc` later on. 

So you need to be inclusive of non-covid-defc rows to capture other DEFCs ... but exclude summing those non-covid DEFC's obs/outs in the total_**covid** values. 

That the case? If so, disregard my previous comment, this works.

----

**_Still ..._**
I wonder if ~a `UNION` might be more efficient here. e.g.~ going after totals separate from DEFCs may be more efficient here, esp. since we're already joining on `financial_accounts_by_awards`:

1) Keep the inner join on `defc.group_name = 'covid_19'`, and this will provide one dijoint subset of file C rows that ARE covid-related, from which you can sum. Have this string of CTEs and query _just_ be about getting those covid totals. Don't bother trying to project the DEF codes here.
2) Then, to get the DEFCs associated with each award, do exactly what you're doing in the `transaction_delta_view` below. That is, end this VIEW with:
```sql
LEFT JOIN (
  SELECT
    faba.award_id,
    ARRAY_AGG(
      DISTINCT CONCAT(
        -- ...
       )
     ) tas_paths,
     ARRAY_AGG(
      DISTINCT CONCAT(
        -- ...
       )
     ) tas_components,
    ARRAY_AGG(faba.disaster_emergency_fund_code) defc
 FROM
   treasury_appropriation_account taa
   INNER JOIN financial_accounts_by_awards faba ON (taa.treasury_account_identifier = faba.treasury_account_id)
   INNER JOIN federal_account fa ON (taa.federal_account_id = fa.id)
   INNER JOIN toptier_agency agency ON (fa.parent_toptier_agency_id = agency.toptier_agency_id)
 WHERE
   faba.award_id IS NOT NULL
 GROUP BY
   faba.award_id
) TREASURY_ACCT ON (TREASURY_ACCT.award_id = vw_award_search.award_id);
```

Would that work ^^ ?? Can you try and experiment with performance?

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -1,6 +1,65 @@
 DROP VIEW IF EXISTS award_delta_view;
-
 CREATE VIEW award_delta_view AS
+with closed_periods as (
+    select
+    	distinct concat(w.submission_fiscal_year::text, lpad(w.submission_fiscal_month::text, 2, '0')) AS fyp,
```
kwhickey, line 5: Can get rid of this line now. See updated CTE sql in sql files attached to https://federal-spending-transparency.atlassian.net/browse/DEV-5283
- it gets rid of messing around with left-padding

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -1,6 +1,65 @@
...
+        w.period_start_date >= '2020-04-01' -- using older date to see data surface
+        and w.submission_reveal_date <= now() -- change "period end" with "window close"
+    order by w.submission_fiscal_year desc, w.submission_fiscal_month desc
+),
+    eligible_submissions AS (
+        SELECT *
+        FROM submission_attributes s
+        WHERE concat(s.reporting_fiscal_year::text, lpad(s.reporting_fiscal_period::text, 2, '0')) IN (
+            SELECT fyp FROM closed_periods
+        )
```
kwhickey, line 20: Add in `w.is_quarter` and join to it in the `eligible_submissions` CTE as per sql attached to https://federal-spending-transparency.atlassian.net/browse/DEV-5283

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -1,6 +1,65 @@
...
+        -- final period's submission (month or quarter), otherwise assume it to be zero
+        SELECT faba.award_id, sum(faba.gross_outlay_amount_by_award_cpe) AS prior_fys_outlay,
+            faba.disaster_emergency_fund_code
+        FROM eligible_file_c_records faba
+        GROUP BY
+            faba.award_id,
+            faba.reporting_fiscal_period,
+            faba.disaster_emergency_fund_code
+        HAVING faba.reporting_fiscal_period = 12
+        AND sum(faba.gross_outlay_amount_by_award_cpe) > 0
```
kwhickey, line 46: These two `having` clauses are actually better suited to just a `where` clauses. See updated SQL attached to https://federal-spending-transparency.atlassian.net/browse/DEV-5283

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -1,6 +1,65 @@
...
+            faba.disaster_emergency_fund_code,
+            sum(faba.gross_outlay_amount_by_award_cpe) AS current_fy_outlay
+        FROM eligible_file_c_records faba
+        GROUP BY
+            faba.award_id,
+            faba.reporting_fiscal_year,
+            faba.reporting_fiscal_period,
+            faba.disaster_emergency_fund_code
+        HAVING concat(faba.reporting_fiscal_year::text, lpad(faba.reporting_fiscal_period::text, 2, '0')) IN
+        (SELECT max(fyp) FROM closed_periods) AND sum(faba.gross_outlay_amount_by_award_cpe) > 0)
```
kwhickey, line 62: Again the `having` is better suited as a `where`. See updated SQL attached to https://federal-spending-transparency.atlassian.net/browse/DEV-5283

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -155,17 +167,15 @@ LEFT JOIN (
...
-        ARRAY_AGG(DISTINCT faba.disaster_emergency_fund_code) AS defc
-    FROM eligible_file_c_records faba
-    LEFT JOIN fy_final_outlay_balances ffy ON ffy.award_id = faba.award_id
-    LEFT JOIN current_fy_outlay_balance cfy
-	    ON cfy.reporting_fiscal_period != 12 -- don't duplicate the year-end period's value if in unclosed period 01
-	    AND cfy.award_id = faba.award_id
-    GROUP BY faba.award_id, total_covid_outlay
+        coalesce(sum(faba.transaction_obligated_amount), 0) as total_covid_obligation,
+        coalesce(sum(ffy.prior_fys_outlay), 0) + coalesce(sum(cfy.current_fy_outlay), 0) as total_covid_outlay,
+        ARRAY_AGG(faba.disaster_emergency_fund_code) AS disaster_emergency_fund_codes
```
tony-sappe, line 174: Isn't the DISTINCT beneficial?

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -2,64 +2,76 @@ DROP VIEW IF EXISTS award_delta_view;
...
+        faba.transaction_obligated_amount,
+        faba.gross_outlay_amount_by_award_cpe,
+        s.reporting_fiscal_year,
+        s.reporting_fiscal_period,
+        s.submission_id
+    from financial_accounts_by_awards faba
+    inner join submission_attributes s on faba.submission_id=s.submission_id
+    inner join disaster_emergency_fund_code defc
+    	on defc.group_name = 'covid_19'
+		and defc.code = faba.disaster_emergency_fund_code
```
kwhickey, line 38: This and further replacements below may have swung _too_ far in the new direction. Since this will only push COVID DEFCs to the ES index. The mock for Adv. Search _does_ only show filtering on covids (LMNOP), so this may be okay for now. 

If you want to keep it open for all DEFCs, I think the way you had it doing a sum of value based on CASE and grouping by award+defc may be the way to go.

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -92,6 +95,80 @@ LEFT JOIN (
...
+            faba.award_id,
+            faba.reporting_fiscal_year,
+            faba.reporting_fiscal_period,
+            faba.disaster_emergency_fund_code
+        HAVING concat(faba.reporting_fiscal_year::text, lpad(faba.reporting_fiscal_period::text, 2, '0')) IN
+        (SELECT max(fyp) FROM closed_periods) AND sum(faba.gross_outlay_amount_by_award_cpe) > 0)
+    SELECT
+        faba.award_id,
+        coalesce(ffy.prior_fys_outlay, 0) + coalesce(cfy.current_fy_outlay, 0) AS total_covid_obligations,
+        sum(faba.transaction_obligated_amount) AS total_covid_outlays,
```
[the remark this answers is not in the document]
tony-sappe: bump

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -167,15 +95,32 @@ LEFT JOIN (
...
+        ARRAY_AGG(DISTINCT disaster_emergency_fund_code) AS disaster_emergency_fund_codes,
+        COALESCE(sum(CASE WHEN latest_closed_period_per_fy.is_quarter IS NOT NULL THEN faba.gross_outlay_amount_by_award_cpe END), 0) AS gross_outlay_amount_by_award_cpe,
+        COALESCE(sum(faba.transaction_obligated_amount), 0) AS transaction_obligated_amount
+    FROM
+        financial_accounts_by_awards faba
+    INNER JOIN disaster_emergency_fund_code defc
+        ON defc.code = faba.disaster_emergency_fund_code
+        AND defc.group_name = 'covid_19'
+    INNER JOIN submission_attributes sa
+        ON faba.submission_id = sa.submission_id
```
kwhickey, line 109: Should be 
```suggestion
    INNER JOIN submission_attributes sa 
        ON sa.submission_id = faba.submission_id
        AND sa.reporting_period_start >= '2020-04-01'
```

Which allows us to pre-filter out faba records that are from "old" submissions where covid data was not present. Because of the `LEFT JOIN` suggestion below, reducing the number of faba records before the left join is ideal.

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -167,15 +95,32 @@ LEFT JOIN (
...
+    INNER JOIN (
```
kwhickey, line 110: Should be 

```suggestion
    LEFT JOIN (
```

Otherwise the INNER JOIN would be cutting out `transaction_obligated_amount` values from rows that were in non-latest, non-year-end submissions.

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -167,15 +95,32 @@ LEFT JOIN (
...
+        SELECT   submission_fiscal_year, is_quarter, max(submission_fiscal_month) AS submission_fiscal_month
+        FROM     dabs_submission_window_schedule
+        WHERE    submission_reveal_date < now() AND period_start_date >= '2020-04-01'
+        GROUP BY submission_fiscal_year, is_quarter
+    ) AS latest_closed_period_per_fy
+        ON latest_closed_period_per_fy.submission_fiscal_year = sa.reporting_fiscal_year
+        AND latest_closed_period_per_fy.submission_fiscal_month = sa.reporting_fiscal_period
+        AND latest_closed_period_per_fy.is_quarter = sa.quarter_format_flag
+GROUP BY
```
kwhickey, line 119:
```suggestion
        AND latest_closed_period_per_fy.is_quarter = sa.quarter_format_flag
WHERE faba.award_id IS NOT NULL
GROUP BY
```

Can limit rows being summed-across in this sub-query even further by dropping-out those that you could never join to because their `award_id` is `NULL`

kwhickey:
Added `Do not merge` label until we figure out why/how the `mv_award_search` matview stuff got into this PR, as it is not deemed necessary yet and shouldn't be propagated to `dev` branch.

PR #2540, opened 2020-06-22
[DEV-5427] Implement Disaster Federal Account Count endpoint

sethstoudenmier:
**Description:**
Implement an endpoint that returns the count of Federal Accounts that belong to the supplied DEF Codes.

**Technical details:**
Implement `v2/disaster/federal_account/count/` which accepts a `def_codes` filter and returns a count of the Federal Accounts that fall under those codes.

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
8. [X] Jira Ticket [DEV-5427](https://federal-spending-transparency.atlassian.net/browse/DEV-5427):

```
```


usaspending_api/disaster/v2/views/federal_account_count.py
```diff
@@ -0,0 +1,37 @@
...
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/disaster/federal_account/count.md"
+
+    @cache_response()
+    def post(self, request: Request) -> Response:
+        filters = [
+            Q(final_of_fy=True),
+            Q(treasury_account__federal_account_id=OuterRef("pk")),
+            Q(disaster_emergency_fund__code__in=self.def_codes),
+            Q(submission__reporting_period_start__gte="2020-04-01"),
```
tony-sappe, line 26: Probably a good idea to make this date an attribute to the `DisasterBase` since it can be used for all views

usaspending_api/disaster/tests/integration/conftest.py
```diff
@@ -0,0 +1,290 @@
...
+        resp = client.post(url, content_type="application/json", data=request_body)
+        return resp
+
+
+@pytest.fixture
+def helpers():
+    return Helpers
+
+
+__all__ = ["disaster_account_data", "helpers"]
```
BrianZito, line 290: I think we should consider splitting up this file. These are the tests for one endpoint and we're already looking at 300 lines.

sethstoudenmier: These aren't the actual tests but fixtures that will be used by multiple tests. For example, the `post_for_count_endpoint()` function will be used by all four count endpoints. Can still break these up and import into conftest.py to break up files and still have the desired functionality.

usaspending_api/disaster/v2/views/federal_account_count.py
sethstoudenmier: Good call, will add that.

usaspending_api/disaster/tests/integration/conftest.py
BrianZito: Misspoke, they are the fixtures for the tests of one group. Even then, I fear this growing into a 1000 line file where nobody knows which fixtures go to which test. I would still favor splitting.

sethstoudenmier: Split the fixtures into different files and imported them into conftest for the same effect without over cluttering conftest.py.

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -49,6 +51,10 @@ def filters(self):
     def def_codes(self):
         return self.filters["def_codes"]
 
+    @property
+    def reporting_period_max(self):
+        return datetime.now().date()
```
tony-sappe, line 56: If max date is always "today" is it necessary?

sethstoudenmier: Wouldn't we want max date to make sure that we only consider submissions that are not part of an active submission window?

tony-sappe: Yes, but that date will come from the future `dabs_submission_window_schedule` table

usaspending_api/submissions/helpers.py
```diff
@@ -0,0 +1,22 @@
+from datetime import date
+
+from usaspending_api.submissions.models import DABSSubmissionWindowSchedule
+
+
+def get_last_closed_submission_date(is_quarter: bool) -> dict:
+    current_date = date.today()
```
tony-sappe, line 7: What prompted the change from datetime.now() to date.today()?

Since these are local times, it is always a good idea to make them TZ-aware like `datetime.now(timezone.utc)`

usaspending_api/disaster/tests/fixtures/disaster_account_data.py
```diff
@@ -0,0 +1,290 @@
...
+@pytest.fixture
+def disaster_account_data():
+    ta1 = mommy.make("references.ToptierAgency", toptier_code="007")
+    ta2 = mommy.make("references.ToptierAgency", toptier_code="008")
+    ta3 = mommy.make("references.ToptierAgency", toptier_code="009")
+    ta4 = mommy.make("references.ToptierAgency", toptier_code="010")
+
+    sub1 = mommy.make(
+        "submissions.SubmissionAttributes",
+        reporting_fiscal_year="2022",
```
tony-sappe, line 15: I'm surprised strings work better for these fields

sethstoudenmier: I can revert this back to integers. Was left as part of my attempts to get tests working for Travis.

usaspending_api/submissions/helpers.py
sethstoudenmier: Didn't think about that but can make the change.

usaspending_api/disaster/tests/fixtures/disaster_account_data.py
tony-sappe: Doesn't matter to me if it works

PR #2552, opened 2020-06-25
[DEV-5291] Disaster/Emergency Spending by Federal Account

tony-sappe:
**Description:**
New endpoints:

- api/v2/disaster/federal_account/spending/
- api/v2/disaster/federal_account/loans/

**Technical details:**
Lots going on in this PR... One liberty I took was assuming that obtaining all of the TAS from the DB and performing the rollups, pagination, etc was faster than running multiple DB queries to obtain all of the needed info. Queries are run against File B or File C. Loan queries run on File C but aggregate `face_value_of_loan` which is from File D2.

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [ ] Necessary PR reviewers:
    - [x] Backend
    - [ ] Frontend
    - [ ] Domain Expert
4. [x] Matview impact assessment completed (N/A)
5. [x] Frontend impact assessment completed
6. [x] Data validation completed (N/A)
7. [x] Appropriate Operations ticket(s) created (N/A)
8. [x] Jira Ticket [DEV-5291](https://federal-spending-transparency.atlassian.net/browse/DEV-5291):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected API
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #2553, opened 2020-06-26
Dev 5424 overview endpoint

BrianZito:
**Description:**
Builds out API for Disaster Overview endpoint.

**Technical details:**

**Performance**
Endpoint now takes 14 seconds in sandbox with a cold cache, or <1 second with a warm cache.

2. [x] API documentation updated (N/A, contracts written in separate PR)
    - [x] Backend
8. [x] Jira Ticket [DEV-5424](https://federal-spending-transparency.atlassian.net/browse/DEV-5424):
    - [x] Performance evaluation of affected (API | Script | Download) (N/A, new endpoint)
    - [x] Before / After data comparison (N/A, new endpoint)

```
```


usaspending_api/disaster/v2/urls.py
```diff
@@ -5,8 +5,11 @@
 from usaspending_api.disaster.v2.views.federal_account.count import FederalAccountCountViewSet
 from usaspending_api.disaster.v2.views.object_class.count import ObjectClassCountViewSet
 from usaspending_api.disaster.v2.views.spending import ExperimentalDisasterViewSet
+from usaspending_api.disaster.v2.views.overview import OverviewViewSet
 
 urlpatterns = [
+    url(r"^spending/$", ExperimentalDisasterViewSet.as_view()),
```
tony-sappe, line 11: `spending/` is duplicated

usaspending_api/disaster/v2/views/overview.py
```diff
@@ -0,0 +1,94 @@
+from rest_framework.response import Response
+from usaspending_api.disaster.v2.views.disaster_base import DisasterBase
+from usaspending_api.common.cache_decorator import cache_response
+from usaspending_api.disaster.v2.views.disaster_base import covid_def_code_strings, latest_gtas_of_each_year
+from usaspending_api.awards.models.financial_accounts_by_awards import FinancialAccountsByAwards
+from django.db.models.functions import Coalesce
+from django.db.models import Sum
+
+# Limits the amount of results the spending explorer returns
+SPENDING_EXPLORER_LIMIT = 500
```
tony-sappe, line 10: Doesn't appear used

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -1,12 +1,57 @@
...
+
+def covid_def_codes():
+    return DisasterEmergencyFundCode.objects.filter(group_name=COVID_19_GROUP_NAME)
+
+
+def covid_def_code_strings():
+    return [code["code"] for code in covid_def_codes().values("code")]
+
+
+def latest_gtas_of_each_year():
```
tony-sappe, line 24: Might be a good idea to add some indication in the function name that this returns a Django queryset. Same with `latest_faba_of_each_year()`

usaspending_api/disaster/v2/views/overview.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+
+    def award_outlays(self):
+        return sum(
+            [
+                elem["amount"]
+                for elem in FinancialAccountsByAwards.objects.filter(
+                    disaster_emergency_fund__in=covid_def_code_strings()
+                )
+                .annotate(amount=Coalesce("gross_outlay_amount_by_award_cpe", 0))
```
tony-sappe, line 73: `gross_outlay_amount_by_award_cpe` is a cumulative value for the current fiscal year. Might be required to do something like this https://github.com/fedspendingtransparency/usaspending-api/pull/2552/files#diff-ca0fba873f0e3d9d237772a4877e7d19R129-R171

BrianZito:
Update: with correction to the award outlays, the endpoint is now no longer able to run before timeout. Working on a solution...

BrianZito:
Update: while the endpoint is still slow with cold cache, it is no longer at risk of gateway timeout, so I'll remove the strict "do not merge"

usaspending_api/disaster/v2/views/overview.py
```diff
@@ -0,0 +1,87 @@
...
+            "total_outlays": self.total_outlays(),
+        }
+
+    def award_obligations(self):
+        return (
+            FinancialAccountsByAwards.objects.filter(disaster_emergency_fund__in=covid_def_code_strings())
+            .values("transaction_obligated_amount")
+            .aggregate(total=Sum("transaction_obligated_amount"))["total"]
+            or 0.0
+        )
```
kwhickey, line 68: If it saves on performance, I believe it was deemed acceptable to limit the look-back of this query to FY 2020 P07 and beyond (April 2020 data and beyond). 

This would need to join in the `submission_attributes` table and sift-out file C records from submissions made for any period prior to that.

usaspending_api/disaster/v2/views/overview.py
```diff
@@ -0,0 +1,87 @@
...
+        return (
+            latest_faba_of_each_year_queryset()
+            .annotate(amount=Coalesce("gross_outlay_amount_by_award_cpe", 0))
+            .aggregate(total=Sum("amount"))["total"]
+        ) or 0.0
+
+    def total_obligations(self, funding):
+        return (
+            sum([elem["amount"] for elem in funding])
+            - latest_gtas_of_each_year_queryset().values("unobligated_balance_cpe").first()["unobligated_balance_cpe"]
```
kwhickey: Why `first()`? Does it get the most-recent? The max?

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -1,11 +1,86 @@
...
+            SubmissionAttributes.objects.values("reporting_fiscal_year")
+            .annotate(fiscal_period_max=Max("reporting_fiscal_period"))
+            .values("reporting_fiscal_year", "fiscal_period_max")
+            .filter(
+                reporting_fiscal_year=OuterRef("reporting_fiscal_year"),
+                reporting_fiscal_year__gte=2020,
+                fiscal_period_max=OuterRef("reporting_fiscal_period"),
+            )
+        )
+    ).filter(include=True)
```
kwhickey, line 68: This needs to reference "latest closed period" by way of the `DABSSubmissionWindowSchedule` model/table. That is the official source of what periods were available, and which of them is considered "closed" (by way of the `submission_reveal_date` field).

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -1,11 +1,86 @@
...
+                )
+                .values("fiscal_year", "fiscal_period_max")
+            )
+        )
+        .annotate(fiscal_period_and_year=Concat("fiscal_year", Value("/"), "fiscal_period", output_field=CharField()))
+        .filter(
+            is_highest_period=True,
+            fiscal_period_and_year__in=[
+                f"{elem['submission_fiscal_year']}/{elem['submission_fiscal_month']}"
+                for elem in DABSSubmissionWindowSchedule.objects.filter(is_quarter=False)
```
kwhickey: I'll have to see the SQL on this, but `DABSSubmissionWindowSchedule` was already interrogated above. Not sure if it needs to be brought into this query again.

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -1,11 +1,86 @@
...
+                    disaster_emergency_fund_code__in=covid_def_code_strings(),
+                )
+                .values("fiscal_year", "fiscal_period_max")
+            )
+        )
+        .annotate(fiscal_period_and_year=Concat("fiscal_year", Value("/"), "fiscal_period", output_field=CharField()))
+        .filter(
+            is_highest_period=True,
+            fiscal_period_and_year__in=[
+                f"{elem['submission_fiscal_year']}/{elem['submission_fiscal_month']}"
```
kwhickey: Is this concat the only way to do a multi-col `IN` clause in Django ORM?

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -1,11 +1,86 @@
...
+def latest_gtas_of_each_year_queryset():
+    return (
+        GTASSF133Balances.objects.annotate(
+            is_highest_period=Exists(
+                GTASSF133Balances.objects.values("fiscal_year")
+                .annotate(
+                    fiscal_period_max=Max("fiscal_period"),
+                    include=Exists(
+                        DABSSubmissionWindowSchedule.objects.filter(
+                            is_quarter=False,
```
kwhickey: This is good here to filter on non-quarters, because GTAS data is only monthly.

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -1,11 +1,86 @@
...
+def latest_submission_of_each_year_queryset():
+    return SubmissionAttributes.objects.annotate(
+        include=Exists(
+            SubmissionAttributes.objects.values("reporting_fiscal_year")
+            .annotate(fiscal_period_max=Max("reporting_fiscal_period"))
+            .values("reporting_fiscal_year", "fiscal_period_max")
+            .filter(
+                reporting_fiscal_year=OuterRef("reporting_fiscal_year"),
+                reporting_fiscal_year__gte=2020,
+                fiscal_period_max=OuterRef("reporting_fiscal_period"),
```
kwhickey, line 65: Further, in this case we _do_ need to factor in `DABSSubmissionWindowSchedule.is_quarter` as part of the uniqueness/group that is returned. This is because we need to grab the latest closed month AND latest closed quarter. 

At end of FYs, e.g. end of FY2020 in nov/dec ... that would be:
1. `submission_fiscal_year=2020, submission_fiscal_quarter=4, submission_fiscal_month=12, is_quarter=True`
1. `submission_fiscal_year=2020, submission_fiscal_quarter=4, submission_fiscal_month=12, is_quarter=False`

Which are two different submission windows/periods. 

But Around Aug 5, 2020, before Q3 has closed by after P 7-9 has closed, it should return:
1. `submission_fiscal_year=2020, submission_fiscal_quarter=2, submission_fiscal_month=6, is_quarter=True`
1. `submission_fiscal_year=2020, submission_fiscal_quarter=3, submission_fiscal_month=9, is_quarter=False`

usaspending_api/disaster/v2/views/overview.py
BrianZito: This has been since updated.

usaspending_api/disaster/v2/views/overview.py
tony-sappe: Could use `reporting_period_end` in `FinancialAccountsByAwards` for the early cutoff date

usaspending_api/disaster/v2/views/overview.py
kwhickey: Updated and pushed? I still see `first()` in the code. Not commenting that it's wrong, just wanted to verify. It looks like you're ordering by FY, Q, P (all desc) ... so `first` would be good here.

usaspending_api/disaster/v2/views/disaster_base.py
kwhickey: This is a corner-case btw, and grabbing only the latest submitted _month_ should cover the vast majority of data that we are after. But it is possible that agencies will still submit COVID data in _quarterly_ submissions -- Broker has nothing in place yet to prevent this.

usaspending_api/disaster/v2/views/overview.py
BrianZito: The new code (which appears after reloading the PR, can't speak to what's happening there) still uses first(), but filters for the current year. We could add a check for length just to make sure the previous query isn't returning more than one result per row.

usaspending_api/disaster/v2/views/overview.py
```diff
@@ -73,10 +75,12 @@ def award_outlays(self):
...
-        return (
-            sum([elem["amount"] for elem in funding])
-            - latest_gtas_of_each_year_queryset().values("unobligated_balance_cpe").first()["unobligated_balance_cpe"]
-        )
+        remaining_balance = latest_gtas_of_each_year_queryset().filter(fiscal_year=current_fiscal_year()).first()
+        if remaining_balance:
+            remaining_balance = remaining_balance.unobligated_balance_cpe
+        else:
+            remaining_balance = Decimal("0.0")
+        return sum([elem["amount"] for elem in funding]) - remaining_balance
```
kwhickey, line 83: Is this accurate for how this is calculated? I would have though total obligation was just a sum of obligations, rather than `budget - unobligated balance`. 

It doesn't appear to be annotated in the Sankey on DEV-5360, but in DEV-5050, we were purposely summing `obligations_incurred_total_cpe` (line 2190) I thought to get this amount directly from GTAS.

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -27,41 +26,27 @@ def covid_def_code_strings():
...
+        .filter(is_quarter=False, submission_reveal_date__lte=datetime.now(timezone.utc), has_gtas=True)
+        .values("submission_fiscal_year")
+        .annotate(max_submission_fiscal_month=Max("submission_fiscal_month"))
+        .values_list("submission_fiscal_year", "max_submission_fiscal_month")
     )
+    q = Q()
+    for final_for_fy in finals_for_fy:
+        q |= Q(fiscal_year=final_for_fy[0]) & Q(fiscal_period=final_for_fy[1])
+    if not q:
+        return GTASSF133Balances.objects.none()
```
kwhickey, line 48: I think this is more of an error-condition rather than a "return none" scenario, yes? We should not run into the scenario where we have no data (no periods to query data for), and if we do, likely want to throw an error.

BrianZito: Is there value in sending the user an error that the devs need to  reload the data? This is still a scenario that makes sense from the API's perspective. It also helps in testing environments where this table could be empty.

usaspending_api/disaster/v2/views/overview.py
BrianZito: The  closest thing I have to business logic for these values is the attachment on https://federal-spending-transparency.atlassian.net/browse/DEV-5424, which says to subtract these values.

PR #2556, opened 2020-06-29
[DEV-4443] AC 1 - Don't carry forward TAS in account downloads

kbard:
**Description:**

Before we were returning the final TAS rows for a fiscal year regardless of the quarter in which it falls.

Now, for Files A and B, we only return TAS rows in the quarter or period requested.  For File C we only return rows in the quarter or period requested OR previous periods if the `transaction_obligated_amount` is non-zero.

Also "fixed" a few typos and bad imports.

And updated the account download endpoint to accepts quarter OR periods.  Is backward compatible so front end should be unaffected until they are ready to support periods.

    - [x] Backend
5. [x] Frontend unaffected (new contract stuff, but backward compatible)
8. [x] Jira Ticket [DEV-4443](https://federal-spending-transparency.atlassian.net/browse/DEV-4443):

### DIFF
--- a/usaspending_api/api_contracts/contracts/v2/download/accounts.md
+++ b/usaspending_api/api_contracts/contracts/v2/download/accounts.md
@@ -95,14 +95,14 @@ Generate files and return metadata using filters on custom account
 + `fy` (required, string)
     The fiscal year to filter by in the format `YYYY`
 + `quarter` (optional, enum[string])
-    Either `quarter` or `period` is required.  Do not supply both.
+    Either `quarter` or `period` is required.  Do not supply both.   Note that both monthly and quarterly submissions will be included in the resulting download file even if only `quarter` is provided.
     + Members
         + `1`
         + `2`
         + `3`
         + `4`
 + `period` (optional, enum[string])
-    Either `quarter` or `period` is required.  Do not supply both.  Agencies cannot submit data for period 1 so it is disallowed as a query filter.
+    Either `quarter` or `period` is required.  Do not supply both.  Agencies cannot submit data for period 1 so it is disallowed as a query filter.   Note that both monthly and quarterly submissions will be included in the resulting download file even if only `period` is provided.
     + Members
         + `2`
         + `3`
--- a/usaspending_api/api_contracts/contracts/v2/download/accounts.md
+++ b/usaspending_api/api_contracts/contracts/v2/download/accounts.md
@@ -94,9 +94,24 @@ Generate files and return metadata using filters on custom account
         + `award_financial`
 + `fy` (required, string)
     The fiscal year to filter by in the format `YYYY`
-+ `quarter` (required, enum[string])
++ `quarter` (optional, enum[string])
+    Either `quarter` or `period` is required.  Do not supply both.
     + Members
         + `1`
         + `2`
         + `3`
         + `4`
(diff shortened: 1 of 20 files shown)


usaspending_api/api_contracts/contracts/v2/download/accounts.md
```diff
@@ -94,9 +94,24 @@ Generate files and return metadata using filters on custom account
...
++ `quarter` (optional, enum[string])
+    Either `quarter` or `period` is required.  Do not supply both.   Note that both monthly and quarterly submissions will be included in the resulting download file even if only `quarter` is provided.
     + Members
         + `1`
         + `2`
         + `3`
         + `4`
++ `period` (optional, enum[string])
+    Either `quarter` or `period` is required.  Do not supply both.  Agencies cannot submit data for period 1 so it is disallowed as a query filter.   Note that both monthly and quarterly submissions will be included in the resulting download file even if only `period` is provided.
+    + Members
```
MaxwellKendall, line 106: @kbard 

Is there no period 1?

Also, can you confirm the return implicitly includes previous periods w/in the same FY?

For example, if I request period 4, it will return data for all periods in that FY less than or equal to period 4.

kbard: There is no period 1.

As for your second question, the logic is fairly complicated.  From in-code comments:
```
# For Files A and B, we only care about the specific period or quarter requested; e.g. "get
# me Period 4 monthly submissions AND Quarter 1 quarterly submissions".  This is because
# monthly and quarterly submissions are on different tracks but both need to be included in
# the download file.

# For File C, we include rows in the specific period/quarter requested AND rows from
# previous periods or quarters where the transaction_obligated_amount is non-zero.
```

So the answer is "it depends".  Some downloads return previous periods some do not.

kbard: So to follow-up on this from a Slack call Max and I had, since our current use case is always "from the beginning of the FY up to and including the quarter/period selected", all the front end needs to return is the FY and quarter or period for all endpoints that use the period/period picker.  The endpoint will handle the logic for deciding whatever needs to be returned.

usaspending_api/common/helpers/fiscal_year_helpers.py
```diff
@@ -46,7 +48,7 @@ def generate_fiscal_year(date):
     return year
 
 
-def generate_fiscal_month(date):
+def generate_fiscal_period(date):
```
tony-sappe, line 51: I tend to prefer "month" over "period" since the "time period" could be in years, quarters, or months. I understand there are inconsistencies with source data, so any normalization we can do it appreciated

usaspending_api/common/helpers/orm_helpers.py
```diff
@@ -145,7 +191,3 @@ def subaward_types_are_valid_groups(type_list):
     is_procurement = set(type_list).difference(set(procurement_type_mapping.keys()))
     is_assistance = set(type_list).difference(set(assistance_type_mapping.keys()))
     return bool(is_procurement) != bool(is_assistance)  # clever XOR logic
-
-
-def category_to_award_materialized_views():
```
tony-sappe, line 150: Thanks, I noticed that isn't used too

PR #2568, opened 2020-07-02
[DEV-5546] New api/v2/disaster/award/amount/ endpoint

tony-sappe:
**Description:**
Provides the File C award outlay and obligation totals. Filterable by DEFC and award type code

**Technical details:**
Also cleaned up previously created tests in the new disaster app

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [x] Necessary PR reviewers:
    - [x] Backend
    - [x] Frontend
4. [x] Matview impact assessment completed (N/A)
5. [x] Frontend impact assessment completed
6. [x] Data validation completed
7. [x] Appropriate Operations ticket(s) created (N/A)
8. [x] Jira Ticket [DEV-5546](https://federal-spending-transparency.atlassian.net/browse/DEV-5546):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected API
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
No edits to database required
```

PR #2567, opened 2020-07-01
[DEV-5291] Adding obligation and outlay to Federal Account Loans response

tony-sappe:
**Description:**
Updates to design mocks surfaced a shortcoming on the initial implementation of the loans endpoint for Federal Account data. Now the loans response includes obligation and outlays from File C. These changes will need to be included in the future loans endpoints.

**Technical details:**
New code is basically a copy/paste from spending endpoint

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated (API Contract updates in #2562)
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [ ] Frontend
4. [x] Matview impact assessment completed (N/A)
5. [x] Frontend impact assessment completed
6. [x] Data validation completed
7. [x] Appropriate Operations ticket(s) created (N/A)
8. [x] Jira Ticket [DEV-5291](https://federal-spending-transparency.atlassian.net/browse/DEV-5291):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected (API)
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
Update to API response doesn't require changes to database
```

PR #2569, opened 2020-07-02
award count endpoint

BrianZito:
**Description:**
Adds count endpoint for awards

**Technical details:**
Includes some refactors for code reuse with other count endpoints where things were obviously duplicated.

    - [x] Backend
8. [x] Jira Ticket [DEV-5541](https://federal-spending-transparency.atlassian.net/browse/DEV-5541):

```
```


usaspending_api/disaster/v2/views/count_base.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def is_non_zero_award_cpe(self):
+        return Q(
+            Q(obligations_incurred_total_by_award_cpe__gt=0)
+            | Q(obligations_incurred_total_by_award_cpe__lt=0)
+            | Q(gross_outlays_delivered_orders_paid_total_cpe__gt=0)
+            | Q(gross_outlays_delivered_orders_paid_total_cpe__lt=0)
+        )
+
+    def is_last_closed_submission_window(self):
```
tony-sappe, line 28: `is_last_closed_submission_window()`is obsolete, see #2567

usaspending_api/disaster/v2/views/count_base.py
```diff
@@ -0,0 +1,60 @@
+from django.db.models import Q
+from usaspending_api.disaster.v2.views.disaster_base import DisasterBase
+
+
+class CountBase(DisasterBase):
+    def is_after_min_date(self):
```
tony-sappe, line 6: `is_after_min_date()` is unnecessary if `self.all_closed_defc_submissions` is used after #2567 is merged

usaspending_api/disaster/v2/views/count_base.py
tony-sappe: Can this method be removed?

usaspending_api/disaster/v2/views/award/count.py
```diff
@@ -0,0 +1,36 @@
...
+    @cache_response()
+    def post(self, request: Request) -> Response:
+        filters = [
+            Q(award=OuterRef("pk")),
+            self.is_in_provided_def_codes(),
+            self.all_closed_defc_submissions,
+            self.is_non_zero_award_cpe(),
+        ]
+        count = (
+            Award.objects.annotate(include=Exists(FinancialAccountsByAwards.objects.filter(*filters).values("pk")))
```
tony-sappe, line 29: I learned on Thursday that for all of the "Award Spending By ___" sections, "All Awards" should return the amounts/counts for all awards in FABA including awards non-linked to File D awards' table. This new understanding will require the endpoint to be implemented differently:
#1: `award_type_codes` needs to be optional (already is)
#2: when `award_type_codes` isn't provided, the logic needs to search for all awards in File C instead of File D
#3: when `award_type_codes` is provided, this current logic can be executed

usaspending_api/disaster/v2/views/award/count.py
```diff
@@ -0,0 +1,36 @@
...
+            .filter(include=True)
+            .filter(self.is_provided_award_type())
```
tony-sappe, line 31: the two filter lines can be combined into a single .filter()

usaspending_api/disaster/v2/urls.py
```diff
@@ -11,6 +12,7 @@
 
 urlpatterns = [
     url(r"^overview/$", OverviewViewSet.as_view()),
+    url(r"^award/count/$", AwardCountViewSet.as_view()),
```
tony-sappe, line 15: Would be nice to add the endpoint in alphabetical order since other devs like to keep lists organized when possible

usaspending_api/disaster/v2/views/award/count.py
BrianZito: Unless there's a python trick I've missed, I can't combine the keyword argument with the positional one.

usaspending_api/disaster/v2/views/award/count.py
tony-sappe: Small comment in API contract to document behavior for API users: https://github.com/fedspendingtransparency/usaspending-api/pull/2568/files#diff-f78d44fe26b3dbca0ed58bde0cc1a853L40-R43

usaspending_api/disaster/v2/views/award/count.py
tony-sappe: Does `.filter(self.is_provided_award_type(), include=True)` not work?

usaspending_api/disaster/v2/views/award/count.py
tony-sappe: Re 1: I noticed that while the filter is optional in TinyShield models, the logic is expecting the key to exist and throws a 500 when the filter is not provided.

BrianZito: I observed the same. Fortunately the same if statement to implement the above logic protects from this issue

usaspending_api/disaster/v2/views/award/count.py
```diff
@@ -19,18 +19,26 @@ class AwardCountViewSet(CountBase):
...
+                .values("pk")
+                .count()
+            )
+        else:
+            filters = [
+                self.is_in_provided_def_codes(),
+                self.all_closed_defc_submissions,
+                self.is_non_zero_award_cpe(),
+            ]
+            count = FinancialAccountsByAwards.objects.filter(*filters).values("award").count()
```
tony-sappe, line 42: This is counting values of the FK to `awards`, which is duplicated and will be null for all records which are not linked to File D.

Here's an example of counting the FABA awards which should work: https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/disaster/v2/views/award/amount.py#L26

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -103,6 +103,10 @@ def filters(self):
     def def_codes(self):
         return self.filters["def_codes"]
 
+    @property
+    def award_type_codes(self):
```
tony-sappe, line 107: This property is defined in `AwardTypeMixin`. I think it will be OK to include in the CountBase definition

usaspending_api/api_contracts/contracts/v2/disaster/award/count.md
```diff
@@ -42,7 +42,8 @@ This endpoint returns a count of Awards
 ## Filter (object)
 + `def_codes` (required, array[DEFC], fixed-type)
 + `award_type_codes` (optional, array[AwardTypeCodes], fixed-type)
-    Defaults to all Award Type Codes.
+    When Award Type Codes are provided, results are only counted if the awards are linked between Financial Account by Awards and Awards (aka C-D Linkage).
```
ebdabbs, line 45:
```suggestion
    When Award Type Codes are provided, results are only counted if the awards are linked between Financial Account by Awards and Awards (File C to File D linkage).
```

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -120,12 +120,14 @@ def all_closed_defc_submissions(self):
...
-                    & Q(submission__reporting_fiscal_period__lte=sub.fiscal_period)
+                q |= Q(
+                    submission__in=SubmissionAttributes.objects.filter(
+                        reporting_fiscal_year=sub.fiscal_year,
+                        quarter_format_flag=sub.is_quarter,
+                        reporting_fiscal_period__lte=sub.fiscal_period,
+                    )
                 )
-        return q & Q(submission__reporting_period_start__gte=str(self.reporting_period_min_date))
+        return q & ~Q(submission__toptier_code="nonsensical string")
```
BrianZito, line 130: Currently a hackish solution for demonstration purposes. For some reason, simply having any inner join on the postgres query dramatically improves performance, so I've added this condition that will always pass that needs the submission table.

tony-sappe: > simply having any inner join on the postgres query dramatically improves performance

For performance, inner joins are preferred over outer joins when the data allows for it. If Django is converting the ORM to SQL with outer joins, it can typically be switched to an inner join if a `__isnull` filter is provided

BrianZito: There don't appear to be any outer joins before or after this modification. Instead the difference is that in the origonal query:

SELECT
   "financial_accounts_by_awards"."award_id" 
FROM
   "financial_accounts_by_awards" 
   INNER JOIN
      "awards" 
      ON ("financial_accounts_by_awards"."award_id" = "awards"."id") 

submission_attributes is skipped (since there's no part of the where clause dependent on it), whereas when modified the joins expand to:

SELECT
   "financial_accounts_by_awards"."award_id" 
FROM
   "financial_accounts_by_awards" 
   INNER JOIN
      "submission_attributes" 
      ON ("financial_accounts_by_awards"."submission_id" = "submission_attributes"."submission_id") 
   INNER JOIN
      "awards" 
      ON ("financial_accounts_by_awards"."award_id" = "awards"."id") 


I'm still  trying to figure out why this would help performance instead of hurting, but the results of performance testing speak for themselves.

BrianZito: Update: It seems the critical difference is that with the join added, a bitmap heap scan:

{{Bitmap Heap Scan on financial_accounts_by_awards  (cost=10043.92..13195.98 rows=144 width=8)
              Recheck Cond: ((submission_id = submission_attributes.submission_id) AND ((disaster_emergency_fund_code)::text = ANY ('{L,M,N,O,P}'::text[])))
              Filter: ((obligations_incurred_total_by_award_cpe > '0'::numeric) OR (obligations_incurred_total_by_award_cpe < '0'::numeric) OR (gross_outlays_delivered_orders_paid_total_cpe > '0'::numeric) OR (gross_outlays_delivered_orders_paid_total_cpe < '0'::numeric))}}

Is converted to an index scan. Still investigating why.

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -120,10 +120,12 @@ def all_closed_defc_submissions(self):
                 sub.fiscal_year == self.reporting_period_min_year
                 and sub.fiscal_period >= self.reporting_period_min_month
             ) or sub.fiscal_year > self.reporting_period_min_year:
-                q |= (
-                    Q(submission__reporting_fiscal_year=sub.fiscal_year)
-                    & Q(submission__quarter_format_flag=sub.is_quarter)
-                    & Q(submission__reporting_fiscal_period__lte=sub.fiscal_period)
+                q |= Q(
```
tony-sappe, line 123: Are these changes necessary? The way the function is written it will run multiple sub-queries, one for each fiscal year (2020 forward)

usaspending_api/disaster/v2/views/count_base.py
tony-sappe: bump

usaspending_api/disaster/v2/views/award/count.py
```diff
@@ -0,0 +1,40 @@
...
+            self.is_in_provided_def_codes(),
+            self.all_closed_defc_submissions,
+            self.is_non_zero_award_cpe(),
+        ]
+        if self.award_type_codes:
+            count = (
+                FinancialAccountsByAwards.objects.filter(*filters)
+                .values("award_id")
+                .filter(award__type__in=self.filters.get("award_type_codes"))
+                .count()
```
tony-sappe, line 33: This count needs to be distinct as well since FABA -> awards is many to one

usaspending_api/disaster/tests/fixtures/helpers.py
```diff
@@ -36,8 +36,10 @@ def post_for_spending_endpoint(client, url, **kwargs):
         return resp
 
     @staticmethod
-    def post_for_count_endpoint(client, url, def_codes=None):
-        if def_codes:
+    def post_for_count_endpoint(client, url, def_codes=None, award_type_codes=None):
+        if award_type_codes:
+            request_body = json.dumps({"filter": {"def_codes": def_codes, "award_type_codes": award_type_codes}})
+        elif def_codes:
```
tony-sappe, line 42: Why check for def_codes if the above line operates as if they are always present?

Perhaps use similar conditionals as `post_for_amount_endpoint()`

usaspending_api/disaster/v2/views/award/count.py
```diff
@@ -0,0 +1,40 @@
...
+
+from usaspending_api.common.cache_decorator import cache_response
+from usaspending_api.disaster.v2.views.count_base import CountBase
+from usaspending_api.awards.models import FinancialAccountsByAwards
+from usaspending_api.disaster.v2.views.disaster_base import FabaOutlayMixin, AwardTypeMixin
+
+
+class AwardCountViewSet(CountBase, FabaOutlayMixin, AwardTypeMixin):
+    """
+    Obtain the count of Agencies related to supplied DEFC filter.
```
tony-sappe, line 13: Docstring is wrong

usaspending_api/disaster/tests/fixtures/helpers.py
BrianZito: They are not always present, as can be seen in https://github.com/fedspendingtransparency/usaspending-api/pull/2569/files#diff-a913bf1e1f3808c8731f33b728e68d3eR89 as well as a few other places.

tony-sappe: I could have rephrased the question, it's currently a non-issue since there aren't tests for when `award_type_codes` is present but `def_codes` is missing. A future dev can duplicate the conditional in `post_for_amount_endpoint()` to handle both if that test case is introduced.

usaspending_api/disaster/v2/views/award/count.py
BrianZito: Is this logic supposed to be unique on award id? The logic used by the amount endpoint (here transposed to https://github.com/fedspendingtransparency/usaspending-api/pull/2569/files#diff-5a427bd0e842bfcd87ac12ab1e696f58R154) uses four different FABA values.

tony-sappe: The four fields in FABA are used in an attempt to determine uniqueness when not joining to the awards table. When joining to the awards table, I assume counting the unique FKs would be more performant

PR #2582, opened 2020-07-09
Mod/dev 5504 es c to d linkages as mods

collinwr:
**Description:**
Creates a new **linked_date** field on the **FinancialAccountsByAward** model. This is set to the current date when a file c to d linkage is made. This new field is taken into consideration in the ElasticSearch delta loaders when determining what is "new"

**Technical details:**
**linked_date** is set when:
- During the **load_file_c** phase of the **load_submission** command. Set to current date if matching award if found, otherwise null.
- In the SQL files of the **update_file_c_linkages** command. Set to NOW() whenever **award_id** is set.
- When FABS/FPDS transactions are deleted. In both cases **linked_date** is set to NOW() when **award_id** is set to NULL. This happens before the transactions are deleted
    - FABS - In **delete_stale_fabs.py**
    - FPDS - In **award_helpers.py** (called by **delete_stale_fabs**)

This new date is used in the Delta Views when calculating the **update_date** field used to determine which rows from the view to load into ElasticSearch. 

    - [ ] Backend
8. [ ] Jira Ticket [DEV-5504](https://federal-spending-transparency.atlassian.net/browse/DEV-5504):


usaspending_api/etl/award_helpers.py
```diff
@@ -320,6 +320,7 @@ def prune_empty_awards(award_tuple: Optional[tuple] = None) -> int:
       UPDATE financial_accounts_by_awards
         SET
           update_date = now(),
+          linked_date = now(),
```
tony-sappe, line 323: Should this be `now()` or `null` ?

usaspending_api/database_scripts/etl/transaction_delta_view.sql
```diff
@@ -341,7 +344,8 @@ LEFT JOIN (
  WHERE
    faba.award_id IS NOT NULL
  GROUP BY
-   faba.award_id
+   faba.award_id,
+   faba.linked_date
```
tony-sappe, line 348: Seems problematic to group by `linked_date`. Would obtaining the max be a viable alternative?

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -207,6 +212,6 @@ LEFT JOIN (
  WHERE
    faba.award_id IS NOT NULL
  GROUP BY
-   faba.award_id
-) TREASURY_ACCT ON (TREASURY_ACCT.award_id = vw_award_search.award_id)
-;
+   faba.award_id,
+   faba.linked_date
```
tony-sappe, line 216: Same concern including in grouping

usaspending_api/broker/helpers/delete_stale_fabs.py
```diff
@@ -42,7 +42,7 @@ def delete_stale_fabs(ids_to_delete):
         )
     if delete_award_ids:
         # Financial Accounts by Awards
-        faba = 'UPDATE "financial_accounts_by_awards" SET "award_id" = null WHERE "award_id" IN ({});'
+        faba = 'UPDATE "financial_accounts_by_awards" SET "award_id" = null, "linked_date" = NOW() WHERE "award_id" IN ({});'
```
tony-sappe, line 45: Should linked_date be `now()` or `null`?

usaspending_api/etl/award_helpers.py
collinwr: If we set this to `null`, there will be no way to indicate to the ElasticSearch loader that this should be reloaded.

usaspending_api/database_scripts/etl/transaction_delta_view.sql
collinwr: I agree. Switching over to the MAX would be better

usaspending_api/etl/award_helpers.py
kwhickey: What's the scenario covered here?

i.e. why would this be called / why is the award_id being set to null?

If because the award is being deleted, then that will be covered by the delete-tracking of awards already baked into the nightly delta (I believe such a thing exists, but haven't looked for it).

Since the nighty delta goes after the award records first, if the award is now null, it won't be able to figure out _which_ **award** data in elasticsearch needs to be changed. While a `linked_date` indicates that faba record was changed, we can't discover what award was affected.

usaspending_api/broker/helpers/delete_stale_fabs.py
kwhickey: Same question as below:

> Assuming the deletion of last fabs transaction led to the award being deleted?..
> If because the award is being deleted, then that will be covered by the delete-tracking of awards already baked into the nightly delta (I believe such a thing exists, but haven't looked for it).
>
>Since the nighty delta goes after the award records first, if the award is now null, it won't be able to figure out _which_ **award** data in elasticsearch needs to be changed. While a `linked_date` indicates that faba record was changed, we can't discover what award was affected.

PR #2586, opened 2020-07-09
[DEV-5233] Create DEFC Filter for Advanced Search

emily-brents:
**Description:**
Adds Filtering on DEF codes to advanced search page.

**Technical details:**
Uses the DEF codes added to the elasticsearch index to allow filtering on DEF codes. This currently only applies to COVID-19 related codes as they are the only codes currently in the index. 
Subawards do not have COVID data related directly to them, so the subawards fields searches the Award index and find subawards related to those awards.

    - [ ] Backend
    - [x] Frontend 
8. [X] Jira Ticket [DEV-5233](https://federal-spending-transparency.atlassian.net/browse/DEV-5233):

```
```


usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
```diff
@@ -195,6 +195,7 @@ List of table columns
 + `treasury_account_components` (optional, array[TreasuryAccountComponentsObject], fixed-type)
 + `object_class` (optional, array[string])
 + `program_activity` (optional, array[number])
++ `def_codes` (optional, array[string])
```
BrianZito, line 198: On the disaster endpoints we use an enum for def_codes: https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/api_contracts/contracts/v2/disaster/award/spending.md#filter-object

Might as well copy it  over here.

usaspending_api/search/filters/postgres/defc.py
```diff
@@ -0,0 +1,17 @@
...
+    underscore_name = "def_codes"
+
+    @classmethod
+    def build_def_codes_filter(cls, queryset, filter_values):
+        query = {"def_codes": filter_values}
+        filter_query = QueryWithFilters.generate_awards_elasticsearch_query(query)
+        search = AwardSearch().filter(filter_query)
+        response = search.handle_execute()
+        award_ids = [res.to_dict()["award_id"] for res in response]
+        return Q(award_id__in=list(award_ids))
```
BrianZito, line 17: Has this been performance tested in sandbox? Using elasticsearch to generate a Postgres query is an interesting idea, but that IN statement on the award ids looks like a lot of trouble.

usaspending_api/search/tests/integration/test_spending_by_award.py
```diff
@@ -1148,6 +1150,71 @@ def _test_correct_response_for_recipient_id(client):
     assert resp.json().get("results") == expected_result, "Recipient ID filter does not match expected result"
 
 
+def _test_correct_response_for_def_codes(client):
```
BrianZito, line 1153: Since the test fixtures are pretty cluttered for these tests, can we add some extra tests that search for a different DEFC to ensure the award is NOT returned?

usaspending_api/search/filters/postgres/defc.py
emily-brents: Haven't tested directly on sandbox, hooked my local API to nonprod ES / DEV database and didn't see any problems

BrianZito:
When testing against sandbox, I was unable to return results for 

{
    "filters": {
        "time_period": [
            {
                "start_date": "2007-10-01",
                "end_date": "2020-09-30"
            }
        ],
        "award_type_codes": ["A","B","C","D"],
        "def_codes": ["L","M","N","O","P"]
    },
    "fields": [
    	"Award ID",
        "Recipient Name",
        "Start Date",
        "End Date",
        "Award Amount",
        "Awarding Agency",
        "Awarding Sub Agency",
        "Contract Award Type",
        "recipient_id"
    ],
    "page": 1,
    "limit": 60,
    "sort": "Award ID",
    "order": "desc",
    "subawards": false
}

This is the correct usage, right?

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
emily-brents: so be it

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
```diff
@@ -295,3 +295,6 @@ List of filterable award types
 - `IDV_C`
 - `IDV_D`
 - `IDV_E`
+
+## DEFC (enum[string])
+List of Disaster Emergency Fund (DEF) Codes (DEFC) defined by legislation at the time of writing
```
BrianZito, line 300: all of it... https://github.com/fedspendingtransparency/usaspending-api/blob/dev/usaspending_api/api_contracts/contracts/v2/disaster/recipient/count.md#members

emily-brents: do i look like i know how to read.

anyway should be fixed now

usaspending_api/api_contracts/contracts/v2/search/spending_by_award.md
```diff
@@ -195,6 +195,7 @@ List of table columns
 + `treasury_account_components` (optional, array[TreasuryAccountComponentsObject], fixed-type)
 + `object_class` (optional, array[string])
 + `program_activity` (optional, array[number])
++ `def_codes` (optional, array[DEFC], fixed-type)
```
ebdabbs, line 198: It looks like the frontend is expecting the `require` / `exclude` pattern used for NAICS, PSC, TAS. cc @MaxwellKendall

MaxwellKendall: True enough! @emily-brents this is a trivial change on the front end, and IIRC we acknowledged that it wasn't actually necessary to provide an `excluded` array for this end point given the shape of the data at the moment.

Let me know what you think and we can conform to just seing y'all an array.

emily-brents: Yeah, I remember we talked about it and I thought we were just using an array. There's no hierarchy to DEF codes so there's no need to have include/exclude, os a plain array makes the most sense

usaspending_api/search/filters/postgres/defc.py
kbard: I think this is always capped at a reasonable page size, isn't it?  Like 100?  So this might not get too out of hand.

PR #2593, opened 2020-07-10
DEV 5571 count toptier agencies from file D when agency codes are provided

BrianZito:
**Description:**

**Technical details:**
The documentation lists the data as coming from file C, since while the actual agencies are coming from D, the link to that file is coming from file C, and all the filters will be applied against file C instead of the file B for when award type filter is not applied.

    - [ ] Backend
    - [ ] Frontend 
8. [x] Jira Ticket [DEV-5571](https://federal-spending-transparency.atlassian.net/browse/DEV-5571):

```
```


usaspending_api/disaster/tests/fixtures/helpers.py
```diff
@@ -43,6 +43,7 @@ def post_for_count_endpoint(client, url, def_codes=None, award_type_codes=None):
             request_body = json.dumps({"filter": {"def_codes": def_codes}})
         else:
             request_body = json.dumps({"filter": {}})
+        print(request_body)
```
sethstoudenmier, line 46: Should remove this print statement

usaspending_api/disaster/v2/views/agency/count.py
```diff
@@ -1,34 +1,48 @@
...
 from usaspending_api.common.cache_decorator import cache_response
 from usaspending_api.disaster.v2.views.count_base import CountBase
+from usaspending_api.awards.models import FinancialAccountsByAwards
 from usaspending_api.financial_activities.models import FinancialAccountsByProgramActivityObjectClass
 from usaspending_api.references.models import ToptierAgency
+from usaspending_api.disaster.v2.views.disaster_base import AwardTypeMixin
 
 
-class AgencyCountViewSet(CountBase):
+class AgencyCountViewSet(CountBase, AwardTypeMixin):
```
sethstoudenmier, line 13: If using this mixin you should order it as `AgencyCountViewSet(AwardTypeMixin, CountBase)` and then you can remove the `required_filters` below since the mixin will override `required_filters` in DisasterBase.

usaspending_api/disaster/tests/integration/test_disaster_agency_count.py
```diff
@@ -26,6 +26,14 @@ def test_agency_count_success(client, monkeypatch, disaster_account_data, helper
     assert resp.data["count"] == 0
 
 
+@pytest.mark.django_db
+def test_agency_count_with_award_types(client, monkeypatch, faba_with_toptier_agencies, helpers):
```
sethstoudenmier, line 30: If you have time I would advise that you add an additional Award to your test data. Right now the test data only has two Awards if I am following correctly so having a count of 2 could be a false positive.

usaspending_api/api_contracts/contracts/v2/disaster/agency/count.md
```diff
@@ -41,6 +41,8 @@ This endpoint returns a count of Agencies
 
 ## Filter (object)
 + `def_codes` (required, array[DEFC], fixed-type)
++ `award_type_codes` (optional, array[AwardTypeCodes], fixed-type)
+    If ANY award type codes are provided, count will be calculated from file C data, while if this parameter is not provided count will be calculated from file B.
```
kwhickey, line 45: I'm don't think if "File C" and "File B" by themselves are in the vocabulary of the USAspending API. Hopefully we have not used those excessively for public API docs. It should refer to the domain that USAspending uses. 

Additionally, the count doesn't exactly come from File C, it comes from the `awards` records linked in File C. 

Maybe:
> ... count will be calculated form the distinct set of toptier agencies, whose subtier agencies funded awards linked to `FinancialAccountsByAwards` records (which are derived from DABS File C). While if this parameter is not provided count will be calculated from the distinct set of toptier agencies "owning" appropriations accounts used in funding this disaster (from `FinancialAccountsByProgramActivityObjectClass` records, which are derived from DABS File B).

usaspending_api/disaster/v2/views/agency/count.py
```diff
@@ -17,18 +19,28 @@ class AgencyCountViewSet(CountBase):
...
-                include=Exists(FinancialAccountsByProgramActivityObjectClass.objects.filter(*filters).values("pk"))
+        if self.award_type_codes:
+            filters = [
+                self.is_in_provided_def_codes(),
+                self.all_closed_defc_submissions,
+                self.is_non_zero_award_cpe(),
+            ]
+            count = FinancialAccountsByAwards.objects.filter(*filters)
+            count = count.values("award_id").filter(award__type__in=self.filters.get("award_type_codes"))
+            count = count.aggregate(count=Count("award__funding_agency__toptier_agency", distinct=True))["count"]
```
kwhickey, line 30: stylistically I think the chaining of filters with just one return reads a bit better. May not be obvious to all readers that the queryset isn't evaluated til an action is taken on it.

e.g. 
```
count = (
    Model.objects.filter(...)
     .this(...)
     .that(...)
     .action(...)
)
```

You do that below, so it'd be good for consistency .

usaspending_api/disaster/v2/views/agency/count.py
```diff
@@ -17,18 +19,28 @@ class AgencyCountViewSet(CountBase):
...
-            self.is_non_zero_object_class_cpe(),
-        ]
-        count = (
-            ToptierAgency.objects.annotate(
-                include=Exists(FinancialAccountsByProgramActivityObjectClass.objects.filter(*filters).values("pk"))
+        if self.award_type_codes:
+            filters = [
+                self.is_in_provided_def_codes(),
+                self.all_closed_defc_submissions,
+                self.is_non_zero_award_cpe(),
```
kwhickey, line 26: Looking at the code for `DisasterBase.is_non_zero_award_cpe()`, it looks wrong:

```
    def is_non_zero_award_cpe(self):
        return Q(
            Q(obligations_incurred_total_by_award_cpe__gt=0)
            | Q(obligations_incurred_total_by_award_cpe__lt=0)
            | Q(gross_outlays_delivered_orders_paid_total_cpe__gt=0)
            | Q(gross_outlays_delivered_orders_paid_total_cpe__lt=0)
        )
```

If this is indeed at the _award_ metrics of account data -- i.e. File C / `FinancialAccountsByAwards` -- then it should be using these fields to yield sums, and therefore these fields should be checked as non-zero:
- obligations: `FinancialAccountsByAwards.transaction_obligated_amount`
- outlay: `FinancialAccountsByAwards.gross_outlay_amount_by_award_cpe`

usaspending_api/disaster/v2/views/agency/count.py
```diff
@@ -17,18 +19,28 @@ class AgencyCountViewSet(CountBase):
...
+            ]
+            count = FinancialAccountsByAwards.objects.filter(*filters)
+            count = count.values("award_id").filter(award__type__in=self.filters.get("award_type_codes"))
+            count = count.aggregate(count=Count("award__funding_agency__toptier_agency", distinct=True))["count"]
+        else:
+            filters = [
+                Q(treasury_account__funding_toptier_agency=OuterRef("pk")),
+                self.is_in_provided_def_codes(),
+                self.all_closed_defc_submissions,
+                self.is_non_zero_object_class_cpe(),
```
kwhickey, line 36: I recommend changing the name from `is_non_zero_object_class_cpe` to `is_non_zero_total_spending` ... where "total" is to indicate the `spending_type` field in play here (one of `total`, `awards`, `loans`, from the api contract and general domain language being used around the different type of disaster spending).

The trouble with the name as it is, is two-fold:
1. It doesn't follow suit with the existing `FinancialAccountsByProgramActivityObjectClass` model it implies reference to, because it only mentions `object_class` and not program activity
2. With just object class there (and no doc strings for the method!) it could easily be implied that this is a filter that has something to do with the object class dimension and not with spending amounts.

usaspending_api/api_contracts/contracts/v2/disaster/agency/count.md
kwhickey: Also, minor type `form` -> `from`

kwhickey: Also, need to include:

```
## AwardTypeCodes (enum[string])
List of procurement and assistance award type codes supported by USAspending.gov

### Members
+ `02`
+ `03`
+ `04`
+ `05`
+ `06`
+ `07`
+ `08`
+ `09`
+ `10`
+ `11`
+ `A`
+ `B`
+ `C`
+ `D`
+ `IDV_A`
+ `IDV_B_A`
+ `IDV_B_B`
+ `IDV_B_C`
+ `IDV_B`
+ `IDV_C`
+ `IDV_D`
+ `IDV_E`
```

in the docs now

BrianZito:
> LGTM. Thanks for adding to the test.

Thanks for reviewing the tests.

usaspending_api/disaster/v2/views/agency/count.py
BrianZito: I used the name `is_non_zero_object_class_cpe` to juxtapose it to `is_non_zero_award_cpe`, naming each vaguely after the table involved. I'm all for a better naming convention, particularly if it's more based on business logic, but I think `is_non_zero_total_spending` vs `is_non_zero_award_cpe` will still be confusing. 
  Possibly `is_non_zero_total_award_spending` vs `is_non_zero_total_object_program_spending`? That seems to bring us back to the original problem though.

BrianZito: On second looks, I think your suggestion makes a lot of sense. I've renamed to `is_non_zero_total_spending` and `is_non_zero_award_spending`, with hopes that a future refactor will bring `is_non_zero_loan_spending` alongside them one day.

PR #2601, opened 2020-07-14
[Unticketed] Remove dump_urls management command

kbard:
**Description:**

`dump_urls` was a handy dandy management command I had created when I first came to this project to dump out URLs recognized by our Django project.  Little did I know, someone had snuck a similar command (called `show_urls`) into the `django-extensions` package.  Since `dump_urls` is obsolete and in need of some maintenance, there's really no need to keep it around.

1. [x] Tests unaffected
    - [ ] Backend
5. [x] Front end unaffected
8. [x] No Jira Ticket - just some basic tech debt cleanup


kbard:
> Makes sense. Let's notify testers about this change, on the off-chance they use it for endpoint testing, but other than that I don't see how this could be being used in an automated way.

Nobody uses it.  It hasn't worked in a while.  Django changed the way they do URLs so it stopped working when we switched to Django 2.  I just never got around to fixing it and since there's a maintained alternative, it's time for it to go bye bye.

PR #2605, opened 2020-07-15
[DEV-5299] Only allow assistance filters and minor cleanup on API contract

sethstoudenmier:
**Description:**
Realized during demos that the CFDA endpoints should only allow the Assistance Award Type Codes. Additionally, needed to cleanup API Contracts where the `award_type_codes` was added for loans endpoints.

**Technical details:**
The `enum_values` for the TinyShield model are now sorted to prevent random order being returned when non-valid award_type_codes are passed back. This helps to make sure that tests don't break because of values not being sorted the same way as expected.

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
8. [X] Jira Ticket [DEV-5299](https://federal-spending-transparency.atlassian.net/browse/DEV-5299):

```
```

### DIFF
--- a/usaspending_api/api_contracts/contracts/v2/disaster/agency/loans.md
+++ b/usaspending_api/api_contracts/contracts/v2/disaster/agency/loans.md
@@ -67,14 +67,17 @@ Returns loan spending details of Agencies receiving supplemental funding budgeta
 
 ## Filter (object)
 + `def_codes` (required, array[DEFC], fixed-type)
-+ `award_type_codes` (optional, array[enum[string]], fixed-type)
-    + Only accepts loan award type `07` or `08` in the array, since this endpoint is specific to loans
-    + If ANY award type codes are provided, loan amounts will be summed for the distinct set of toptier agencies,
++ `award_type_codes` (optional, array[enum[string]], fixed-type)  
+    Only accepts loan award type `07` or `08` in the array, since this endpoint is specific to loans.
+
+    If ANY award type codes are provided, loan amounts will be summed for the distinct set of toptier agencies,
     whose subtier agencies funded loan awards linked to `FinancialAccountsByAwards` records (which are derived from DABS File C).
-    + If this parameter is not provided, loan amounts will be summed for a different set of agencies: 
+
+    If this parameter is not provided, loan amounts will be summed for a different set of agencies: 
     the distinct set of toptier agencies "owning" appropriations accounts used in funding _any_ award spending
     for this disaster (i.e. from agencies "owning" Treasury Account Symbol (TAS) accounts on `FinancialAccountsByAwards`
     records, which are derived from DABS File C).
+
     + Members
         + `07`
         + `08`
--- a/usaspending_api/api_contracts/contracts/v2/disaster/agency/loans.md
+++ b/usaspending_api/api_contracts/contracts/v2/disaster/agency/loans.md
@@ -67,7 +67,7 @@ Returns loan spending details of Agencies receiving supplemental funding budgeta
 
 ## Filter (object)
 + `def_codes` (required, array[DEFC], fixed-type)
-+ `award_type_codes` (optional, array[string], fixed-type)
++ `award_type_codes` (optional, array[enum[string]], fixed-type)
     + Only accepts loan award type `07` or `08` in the array, since this endpoint is specific to loans
     + If ANY award type codes are provided, loan amounts will be summed for the distinct set of toptier agencies,
     whose subtier agencies funded loan awards linked to `FinancialAccountsByAwards` records (which are derived from DABS File C).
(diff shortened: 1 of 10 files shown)


usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -112,11 +112,21 @@ def filters(self):
                 "name": "award_type_codes",
                 "type": "array",
                 "array_type": "enum",
-                "enum_values": list(loan_type_mapping.keys()),
+                "enum_values": sorted(list(loan_type_mapping.keys())),
```
kbard, line 115: I see this a lot.  There's no need to change it, but `sorted` returns a list.  Also, the default iterator behavior of a dictionary returns only the keys so this could be simplified down to `sorted(loan_type_mapping)`.

usaspending_api/api_contracts/contracts/v2/disaster/cfda/loans.md
```diff
@@ -67,7 +67,7 @@ Returns loan spending details of CFDA Programs receiving supplemental funding bu
 
 ## Filter (object)
 + `def_codes` (required, array[DEFC], fixed-type)
-+ `award_type_codes` (optional, array[string], fixed-type)
++ `award_type_codes` (optional, array[enum[string]], fixed-type)
     + Only accepts loan award type `07` or `08` in the array, since this endpoint is specific to loans
```
kbard, line 71: These contracts aren't passing Dredd.  It looks like we've done descriptions wrong in a whole mess of places.  Turns out, any descriptive text should not be preceded by a `+`.  The only reason these failed is because they're mixing the text description with integer `07` and `08` values which caused Python to throw a fit.

You can see a description of how enums should work here: https://apiblueprint.org/documentation/specification.html#uri-parameters-section

Even though it's talking about parameters, the syntax is the same.

kbard: Ok, not as many as I thought.  These three contracts and `disaster/agency/spending.md`.  Would you mind horribly correcting it there as well?

kbard: ALSO, and this is an issue from before, `sort` is not accepting `id` as a sort order.  I don't know if that's an issue with the contract or the endpoint.  Should it sort on `id`?

`Field 'sort' is outside valid values ['description', 'count', 'obligation', 'outlay', 'face_value_of_loan']`

usaspending_api/api_contracts/contracts/v2/disaster/agency/loans.md
```diff
@@ -67,7 +67,7 @@ Returns loan spending details of Agencies receiving supplemental funding budgeta
 
 ## Filter (object)
 + `def_codes` (required, array[DEFC], fixed-type)
-+ `award_type_codes` (optional, array[string], fixed-type)
++ `award_type_codes` (optional, array[enum[string]], fixed-type)
```
kbard, line 70: This contract has other issues, presumably because two different people worked on it.  The two different flavors of the endpoint do not accept the same sort fields.  Do we want to take least common denominator?

usaspending_api/api_contracts/contracts/v2/disaster/cfda/loans.md
sethstoudenmier: I can go through and clean these up. The `id` field is not sortable due to limitations of us using pre-aggregated "agg_key" to make Elasticsearch more performant so I will remove that.

usaspending_api/disaster/v2/views/disaster_base.py
sethstoudenmier: Will make the change since I will be making other changes anyways

usaspending_api/api_contracts/contracts/v2/disaster/agency/loans.md
sethstoudenmier: It's really a lose-lose situation unless we change up the endpoint some. If we don't show `id` on the API Contract then the non-award type codes contract is wrong because the default would also be off. But if we leave it then the API Contract fails if you run it.

tony-sappe: Unfortunately we can't drop the `id` field because it is necessary for links to the agency page. I think the inconsistencies do need to be fixed this sprint. This ticket highlights a problem with the `id` values being returned 
https://federal-spending-transparency.atlassian.net/browse/DEV-5635

sethstoudenmier: @tony-sappe Since you will be touching the endpoints do you mind handling the sort portion of this contract not passing Dredd when you actually run it (instead of a dry-run) and I will make sure that the `award_type_codes` portion passes as part of this PR?

tony-sappe: I might get pulled into something else, but I will add a note to the Jira ticket for whoever does work on it

PR #2610, opened 2020-07-16
[DEV-5635] Fix Agency ID response values

tony-sappe:
**Description:**
Original implementation returned the values from `id` in `toptier_agency` instead of `agency`

**Technical details:**
Required ES re-index of awards index

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [x] Necessary PR reviewers:
    - [x] Backend
4. [x] Matview impact assessment completed (N/A)
5. [x] Frontend impact assessment completed
6. [x] Data validation completed
7. [x] Appropriate Operations ticket(s) created
8. [x] Jira Ticket [DEV-5635](https://federal-spending-transparency.atlassian.net/browse/DEV-5635):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected API
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
No changes to matviews
```

PR #2624, opened 2020-07-22
More fixes to disaster endpoints

tony-sappe:
**Description:**
Follows #2619
.... and #2610

**Technical details:**
- Fixing Agency IDs _again_ for ES Awards index
- Removed duplicates from Object Class count endpoint
- Fixed Total Budgetary Resources from agency/spending
- Added tests to cover these above scenarios
- Improved code re-use for queryset filters

    - [ ] Backend
    - [ ] Frontend
    - [ ] Domain Expert
8. [x] Jira Ticket (N/A)

```
```


usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -71,7 +71,7 @@ SELECT
       THEN CONCAT(
         '{"name":"', vw_award_search.funding_toptier_agency_name,
         '","code":"', vw_award_search.funding_toptier_agency_code,
-        '","id":"', vw_award_search.funding_agency_id, '"}'
+        '","id":"', (SELECT aa.id FROM Agency aa WHERE aa.toptier_agency_id = (SELECT aaa.toptier_agency_id FROM Agency aaa WHERE aaa.id = vw_award_search.funding_agency_id) AND aa.toptier_flag = True), '"}'
```
BrianZito, line 74: Wouldn't it make more sense to select the toptier_agency_id instead of the agency id? I'm pretty sure that this will still group by the agency table instead of the toptier agency table.

usaspending_api/disaster/v2/views/agency/count.py
```diff
@@ -19,25 +19,24 @@ class AgencyCountViewSet(AwardTypeMixin, CountBase):
...
+            self.is_in_provided_def_codes,
+        ]
+
         if self.award_type_codes:
-            filters = [
-                self.is_in_provided_def_codes(),
-                self.all_closed_defc_submissions,
-                self.is_non_zero_award_spending(),
-            ]
+            filters.append(self.is_non_zero_award_spending)
```
BrianZito, line 29: minor, but this can probably move to before the if statement, since this filter is also added in the else condition.

usaspending_api/database_scripts/etl/award_delta_view.sql
tony-sappe: The agency IDs need to be the PK from the agency table. The rub is they need to be the _same_ id for the records

usaspending_api/disaster/v2/views/agency/count.py
tony-sappe: Look closer at the property names

BrianZito: I see now. I have been a fool.

usaspending_api/database_scripts/etl/award_delta_view.sql
BrianZito: I see, so the strategy here is to filter on the toptier flag, which should only be true once per toptier ID. That looks consistent to me.

tony-sappe: Yeah, it appears to work well enough for the postgres api responses. There are a few edgecases so it isn't perfect.

tony-sappe: Improved SQL. Thanks @kbard

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -71,7 +71,7 @@ SELECT
       THEN CONCAT(
         '{"name":"', vw_award_search.funding_toptier_agency_name,
         '","code":"', vw_award_search.funding_toptier_agency_code,
-        '","id":"', vw_award_search.funding_agency_id, '"}'
+        '","id":"', (SELECT aa.id FROM Agency aa WHERE aa.toptier_agency_id = (SELECT aaa.toptier_agency_id FROM Agency aaa WHERE aaa.id = vw_award_search.funding_agency_id) ORDER BY aa.toptier_flag DESC, aa.id LIMIT 1), '"}'
```
kbard, line 74: Wow.  That is some fantastic SQL.  Good job.  👍

kbard: Although `Agency` with a capital `A` is messing with my OCD.

kbard: It dawns on me that another way to do this would be to resolve all of the agencies in a subquery or cte so that they  don't have to be correlated subqueries.  Might help with performance.  Not sure what performance is like on this but the subqueries couldn't have helped.

tony-sappe: Alt Text

tony-sappe: (in case anyone is confused, that is Kirk ^^^^^^^^^^^^)

kbard: No one was confused, they've just forgotten what I look like.

usaspending_api/disaster/v2/views/agency/spending.py
```diff
@@ -121,13 +113,35 @@ def total_queryset(self):
...
+                .order_by("-toptier_flag", "id")
+                .values("id")[:1]
+            ),
+            "code": F("toptier_code"),
+            "description": F("name"),
+            # Currently, this endpoint can never have children w/o type = `award` & `award_type_codes`
+            "children": Value([], output_field=ArrayField(IntegerField())),
+            "count": Value(0, output_field=IntegerField()),
+            "obligation": cte.col.obligation,
+            "outlay": cte.col.outlay,
```
BrianZito, line 137: Does calling cte twice here make two ctes? `With` looks like a class instead of a function, so I think this is okay?

tony-sappe: CTEs are treated like a model, so it's just referencing the attributes in the CTEs rather than executing. Unlike sub-queries which are separate queries executed independently

PR #2623, opened 2020-07-21
Mod/dev 5621 add award count

BrianZito:
**Description:**
Replaces the "count" variable with "award_count", which always counts the number of unique file C records

**Technical details:**
Elasticsearch endpoints are not updated in this PR.

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [x] Necessary PR reviewers:
    - [x] Backend
    - [x] Frontend 
4. [ ] Matview impact assessment completed
5. [ ] Frontend impact assessment completed
6. [ ] Data validation completed
7. [ ] Appropriate Operations ticket(s) created
8. [x] Jira Ticket [DEV-5621](https://federal-spending-transparency.atlassian.net/browse/DEV-5621):
    - [x] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #2630, opened 2020-07-23
[DEV-5706] Improving Sorting and fixed items from DEV-5621

tony-sappe:
**Description:**
1. Added a secondary sort on `id` for all disaster endpoints which support pagination to provide deterministic result order when results are sorted on fields with the same value
1. Changed the sort field from `count` to `award_count` for endpoints modified by #2623
1. Changed Response and sort field for other endpoints from `count` -> `award_count` not addressed in #2623
    - ` /api/v2/disaster/award/amount/` (response only)
    - `/api/v2/disaster/cfda/loans/`
    - `/api/v2/disaster/cfda/spending/`
    - `/api/v2/disaster/recipient/loans/`
    - `/api/v2/disaster/recipient/spending/`
    - `/api/v2/disaster/agency/spending/` (previously inconsistent)
    - `/api/v2/disaster/agency/loans/` (previously inconsistent)

**Technical details:**
N/A

    - [ ] Backend
    - [ ] Frontend
8. [x] Jira Ticket [DEV-5706](https://federal-spending-transparency.atlassian.net/browse/DEV-5706):

```
No changes to database or ES necessary
```

### DIFF
--- a/usaspending_api/common/data_classes.py
+++ b/usaspending_api/common/data_classes.py
@@ -10,7 +10,18 @@ class Pagination:
     upper_limit: int
     sort_key: Optional[str] = None
     sort_order: Optional[str] = None
+    secondary_sort_key: Optional[str] = None
+
+    @property
+    def _sort_order(self):
+        if self.sort_order == "desc":
+            return "-"
+        return ""
 
     @property
     def order_by(self):
-        return self.sort_key if self.sort_order == "asc" else "-" + self.sort_key
+        return f"{self._sort_order}{self.sort_key}"
+
+    @property
+    def robust_order_by_fields(self):
+        return (self.order_by, f"{self._sort_order}{self.secondary_sort_key}")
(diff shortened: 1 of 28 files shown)


usaspending_api/common/data_classes.py
```diff
@@ -10,7 +10,18 @@ class Pagination:
     upper_limit: int
     sort_key: Optional[str] = None
     sort_order: Optional[str] = None
+    secondary_sort_key: Optional[str] = None
+
+    @property
+    def _sort_order(self):
```
BrianZito, line 16: I would advise against a function with the same name as a variable. In this case it might be easiest just to leave the logic within order_by()

tony-sappe: It's not the same name

BrianZito: A future reader will mistake the two for each other.

tony-sappe: That will teach them for using a property with a naming signal to denote internal use only!

PR #2639, opened 2020-07-25
Production Deploy

tony-sappe:
## USAspending Sprint 111 & 112 Deploy!!!

Contains a significant volume of new functionality, alterations, and fixes.


usaspending_api/submissions/management/commands/load_dabs_submission_window_schedule.py
```diff
@@ -27,7 +27,7 @@
 from
     submission_window_schedule
 where
-    period_start >= '2020-07-17'
+    period_start >= '2020-05-15'
```
tony-sappe, line 30: Does this need to be reverted?

kwhickey: @tony-sappe I talked to Wil and tested the sql out against Broker with both dates in. Since there are no monthlies other than the 3 starting 2020-04-01, the sql returns the same results when either date is used. So this change his harmless and can stay.

PR #2644, opened 2020-07-26
Fixing views found to be returning a Queryset

tony-sappe:
**Description:**
Some views were returning a Queryset (either at the top-level or buried in the object) which wasn't a "problem" for the existing endpoints but isn't optimal in any case. Introducing Django-CTE surfaced this lingering issue which was a problem for setting a new value in the API cache, effectively braking the cache. This PR replaces the "band-aid" fix which materialized the queryset into Python primitives in limited situations. Now a more robust check is in place and the offending views were modified to no longer return a queryset.

**Technical details:**
- Introduced a new check in the api cache_decorator.py function wrapper to check if a Queryset object is in the response.data field. It will only run in local mode.
- The check function was a generator until an API responses hit the recursion depth limit :-/
- It might be better to embed the check in a new view class but that can be a future exercise if the need presents itself.

    - [x] Backend
8. [x] Jira Ticket (N/A)

```
These changes shouldn't impact observable behaviors
```


usaspending_api/common/cache_decorator.py
```diff
@@ -1,13 +1,31 @@
...
 
+def contains_queryset(data: Any) -> bool:
+    """Boring non-generator to traverse a complex object and flag any Querysets"""
+    if isinstance(data, QuerySet):
+        return True
+    elif isinstance(data, dict):
+        for v in data.values():
+            if contains_queryset(v):
+                return True
+    elif isinstance(data, list):
```
kbard, line 21: tuples?  sets?  other iterables?

usaspending_api/common/cache_decorator.py
```diff
@@ -30,32 +48,16 @@ def process_cache_response(self, view_instance, view_method, request, args, kwar
...
-                response.data["results"] = list(response.data["results"])
+            # While returning a Queryset is functional most of the time, it isn't
+            # fully supported by Django Rest Framework. This check was inserted
+            # in local mode to catch if a Queryset is being returned by the view
+            # which could cause an exception when setting the cache
+            if settings.IS_LOCAL and response and not response.is_rendered and contains_queryset(response.data):
+                raise RuntimeError(
+                    "Your view is returning a QuerySet.  QuerySets are not really designed to "
+                    "be pickled and can cause caching issues.  Please materialize the QuerySet "
+                    "using a List or some other more primitive data structure."
```
kbard, line 59: "Thank you, have a lovely day, and don't forget to wash your hands regularly!" ... too much?

tony-sappe: Over the top

kbard: Aww

usaspending_api/common/cache_decorator.py
tony-sappe: Good catch. At first the correction caused problems by significantly increasing recursion depth and hitting the limit. Then I realized strings are Iterables and were blowing up the function calls. It's so obvious in hindsight

kbard: lol... stupid strings... sorry for causing a headache

PR #2684, opened 2020-07-31
[DEV-5696] Reduce submission loader false negatives (QAT)

kbard:
- Reduce the number of Slack false negatives.  Only the last process standing reports unstable status if necessary.
- Didn't like that the queue status was pinging the database 5 times.
- Mentions when the database commit is starting so that anyone monitoring the logs isn't left wondering why nothing appears to be happening.
- Logs queue status after every submission is processed.  I found myself querying the database a lot trying to get a feel for how much was left in the queue.

### DIFF
--- a/usaspending_api/etl/management/commands/load_submission.py
+++ b/usaspending_api/etl/management/commands/load_submission.py
@@ -152,6 +152,8 @@ class Command(load_base.Command):
         # Cleanup not specific to this submission is run in the `.handle` method
         logger.info(f"Successfully loaded submission {self.submission_id}.")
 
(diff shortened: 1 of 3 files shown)


usaspending_api/etl/management/commands/load_submission.py
```diff
@@ -152,6 +152,8 @@ def signal_handler(signal, frame):
         # Cleanup not specific to this submission is run in the `.handle` method
         logger.info(f"Successfully loaded submission {self.submission_id}.")
 
+        logger.info("Committing transaction...")
```
tony-sappe, line 155: 💯

PR #2700, opened 2020-08-04
[DEV-5814] Logs download_count when downloading csv regardless of --skip-counts flag

collinwr:
Description:
The number of rows downloaded into a CSV during the ElasticSearch ETL should be logged whether or not the --skip-counts flag is enabled. This flag is used to determine whether an additional sql script should be run to provide a secondary count to compare against the download count.

Additionally, it logs the total count of records to update before the CSV download phase begins. 

This PR makes the change described above.

Technical details:
The only non-cosmetic change is that the download_db_records method now returns the number of rows in the downloaded csv. It previously returned None if --skip-counts was enabled and the count from the count_sql script if disabled.

Requirements for PR merge:

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [DEV-5814](https://federal-spending-transparency.atlassian.net/browse/DEV-5814):

### DIFF
--- a/usaspending_api/etl/es_etl_helpers.py
+++ b/usaspending_api/etl/es_etl_helpers.py
@@ -391,22 +391,24 @@ def download_db_records(fetch_jobs, done_jobs, config):
 
 
 def download_csv(count_sql, copy_sql, filename, job_id, skip_counts, verbose):
-    if skip_counts:
-        count = None
-        printf({"msg": "Skipping count checks. Writing file: {}".format(filename), "job": job_id, "f": "Download"})
-    else:
-        count = execute_sql_statement(count_sql, True, verbose)[0]["count"]
-        printf({"msg": "Writing {} to this file: {}".format(count, filename), "job": job_id, "f": "Download"})
+
+    # Execute Copy SQL to download records to CSV
     # It is preferable to not use shell=True, but this command works. Limited user-input so risk is low
     subprocess.Popen("psql {} -c {}".format(get_database_dsn_string(), copy_sql), shell=True).wait()
+    download_count = count_rows_in_delimited_file(filename, has_header=True, safe=False)
+    printf({"msg": "Wrote {} to this file: {}".format(download_count, filename), "job": job_id, "f": "Download"})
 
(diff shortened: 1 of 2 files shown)


usaspending_api/etl/es_etl_helpers.py
```diff
@@ -327,6 +335,19 @@ def configure_sql_strings(config, filename, deleted_ids):
     return copy_sql, id_sql, count_sql
 
 
+def get_updated_record_count(config):
+    update_date_str = UPDATE_DATE_SQL.format(config["starting_date"].strftime("%Y-%m-%d"))
```
tony-sappe, line 339: Was there an issue including the time?

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -194,18 +194,24 @@
     "total_covid_outlay",
 ]
 
-UPDATE_DATE_SQL = " AND update_date >= '{}'"
+UPDATE_DATE_SQL = "update_date >= '{}'"
```
tony-sappe, line 197: If not providing the `update_date` sql snippet for the below SQL results in an error now, is there value keeping this string defined separately?

usaspending_api/etl/rapidloader.py
```diff
@@ -28,6 +29,15 @@ def run_load_steps(self) -> None:
         download_queue = Queue()  # Queue for jobs which need a csv downloaded
         es_ingest_queue = Queue(20)  # Queue for jobs which have a csv and are ready for ES ingest
 
+        updated_record_count = get_updated_record_count(self.config)
+        printf(
+            {
+                "msg": "Found {} new {} records to update in ElasticSearch".format(
```
tony-sappe, line 35: f strings are preferable now since we're using Python3.7, it might allow this to be a single line.

Minor gripe: when creating a new index the log state says "update"

tony-sappe: Additionally, might as well format the number to include comma separators since we expect every load to be over 1K and often orders of magnitude larger

PR #2703, opened 2020-08-05
[DEV-5825] Malformed json strings from Elasticsearch causing HTTP500s

tony-sappe:
**Description:**
Title says it all. Recipient name contains three leading backslash characters. After the issue appeared in the COVID-19 page I tracked it down in other places across the website.

**Technical details:**
Escaping the slashes with another slash each fixed the issue. API responses will contain `\\\\\\STRIPES/// PARKING LOT SERVICE, INC.` which is correctly parsed by JSON readers, such as the website.

    - [x] Backend
8. [x] Jira Ticket [DEV-5825](https://federal-spending-transparency.atlassian.net/browse/DEV-5825):
    - [x] Performance evaluation of affected API (No impact)

```
```

### DIFF
--- a/usaspending_api/disaster/v2/views/agency/loans.py
+++ b/usaspending_api/disaster/v2/views/agency/loans.py
@@ -117,7 +117,7 @@ class LoansBySubtierAgencyViewSet(ElasticsearchLoansPaginationMixin, Elasticsear
         return results
 
     def _build_json_result(self, bucket: dict):
-        info = json.loads(bucket.get("key").encode('unicode_escape'))
+        info = json.loads(bucket.get("key").encode("unicode_escape"))
         return {
             "id": int(info["id"]),
             "code": info["code"],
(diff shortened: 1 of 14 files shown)


usaspending_api/disaster/v2/views/agency/loans.py
```diff
@@ -117,7 +117,7 @@ def build_elasticsearch_result(self, response: dict) -> List[dict]:
         return results
 
     def _build_json_result(self, bucket: dict):
-        info = json.loads(bucket.get("key"))
+        info = json.loads(bucket.get("key").encode("unicode_escape"))
```
kbard, line 120: These should probably have a default value if there's even the remotest chance "key" can be None.

`json.loads(bucket.get("key", "").encode("unicode_escape")) or None`

If there's no chance "key" can be None, then probably

`json.loads(bucket["key"].encode("unicode_escape"))`

tony-sappe: I thought about that, but it appeared the existing code always expects/assumes data. The original code would have run `json.loads(None)` which raises an exception. Does adding .encode() make it more fragile?

kbard: Apparently not.

PR #2709, opened 2020-08-06
[DEV-5825] Fixing issue introduced by first PR

tony-sappe:
**Description:**
Apparently, encoding a string of json with `unicode_escape` will break the json loader if there are unicode characters in the string. Who knew?

**Technical details:**
New ugly helper function to handle json loading from str

    - [x] Backend
8. [x] Jira Ticket [DEV-5825](https://federal-spending-transparency.atlassian.net/browse/DEV-5825):
    - [x] Performance evaluation of affected API (None observed)

```
fixing code
```

### DIFF
--- a/usaspending_api/common/elasticsearch/json_helpers.py
+++ b/usaspending_api/common/elasticsearch/json_helpers.py
@@ -1,0 +1,16 @@
+import json
+
+
+def json_str_to_dict(string: str) -> dict:
+    if not str:
+        return {}
+
+    try:
+        return json.loads(string)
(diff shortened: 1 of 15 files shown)


tony-sappe:
> Confirmed locally that this works as expected for the cases we have run into. However, if a string happens to have both escape characters and unicode characters then it will still fail.

Yep, that is true. I wasn't sure how to get that situation to work. Perhaps there is something possible on the Elasticsearch side of things to not send response strings with unescaped JSON

PR #2723, opened 2020-08-13
[DEV-5855] Efficiency Improvements to Elasticsearch ETL

tony-sappe:
**Description:**

- Using postgres `id` field for Elasticsearch's `_id` document field
    - This allows the bulk loader to effectively perform upsert operations
    - Removed costly delete calls before loading a batch of new data
- Replaced  `streaming_bulk ` with `parallel_bulk` (es_etl_helpers.py LN 479) which will use 4 (default) threads to stream the 500 (default) records per message
- Fixed flag incorrectly using underscore `--load_type`  ->  `--load-type`
- Cleaned up strings and log statements

**Technical details:**
- https://www.elastic.co/guide/en/elasticsearch/reference/7.0/docs-bulk.html
- https://elasticsearch-py.readthedocs.io/en/master/helpers.html#elasticsearch.helpers.parallel_bulk

    - [x] Backend
    - [x] Operations
8. [ ] Jira Ticket [DEV-5855](https://federal-spending-transparency.atlassian.net/browse/DEV-5855):

```
```


usaspending_api/etl/es_etl_helpers.py
```diff
@@ -452,43 +446,47 @@ def csv_chunk_gen(filename, chunksize, job_id, load_type):
...
-    printf({"msg": "Completed Elasticsearch data load", "f": "ES Ingest"})
+    printf({"msg": "Completed Elasticsearch data load", "f": "ES Index"})
     return
 
 
 def streaming_post_to_es(client, chunk, index_name: str, type: str, job_id=None):
     success, failed = 0, 0
     try:
-        for ok, item in helpers.streaming_bulk(client, chunk, index=index_name):
+        for ok, item in helpers.parallel_bulk(client, chunk, index=index_name):
```
kwhickey, line 481: Some minor quibbles about the name of our wrapper function here. What makes it "streaming"? To me streaming would imply that as soon as data is changing some event handling is occurring to stream the event to its final destination. 
Other options for perhaps more accurate is batching, or microbatching. Probably batching is still most accurate since microbatching is a way to pile-up streaming events into small batches. And then use of parallel is a good qualifier in the name if executing batches in parallel.

kwhickey, line 481: How bout trying to change a couple params for better throughput? Or are you reserving that experimentation for the follow-on ticket?

```
thread_count – size of the threadpool to use for the bulk requests
chunk_size – number of docs in one chunk sent to es (default: 500)
max_chunk_bytes – the maximum size of the request in bytes (default: 100MB)
...
queue_size – size of the task queue between the main thread (producing chunks to send) and the processing threads.
```

I'm thinking at least 8 worker threads, a chunk size closer to 4000, and then putting a limiting factor of 1MB (1048576 bytes) on `max_chunk_bytes`. I haven't quite worked out how `queue_size` and `thread_count` play on each other.

tony-sappe: I can understand the ambiguity. From the elasticsearch library method perspective, `streaming_bulk()` and `parallel_bulk()` are streaming a batch load. The chunk size is 250K, but each load iteration is only 500 records (the functions are iterators/generators, which is why they are looping in the for statement).

tony-sappe: I thought about it and tested it locally a little. I couldn't find a better combination than the defaults. And that is just local to my machine, no idea what the behaviors will be like from the jenkins worker on the internal AWS network. My instict is the leave the defaults as-is for now and spend time with A-B tests in the future if there is a need. Just goin from 1 to 4 threads is going to significantly help

usaspending_api/etl/es_etl_helpers.py
```diff
@@ -211,7 +208,7 @@
 COPY_SQL = """"COPY (
```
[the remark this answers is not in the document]
tony-sappe: I did the same thing 😄

tony-sappe:
Waiting to complete testing in a deployed cluster before merging.

usaspending_api/etl/tests/test_es_rapidloader.py
```diff
@@ -108,19 +110,29 @@ def _sleep(seconds):
...
+#  which would be laborious and fragile. Leaving tests in-place to document this
+#  testing shortcoming. It may be addressed in the near future (time-permitting)
+#  if some refactoring occurs and allows more flexibility. -- from Aug 2020
+################################################################################
 
-def test_es_award_loader_class(award_data_fixture, elasticsearch_award_index, baby_sleeps):
+
+@pytest.mark.skip
+def test_es_award_loader_class(award_data_fixture, elasticsearch_award_index, baby_sleeps, monkeypatch):
+    monkeypatch.setattr("usaspending_api.etl.es_etl_helpers.execute_sql_statement", mock_execute_sql)
```
sethstoudenmier, line 135: Almost feel that the monkeypatch step should not be included in the test so that when we go back around to fix this they are in their original state. I think the comment above is enough regarding monkeypatch.

Same for the test below.

tony-sappe: Adding in the monkeypatch made the test closer to functional, without it the code didn't see the mock award data. There needs to be more monkeypatch or mocking the CSVs for the test to be functional again because the script uses psql which opens its own DB connections outside of the test connection.

PR #2737, opened 2020-08-21
Mod/dev 5551 add unlinked items

BrianZito:
**Description:**
Appends `totals` section to disaster page endpoints, which can be subtracted from the totals from award/amount to derive unlinked data amounts.

**Technical details:**

    - [x] Backend
8. [x] Jira Ticket [DEV-5551](https://federal-spending-transparency.atlassian.net/browse/DEV-5551):

```
```


usaspending_api/api_contracts/contracts/v2/disaster/cfda/loans.md
```diff
@@ -26,10 +26,15 @@ Returns loan spending details of CFDA Programs receiving supplemental funding bu
         + `results` (required, array[Result], fixed-type)
```
tony-sappe, line 26: `+ `totals` (required, Totals, fixed-type)` here?

usaspending_api/api_contracts/contracts/v2/disaster/cfda/spending.md
```diff
@@ -31,10 +31,14 @@ Returns spending details of CFDA receiving supplemental funding budgetary resour
         + `results` (required, array[Result], fixed-type)
```
tony-sappe, line 31: `+ `totals` (required, Totals, fixed-type)` here?

usaspending_api/api_contracts/contracts/v2/disaster/recipient/loans.md
```diff
@@ -26,10 +26,15 @@ Returns loan spending details of recipients receiving supplemental funding budge
         + `results` (required, array[Result], fixed-type)
```
tony-sappe, line 26: `+ `totals` (required, Totals, fixed-type)` here?

sethstoudenmier:
>Minor changes for API contracts.

Made necessary updates to API Contracts.

>Any performance differences with Elasticsearch-based views?

Had to go a different route with Elasticsearch based endpoints in the most recent commit. The endpoints now retrieve all aggregation buckets (up to a max of 10k) and then piece together the "totals". When testing against QAT performance for all endpoints is still under 10 seconds and the change from only 10 buckets to all buckets does not appear to affect query / aggregation performance much.

Part of the reason for this switch is because by aggregating the sum of outlay, obligation, and face_value_of_loan for all Awards that match the query the NULL agg keys are not being taken into consideration. Now the logic is similar to the Postgres implementation. The only outstanding piece is Recipient table because it has both a search and is capped at 10k. For right now the logic for Recipient is the same while waiting on clarification.

sethstoudenmier:
Update:

Due to handling recipient with Elasticsearch and capping the results at 10k there was a need to use Postgres to generate the totals. Thankfully, the COVID matview that was used for Recipient Count is able to be used here and on QAT only takes about 3.5 seconds to generate the needed sums/count. All together when testing against QAT the endpoint takes less than 10 seconds even with the additional SQL query.

While referencing Recipient Count approach it was noticed that the endpoint was doing an unnecessary INNER JOIN to Awards table. That has been updated to avoid that INNER JOIN which saves time on that endpoint.

tony-sappe (approved):

PR #2756, opened 2020-08-31
[DEV-5943] new distinct award field in FinancialAccountsByAwards

tony-sappe:
**Description:**
Creating a new column in `financial_accounts_by_awards` for faster aggregations when counting "distinct file C award records"

**Technical details:**
This change is introducing two migrations, the first to create the new field, the second to add the not null constraint.

    - [x] Backend
8. [x] Jira Ticket [DEV-5943](https://federal-spending-transparency.atlassian.net/browse/DEV-5943):

```
```


usaspending_api/disaster/v2/views/award/amount.py
```diff
@@ -63,7 +63,7 @@ def _file_c_aggregation(self):
             self.is_in_provided_def_codes,
         ]
 
-        group_by_annotations = {"award_identifier": self.unique_file_c_awards}
+        group_by_annotations = {"award_identifier": F("distinct_award_key")}
```
BrianZito, line 66: Will we ever need the old behavior for `unique_file_c_awards`? If not, we should replace the functionality in that function in order to make this behavior reusable, and prevent misuse of the old algorithm.

BrianZito: nevermind, I see the function was deleted altogether

usaspending_api/awards/migrations/0075_financialaccountsbyawards_distinct_award_key.py
```diff
@@ -0,0 +1,18 @@
...
+        ('awards', '0074_financialaccountsbyawards_distinct_award_key'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='financialaccountsbyawards',
+            name='distinct_award_key',
+            field=models.TextField(db_index=True),
+        ),
+    ]
```
BrianZito, line 18: Will this set some kind of default value before the update is run? In my experience running these kinds of migrations locally prompts the user for a default value, and I'm unsure how that would work when running this migration via Jenkins.

tony-sappe: There are two migrations, the first allows nulls (so the default value is null) and this migration adds the not null constraint. It is expected that a backfill script would run between the migrations. This is a common pattern for adding not null columns to production databases

usaspending_api/etl/submission_loader_helpers/file_c.py
```diff
@@ -198,10 +181,15 @@ def _save_file_c_rows(certified_award_financial, total_rows, start_time, skipped
...
 
         save_manager.append(
             load_data_into_model(award_financial_data, row, value_map=value_map_faba, save=False, reverse=reverse)
         )
 
     save_manager.save_stragglers()
+
+
+def create_distinct_award_key(row):
+    return f"{row.get('piid') or ''}|{row.get('parent_award_id') or ''}|{row.get('fain') or ''}|{row.get('uri') or ''}"
```
kwhickey, line 195: Should standardize on case here too, e.g. uppercase everything.

usaspending_api/etl/submission_loader_helpers/file_c.py
```diff
@@ -171,23 +171,6 @@ def _save_file_c_rows(certified_award_financial, total_rows, start_time, skipped
...
-            filters["parent_piid"] = row.get("parent_award_id")
-        else:
-            if row.get("fain") and not row.get("uri"):
-                filters["fain"] = row.get("fain")
-            elif row.get("uri") and not row.get("fain"):
-                filters["uri"] = row.get("uri")
-            else:
-                filters["fain"] = row.get("fain")
-                filters["uri"] = row.get("uri")
-
```
kwhickey, line 190: We able to get rid of this because we do the C-to-D linkage from SQL now at the end of the load?

Was this just accidentally left in here as dead code? -- i'm not seeing where these filters dict would have been used.

usaspending_api/database_scripts/job_archive/backfill_faba_distinct_award_key.py
```diff
@@ -0,0 +1,111 @@
...
+
+CHUNK_SIZE = 50000
+TOTAL_UPDATES = 0
+
+SQLS = [
+    """
+update  financial_accounts_by_awards
+set     distinct_award_key = concat(piid, '|', parent_award_id, '|', fain, '|', uri)
+where   financial_accounts_by_awards_id between {minid} and {maxid}
+    and distinct_award_key is distinct from concat(piid, '|', parent_award_id, '|', fain, '|', uri)
```
kwhickey, line 36: I think you need to standardize on casing here, and e.g. uppercase everything, since Broker DABS submitters may use mixed casing for things like FAIN and URI on different submissions.

usaspending_api/disaster/v2/views/disaster_base.py
```diff
@@ -325,13 +324,8 @@ def when_non_zero_award_spending(self, query):
...
         ).exclude(total_outlay=0, total_obligation=0)
 
-    @property
-    def unique_file_c_awards(self):
-        delimiter = Value("|")
-        return ConcatAll("piid", delimiter, "parent_award_id", delimiter, "fain", delimiter, "uri")
-
     def unique_file_c_award_count(self):
-        return Count(self.unique_file_c_awards, distinct=True)
+        return Count("distinct_award_key", distinct=True)
```
kwhickey, line 328: Again pleased but maybe surprised that this is the only Django query clause that needed refactoring. All that trouble with the `award/amount` endpoint, and this is the line that was doing it?

Actually, isn't there obligation and outlay sums that also utilize this "award_identifer" concatenation? These values that aren't being populated:
image

usaspending_api/etl/submission_loader_helpers/file_c.py
tony-sappe: dead code

usaspending_api/awards/migrations/0075_financialaccountsbyawards_distinct_award_key.py
tony-sappe: > Since all 4 contributing fields can be null

Are we sure that is the case? Currently there are no records missing all 4 fields. I haven't checked the DAIMS to confirm.

AFAIK the a fresh DB will not contain any records in `submission_attributes`, so what would the migration do in this case? When pre-existing data exists, there is a backfill script in the PR to run.

I strongly recommend never having data be populated by a migration unless is it **very** static data, like status codes for a process. The concern for a developer's local DB is a one-time event. But these migrations are run hundreds to thousands of times by CI tests each week. We're already having issues with the DABS submission window data inserted by a migration. It adds time and messes with the idea that the test environment db is a "clean slate"

usaspending_api/disaster/v2/views/disaster_base.py
tony-sappe: Your surprise is surprising to me. It's a function, define once and reuse is the goal, right?

Counting File C awards was in more endpoints, but the award count was changed to be counts of File D awards.

usaspending_api/database_scripts/job_archive/backfill_faba_distinct_award_key.py
```diff
@@ -0,0 +1,111 @@
...
+CONNECTION_STRING = environ["DATABASE_URL"]
+
+
+CHUNK_SIZE = 50000
+TOTAL_UPDATES = 0
+
+SQLS = [
+    """
+update  financial_accounts_by_awards
+set     distinct_award_key = concat(piid, '|', parent_award_id, '|', fain, '|', uri)
```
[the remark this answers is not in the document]
tony-sappe: Based on communication with Broker, that situation will never exist: `piid`, `fain`, or `uri` are required for C14.1 validation

usaspending_api/etl/submission_loader_helpers/file_c.py
tony-sappe: fair point

usaspending_api/database_scripts/job_archive/backfill_faba_distinct_award_key.py
tony-sappe: Done

usaspending_api/etl/submission_loader_helpers/file_c.py
tony-sappe: Done

usaspending_api/awards/migrations/0075_financialaccountsbyawards_distinct_award_key.py
kwhickey: In my mind, migrations should be migrations of a production system ... not a developer first-time setup. So they should consider how to migrate a db from x -> y if there is production data in state "x". This case wouldn't be adding any data to the system, it is a backfill only operating on existing data. On a clean unit test database, the migration would be a no-op since there's no `financial_accounts_by_awards` data at migration-time in that scenario.

As you mentioned this is a common scenario: "how to add a `NOT NULL` column to an existing table with data". I usually see the migrations for this orchestrating all 3 steps though, ideally in a transaction if possible in case any one step fails.
1. add the column as NULL allowed
2. add the missing data
3. make the column NOT NULL

Still, if this is the pattern we have, where migrations are strictly about schema and not about data, **then we should just stick with it**. Benefit: consistency. Downside: can't just run all migrations on an existing and have to consult one or more deployment coordination guides.

PR #2761, opened 2020-09-02
Ftr/dev 6005 add cfda details endpoint

BrianZito:
**Description:**
Adds reference endpoint to supply further details about CFDAs from the grants API.

**Technical details:**

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-6005](https://federal-spending-transparency.atlassian.net/browse/DEV-6005):

```
```


BrianZito:
"Do not merge" added since whitelisting has not been testing in Sandbox yet.

usaspending_api/api_contracts/contracts/v2/references/cfda/totals/cfda.md
```diff
@@ -0,0 +1,53 @@
...
+                        "closed": 234,
+                        "archived": 1685,
+                        "forecasted": 208
+                    }
+                ]
+            }
+
+# Data Structures
+## CFDA (object)
++ `cfda` (required, string, nullable)
```
tony-sappe, line 44: Is null a valid CFDA?

BrianZito: it is not; corrections have been made

usaspending_api/references/v2/views/cfda.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class CFDAViewSet(APIView):
+    """
+    Return an agency name and active fy.
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/references/agency/id.md"
+
+    def get(self, request, cfda=None):
```
tony-sappe, line 17: No API cache?

usaspending_api/references/v2/views/cfda.py
```diff
@@ -0,0 +1,48 @@
...
+
+ALL_CFDAS = None
+
+
+class CFDAViewSet(APIView):
+    """
+    Return an agency name and active fy.
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/references/agency/id.md"
```
tony-sappe: endpoint doc string is wrong

usaspending_api/references/v2/views/cfda.py
```diff
@@ -0,0 +1,48 @@
...
+        global ALL_CFDAS
+        if not ALL_CFDAS:
+            response = self._request_from_grants_api()
+
+            #  grants API is brittle in practice, so if we don't get results retry at a polite rate
+            while not response:
+                sleep(2)
+                response = self._request_from_grants_api()
+
+            ALL_CFDAS = response.values()
```
tony-sappe, line 42: Wouldn't it be helpful to store the "cfdas" as the dictionary for looking up a specific CFDA? I would still be easy to  returning all CFDA records using `.values()`

usaspending_api/references/v2/views/cfda.py
```diff
@@ -0,0 +1,48 @@
...
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/references/agency/id.md"
+
+    def get(self, request, cfda=None):
+        """
+        Return the view's queryset.
+        """
+        self._populate_cfdas_if_needed()
+
+        if cfda:
+            results = [result for result in ALL_CFDAS if result["cfda"] == cfda]
```
tony-sappe, line 24: Since this is a GET request, shouldn't requesting a specific cfda which doesn't exist result in a 404 instead of an empty results response?

usaspending_api/settings.py
```diff
@@ -122,6 +122,9 @@
 ES_REPOSITORY = ""
 ES_ROUTING_FIELD = "recipient_agg_key"
 
+# Grants API
+GRANTS_API_KEY = os.environ.get("GRANTS_API_KEY")
```
tony-sappe, line 126: Make sure you work with OPS on this, the older pattern was to not use env vars but inject the config into the settings,py file using regex. I never learned why....

usaspending_api/references/v2/views/cfda.py
BrianZito: This was probably updated while you were reviewing.

usaspending_api/references/v2/views/cfda.py
BrianZito: If the endpoint was always returning a single result instead of a list, I would absolutely agree with you. With there being a list (that could have multiple values), I think there's a grey area. An alternative would be to have the endpoint not return a list when filtered down to one CFDA code (since this should only be getting one result anyway), then returning 404 if the CFDA isn't found, but that would be changing the contract underneath frontend mid-sprint, so I'd  want to talk to John first.

usaspending_api/references/v2/views/cfda.py
```diff
@@ -0,0 +1,48 @@
...
+
+        return Response(response)
+
+    def _populate_cfdas_if_needed(self):
+        global ALL_CFDAS
+        if not ALL_CFDAS:
+            response = self._request_from_grants_api()
+
+            #  grants API is brittle in practice, so if we don't get results retry at a polite rate
+            while not response:
```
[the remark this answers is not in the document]
BrianZito: As for the infinite loop, yes, it would be possible if the grants API went down indefinitely, although our daily redeploy means that no request would hog resources permanently. Either way, it would be easy enough to add a counter to just "give up" after a minute since we wouldn't be returning anything to the user after that point anyway.

It is true that if Grants API gives a different response structure than what was expected, this code would throw an exception. In that case, however, I would argue that the correct response would be to fail, and 500 would be the most appropriate response code, since a service used on the server side of the architecture failed.

usaspending_api/references/v2/views/cfda.py
tony-sappe: There are two endpoints in this view, the responses don't need to be identical between them.

Can multiple CFDAs be returned when the client provides a specific CFDA `/api/v2/references/cfda/totals/{?cfda_code}/` ?

BrianZito: It never should, since Grants API returns a JSON object instead of a JSON array, which guarantees unique keys. The only reason to keep this as a list is to avoid adding more work to @jonhill13 late in the sprint. John, do you forsee a problem if the api only returned results for a single CFDA instead of a results array when filtering on a single CFDA? I can get you a sample of what that would look like for you to develop against.

usaspending_api/references/v2/views/cfda.py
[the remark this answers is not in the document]
tony-sappe: > Additionally, it seems there is an expectation an exception will be raised in _request_from_grants_api() from .json()["cfdas"] since that is making hard assumptions about the response.

The context of that comment was for breaking out of the infinite loop. There was no explicit checking or error handling to help end the infinite loop.

In regards to your comment, it would be a good idea to add helpful logging to know if the issue was unreachable server, unexpected API response, authentication/missing api, or whatever likely situation you expect.

usaspending_api/settings.py
bminahan-kc: we can make this work. We will handle the same way as the database secrets. We should probalby rename the uwsgi file containing secrets from `uwsgi_db.ini` to something else now that there is a secret that is _not_ related to the databsae.

usaspending_api/references/v2/views/cfda.py
jonhill13: Yes, i would recommend returning just an object when returning a specific result and if an error or any other message is returned from the grants api return that.

The only reason I wrote the contract like that was to make it easier on the backend, but in hindsight, I do not know why I did that. I guess trying to be nice to the backend but we should be following best practice @BrianZito @tony-sappe

usaspending_api/settings.py
bminahan-kc: [this PR](https://github.com/fedspendingtransparency/data-act-build-tools/pull/269) will make sure the API key is added. Once that is merged in we should be good to go there.

jonhill13 (approved):

dpb-bah (approved):

dpb-bah:
Approving but with a couple caveats:
* https://github.com/fedspendingtransparency/data-act-build-tools/pull/269 gets merged
* There's an underlying flaw with this approach in that it makes the endpoint dependent on an external API which could go down for any reason for any length of time out of our control. Suggestion would be to make a couple tickets to rework this on the backend such that 1) a nightly loader gathers the data needed from the Grants API and stores it in a table/view and 2) the endpoint the FE calls simply refers to that table/view.

tony-sappe:
> Approving but with a couple caveats:
> 
> * [fedspendingtransparency/data-act-build-tools#269](https://github.com/fedspendingtransparency/data-act-build-tools/pull/269) gets merged
> * There's an underlying flaw with this approach in that it makes the endpoint dependent on an external API which could go down for any reason for any length of time out of our control. Suggestion would be to make a couple tickets to rework this on the backend such that 1) a nightly loader gathers the data needed from the Grants API and stores it in a table/view and 2) the endpoint the FE calls simply refers to that table/view.
>   @tony-sappe @jonhill13

@dpb-bah Thanks for reviewing the PR. There were a few discussions on the approach and when the sprint started it was expected the website react app would call the grants API directly. There were last-minute changes so this was settled on to complete the ticket. It's definitely a weakness, not sure I would call it a design flaw.

PR #2767, opened 2020-09-04
[DEV-5039] Increases the pagination limit on the Spending by Category endpoint to 9999

collinwr:
**Description:**
This PR increases the max number of results returned from the Spending by Category endpoint from 100 to 9999. This endpoint was recently modified to use Elasticsearch, which allowed for more records to be returned while maintaining good performance. A performance investigation was performed to ensure reasonable execution times. 

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
8. [ ] Jira Ticket [DEV-5039](https://federal-spending-transparency.atlassian.net/browse/DEV-5039):

```
```

### DIFF
--- a/usaspending_api/api_contracts/contracts/v2/search/spending_by_category.md
+++ b/usaspending_api/api_contracts/contracts/v2/search/spending_by_category.md
@@ -56,7 +56,7 @@ This endpoint returns a list of the top results of specific categories sorted by
     + Attributes (object)
         + `category` (required, string)
         + `results` (required, array[CategoryResult], fixed-type)
-        + `limit` (required, number)
+        + `limit` (required, number) Maximum - 9,999
         + `page_metadata` (PageMetadataObject)
         + `messages` (optional, array[string])
             An array of warnings or instructional directives to aid consumers of this endpoint with development and debugging.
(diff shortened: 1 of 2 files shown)


kwhickey (changes requested):
@collinwr While a single requester could request 9999 items at once, I know that our UI could not display that adequately. My impression was that, since the the UI currently only puts 10 items on a single page for this endpoint, as seen in the image below, that it was what our backend was currently limiting it to. 

If our backend already was allowing up to 100 items at a time, then I think that is sufficient for UI needs, and we should not increase that limit. I don't think showing more than 100 on one page is being considered. While they are thinking of upping it, I think it would be perhaps to 25 or 50 or 100 at most per page.  

So in summary: 
- If it already allows 100 items to be requested, go ahead and close this. 
- If it limits it to something less 100 currently, update this PR to change that limit to 100 max.

collinwr:
> So in summary:
> 
> * If it already allows 100 items to be requested, go ahead and close this.
> * If it limits it to something less 100 currently, update this PR to change that limit to 100 max.

@kwhickey The current limit 100. I'll close the PR and update the ticket

PR #2780, opened 2020-09-16
[DEV-5944] Adding partial index to support index-only query for disaster award endpoint

klrBAH:
**Description:**
This PR adds a multi-column partial index that improves the performance of the counts and totals query called by the api/v2/disaster/award/amount/ endpoint

**Technical details:**
Partial index filtered to disaster awards (disaster_emergency_fund_code" IN ('L', 'M', 'N', 'O', 'P')). Further evaluation of existing indexes on FABA table needed and we may eventually be able to remove the existing financial_accounts_by_aw_disaster_emergency_fund__a84f52cd_idx index to improve data load into the table since we now have the distinct_award_key.

1. [ ] N/A Unit & integration tests updated
2. [ ] N/A API documentation updated
    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
4. [ ] N/A Matview impact assessment completed
5. [ ] N/A Frontend impact assessment completed
6. [ ] N/A Data validation completed
7. [ ] N/A Appropriate Operations ticket(s) created
8. [ ] Jira Ticket [DEV-5944](https://federal-spending-transparency.atlassian.net/browse/DEV-123):

```
```


usaspending_api/awards/models/financial_accounts_by_awards.py
```diff
@@ -148,3 +149,9 @@ class Meta:
                 "gross_outlay_amount_by_award_cpe",
             ]
         ]
+        indexes = [
+            models.Index(
+                fields=['submission','distinct_award_key', 'transaction_obligated_amount','gross_outlay_amount_by_award_cpe'],
+                name='faba_subid_awardkey_sums_idx',
+                condition=Q(disaster_emergency_fund__in=['L','M','N','O','P']))
```
kwhickey, line 156: One thing we'll have to watch out for is this hard-coding. If a new set of disaster codes comes around ==> new index. Or if COVID expands to incorporate another disaster code, we'll need to update this index.

usaspending_api/awards/migrations/0076_auto_20200916_0312.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('awards', '0075_financialaccountsbyawards_distinct_award_key'),
+    ]
+
+    operations = [
+        migrations.AddIndex(
+            model_name='financialaccountsbyawards',
+            index=models.Index(condition=models.Q(disaster_emergency_fund__in=['L', 'M', 'N', 'O', 'P']), fields=['submission', 'distinct_award_key', 'transaction_obligated_amount', 'gross_outlay_amount_by_award_cpe'], name='faba_subid_awardkey_sums_idx'),
```
kwhickey, line 15: One other thing that came to my attention when looking at QAT is that I think the counts of awards for CFDA uses a different query, that may not benefit as much from the `distinct_award_key` or this index. Because it first tries to narrow down to "All Assistance Awards". I think it does that by checking FAIN or URI is not null. @tony-sappe can you clarify how that query is run and how it's alike or different from the one for "All Awards" in its use of award identifiers?

If that's the case, and it is isolating on "NULLness" of FAIN/URI, should we add those to this index? Or add another partial index similar to this that has a partial clause about whether those are NULL?

kwhickey (approved):
Code as is looks fine. One thing to consider for CFDA and another potential note-to-self around the partial index.

Also Travis wants you to _black_ the code: https://github.com/fedspendingtransparency/usaspending-api/blob/dev/CONTRIBUTING.md#code-style

usaspending_api/awards/migrations/0076_auto_20200916_0312.py
tony-sappe: @kwhickey the query to count assistance awards check if `piid` is null or not-null

kwhickey: Thanks. I had it backwards. 
@tony-sappe is it using `distinct_award_key`, like?
```sql
SELECT ..., count(distinct distinct_award_key)
WHERE faba.piid IS NULL
GROUP BY faba.distinct_award_key
```
I'm wondering if that query will or will not use this index. And if that query could be much more improved if a partial index was added with condition of `piid IS NULL and disaster_emergency_fund_code IN ('L','M','N','O','P')`

tony-sappe: I don't see how that query using `piid` can use the new index for an **index-only scan**. It could still use the index depending on what the planner determines to be the best path.

tony-sappe (approved):

PR #2807, opened 2020-09-29
[DEV-6153] Adding server-side timeout to Download Jobs

tony-sappe:
**Description:**
There are suspicions some download COPY SQL statements continue to run past the desired timeout. These changes add a server-side timeout to the psql connection.

Leaving the original timeout code in-place since the error report is more helpful.

**Technical details:**

    - [ ] Backend
    - [ ] Operations
8. [x] Jira Ticket [DEV-6153](https://federal-spending-transparency.atlassian.net/browse/DEV-6153):
    - [x] Performance evaluation of affected Download (Same)

```
No changes to API, matviews, DB, or changes to desired behavior expectations
```


BrianZito (approved):
Tested by setting a very short timeout window. This does have the side effect of returning a 500 error if the DB query times out, but I would argue that's the correct code to send back in that case, and it only applies if we made the timeout unreasonably small.

PR #2817, opened 2020-10-02
Ftr/dev 5892 chunk utm refresh

collinwr:
**Description:**
This changes the `universal_transaction_matview` to a `universal_transaction_table`. This is done by creating "chunked" matviews containing a subset of the full `universal_transaction_matview` then combining them into a table

**Technical details:**
This is accomplished with the following pieces:
- `matview_sql_generator.py` now contains a new arg, `chunks`, that breaks a matview into provided number of chunks
- A new `combine_universal_transaction_matview_chunks` command combines the chunks into a table

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [DEV-5892](https://federal-spending-transparency.atlassian.net/browse/DEV-5892):

```
```


usaspending_api/database_scripts/matview_generator/special_matviews/universal_transaction_matview.json
```diff
@@ -156,56 +156,6 @@
     "      OR rl_country_lookup.country_name = COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code))",
```
tony-sappe, line 156: Might be better to keep this file in its original location

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,158 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument("--chunk-count", default=10, help="Number of chunked matviews to read from", type=int)
+        parser.add_argument(
+            "--analyze", action="store_true", help="Indicates whether table should be analyzed", type=int
+        )
+
+    def handle(self, *args, **options):
+        chunk_count = options["chunk_count"]
+
+        logger.info("Chunk Count: {}".format(chunk_count))
```
tony-sappe, line 93: Do you not like f-strings?

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,158 @@
...
+ALTER INDEX idx_ut_pop_zip5_temp RENAME TO idx_ut_pop_zip5;
+ALTER INDEX idx_ut_recipient_unique_id_temp RENAME TO idx_ut_recipient_unique_id;
+ALTER INDEX idx_ut_parent_recipient_unique_id_temp RENAME TO idx_ut_parent_recipient_unique_id;
+ALTER INDEX idx_ut_simple_pop_geolocation_temp RENAME TO idx_ut_simple_pop_geolocation;
+ALTER INDEX idx_ut_recipient_hash_temp RENAME TO idx_ut_recipient_hash;
+ALTER INDEX idx_ut_action_date_pre2008_temp RENAME TO idx_ut_action_date_pre2008;
+"""
+
+ANALYZE_TABLE_SQL = """
+ANALYZE VERBOSE universal_transaction_table;
```
tony-sappe, line 69: You'll need a ton of code changes if you're going to change the name

usaspending_api/database_scripts/matview_generator/matview_sql_generator.py
```diff
@@ -371,6 +372,18 @@ def create_monolith_file(sql_json):
     write_sql_file(sql_strings, os.path.join(DEST_FOLDER, sql_json["final_name"]))
 
 
+def add_chunk_strings(sql_json, chunk):
+    chunked_sql_json = copy.deepcopy(sql_json)
+
+    chunk_count = GLOBAL_ARGS.chunks
+
+    if chunk_count > 1:
+        chunked_sql_json["final_name"] += "_{}".format(chunk)
```
sethstoudenmier, line 381: If you are still working this would be nice to see f-strings here (e.g. `f"_{chunk}"`) and below.

usaspending_api/database_scripts/matview_generator/matview_sql_generator.py
```diff
@@ -371,6 +372,18 @@ def create_monolith_file(sql_json):
     write_sql_file(sql_strings, os.path.join(DEST_FOLDER, sql_json["final_name"]))
 
 
+def add_chunk_strings(sql_json, chunk):
+    chunked_sql_json = copy.deepcopy(sql_json)
+
+    chunk_count = GLOBAL_ARGS.chunk_count
+
+    if chunk_count > 1:
+        chunked_sql_json["final_name"] += f"_{chunk}"
```
tony-sappe, line 381: Minor suggestion to simplify reading the line. Only todo if there are other changes.

```suggestion
        chunked_sql_json["final_name"] = f"{chunked_sql_json['final_name']}_{chunk}"
```

usaspending_api/common/management/commands/matview_runner.py
```diff
@@ -13,6 +13,7 @@
     DEPENDENCY_FILEPATH,
     DROP_OLD_MATVIEWS,
     MATERIALIZED_VIEWS,
+    CHUNKED_MATERIALIZED_VIEWS,
```
tony-sappe, line 16: Minor detail, these imports were listed alphabetically. Not all in the codebase are, but it is moving in that direction

usaspending_api/common/management/commands/matview_runner.py
```diff
@@ -75,21 +79,37 @@ def generate_matview_sql(self):
...
+        # Create Matviews
         for matview, config in self.matviews.items():
-            logger.info("Creating Future for {}".format(matview))
+            logger.info(f"Creating Future for matview {matview}")
             sql = (self.matview_dir / config["sql_filename"]).read_text()
             tasks.append(asyncio.ensure_future(async_run_creates(sql, wrapper=Timer(matview)), loop=loop))
 
+        # Create Chunked Matviews
+        for matview, config in self.chunked_matviews.items():
+            for current_chunk in range(0, self.chunk_count):
```
tony-sappe, line 107: Including the default starting index of 0 isn't typical, unless if a different step size is also provided.

```suggestion
            for current_chunk in range(self.chunk_count):
```

usaspending_api/database_scripts/matview_generator/matview_sql_generator.py
```diff
@@ -383,8 +396,13 @@ def main(source_file):
         print("Error on Matview source JSON file: {}".format(source_file))
         print(e)
         raise SystemExit(1)
-    create_monolith_file(sql_json)
-    create_componentized_files(sql_json)
+
+    for chunk in range(0, GLOBAL_ARGS.chunk_count):
```
tony-sappe, line 400: same comment about range here too

usaspending_api/database_scripts/tests/test_matview_sql_generator.py
```diff
@@ -1,8 +1,11 @@
 import pytest
 
 from django.db import connection
+from collections import OrderedDict
 from usaspending_api.common.helpers.generic_helper import generate_matviews
-from usaspending_api.common.matview_manager import MATERIALIZED_VIEWS
+from usaspending_api.common.matview_manager import MATERIALIZED_VIEWS, CHUNKED_MATERIALIZED_VIEWS
+
+ALL_MATVIEWS = OrderedDict(list(MATERIALIZED_VIEWS.items()) + list(CHUNKED_MATERIALIZED_VIEWS.items()))
```
tony-sappe, line 8: Dictionaries already preserve order in Python 3.6+ (happy side-effect from other changes in 3.6, added to language specification for 3.7 and forward)

tony-sappe: Perhaps this will satisfy the combining of dictionaries.

```suggestion
ALL_MATVIEWS = {**MATERIALIZED_VIEWS, **CHUNKED_MATERIALIZED_VIEWS}
```

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+DROP {old_object_type} IF EXISTS universal_transaction_matview_temp;
+CREATE TABLE universal_transaction_matview_temp AS SELECT * from universal_transaction_matview_0 WITH NO DATA;
+"""
+
+INSERT_INTO_TABLE_SQL = """
+INSERT INTO universal_transaction_matview_temp SELECT * FROM universal_transaction_matview_{current_chunk}
+"""
+
+TABLE_INDEX_SQL = """
+CREATE UNIQUE INDEX idx_ut_transaction_id_temp ON universal_transaction_matview_temp USING BTREE(transaction_id ASC NULLS LAST) WITH (fillfactor = 97);
```
tony-sappe, line 22: Not a fan of specifying the DDL of the table in separate files. A management command wouldn't be my first guess. I can see people forgetting where the indexes are defined.

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+CREATE INDEX idx_ut_recipient_unique_id_temp ON universal_transaction_matview_temp USING BTREE(recipient_unique_id) WITH (fillfactor = 97) WHERE recipient_unique_id IS NOT NULL AND action_date >= '2007-10-01';
+CREATE INDEX idx_ut_parent_recipient_unique_id_temp ON universal_transaction_matview_temp USING BTREE(parent_recipient_unique_id) WITH (fillfactor = 97) WHERE parent_recipient_unique_id IS NOT NULL AND action_date >= '2007-10-01';
+CREATE INDEX idx_ut_simple_pop_geolocation_temp ON universal_transaction_matview_temp USING BTREE(pop_state_code, action_date) WITH (fillfactor = 97) WHERE pop_country_code = 'USA' AND pop_state_code IS NOT NULL AND action_date >= '2007-10-01';
+CREATE INDEX idx_ut_recipient_hash_temp ON universal_transaction_matview_temp USING BTREE(recipient_hash) WITH (fillfactor = 97) WHERE action_date >= '2007-10-01';
+CREATE INDEX idx_ut_action_date_pre2008_temp ON universal_transaction_matview_temp USING BTREE(action_date) WITH (fillfactor = 97) WHERE action_date < '2007-10-01';
+"""
+
+SWAP_TABLES_SQL = """
+DROP {old_object_type} IF EXISTS universal_transaction_matview_old;
```
tony-sappe, line 37: I understand why it was done this way. This is similar to the first SQL scripts for the materialized views. Updating and managing those files was a massive headache.

Is there any other solution to make the renaming more programatic so they are defined once and devs don't need to make three changes?

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+ALTER INDEX idx_ut_pop_zip5_temp RENAME TO idx_ut_pop_zip5;
+ALTER INDEX idx_ut_recipient_unique_id_temp RENAME TO idx_ut_recipient_unique_id;
+ALTER INDEX idx_ut_parent_recipient_unique_id_temp RENAME TO idx_ut_parent_recipient_unique_id;
+ALTER INDEX idx_ut_simple_pop_geolocation_temp RENAME TO idx_ut_simple_pop_geolocation;
+ALTER INDEX idx_ut_recipient_hash_temp RENAME TO idx_ut_recipient_hash;
+ALTER INDEX idx_ut_action_date_pre2008_temp RENAME TO idx_ut_action_date_pre2008;
+"""
+
+REMOVE_OLD_TABLE_SQL = """
+DROP TABLE IF EXISTS universal_transaction_matview_old;
```
tony-sappe, line 72: I would prefer short strings to be single-line

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+        with Timer("Creating table indexes"):
+            self.create_indexes()
+
+        with Timer("Swapping Tables/Indexes"):
+            self.swap_matviews()
+
+        if not options["keep_old_data"]:
+            self.remove_old_data(chunk_count)
+
+        if options["analyze"]:
```
tony-sappe, line 125: I can't think of a good reason to not analyze the new table after it is created and indexed. Perhaps it can always occur and the cli argument can be removed

tony-sappe (changes requested):
Biggest concern is separating the matview/table definition across different files.

Also a few minor comments which aren't blockers.

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
kwhickey: What's the alternative @tony-sappe? 

Or maybe you can expand upon the comment? 
- Is it that the DDL went from being stored in a JSON file to a Python file? If so, I kindof like that and have always thought managing SQL strings in JSON was really tough.
- Or is it that this DDL in this python file is one part, and there's more DDL in another Python (or JSON) file? -- disclosure: haven't done a full review of the code yet

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+        with connection.cursor() as cursor:
+            # This table was previously a Matview, so the DROP may fail
+            try:
+                cursor.execute(SWAP_TABLES_SQL.format(old_object_type="TABLE"))
+            except Exception as e:
+                if "is not a table" in str(e):
+                    logger.info(
+                        "universal_transaction_matview_temp existed, but not as table. Trying to drop as Matview"
+                    )
+                    cursor.execute(SWAP_TABLES_SQL.format(old_object_type="MATERIALIZED VIEW"))
```
kwhickey, line 179: This batch of SQL statements in `SWAP_TABLES_SQL` needs to be run in a single atomic database transaction. 

- It all either needs to succeed
- Or it all needs to fail

For instance, done this way, there will be a period of time where `universal_transaction_matview` doesn't exist at all. Between the rename to `_old` and the rename from `_temp` while the indexes are being altered. 

And then the indexes are coming online for the new one 1 index at a time.

Since this is a management command, the `connection` is a django managed connection, so you _could_ using the 

```python
from django.db import transaction

@transaction.atomic
def swap_matviews(self):
```

See: https://docs.djangoproject.com/en/2.2/topics/db/transactions/#controlling-transactions-explicitly

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def remove_old_data(self, chunk_count):
```
kwhickey, line 181: This will get called by default unless someone supplies `--keep-old-data`

The removal of table with suffix `_old` is fine. But this:
`REFRESH MATERIALIZED VIEW universal_transaction_matview_{current_chunk} WITH NO DATA;`

Getting called will _empty_ out each of the matview chunks. Meaning the next night, they need to be created from scratch again rather than updated via a `REFRESH` (which under the hood compares old rows to new rows).

Is that what we want ☝️ ??
- I would have though that doing a refresh from night to night rather than a recreate on the matviews would save on execution time of "re-hydrating" the matview chunks with new data.

kwhickey (changes requested):
Some things to change and a handful of clarification/validation questions to answer

- ⚠️ swaps need to be done in a transaction
- Some issues/questions with the args and what their default settings will lead to (empty matviews?)
- Other questions/clarifications and things to try for performance

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+class Command(BaseCommand):
+
+    help = """
+    This script combines the chunked Universal Transaction Matviews and
+    combines them into a single table.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument("--chunk-count", default=10, help="Number of chunked matviews to read from", type=int)
+        parser.add_argument("--analyze", action="store_true", help="Indicates whether table should be analyzed")
```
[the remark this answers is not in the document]
collinwr: Based on some previous feedback, I removed the argument. The command now analyzes the tables during every run.

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
tony-sappe: My comment is about parts of the matview DDL being stored in separate files not located near each other in the codebase.

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
tony-sappe: > Is that what we want ☝️ ??
@kwhickey yes. Why not? @collinwr said the times of the mativew creation was very similar to a refresh. Are there more than these three options?

1. Leave the "chunked" matviews in-place during the day, consuming approx 120GB but useless for any queries
2. Refresh with no data which recovered the storage and leaves the matviews around for simpler pipeline operations
3. Drop the matviews afterward, which would be closer to "created from scratch".


@collinwr currently the refresh SQL will fail. A materialized view with no data cannot be concurrently refreshed. Additionally, even with data a concurrent refresh requires a unique index on the matview

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+from django.db import connection
+from django.core.management.base import BaseCommand
+from usaspending_api.common.data_connectors.async_sql_query import async_run_creates
+from usaspending_api.common.helpers.timing_helpers import ConsoleTimer as Timer
+
+logger = logging.getLogger("script")
+
+RECREATE_TABLE_SQL = """
+DROP {old_object_type} IF EXISTS universal_transaction_matview_temp;
+CREATE TABLE universal_transaction_matview_temp AS SELECT * from universal_transaction_matview_0 WITH NO DATA;
```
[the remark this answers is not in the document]
collinwr: I did a schema comparison between all three options: 
- Chunked Matview
- Table from `CREATE TABLE ... AS`
- Table from `CREATE TABLE ... (LIKE`

And all three had the same:
- Constraints (none)
- Foreign keys (none)
- Not NULL constraints (none)
- Column Data types and precision

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+                    logger.info(
+                        "universal_transaction_matview_temp existed, but not as table. Trying to drop as Matview"
+                    )
+                    cursor.execute(RECREATE_TABLE_SQL.format(old_object_type="MATERIALIZED VIEW"))
+
+    def insert_matview_data(self, chunk_count):
+        loop = asyncio.new_event_loop()
+        tasks = []
+
+        for current_chunk in range(0, chunk_count):
```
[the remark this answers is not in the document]
collinwr: I just tested this. There are not records in the table after running `CREATE TABLE universal_transaction_matview_temp AS SELECT * from universal_transaction_matview_0 WITH NO DATA;`

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+    def recreate_matview(self):
+        with connection.cursor() as cursor:
+            # This table was previously a Matview, so the DROP may fail
+            try:
+                cursor.execute(RECREATE_TABLE_SQL.format(old_object_type="TABLE"))
+            except Exception as e:
+                if "is not a table" in str(e):
+                    logger.info(
+                        "universal_transaction_matview_temp existed, but not as table. Trying to drop as Matview"
+                    )
```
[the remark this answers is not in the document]
collinwr: Is something like this better?
"universal_transaction_matview_temp existed, but not as table. This may be because this command is being run for the first time. Trying to drop as Matview instead...

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+    def swap_matviews(self):
+        with connection.cursor() as cursor:
+            # This table was previously a Matview, so the DROP may fail
+            try:
+                cursor.execute(SWAP_TABLES_SQL.format(old_object_type="TABLE"))
+            except Exception as e:
+                if "is not a table" in str(e):
+                    logger.info(
+                        "universal_transaction_matview_temp existed, but not as table. Trying to drop as Matview"
+                    )
```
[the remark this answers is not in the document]
collinwr: Updated this with the same message as above

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
collinwr: This has been added

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
collinwr: I'm not sure we need to perform the refresh concurrently because the matviews are not live (being used by the application). They are only used to insert data into the table. Is there another reason to do the refresh concurrently?

tony-sappe: @collinwr we don't need concurrent refreshes. But unless I am mistaken the SQL being stored in `*__refresh.sql` for the chunked materialized views is
```sql
REFRESH MATERIALIZED VIEW CONCURRENTLY universal_transaction_matview_9 WITH DATA;
VACUUM ANALYZE VERBOSE universal_transaction_matview_9;
```

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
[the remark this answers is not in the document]
kwhickey: that works

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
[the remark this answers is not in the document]
kwhickey: 👍 leave as is. thanks for checking.

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
kwhickey: We should chat about this. If we're recreating the mativiews from scratch (going from NO DATA to DATA everyday) ... why are they matviews? The chunks should just be tables that are truncated and inserted into created from scratch each day. 

Matviews have the benefit of a REFRESH to only update _part_ of them. And we're not using that benefit.

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,200 @@
...
+
+        with Timer("Recreating table"):
+            self.recreate_matview()
+
+        with Timer("Inserting data into table"):
+            self.insert_matview_data(chunk_count)
+
+        with Timer("Creating table indexes"):
+            self.create_indexes()
+
```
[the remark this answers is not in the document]
kwhickey: Can try as a follow on. Can you create a ticket which covers this and perhaps covers testing performance comparisons of 1) working on empty matview chunks each day, 2) refreshing full matview chunks each day (data from the day prior), 3) recreating or rehydrating empty tables each day.

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
tony-sappe: Open and available to chat.

> why are they matviews

Why not? Matviews have several benefits, and theses scripts _are_ leveraging matview capabilities. I am curious to hear why recovering the data is concerning to you

tony-sappe (commented):
`matview_runner.py` and `combine_universal_transaction_matview_chunks` worked locally. I'm including minor requests in comments. Didn't complete a full review

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
```diff
@@ -0,0 +1,162 @@
...
+            "--keep-old-data",
+            action="store_true",
+            default=False,
+            help="Indicates whether or not to drop old table at end of command",
+        )
+        parser.add_argument(
+            "--keep-matview-data",
+            action="store_true",
+            default=False,
+            help="Indicates whether or not to empty data from chunked matviews at the end of command",
```
kwhickey, line 40: So just to be clear, we're going with the default of emptying the matview data out each time the table is generated, and then the refresh will refresh it from empty to full on each run of the pipeline?
(this is what `default=False` will lead to).

kwhickey (approved):
Looks mostly good to me. Just a few questions in the comments provided. 

I think the extraction of common code is fine. Favoring composition over inheritance, and the new "makes" for the chunked one follow the same naming patterns.

Have you been able to test in a lower env using the parameterized pipeline? E.g. running just the changed-stages in a pipeline that "ingests" from broker sandbox to usaspending sandbox?

usaspending_api/etl/management/commands/combine_universal_transaction_matview_chunks.py
collinwr: That's correct @kwhickey

usaspending_api/common/management/commands/matview_runner.py
```diff
@@ -74,25 +90,50 @@ def generate_matview_sql(self):
...
+            for current_chunk in range(self.chunk_count):
+                chunked_matview = f"{matview}_{current_chunk}"
+                logger.info(f"Creating Future for chunked matview {chunked_matview}")
+                sql = (self.matview_chunked_dir / f"{chunked_matview}.sql").read_text()
+                tasks.append(asyncio.ensure_future(async_run_creates(sql, wrapper=Timer(chunked_matview)), loop=loop))
+
         loop.run_until_complete(asyncio.gather(*tasks))
         loop.close()
 
+        if "universal_transaction_matview" in self.chunked_matviews:
```
tony-sappe, line 133: It seems like universal_transaction_matview will always be in that list. Is there a way to not run `universal_transaction_matview`?

Originally, using the --only flag to select a specific matview to run would prevent the others from running. I don't see any logic to prevent UTM from running in this script

collinwr: Ah, I see what you meant by your original comment. I'll get this addressed

tony-sappe (approved):
Seems good enough to test in QAT

PR #2831, opened 2020-10-27
[DEV-6052] Simplifying on-the-fly SQL for transactions ES ETL

tony-sappe:
**Description:**
The SQL view used for the ES ETL, `transaction_delta_view`, became much too complicated and no longer looked remotely close to its humble origins. A number of derived columns were move to the underlying relations, universal_transaction_matview (a table) and associated materialized views.

**Technical details:**
- One column is new to transaction_delta_view: `etl_update_date`
- Two columns were renamed in Elasticsearch: `award_description` -> `transaction_description`, `transaction_fiscal_year` -> `fiscal_year`

    - [x] Backend
8. [x] Jira Ticket [DEV-6052](https://federal-spending-transparency.atlassian.net/browse/DEV-6052):

```
No changes to the API
```


usaspending_api/etl/es_transaction_template.json
```diff
@@ -47,6 +47,11 @@
         "format": "yyyy-MM-dd HH:mm:ss||yyyy-MM-dd||epoch_millis",
         "index": false
       },
+      "etl_update_date": {
+        "type": "date",
+        "format": "yyyy-MM-dd HH:mm:ss||yyyy-MM-dd||epoch_millis",
+        "index": false
```
kwhickey, line 53: why `"index": false`? 

I haven't see yet where this is used, but if it's used at all in ES queries (i.e. to find records for delete/upsert), it should be indexed.

usaspending_api/etl/management/commands/elasticsearch_indexer.py
```diff
@@ -227,7 +227,7 @@ def set_config(passthrough_values: list, arg_parse_options: dict) -> dict:
             "field_for_es_id": "transaction_id",
             "initial_datetime": default_datetime,
             "max_query_size": settings.ES_TRANSACTIONS_MAX_RESULT_WINDOW,
-            "optional_predicate": """WHERE "update_date" >= '{starting_date}'""",
+            "optional_predicate": """WHERE "etl_update_date" >= '{starting_date}'""",
```
kwhickey, line 230: Why use this one here? 
In the `es_rapidloader.py`, it sticks to just comparing to `UTM.update_date`, which is the update date of the `transaction_normalized` record. By using `etl_update_date` it may end up pulling in transactions even if there was no change to the transaction, but because the transaction's award was updated -- which seems to not be the desired outcome (and is avoided in the rapidloader).

kwhickey (changes requested):
Questions on use and indexing of `etl_update_date`, but otherwise good.

usaspending_api/etl/es_transaction_template.json
tony-sappe: Why does it need to be indexed? The current field `update_date` is not indexed

usaspending_api/etl/management/commands/elasticsearch_indexer.py
tony-sappe: > By using etl_update_date it may end up pulling in transactions even if there was no change to the transaction, but because the transaction's award was updated 

Yes, that is the needed behavior

kwhickey: Spoke about this briefly -- looks like this is the one we want to use (since we're pulling fields from the award down into the transaction-level ES index record). 

That being the case, the `es_rapidloader.py` should be updated to use `etl_update_date` to get the next batch of changes. It'd be helpful but not required too if the python param in the function was also renamed from `update_date=` to `etl_update_date=`

@tony-sappe

tony-sappe: already done

kwhickey (approved):
Looks good!

PR #2842, opened 2020-11-09
[DEV-6078] Elasticsearch-backed disaster endpoints for COVID Page Budget Categories refactor

emily-brents:
**Description:**
Continuation of 1st-round PR: https://github.com/fedspendingtransparency/usaspending-api/pull/2837

Re-writes the portions of the `/disaster/agency/spending/`, `/disaster/federal_account/spending/` and `/disaster/object_class/spending/` endpoints to use the new ES Accounts index when `spending_type: "award"`. Also rewrites the entirety of the `/disaster/agency/loans/`, `disaster/federal_account/loans/` and `/disaster/object_class/loans/` to use the same account index.

**Technical details:**
Creates a new `elasticsearch_account_base` class to support the new endpoints. Skips doing the bucket aggregation due to performance concerns.
Note: Agency endpoints use the awards index to do the subtier agency queries as the accounts index does not contain data for subtier agencies.

    - [ ] Backend
8. [X] Jira Ticket [DEV-6294](https://federal-spending-transparency.atlassian.net/browse/DEV-6294)
[DEV-6295](https://federal-spending-transparency.atlassian.net/browse/DEV-6295)
[DEV-6296](https://federal-spending-transparency.atlassian.net/browse/DEV-6296)
[DEV-6297](https://federal-spending-transparency.atlassian.net/browse/DEV-6297)
[DEV-6298](https://federal-spending-transparency.atlassian.net/browse/DEV-6298)
[DEV-6299](https://federal-spending-transparency.atlassian.net/browse/DEV-6299):

```
```


usaspending_api/disaster/tests/integration/test_disaster_agency_spending.py
```diff
@@ -1,16 +1,22 @@
...
 url = "/api/v2/disaster/agency/spending/"
 
 
 @pytest.mark.django_db
-def test_basic_success(client, disaster_account_data, elasticsearch_award_index, monkeypatch, helpers):
-    setup_elasticsearch_test(monkeypatch, elasticsearch_award_index)
+def test_basic_success(client, disaster_account_data, elasticsearch_account_index, monkeypatch, helpers):
+
     helpers.patch_datetime_now(monkeypatch, 2022, 12, 31)
+    bad_date_window = DABSSubmissionWindowSchedule.objects.get(id=2022071)
```
tony-sappe, line 16: There has got to be a better way to setup the mock data

usaspending_api/disaster/tests/integration/test_disaster_agency_spending.py
```diff
@@ -97,17 +103,16 @@ def test_basic_success(client, disaster_account_data, elasticsearch_award_index,
             "code": "007",
             "description": "Agency 007",
             "children": [],
-            "award_count": 1,
+            "award_count": 2,
```
tony-sappe, line 106: This bothers me, I didn't see a new award added to the mock data but I could have missed it. Was the original API or test wrong?

usaspending_api/disaster/tests/integration/test_federal_account_award.py
```diff
@@ -39,16 +42,14 @@ def test_federal_account_award_success(client, generic_account_data, monkeypatch
...
         {
+            "id": 21,
+            "code": "000-0000",
+            "description": "gifts",
+            "award_count": 4,
+            "obligation": 304.0,
+            "outlay": 667.0,
+            "total_budgetary_resources": None,
             "children": [
-                {
```
tony-sappe, line 43: If the order changed, does that mean the sorting behavior is different between the original and new?

tony-sappe: If I have time I'll investigate. The original sorted always added a secondary sort on an id field to provide deterministic inverse asc/desc ordering. But I would assume the default sort field to be obligation or outlay, which appears broken in these results

usaspending_api/disaster/tests/integration/test_federal_account_loans.py
```diff
@@ -16,15 +19,6 @@ def test_federal_account_loans_success(client, generic_account_data, monkeypatch
     expected_results = [
         {
             "children": [
-                {
-                    "code": "2020/52",
-                    "award_count": 1,
-                    "description": "ferns",
```
tony-sappe, line 22: same concern with this order

usaspending_api/disaster/tests/integration/test_object_class_spending_award.py
```diff
@@ -52,18 +60,22 @@ def test_object_class_counts_awards(client, faba_with_object_class_and_two_award
...
 def test_object_class_groups_by_object_classes(
-    client, faba_with_two_object_classes_and_two_awards, monkeypatch, helpers
+    client, elasticsearch_account_index, faba_with_two_object_classes_and_two_awards, monkeypatch, helpers
 ):
+    setup_elasticsearch_test(monkeypatch, elasticsearch_account_index)
     helpers.patch_datetime_now(monkeypatch, 2022, 12, 31)
     helpers.reset_dabs_cache()
 
-    resp = helpers.post_for_spending_endpoint(client, url, def_codes=["M"], spending_type="award")
+    resp = helpers.post_for_spending_endpoint(client, url, def_codes=["M", "N"], spending_type="award")
```
tony-sappe, line 69: Concerned about this change being necessary if the `faba_with_two_object_classes_and_two_awards` fixture changes are not obviously tied

usaspending_api/disaster/v2/views/agency/spending.py
```diff
@@ -57,16 +56,33 @@ def route_agency_spending_backend(request, *args, **kwargs):
     return route_agency_spending_backend
 
 
-class SpendingByAgencyViewSet(PaginationMixin, SpendingMixin, FabaOutlayMixin, DisasterBase):
+class SpendingByAgencyViewSet(
+    PaginationMixin,
+    SpendingMixin,
+    FabaOutlayMixin,
+    ElasticsearchAccountDisasterBase,
+    ElasticsearchSpendingPaginationMixin,
```
tony-sappe, line 64: Can `PaginationMixin` and `DisasterBase` be removed?

tony-sappe (changes requested):
Leaving some comments, I haven't thoroughly reviewed everything. There are concerns with changes to tests for non-obvious reasons. Typically best practice refactors requires tests on the old code. Then replace the old code with new code but still have the original tests pass. It is possible there were problems with the old tests, so the comments are more of a conversation topic than "pls fix" for tests.

usaspending_api/disaster/tests/integration/test_federal_account_award.py
emily-brents: The default sort behavior is to sort on descending obligated amount. I believe the previous behavior was to default to the id field, but this can easily be changed.

usaspending_api/conftest_helpers.py
```diff
@@ -60,28 +64,52 @@ def _add_contents(self, **options):
...
+        for record in records:
             # Special cases where we convert array of JSON to an array of strings to avoid nested types
             routing_key = options.get("routing", settings.ES_ROUTING_FIELD)
-            routing_value = transaction.get(routing_key)
+            routing_value = record.get(routing_key)
             if self.index_type == "transactions":
-                transaction["federal_accounts"] = self.convert_json_arrays_to_list(transaction["federal_accounts"])
+                record["federal_accounts"] = self.convert_json_arrays_to_list(record["federal_accounts"])
+            elif self.index_type == "covid19_faba_":
+                es_id = record.pop("_id")
```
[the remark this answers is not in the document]
emily-brents: This was some odd behavior with the `transform_covid19_faba_data` - it manually sets the `_id` field, but the way that we index the records in this test setup will throw an error if the `_id` field is set, so we pop it to prevent that error.

usaspending_api/disaster/tests/fixtures/object_class_data.py
```diff
@@ -20,13 +20,13 @@ def basic_faba_with_object_class(award_count_sub_schedule, award_count_submissio
         disaster_emergency_fund=DisasterEmergencyFundCode.objects.filter(code="M").first(),
         submission=SubmissionAttributes.objects.all().first(),
         object_class=basic_object_class[0],
+        transaction_obligated_amount=1,
```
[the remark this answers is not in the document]
emily-brents: We prevent results from coming up in the API if the obligated and outlay amounts are zero, I thought that the old code had a similar check, but based on the fact that I had to change this it seems that it did not.

usaspending_api/disaster/tests/integration/test_disaster_agency_spending.py
emily-brents: This is due to changing from using `award_id` to the `financial_account_distinct_award_key`

usaspending_api/disaster/v2/views/agency/spending.py
emily-brents: yeah

usaspending_api/disaster/v2/views/agency/loans.py
```diff
@@ -44,25 +43,49 @@ def route_agency_loans_backend(request, *args, **kwargs):
...
+    agg_key = "financial_accounts_by_award.funding_toptier_agency_id"  # primary (tier-1) aggregation key
+    nested_nonzero_fields = {"outlay": "gross_outlay_amount_by_award_cpe", "obligation": "transaction_obligated_amount"}
+    top_hits_fields = [
+        "financial_accounts_by_award.funding_toptier_agency_code",
+        "financial_accounts_by_award.funding_toptier_agency_name",
+    ]
 
     @cache_response()
     def post(self, request):
+        self.filters.update({"award_type": ["07", "08"]})
```
kwhickey, line 66: See https://github.com/fedspendingtransparency/usaspending-api/pull/2837#discussion_r520177339

usaspending_api/disaster/tests/integration/test_disaster_agency_spending.py
kwhickey: Is this just pulling in a `DABSSubmissionWindowSchedule` that was created elsewhere in a pytest fixture? If so, seems ok (the ID format uses a deterministic date+boolean structure)

usaspending_api/disaster/v2/views/agency/loans.py
```diff
@@ -44,25 +43,49 @@ def route_agency_loans_backend(request, *args, **kwargs):
     return route_agency_loans_backend
 
 
-class LoansByAgencyViewSet(LoansPaginationMixin, LoansMixin, FabaOutlayMixin, DisasterBase):
+class LoansByAgencyViewSet(
+    LoansPaginationMixin, ElasticsearchAccountDisasterBase, LoansMixin, FabaOutlayMixin, DisasterBase
```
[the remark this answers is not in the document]
kwhickey: I think it was because this agency endpoint has a dual impl. If `award_type_codes` provided - hit awards data using subtier agency. If not hit postgres data to get top-tier agency sum. 

But, now that toptier is done in ES, is `DiasterBase` still needed? May it is if this impl handles when "Total" awards (file b data) is also selected?

usaspending_api/disaster/v2/views/federal_account/loans.py
```diff
@@ -1,38 +1,103 @@
...
-from usaspending_api.disaster.v2.views.federal_account.spending import construct_response
+from usaspending_api.disaster.v2.views.elasticsearch_account_base import ElasticsearchAccountDisasterBase
 
 
-class LoansViewSet(LoansMixin, LoansPaginationMixin, FabaOutlayMixin, DisasterBase):
+class LoansViewSet(LoansMixin, LoansPaginationMixin, FabaOutlayMixin, ElasticsearchAccountDisasterBase):
     """ Returns loan disaster spending by federal account. """
 
     endpoint_doc = "usaspending_api/api_contracts/contracts/v2/disaster/federal_account/loans.md"
+    agg_key = "financial_accounts_by_award.treasury_account_id"  # "financial_accounts_by_award.federal_account_id"  primary (tier-1) aggregation key
```
kwhickey, line 18: Is this correct? Why is this part here (looks like commented-out previous impl):

> `# "financial_accounts_by_award.federal_account_id"`

I would think that the top-tier agg is the fed account, and then `sub_agg_key` should be `treasury_account_id`

usaspending_api/disaster/v2/views/federal_account/spending.py
```diff
@@ -31,16 +34,41 @@ def construct_response(results: list, pagination: Pagination):
...
+    SpendingMixin,
+    FabaOutlayMixin,
+    ElasticsearchAccountDisasterBase,
+    ElasticsearchSpendingPaginationMixin,
+    DisasterBase,
+):
     """ Returns disaster spending by federal account. """
 
     endpoint_doc = "usaspending_api/api_contracts/contracts/v2/disaster/federal_account/spending.md"
+    agg_key = "financial_accounts_by_award.treasury_account_id"  # "financial_accounts_by_award.federal_account_id"  # primary (tier-1) aggregation key
```
kwhickey, line 48: Same as previous comment about `federal_account_id` or `treasury_account_id` being the `agg_key` vs. `sub_agg_key`

usaspending_api/search/tests/data/utilities.py
```diff
@@ -5,6 +5,9 @@ def setup_elasticsearch_test(monkeypatch, index_fixture, **options):
     if index_fixture.index_type == "awards":
         search_wrapper = "AwardSearch"
         query_alias = settings.ES_AWARDS_QUERY_ALIAS_PREFIX
+    elif index_fixture.index_type == "covid19_faba_":
+        search_wrapper = "AccountSearch"
+        query_alias = settings.ES_COVID19_FABA_QUERY_ALIAS_PREFIX
     else:
```
kwhickey, line 11: Similar to a previous comment - now that we're expanding our es ETL code to handle arbitrary ES data types, we shouldn't have conditional statements ending on `else: ...`. Should be more like a case statement: either it matches one or fails.

kwhickey (changes requested):
Well cleaned up. A few lingering issues to finalize from the last review (#2837). 

Need to be careful about the weird dual impl for agency loan endpoint based on whether `award_type_codes` are provided or not.

usaspending_api/disaster/v2/views/federal_account/loans.py
emily-brents: Originally yes, but the implementation was changed for performance reasons to use the lower-tier as the agg key and get the toptier data, then combine the results.

usaspending_api/disaster/v2/views/elasticsearch_account_base.py
```diff
@@ -0,0 +1,241 @@
...
+            filters["nested_query"] = {"text": query, "fields": self.query_fields}
+
+        # Ensure that only non-zero values are taken into consideration
+        filters["nested_nonzero_fields"] = list(self.nested_nonzero_fields.values())
+        filters["nonzero_fields"] = self.nonzero_fields
+        self.filter_query = QueryWithFilters.generate_accounts_elasticsearch_query(filters)
+        # using a set value here as doing an extra ES query is detrimental to performance
+        # And the dimensions on which group-by aggregations are performed so far
+        # (agency, TAS, object_class) all have cardinality less than this number
+        self.bucket_count = 1000
```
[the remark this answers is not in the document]
emily-brents: We have 176 TAS currently, so I don't anticipate it being a problem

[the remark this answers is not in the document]
tony-sappe: Perhaps 176 TAS associated with DEFC. A quick count showed 18,434 in the database. Not sure what projections we have for TAS in future COVID spending

usaspending_api/disaster/v2/views/agency/loans.py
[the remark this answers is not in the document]
tony-sappe: There were other ways that dual-implementation was handled. Creating the view class from two "base" classes really muddies the water

usaspending_api/etl/elasticsearch_loader_helpers/transform_data.py
```diff
@@ -38,18 +38,24 @@ def transform_covid19_faba_data(worker: TaskSpec, records: List[dict]) -> List[d
...
         if temp_key not in results:
             results[temp_key] = {
                 "financial_account_distinct_award_key": disinct_award_key,
                 "award_id": award_id,
-                "award_type": award_type,
+                "type": award_type,
                 "generated_unique_award_id": generated_unique_award_id,
                 "total_loan_value": total_loan_value,
                 "financial_accounts_by_award": list(),
+                "obligated_sum": obligated_sum,
```
tony-sappe, line 52: I recommend following the pattern for `financial_accounts_by_award`. For these two fields simply set the initial value to 0 in the dict, then add the new values below. Doesn't require the else

usaspending_api/etl/elasticsearch_loader_helpers/transform_data.py
```diff
@@ -38,18 +38,24 @@ def transform_covid19_faba_data(worker: TaskSpec, records: List[dict]) -> List[d
         award_type = record.pop("award_type")
```
tony-sappe, line 38: Might as well remove the alias in the SQL view so it is `type` all the way through the ETL pipeline

usaspending_api/disaster/v2/views/federal_account/loans.py
kwhickey: Ok. probably should remove the  `"financial_accounts_by_award.federal_account_id"` part from the comment then in both places.

usaspending_api/disaster/v2/views/elasticsearch_account_base.py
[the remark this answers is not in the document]
kwhickey: I guess there's really no good option here. The alternative would be setting it to the upper limit of 10,000, which would still truncate-silently the results if over 10k. The cardinality query was there to fail-early if over 10k.

Maybe just adding to the comment that if data is known to be greater than this amount, this needs to change.

kwhickey (approved):
Looks good. Only minor issue is consistency in the es mapping field names for total/sum of outlay and obligation.

usaspending_api/disaster/tests/integration/test_disaster_agency_spending.py
tony-sappe: The concern wasn't based on confusion but how it is altering a record. It isn't very robust and leaves the test open to testing behavior the code isn't expecting and/or break for potential valid future changes. Also isn't a pattern worth repeating. 

Not a blocker to this PR

usaspending_api/common/query_with_filters.py
```diff
@@ -477,17 +545,28 @@ def _generate_elasticsearch_query(cls, filters: dict, query_type: _QueryType) ->
...
+                list_pointer.extend(query)
             else:
-                must_queries.append(query)
+                list_pointer.append(query)
+
+        nested_query = ES_Q("nested", path="financial_accounts_by_award", query=ES_Q("bool", must=nested_must_queries))
+        if must_queries and nested_must_queries:
+            must_queries.append(nested_query)
+        elif nested_must_queries:
+            must_queries = nested_query
```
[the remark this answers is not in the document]
tony-sappe: Bump, this still doesn't make sense. Is `list_pointer` redundant?

tony-sappe (changes requested):

usaspending_api/common/query_with_filters.py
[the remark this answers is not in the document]
emily-brents: No, the list pointer switches between appending the query being created in the loop to the `must_queries` array and the `nested_must_queries` array. It's sort of confusing, but it's either that or having the ` if isinstance(query, list)` code duplicated in both branches of the `if "nested_" in filter_type:` conditional.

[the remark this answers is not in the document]
tony-sappe: @emily-brents can you share where the _value_ of `list_pointer` is used? When I search for the variable I only see it being assigned or appended to, not used

image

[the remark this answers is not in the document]
emily-brents: So `list_pointer` isn't ever called, but the arrays it is pointing to are - lines 561-565 
```        
    if must_queries and nested_must_queries:
            must_queries.append(nested_query)
        elif nested_must_queries:
            must_queries = nested_query
        return ES_Q("bool", must=must_queries)
```
`list_pointer` points to either `must_queries` or `nested_must_queries`, which are appended to using `list_pointer` and then used in the code above.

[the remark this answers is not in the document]
tony-sappe: OK, it is finally starting to click for me. Sorry about that.

Wouldn't something like this be cleaner? I think it works the same without requiring the implicit Python shared variable references or the type checking.

```python
            # Generate the query for a filter
            if "nested_" in filter_type:
                query = cls.nested_filter_lookup[filter_type].generate_query(filter_values, query_type, nested_path)
               nested_must_queries.extend(list(query))
            else:
                query = cls.filter_lookup[filter_type].generate_query(filter_values, query_type)
                must_queries.extend(list(query))

        nested_query = ES_Q("nested", path="financial_accounts_by_award", query=ES_Q("bool", must=nested_must_queries))

```

image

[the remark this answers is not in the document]
emily-brents: yeah, I think that should work

tony-sappe:
closing since another PR, #2845, was made and merged into `dev`

PR #2867, opened 2020-12-08
[DEV-6465] Add Missing GTAS Fields

collinwr:
**Description:**
New fields are being added to the `gtas_sf133_balances` table based on aggregations from the broker table: `sf_133`.

**Technical details:**
This ticket involves a new migration to add the fields to the table/model and an update to the management command that populates the table. The command does a full refresh during each run by truncating the existing data and recalculating the complete dataset. The same aggregation logic is being used on the new fields as those already in the script. 

**Note**
The data change ticket is written with the understanding that the migration is backwards compatible with the current code. Adding new fields to the table, should not break the current code: https://federal-spending-transparency.atlassian.net/browse/DEV-6514

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [DEV-6465](https://federal-spending-transparency.atlassian.net/browse/DEV-6465):

```
```


usaspending_api/references/management/commands/load_gtas.py
```diff
@@ -10,16 +10,33 @@
 logger = logging.getLogger("console")
```
tony-sappe, line 10: Can you change the log formatter to "script" for more details in the log statements? (All ETL scripts should use "script" instead of "console" to provide datetimestamps (with time zone) and log level.

```suggestion
logger = logging.getLogger("script")
```

usaspending_api/references/management/commands/load_gtas.py
```diff
@@ -40,7 +57,7 @@ def handle(self, *args, **options):
 
         logger.info("Inserting GTAS total obligations records into website")
         total_obligation_objs = [GTASSF133Balances(**values) for values in total_obligation_values]
-        GTASSF133Balances.objects.bulk_create(total_obligation_objs)
+        GTASSF133Balances.objects.bulk_create(total_obligation_objs, batch_size=100000)
```
tony-sappe, line 60: How did you determine 100K record per batch was optimal? Was it memory utilization or something else which prompted this?

usaspending_api/references/management/commands/load_gtas.py
```diff
@@ -72,16 +89,29 @@ def broker_fetch_sql(self):
...
+                [
+                    f"""COALESCE(SUM(CASE
+                        WHEN line IN ({','.join([str(elem) for elem in val["before_year"]])}) AND fiscal_year < {val["change_year"]} THEN sf.amount * -1
+                        WHEN line IN ({','.join([str(elem) for elem in val["year_and_after"]])}) AND fiscal_year >= {val["change_year"]} THEN sf.amount * -1
+                        ELSE 0
+                    END), 0.0) AS {key},"""
+                    for key, val in DERIVED_COLUMNS_DYNAMIC.items()
+                ]
+            )
         )
```
tony-sappe, line 116: Do you think splitting this "one-liner" into separate statements is cleaner?

```suggestion
        simple_fields = [
            f"COALESCE(SUM(CASE WHEN line IN ({','.join([str(elem) for elem in val])}) THEN sf.amount ELSE 0 END), 0.0) AS {key},"
            for key, val in DERIVED_COLUMNS.items()
        ]
        inverted_fields = [
            f"COALESCE(SUM(CASE WHEN line IN ({','.join([str(elem) for elem in val])}) THEN sf.amount * -1 ELSE 0 END), 0.0) AS {key},"
            for key, val in INVERTED_DERIVED_COLUMNS.items()
        ]
        year_specific_fields = [
            f"""COALESCE(SUM(CASE
                        WHEN line IN ({','.join([str(elem) for elem in val["before_year"]])}) AND fiscal_year < {val["change_year"]} THEN sf.amount * -1
                        WHEN line IN ({','.join([str(elem) for elem in val["year_and_after"]])}) AND fiscal_year >= {val["change_year"]} THEN sf.amount * -1
                        ELSE 0
                    END), 0.0) AS {key},"""
            for key, val in DERIVED_COLUMNS_DYNAMIC.items()
        ]
        return "\n".join(simple_fields + inverted_fields + year_specific_fields)
```

usaspending_api/references/tests/data/broker_gtas.json
```diff
@@ -1,11 +1,18 @@
...
     {
       "fiscal_year": 1600,
       "fiscal_period": "-1",
+      "budget_authority_unobligated_balance_brought_forward_cpe": "-11",
```
tony-sappe, line 6: I realize this PR is a continuation of earlier work. Reading over the tests makes me wonder if it is testing the loader as much as it could. The canned SQL query + response JSON is a much older style of tests. We now have the ability to load data into a test broker DB and run legit queries on the DB. That would allow the aggregations of multiple rows of GTAS source data to be loaded into the USAspending table and then compared to expected results.

Might not be feasible with the time left in the sprint

usaspending_api/references/management/commands/load_gtas.py
```diff
@@ -10,16 +10,33 @@
...
     "total_budgetary_resources_cpe": [1910],
 }
+
 INVERTED_DERIVED_COLUMNS = {
     "gross_outlay_amount_by_tas_cpe": [3020],
 }
 
+# The before_year list of items is applied to records before the change_year fiscal year.
+# The year_and_after list is applied to the change_year and subsequent fiscal years.
+DERIVED_COLUMNS_DYNAMIC = {
```
tony-sappe, line 32: Nice idea

tony-sappe (changes requested):

usaspending_api/references/management/commands/load_gtas.py
collinwr: Yes, can do

usaspending_api/references/management/commands/load_gtas.py
collinwr: Yes, I am good with this change. It does look cleaner

tony-sappe (dismissed):
Approving the code. Improving the tests would be good but are not necessary for this PR

usaspending_api/references/management/commands/load_gtas.py
collinwr: This was originally added to ensure that the command could be run locally with a Postgres instance running in Docker. After upgrading to Docker 3.0.1, I have determined it is no longer necessary. I have removed the `batch_size` argument from the PR

tony-sappe (approved):

PR #2884, opened 2020-12-23
Improve Docker Compose Services and Update Dev Documentation

tony-sappe:
**Description:**
What started out as a minor upkeep of CONTRIBUTING.md turned into more substantial changes including README.md and fixing/improving the docker-compose services.

**Technical details:**

1. New Docker compose services:
    - `usaspending-test` : easy way to run pytest
    - `usaspending-ci` : one-liner to run (most) of the travis-ci checks
2. Fixed creating materialized views and running pytest in docker containers
3. Switched Docker containers' `bind` to `volume` for faster disk performance 
    - These were unfortunately used for the data directories for `usaspending-db` and `usaspending-es`
4. Updated PostgreSQL image from 10.6 to 10.13
5. Added a few missing file cleanups to the matview scripts

    - [x] Backend
    - [x] Operations
    - [x] Domain Expert
8. [x] Jira Ticket (N/A)

```
No changes to API or database
Few minor changes to tests and operation scripts (some only used locally)
Overwhelming majority of the changes were docs and docker-compose
```

### DIFF
--- a/.env
+++ b/.env
@@ -16,9 +16,6 @@ MATVIEW_SQL_DIR=matview_sql
 ES_HOST=http://usaspending-es
 ES_PORT=9200
 
-# ES_CLUSTER_DIR should point to a path outside, or it will be mounted w/Docker
-ES_CLUSTER_DIR=../../docker_es
-
 # Configuration values for a connection string to a Broker database
 #    Only necessary for some management commands
 BROKER_USER=root
--- a/.env
+++ b/.env
@@ -21,9 +21,9 @@ ES_CLUSTER_DIR=../../docker_es
 
 # Configuration values for a connection string to a Broker database
 #    Only necessary for some management commands
-BROKER_USER=usaspending
-BROKER_PASSWORD=usaspender
-BROKER_HOST=usaspending-db
+BROKER_USER=root
+BROKER_PASSWORD=password
+BROKER_HOST=broker-db
 BROKER_PORT=5432
 
 # Use env var to toggle Django's DEBUG mode for local docker envs
--- a/.env
+++ b/.env
@@ -8,9 +8,6 @@ POSTGRES_PASSWORD=usaspender
 POSTGRES_HOST=usaspending-db
 POSTGRES_PORT=5432
 
-# POSTGRES_CLUSTER_DIR should point to a path outside, or it will be mounted w/Docker
-POSTGRES_CLUSTER_DIR=../docker_db
-
 # MATVIEW_SQL_DIR has to be inside of the project (check the docker-compose file)
 MATVIEW_SQL_DIR=matview_sql
 
@@ -20,14 +17,14 @@ ES_HOST=http://usaspending-es
 ES_PORT=9200
 
 # ES_CLUSTER_DIR should point to a path outside, or it will be mounted w/Docker
-ES_CLUSTER_DIR=../docker_es
+ES_CLUSTER_DIR=../../docker_es
 
 # Configuration values for a connection string to a Broker database
 #    Only necessary for some management commands
-BROKER_USER=root
-BROKER_PASSWORD=password
-BROKER_HOST=broker-db
+BROKER_USER=usaspending
+BROKER_PASSWORD=usaspender
+BROKER_HOST=usaspending-db
 BROKER_PORT=5432
 
-# Default to having Django settings.py DEBUG=True for local docker envs
-DJANGO_DEBUG=True
+# Use env var to toggle Django's DEBUG mode for local docker envs
+DJANGO_DEBUG=False
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,16 +1,19 @@
 version: '3.7'
 
+volumes:
+  local_pg_data:
+    driver: local
+
 # See run instructions in accompanying Dockerfile
 services:
 
   usaspending-db:
     container_name: usaspending-db
-    image: postgres:10.6-alpine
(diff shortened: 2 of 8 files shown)


brett-varney:
Windows note: all 3 containers open and don't fail, which is a change.

.env
```diff
@@ -19,15 +16,12 @@ MATVIEW_SQL_DIR=matview_sql
...
 BROKER_USER=root
 BROKER_PASSWORD=password
 BROKER_HOST=broker-db
 BROKER_PORT=5432
 
-# Default to having Django settings.py DEBUG=True for local docker envs
-DJANGO_DEBUG=True
\ No newline at end of file
+# Use env var to toggle Django's DEBUG mode for local docker envs
+DJANGO_DEBUG=False
```
kwhickey, line 27: I still think the initial default for this ought to be `True`. Pretty much every developer starting out on this project has gotten hung up with local dev envs that "don't start"... and often the reason why is something that Django logs would have brought to light. But the logs settings we have operate on whether DEBUG is true and DEBUG is only true if this env var is true.

docker-compose.yaml
```diff
@@ -22,13 +27,53 @@ services:
...
       DATA_BROKER_DATABASE_URL: postgresql://${BROKER_USER}:${BROKER_PASSWORD}@${BROKER_HOST}:${BROKER_PORT}/data_broker
 
+  usaspending-test:
+    build: .
+    volumes:
+     - .:/dockermount
+    depends_on:
+      - usaspending-db
+      - usaspending-es
+    command: pytest
```
kwhickey, line 46: We may need to let this one sleep a few seconds on a totally fresh/clean setup if it truly depends on the DB and ES. The `depends_on` directive doesn't really have readiness checks built-in by default. 

There _was_ a cool way in docker-compose v2 to do this with `healthcheck` and `depends_on` with a healthy condition. See:  https://github.com/peter-evans/docker-compose-healthcheck ... 

But since v3 of docker-compose they are getting away from that (see note in the beginning of the above link). So you're left with either a rudimentary sleep, or one of the `wait-for-it.sh` scripts that are out there.

docker-compose.yaml
```diff
@@ -22,13 +27,53 @@ services:
...
+    environment:
+      DATABASE_URL: postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/data_store_api
+      ES_HOSTNAME: ${ES_HOST}:${ES_PORT}
+      DATA_BROKER_DATABASE_URL: postgresql://${BROKER_USER}:${BROKER_PASSWORD}@${BROKER_HOST}:${BROKER_PORT}/data_broker
+
+  usaspending-ci:
```
kwhickey, line 52: ➕ for giving devs a way to preempt travis problems

docker-compose.yaml
```diff
@@ -22,13 +27,53 @@ services:
...
+    build: .
+    volumes:
+     - .:/dockermount
+    depends_on:
+      - usaspending-db
+      - usaspending-es
```
kwhickey, line 58: Same potential dependency problem ... 
Since this whole ci check is a series of shell commands, you could just add the waits up front.

kwhickey (approved):
Approving with suggestions. 

- still think default of having DEBUG to on is good for first-time-devs. Experienced devs can set an env var to keep debug off
- potential stumbling block of waiting for DB or ES for tests/ci in a fresh dev env

.env
brett-varney: Good point, but I think the default should be what's most common for all devs. It can be set by new devs by env, then removed when it's working, or updates mess things up again.

tony-sappe: I don't recall a time debug logs reveal the issue with a local env. Usually the django error output reveals enough info.

Regardless, if DEBUG mode should be enabled as the default I suggest it is set in `settings.py` instead of here so it is consistent across local envs

docker-compose.yaml
tony-sappe: It seems like an edge case which has a very straightforward and convenient workaround: try again. 

If a decision needs to be made between attempting to preemptively protect a first-time user or making it more efficient for repeated use I will lean toward the repeated use. A number of the changes in this PR are to make the docker-compose managed env _convenient_ for developers (backend and frontend) so there is more buy-in of the tool and it becomes the de facto local env across the team. Adding a small delay doesn't sound like much, but it would be unnecessary for over 95% of the usage and add a strike against a migration to the docker managed env.

I am certainly not arguing to ignore the needs of first-time contributors. The dockerization and compose services should be enough support for first-time contributors especially if the docker-compose script is widely used and maintained moving forward.

.env
kwhickey: I think that it's not so much an issue of the log _level_ needing to be at debug, but rather redirecting log _output_ to the `console` when developing locally (rather than just to a file). 

I think this part of `settings.py` forces it into the console when otherwise these django error output would otherwise not show up in the console. 

https://github.com/fedspendingtransparency/usaspending-api/blob/0fc869f051ea7c1d56ee58a66d5aef7fdd58c92c/usaspending_api/settings.py#L378-L382

kwhickey: i think it's the `server` log that has useful info from Django API which is only sent to `server.log` and not to the console that devs run when invoking `docker-compose`

tony-sappe: A bit anecdotal, but `DEBUG=False` is a better default after my personal and observed experience of developing locally.

collinwr (approved):

PR #2894, opened 2021-01-05
[DEV-6435] API Contract Update

MaxwellKendall:
`/api/v2/reporting/agencies/publish_dates` & `/api/v2/reporting/agencies/overview/` params update:

Renaming ?search to ?filter in query params

- [x] Code Review

```
Very trivial contract update
```


emily-brents (approved):
LGTM, my bad for not updating these contracts, the base class we use for agency endpoints calls the param `filter` not `search`.

tony-sappe (approved):

PR #2911, opened 2021-01-21
[DEV-6526] Adds --deletes-only argument to elasticsearch_indexer.py

collinwr:
**Description:**
https://federal-spending-transparency.atlassian.net/browse/DEV-6526 requires that the `elasticsearch_indexer.py` command has an option to only perform deletes. The command currently has the ability to perform deletes, but upserts are also performed. This PR adds a new parameter `--deletes-only` to support this requirement.

**Technical details:**
Including the `--deletes-only` flag skips the `prepare_for_etl()` and `dispatch_tasks()` steps of the command. 

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [DEV-6526](https://federal-spending-transparency.atlassian.net/browse/DEV-6526):

```
```


sethstoudenmier (dismissed):
Did not test locally, but code seems really straight forward. LGTM.

tony-sappe (changes requested):
Problem observed

The date stored in the database after a successful incremental run is modified when this is run with `--deletes-only`. This will effectively break the subsequent run to load the new and modified records.

A delete-only run should not be categorized as an incremental run

usaspending_api/etl/management/commands/elasticsearch_indexer.py
```diff
@@ -180,7 +186,7 @@ def parse_cli_args(options: dict, es_client) -> dict:
             logger.error(f"Write alias '{config['write_alias']}' is missing")
             raise SystemExit(1)
     else:
-        if es_client.indices.exists(config["index_name"]):
+        if config["index_name"] and es_client.indices.exists(config["index_name"]):
             logger.error(f"Data load into existing index. Change index name or run an incremental load")
             raise SystemExit(1)
```
tony-sappe, line 191: Comment is for code LN 181 - 191. Don't the delete operations require the write alias to be present?

Self-investigative answer: No, because the delete code is very oddly using the `query_alias_prefix` config value. Odd.

Still, I think the delete-only workflow should still ensure the target index or alias exists rather than attempt to delete records from a non-existent index

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
```diff
@@ -99,7 +97,7 @@ def complete_process(self) -> None:
 
         close_all_django_db_conns()
 
-        if self.config["is_incremental_load"]:
+        if self.config["is_incremental_load"] and not self.config["deletes_only"]:
```
tony-sappe, line 100: Isn't `self.config["is_incremental_load"]` False if --deletes-only was provided per LN 178 in usaspending_api/etl/management/commands/elasticsearch_indexer.py?

collinwr: You're right. I believe I added this before I decided to update the `parse_cli_args` method.

sethstoudenmier (approved):

tony-sappe (approved):

PR #2924, opened 2021-01-26
Mod/dev 6087 improve cfda responses

brett-varney:
**Description:**
provide clearer exception responses to client and log than 404 for all problems

**Technical details:**
analyze response from grants.gov API call for various known possible problems, raise appropriate errors to client

2. [x] API documentation updated-N/A API not documented
    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>-N/A
    - [x] Domain Expert <OPTIONAL>-N/A
4. [x] Matview impact assessment completed-N/A no matviews
5. [x] Frontend impact assessment completed-N/A frontend updated in related PR
6. [x] Data validation completed-N/A
7. [x] Appropriate Operations ticket(s) created-N/A
8. [x] Jira Ticket [DEV-6087](https://federal-spending-transparency.atlassian.net/browse/DEV-6087):
    - [x] Performance evaluation of affected (API | Script | Download)-N/A
    - [x] Before / After data comparison-N/A

```
```


usaspending_api/references/v2/views/cfda.py
```diff
@@ -26,15 +29,19 @@ def get(self, request, cfda=None):
...
+            try:
+                response = {
+                    "cfda": result["cfda"],
+                    "posted": result["posted"],
+                    "closed": result["closed"],
+                    "archived": result["archived"],
+                    "forecasted": result["forecasted"],
+                }
+            except KeyError:
+                raise InternalServerError(f"Data from {CFDA_URL} not in expected format: {result}")
```
tony-sappe, line 43: I discourage returning a HTTP 500 error. I definitely would not publicly return the CFDA URL and response in our endpoint response. The main purpose of this ticket was A) better logging and B) robust client code to handle potential problems accessing the external service

usaspending_api/references/v2/views/cfda.py
```diff
@@ -26,15 +29,19 @@ def get(self, request, cfda=None):
...
-                return Response(status=404)
-
-            response = {
-                "cfda": result["cfda"],
-                "posted": result["posted"],
-                "closed": result["closed"],
-                "archived": result["archived"],
-                "forecasted": result["forecasted"],
-            }
+                raise NoDataFoundException(f"No grant records found for '{cfda}'.")
```
tony-sappe, line 32: Nice usage of NoDataFoundException

usaspending_api/references/v2/views/cfda.py
```diff
@@ -57,7 +64,13 @@ def _populate_cfdas_if_needed(self):
...
     def _request_from_grants_api(self):
-        return post(
-            "https://www.grants.gov/grantsws/rest/opportunities/search/cfda/totals",
+        cfda_response = post(
+            CFDA_URL,
             headers={"Authorization": f"APIKEY={settings.GRANTS_API_KEY}"},
-        ).json()["cfdas"]
+        )
+        if cfda_response.status_code == status.HTTP_503_SERVICE_UNAVAILABLE:
+            raise ServiceUnavailable(f"{CFDA_URL} not available (status 503)")
```
tony-sappe, line 72: I wouldn't return a 503 since that would imply the USAspending API is unavailable

usaspending_api/references/v2/views/cfda.py
```diff
@@ -57,7 +64,13 @@ def _populate_cfdas_if_needed(self):
...
 
     def _request_from_grants_api(self):
-        return post(
-            "https://www.grants.gov/grantsws/rest/opportunities/search/cfda/totals",
+        cfda_response = post(
+            CFDA_URL,
             headers={"Authorization": f"APIKEY={settings.GRANTS_API_KEY}"},
-        ).json()["cfdas"]
+        )
+        if cfda_response.status_code == status.HTTP_503_SERVICE_UNAVAILABLE:
```
tony-sappe, line 71: Perhaps check if the status code is not a 200? If not 200 then log the status code and response and return either an empty array or raise NoDataFoundException

usaspending_api/references/v2/views/cfda.py
```diff
@@ -57,7 +64,13 @@ def _populate_cfdas_if_needed(self):
...
+            raise ServiceUnavailable(f"{CFDA_URL} not available (status 503)")
+        if cfda_response.json()["errorMsgs"] != []:
+            raise InternalServerError(f"Error returned by {CFDA_URL}: {cfda_response.json()['errorMsgs']}")
```
tony-sappe, line 74: Again, I discourage returning a 500. Raised Django API exceptions should be specific

usaspending_api/references/v2/views/cfda.py
```diff
@@ -57,7 +64,13 @@ def _populate_cfdas_if_needed(self):
...
+
+        return cfda_response.json()["cfdas"]
```
tony-sappe, line 76: When I helped with monkeypatching troubleshooting this line surfaced a problem with the code if this is an empty array (or other falsy value) which causes the API to be retried until the retries expire.

If it was my ticket I would have re-written most of this view class needs to be more robust, or at least defensively programmed, since we're dealing with an external API and cannot assume a valid response with

tony-sappe (changes requested):
Strongly recommend against using HTTP 500s as a defined pattern and would not include as many specific details in the responses (no need for our API consumers to know this endpoint is an API client).

I would recommend refactoring the logic to be robust but that is open for discussion.

usaspending_api/common/exceptions.py
```diff
@@ -49,6 +49,14 @@ class UnprocessableEntityException(APIException):
     default_code = "invalid_request"
 
 
+class InternalServerError(APIException):
```
tony-sappe, line 52: If these two new HTTP error response classes aren't going to be used they can be removed

usaspending_api/references/v2/views/cfda.py
```diff
@@ -1,32 +1,33 @@
...
 
     @cache_response()
     def get(self, request, cfda=None):
         """
-        Return the view's queryset.
+        Return the CFDA's opportunity totals from grants.gov, or totals for all current CFDAs.
         """
-        self._populate_cfdas_if_needed()
+
+        CFDA_DICTIONARY = self.get_cfdas()
```
tony-sappe, line 25: It appears this code will not longer "cache" the CFDA data as it did before using the global variable which persists as long and the uWSGI worker is running the django app

usaspending_api/references/v2/views/cfda.py
```diff
@@ -35,29 +36,37 @@ def get(self, request, cfda=None):
...
+                raise NoDataFoundException("Grant data is not currently available.")
+            response = response.json()
+            if response.get("errorMsgs") and response["errorMsgs"] != []:
+                logger.error(f"Error returned by grants API: {response['errorMsgs']}")
+                raise NoDataFoundException("Grant data is not currently available.")
+
+            # check response for expected details
+            if response.get("cfdas"):
+                cfdas = response["cfdas"]
+                key_check = {"cfda": None, "posted": None, "closed": None, "archived": None, "forecasted": None}
```
tony-sappe, line 66: If this isn't going to be used as a dictionary, it could be created as a set:
```suggestion
                key_check = {"cfda", "posted", "closed", "archived", "forecasted"}
```

usaspending_api/references/v2/views/cfda.py
```diff
@@ -35,29 +36,37 @@ def get(self, request, cfda=None):
...
+                for cfda in cfdas.values():
+                    if set(key_check) - set(cfda):
+                        logger.error(f"Data from grants API not in expected format: {cfdas}")
+                        raise NoDataFoundException("Grant data is not currently available.")
+
+        return cdfas
```
tony-sappe, line 72: So if the `cfdas` key is missing from the grants API response, the code will fetch the same data from the API 10 times then return an empty dictionary?

usaspending_api/references/v2/views/cfda.py
```diff
@@ -35,29 +36,37 @@ def get(self, request, cfda=None):
...
+                "https://www.grants.gov/grantsws/rest/opportunities/search/cfda/totals",
+                headers={"Authorization": f"APIKEY={settings.GRANTS_API_KEY}"},
+            )
+            if response.status_code != 200:
+                logger.error(f"Status returned by grants API: {response.status_code}")
+                raise NoDataFoundException("Grant data is not currently available.")
+            response = response.json()
+            if response.get("errorMsgs") and response["errorMsgs"] != []:
+                logger.error(f"Error returned by grants API: {response['errorMsgs']}")
+                raise NoDataFoundException("Grant data is not currently available.")
```
tony-sappe, line 61: I'm still not sure it is wise to raise an exception based on `errorMsgs` without even checking if `cfdas` is present and contains acceptable data. Based on the current logic the API response would need to have a HTTP 200 before checking error messages.

usaspending_api/references/v2/views/cfda.py
```diff
@@ -35,29 +36,37 @@ def get(self, request, cfda=None):
...
+        remaining_tries = 10
+
+        while not cdfas and remaining_tries:
+            remaining_tries -= 1
+            response = post(
+                "https://www.grants.gov/grantsws/rest/opportunities/search/cfda/totals",
+                headers={"Authorization": f"APIKEY={settings.GRANTS_API_KEY}"},
+            )
+            if response.status_code != 200:
+                logger.error(f"Status returned by grants API: {response.status_code}")
```
tony-sappe, line 56: If we're terminating the view business logic here, might as well include the entire response body in the log statement for debugging

tony-sappe (changes requested):

usaspending_api/references/v2/views/cfda.py
brett-varney: @tony-sappe yes; what was the purpose of trying more than once?

usaspending_api/references/v2/views/cfda.py
brett-varney: makes sense; done

usaspending_api/references/v2/views/cfda.py
brett-varney: I would assume that if there is an error from them, any data would be suspect. The API doc (https://www.grants.gov/web/grants/s2s/grantor/restful-apis/opportunity-totals-by-cfda.html) doesn't specify much.

usaspending_api/references/v2/views/cfda.py
tony-sappe: Sorry if my comment was confusing. I was pointing out that if the `cfdas` key is missing from the grants.gov response, an exception is not raised by the code. Instead the function will return an empty dictionary. This seems to be problematic

usaspending_api/references/v2/views/cfda.py
```diff
@@ -35,29 +38,37 @@ def get(self, request, cfda=None):
...
+            if response.get("errorMsgs") and response["errorMsgs"] != []:
+                logger.error(f"Error returned by grants API: {response['errorMsgs']}")
+                raise NoDataFoundException("Grant data is not currently available.")
+
+            # check response for expected details
+            if response.get("cfdas"):
+                cfdas = response["cfdas"]
+                key_check = {"cfda", "posted", "closed", "archived", "forecasted"}
+                for cfda in cfdas.values():
+                    if set(key_check) - set(cfda):
```
tony-sappe, line 70: `key_check` is already a set

usaspending_api/references/v2/views/cfda.py
tony-sappe: > I would assume that if

Yeah, we're on the same page of having uncertainty. I am confused how a HTTP 200 response would contain error messages and not CFDA details. I was thinking if the response is a 200 and provides a list of `cfdas` in the expected schema, then the code could go ahead and provide those results even if `errorMsgs` is populated.

usaspending_api/references/v2/views/cfda.py
brett-varney: yep, fixed

usaspending_api/references/v2/views/cfda.py
brett-varney: So should this return whatever data there is, and pass along the error message to the FE page?

usaspending_api/references/v2/views/cfda.py
brett-varney: No, I just wasn't reading it correctly, updated to check

usaspending_api/references/v2/views/cfda.py
tony-sappe: I recommend logging the details but not piping the API response directly through our API.

If you prefer to leave the check for `errorMsgs` here and raise an exception that is OK with me. I see your perspective and without understanding the grants.gov api behavior I do not know which assumption is more correct.

tony-sappe (approved):

PR #2919, opened 2021-01-25
[DEV-6724] Integration tests ES ETL deletes

kwhickey:
**Description:**
Creates automated integration tests to exercise the delete behavior that happens at the beginning of the ES ETLs, where deletes in the DB and logged in the S3 delete log CSV file are synced to the ES indices.

**Technical details:**
This will and _should_ fail, until hotfix PR #2916 is downmerged to the `dev` branch.
- Once downmerged, minor edits need to be made to this branch to catch-up to naming conventions there
- DO NOT MERGE ... until then

- This adds to `TestElasticsearchIndex` additional ES ETL `config` options and the additional step of running `es_configure` after the test index is setup
- The tests patch the function that retrieves and parses the s3 csv delete log file to return a dummy dict of results

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [NA] API documentation updated
3. [x] Necessary PR reviewers:
    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
    - [ ] Domain Expert <OPTIONAL>
4. [NA] Matview impact assessment completed
5. [NA] Frontend impact assessment completed
6. [x] Data validation completed
7. [NA] Appropriate Operations ticket(s) created
8. [x] Jira Ticket [DEV-6724](https://federal-spending-transparency.atlassian.net/browse/DEV-6724):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected (API | Script | Download)
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #2929, opened 2021-01-28
[DEV-6756] Speedup ES ETL delete sync using terms query and keyword fields

kwhickey:
**Description:**
Cut down on chatter between ETL code and ES cluster during synchronization of DB deletes with ES indexes by doing a single `_delete_by_query` for each chunk of docs to be deleted that builds on a `terms` query. 

NOTE: Do not merge without an ES reindex. See linked Data Change subtasks
- Also needs performance- and data-testing in sandbox

Also builds on branch in https://github.com/fedspendingtransparency/usaspending-api/pull/2919

**Technical details:**
- This required that the fields being used with the terms query be converted to `keyword` field
    - this included `generated_unique_transaction_id` on the transaction index
    - it also included `generated_unique_award_id` on the award index. This _had_ a `.keyword` sub-field, but it wasn't in use. So just changed the whole field to keyword type.
    - Also changed the two FABS/FPDS unique key fields to keyword on the transaction index
- Did some function renames and docstrings to clarify usage and purpose of functions
- Found and fixed a danger of leaving un-refreshed deletes if an error happens between delete and refresh
- Found and fixed a bug where awards may never have deleted more than 1000 in one go.
    - may make a new Bug and hotfix for this
    - the awards to be deleted were overwriting in the assignment with `awards = [...]` rather than `awards += [...]`
- Found and fixed a bug which loses the upper-bound ID when the total number of IDs is divisible by the partition count
- Added new unit tests for initial load and incremental load of index

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [DEV-6756](https://federal-spending-transparency.atlassian.net/browse/DEV-6756), as well as two Data Change sub-tasks for reindexing ES: [DEV-6760](https://federal-spending-transparency.atlassian.net/browse/DEV-6760) [DEV-6761](https://federal-spending-transparency.atlassian.net/browse/DEV-6761)

```
```


usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -115,6 +140,17 @@ def delete_docs_by_unique_key(client: Elasticsearch, key: str, value_list: list,
     if not index:
         raise RuntimeError("index name must be provided")
 
+    if use_aliases:
+        index = f"{index}-*"
```
tony-sappe, line 144: I'm not following the usecase of `use-aliases`. Deletes cannot occur on an alias that points to multiple indexes. (https://discuss.elastic.co/t/delete-a-document-from-an-alias/55789/2). Additionally, a delete function that uses a wildcard for the name seems to be asking for trouble.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -142,46 +186,92 @@ def delete_docs_by_unique_key(client: Elasticsearch, key: str, value_list: list,
...
-            ]
 
-        config (dict): collection of key-value pairs that encapsulates runtime arguments for this ES management task
+def _lookup_deleted_award_keys(
+    client: Elasticsearch,
+    lookup_key: str,
+    value_list: list,
+    config: dict,
+    index: Optional[str] = None,
+    lookup_chunk_size=50000,
```
tony-sappe, line 216: Does the lookup chunk size need to be variable? Seems like it adds unnecessary complexity from the conditional checks below for only a potentially minuscule value

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -142,46 +186,92 @@ def delete_docs_by_unique_key(client: Elasticsearch, key: str, value_list: list,
...
+        )
+        logger.error(format_log(msg=msg, action="Delete"))
+        raise RuntimeError(msg)
+
+    award_key_list = []
+    values_generator = chunks(value_list, lookup_chunk_size)
+    for chunk_of_values in values_generator:
+        # Creates an Elasticsearch query criteria
+        q = ES_Q("terms", **{lookup_key: chunk_of_values})
+        response = Search(using=client, index=index).filter(q)[0 : config["max_query_size"]].execute()
```
tony-sappe, line 268: This syntax is crazy, surprised it works.

Is the response limit necessary? If yes, will the records above `max_query_size` be ignored?

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -199,53 +289,73 @@ def delete_awards(client: Elasticsearch, config: dict) -> None:
...
-    if (len(award_ids)) == 0:
+    award_keys = _lookup_deleted_award_keys(
+        client,
+        ES_TRANSACTIONS_UNIQUE_KEY_FIELD,
+        [*deleted_tx_keys],
+        config,
+        settings.ES_TRANSACTIONS_QUERY_ALIAS_PREFIX + "-*",
+    )
+    if (len(award_keys)) == 0:
         logger.info(format_log(f"No related awards require deletion", action="Delete"))
```
tony-sappe, line 304: Either this log statement or LN 309 should be updated so at least one of them is different to help when reading the logs

tony-sappe (commented):
Leaving initial comments while I dig deeper into the changes

usaspending_api/etl/management/commands/elasticsearch_indexer.py
```diff
@@ -149,7 +149,7 @@ def handle(self, *args, **options):
 
 
 def parse_cli_args(options: dict, es_client) -> dict:
-    passthrough_values = (
+    passthrough_values = [
```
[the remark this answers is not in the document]
kwhickey: The type-hint on the function that was taking this in was specified as a `list`, so this is just bringing it in line with the expected type.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -50,6 +58,7 @@ def _delete_from_es(
         ...
     ]
     """
+    warnings.warn("_delete_from_es is deprecated; use delete_docs_by_unique_key", DeprecationWarning)
```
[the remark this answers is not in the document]
kwhickey: I was thinking we may find reason to use _this_ version of delete, based on `match_prefix` at some point in the future. That or need to hastily revert back to using this if a bug is found in production. 

But I suppose that's what version control is for. 

I can remove.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: At some point I thought I saw transactions having a query size (results window) of only 10,000 and awards of 50,000. And if not, that seems a possible case where many of the indices we have may have different max_result_window settings.

This also helps us test boundaries and chunks-processing in unit test with small datasets (e.g. setting `lookup_chunk_size=1`).

The conditionals would should still be there, as checks, to compare max_result_window (query size) to the chunk size. This is just having it compared to a variable with a default, rather than a hard-coded `50000` value.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: yeah, it's a stupid syntax but apparently THE syntax for `elasticsearch-dsl`. You use slicing (which they must have overloaded operators) to tell it what. you want the `"size": num` value to be in the request. And yes, above max_query_size are ignored -- i ran into that problem in testing. Truncates to the default of 10

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
tony-sappe: > The conditionals would should still be there, as checks, to compare max_result_window (query size) to the chunk size. This is just having it compared to a variable with a default, rather than a hard-coded 50000 value.

Or use `max_result_window` as the chunk size if there is a concern the max result window sizes will be different for the various indexes in the future. I'm not sure what benefit it is to test smaller chunk sizes

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
```diff
@@ -142,7 +142,7 @@ def configure_task(self, partition_number: int, name_gen: Generator, is_null_par
         )
 
     def get_id_range_for_partition(self, partition_number: int) -> Tuple[int, int]:
-        range_size = ceil((self.max_id - self.min_id) / self.config["partitions"])
+        range_size = ((self.max_id - self.min_id) // self.config["partitions"]) + 1
```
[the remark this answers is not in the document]
kwhickey: Yes - glad this caught your eye. Likely be a production bug we may consider hotfixing. Any time the range size is equally divisible by number of partitions, it leaves off the max id from the records to process. 

Example that failed in unit tests: 
- 2 awards, ids = `[1, 2]`
- with default (large) partition size, yields `config["partitions"] = 1`
- `range_size = ceil((self.max_id - self.min_id) / self.config["partitions"]) # range_size = 1`
- `lower_bound = min_id + (range_size * partition_number)`
- => `lower_bound = 1 + (1 * 0) = 1`
- `upper_bound = min(min_id + ((range_size * (partition_number + 1) - 1)), max_id)`
- => `upper_bound = min(1 + ((1 * (0 + 1) - 1)), 2)` = 1
- yields partition SQL of `BETWEEN 1 and 1`

Example with larger datasets:
- `partition-size = 2000`
- award IDs between `4...6004`
- assuming no gaps, i.e. `record_count` === (6004-4) = 6000
- `config["partitions"] = ceil(max((self.max_id - self.min_id), self.record_count) / (self.config["partition_size"]))`
- => `config["partitions"] = ceil(max((6004 - 4), 6000) / (2000)) = 3`
- `range_size = ceil((self.max_id - self.min_id) / self.config["partitions"])`
- => `range_size = ceil((6004 - 4) / 3) = 2000`
- ... go to last partition (`partition_number = 2`)...
- `lower_bound = min_id + (range_size * partition_number)`
- => `lower_bound = 4 + (2000 * 2) = 4004`
- `upper_bound = min(min_id + ((range_size * (partition_number + 1) - 1)), max_id)`
- => `upper_bound = min(4 + ((2000 * (2 + 1) - 1)), 6004) = 6003`
- yields partition SQL of `BETWEEN 4 and 6003` --> ⚠️  Misses last ID of `6004`

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
tony-sappe: wow, crazy. Thanks for the explanation

usaspending_api/etl/elasticsearch_loader_helpers/load_data.py
```diff
@@ -72,7 +72,7 @@ def streaming_post_to_es(
     try:
         if delete_before_index:
             value_list = [doc[delete_key] for doc in chunk]
-            delete_docs_by_unique_key(client, delete_key, value_list, job_name, index_name)
+            delete_docs_by_unique_key(client, delete_key, value_list, job_name, index_name, refresh_after=False)
```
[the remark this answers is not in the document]
kwhickey: The new unit test for incremental load would indicate that this works. Meaning, you can delete a doc and re-add the updated version of it without doing a refresh in-between. That's the assumption I was going on, but merits more real-testing against larger datasets and incremental loads.

The thinking / motivation here is that if we are pushing batches of 4000 or 10000 docs to ES at a time, and having to do and covering A LOT of records (say 1M or 2M updates) that's hundreds of refresh calls during the processing, covering thousands of documents each time to refresh, which adds drag to the overall process. So if we can **defer** the refresh of each batch and do only a refresh one all batches are done (or when an error occurs), it would streamline the data ingest. 

That ^ is based on the assumption that you can _add_ a new doc the same as the one just deleted _without_ having to do a refresh in between. I'll test to be sure that's the case.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: I'll have to look into this. During unit testing, the index passed in _was_ the alias here, which caused complications until this was added -- following the pattern that was in the other delete method `_delete_from_es`:

https://github.com/fedspendingtransparency/usaspending-api/blob/45429fa55fc2984b58ba06f2ea8bd8f915094aa4/usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py#L58-L59

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: The testing of smaller chunks is what revealed the hotfix bug that was fixed in https://github.com/fedspendingtransparency/usaspending-api/pull/2931

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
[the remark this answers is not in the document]
kwhickey: I'm open to a cleaner re-write of the boundary detection ... but this seemed to work. The issue is that in cases where the division by partition number is fractional, the `ceil(...)` effectively added one to the whole-part. So instead, I trim off the fractional part (`//` is basically `floor(...)`) and _always_ add 1.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: I concur that'll be a helpful distinction. Changing

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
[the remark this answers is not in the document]
tony-sappe: The number of records is 6001 if the min ID was 4 and the max id was 6004. Still ends up with the same final ID though. You're right.

image

=> config["partitions"] = ceil(max((6004 - 4), 6001) / (2000)) = 4
=> range_size = ceil((6004 - 4) / 4) = 1500
=> lower_bound = 4 + (1500 * 3) = 4504
=> upper_bound = min(4 + ((1500 * (4 + 1) - 1)), 6004) = 6003


Using the new logic
=> config["partitions"] = ceil(max((6004 - 4), 6001) / (2000)) = 4
=> range_size = ((6004 - 4) // 4) + 1 = 1501

partition 0
=> lower_bound = 4 + (1501 * 0) = 4
=> upper_bound = min(4 + ((1501 * (0 + 1) - 1)), 6004) = 1,504

...

partition 3
=> lower_bound = 4 + (1501 * 3) = 4507
=> upper_bound = min(4 + ((1501 * (4 + 1) - 1)), 6004) = 6004 ( 6007 capped to max_id)



I don't think it is hotfix-worthy. The scenario is very specific.

usaspending_api/etl/elasticsearch_loader_helpers/load_data.py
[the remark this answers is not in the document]
tony-sappe: Was your test cluster multi-node or a single node?

I definitely understand the performance hit and the annoyance of calling refresh. Depends if a delete is eventually consistent or consistent. It might be a price for horizontal scaling technology.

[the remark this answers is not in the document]
kwhickey: Seems to have done okay with processing about 2M transaction updates from the `transaction_search` table to the transaction ETL (deletes before re-insert on transaction index) `elasticsearch_indexer_etl` job number `48`. Looks like the refreshes were held off, but ES still seems to trigger some kind of refresh as needed at intervals of 500k docs or so based on the stair-step of deleted docs during the run. 
image

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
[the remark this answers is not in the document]
kwhickey: I still need to look at my code changes that "solves" this. With a partition set to 4000, it is annoying processing 4001 docs at a time. Need to get it to use the partition size given.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: Figured out how to force it to use the write alias (like for incremental load) and therefore needs no wildcard. Removed this.

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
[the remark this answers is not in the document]
tony-sappe: The "desired range size +1" was included in the above comment. The current code was tested in many iterations and worked very well except for the scenario you found with evenly divisible ID ranges

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -19,89 +19,46 @@
...
+    refresh_after: bool = True,
+    delete_chunk_size: int = 10000,
+) -> int:
     """
     Bulk delete a batch of documents whose field identified by ``key`` matches any value provided in the
     ``values_list``.
 
+    NOTE: This delete routine looks at just the index name given. If there are duplicate records across
+    multiple indexes, an alias or wildcard should be provided for ``index`` param that covers multiple
+    indices, or this will need to be run once per index.
```
tony-sappe, line 42: Do deletes work for an alias with multiple indices?

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -19,89 +19,46 @@
...
-
-            NOTE: This delete routine looks at just the index name given. If there are duplicate records across
-            multiple indexes, an alias or wildcard should be provided for ``index`` param that covers multiple
-            indices, or this will need to be run once per index.
+        refresh_after (bool): Whether to call ``_refresh`` on the index when all of the provided values in
+            ``value_list`` have been processed for delete; defaults to ``True``. If many small deletes happen at a
+            rapid rate, it may be best to set this ``False`` and await a deferred refresh afterward in the calling
+            code. However it is important to not retry deletes that have already been deleted until _refresh is
+            called, or a elasticsearch.exceptions.ConflictError will be raised due to an ES doc version_conflict
+            error response. NOTE: This param will be ignored and a refresh will be attempted if this function
```
tony-sappe, line 56: > However it is important to not retry deletes that have already been deleted until _refresh is called, or a elasticsearch.exceptions.ConflictError will be raised due to an ES doc version_conflict error response

Is this still true with `conflicts="proceed"` used below?

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -115,73 +72,171 @@ def delete_docs_by_unique_key(client: Elasticsearch, key: str, value_list: list,
...
+        values_generator = chunks(value_list, delete_chunk_size)
         for chunk_of_values in values_generator:
-            # Creates an Elasticsearch query criteria for the _delete_by_query call
-            q = ES_Q("terms", **{key: chunk_of_values})
             # Invoking _delete_by_query as per the elasticsearch-dsl docs:
             #   https://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html#delete-by-query
-            response = Search(using=client, index=index).filter(q).delete()
-            chunk_deletes = response["deleted"]
-            deleted += chunk_deletes
+            # _refresh is deferred til the end of chunk processing
```
tony-sappe, line 100: Recommend replacing with `until` or `'til`

```suggestion
            # _refresh is deferred until the end of chunk processing
```

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -115,73 +72,171 @@ def delete_docs_by_unique_key(client: Elasticsearch, key: str, value_list: list,
...
+            q = Search(using=client, index=index).filter("terms", **{key: chunk_of_values})  # type: Search
```
tony-sappe, line 101: I'm not quite understanding the mixture of `# type` and `# params:` together, is there a good resource to understand the syntax?

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -115,73 +72,171 @@ def delete_docs_by_unique_key(client: Elasticsearch, key: str, value_list: list,
...
+    client: Elasticsearch,
+    lookup_key: str,
+    value_list: list,
+    config: dict,
+    index: Optional[str] = None,
+    lookup_chunk_size: int = 50000,
+) -> list:
+    """Derive a list of award keys given a target index, Lookup field, and lookup values
+
+    This returns a list of all unique award keys, which  arecompiled from the ``ES_AWARDS_UNIQUE_KEY_FIELD`` field of
```
tony-sappe, line 182:
```suggestion
    This returns a list of all unique award keys, which are compiled from the ``ES_AWARDS_UNIQUE_KEY_FIELD`` field of
```

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -115,73 +72,171 @@ def delete_docs_by_unique_key(client: Elasticsearch, key: str, value_list: list,
...
+            f"{lookup_chunk_size} is greater {config['max_query_size']}, which is the max number of query "
+            f"results returnable from this index. Use a smaller chunk or increase max_result_window for this index."
+        )
+        logger.error(format_log(msg=msg, action="Delete"))
+        raise RuntimeError(msg)
+
+    award_key_list = []
+    values_generator = chunks(value_list, lookup_chunk_size)
+    for chunk_of_values in values_generator:
+        q = Search(using=client, index=index).filter("terms", **{lookup_key: chunk_of_values})  # type: Search
```
tony-sappe, line 231: is `# type: Search` helpful information here?

tony-sappe (approved):
PR looks good and adds a ton of useful information to logging (not to mention the bugs fixed)

Leaving a few PR comments which aren't blockers but could be addressed

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: It's a form of type hint convention introduced by [PEP 484](https://www.python.org/dev/peps/pep-0484/#type-comments). Supporting IDEs like PyCharm will then know what type is returned by the last function in a chained call like that, and then provide api-autocomplete as well as static code analysis thereafter.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: It's a form of type hint convention introduced by [PEP 484](https://www.python.org/dev/peps/pep-0484/#type-comments). Supporting IDEs like PyCharm will then know what type is returned by the last function in a chained call like that, and then provide api-autocomplete as well as static code analysis thereafter.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: A couple searches would indicate that it can be done. Also, this was previous comment that already existed for the prior delete functionality. 

- https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html mentions use of an alias as the target
- old post when delete_by_query was a plugin shows support: https://discuss.elastic.co/t/delete-by-query-on-alias-or-multiple-indexes/43102/2
- alias docs: 
    - > Aliases with filters provide an easy way to create different "views" of the same index. The filter can be defined using Query DSL and is applied to all Search, Count, Delete By Query and More Like This operations with this alias.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
tony-sappe: I couldn't find any references to `# params` in PEP 484 when I looked the other day

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
tony-sappe: cool, thanks. I knew indexing documents will not work with multi-index aliases and now I know deletes are fine

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
```diff
@@ -189,7 +190,7 @@ def extract_transform_load(task: TaskSpec) -> None:
             msg = f"Partition #{task.partition_number} failed after an error was previously encountered"
             logger.warning(format_log(msg, name=task.name))
         else:
-            logger.error(format_log(f"{task.name} failed!", name=task.name))
+            logger.exception(format_log(f"{task.name} failed!", name=task.name))
```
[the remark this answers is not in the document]
kwhickey: Here's the full snippet
```python
    try:
        records = task.transform_func(task, extract_records(task))
        if abort.is_set():
            f"Prematurely ending partition #{task.partition_number} due to error in another process"
            logger.warning(format_log(msg, name=task.name))
            return
        if len(records) > 0:
            success, fail = load_data(task, records, client)
        else:
            logger.info(format_log("No records to index", name=task.name))
            success, fail = 0, 0
        with total_doc_success.get_lock():
            total_doc_success.value += success
        with total_doc_fail.get_lock():
            total_doc_fail.value += fail
    except Exception:
        if abort.is_set():
            msg = f"Partition #{task.partition_number} failed after an error was previously encountered"
            logger.warning(format_log(msg, name=task.name))
        else:
            logger.exception(format_log(f"{task.name} failed!", name=task.name))
            abort.set()
```

I think at most it will record the exception twice. Since once it logs the exception, it sets `abort.set()` and the caught exception will just log a warning if it enters there again for a different process, by finding that `if abort.is_set():` is `True`. 

So that has one of two outcomes: 
1. The worker process already caught and logged this exception, and then raised it up to the controller process, where it will be logged a second time. 
1. It was an _unhandled_ exception (e.g. happening somewhere that was not expected, and therefore had not `try-except` block wrapping it) and so it bubbles up to the controller, where it is logged for the first and only time. 

This was added more for the latter scenario to make sure that we log exceptions that were not expected, rather than just logging that an error occurred, but not seeing the traceback.

To do further cleanup what we _could_ do is always just log an ERROR in subprocesses, and ensure that this controller `try-except` logs the actual exception. There are cases where I saw `SystemExit` though so i'm not sure it will always make it up to the top controller process. For now might just be best to err on the side of caution and easiest to ride with the 2x redundancy and revisit if the logs get really painful to read.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: Nope, removed that part of the comment.

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
```diff
@@ -105,11 +105,11 @@ def complete_process(self) -> None:
             update_last_load_date(f"{self.config['stored_date_key']}", self.config["processing_start_datetime"])
 
     def determine_partitions(self) -> int:
-        """Create partition size less than or equal to max_size for more even distribution"""
-        if self.config["partition_size"] > (self.max_id - self.min_id):
+        """Simple strategy of partitions that cover the id-range in in an even distribution"""
```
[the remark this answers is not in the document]
kwhickey: Fixed

usaspending_api/etl/elasticsearch_loader_helpers/controller.py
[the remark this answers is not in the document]
tony-sappe: I think it is possible to have dozens of processes simultaneously fail due to a cascading issue. And as far as I can tell by observed behavior and the docs, it is possible for a number of processes to check `abort.is_set()` before the first process can set it using `abort.set()`. I think the logging statements can slow down the execution enough and I don't think the call to set actually obtains a lock.

Feel free to read the notes yourself: https://docs.python.org/3.7/library/threading.html#threading.Event

I assume that most of these unfortunate scenarios will occur locally, so I won't drag the PR over this.

tony-sappe (approved):
re-adding approval

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
kwhickey: Oh. That is just like a "heading" to the 2 comment lines that follow. describing the `q.params(...)` line:

i.e. "the two params used mean: ..."

PR #2963, opened 2021-02-12
[FIX] Augment tempfile.SpooledTemporaryFile() for expected behavior

tony-sappe:
**Description:**
This PR creates a new class to give tempfile.SpooledTemporaryFile an uplift. The current implementation in the standard library isn't satisfying the expected API. Augmenting the class methods and importing io.IOBase should get the class close to where it needs to be. Until https://github.com/python/cpython/pull/3249 is merge, released, and used by this project we will need to keep this augmented class around.

**Technical details:**
While this issue has existed since Python 3.0, it was only recently felt in this project with the upgrade from pandas v0.45.x to pandas v1.0.2+ (pandas v1.0.0 and v1.0.1 introduced another bug which they fixed in v1.0.2, relying on API missing in the SpooledTemporaryFile implementation). See https://bugs.python.org/issue26175 and https://github.com/python/cpython/pull/3249

Locally `loadcfda`, `load_rosetta`, `load_city_county_state_code`, and downloads were tested to ensure the changes did not break other areas. FABS and FPDS ETL scripts also use the class for obtaining delete records and should be tested. Several other ETL scripts (load_tas, load_dabs_submission_window_schedule)  contain a path for loading a file which isn't used by the nightly pipeline which connects directly to broker for refreshing data.

    - [x] Backend
    - [ ] Operations
8. [x] Jira Ticket (N/A)

```
PR is only a change to class used for retrieving temporary files
```

### DIFF
--- a/usaspending_api/common/retrieve_file_from_uri.py
+++ b/usaspending_api/common/retrieve_file_from_uri.py
@@ -1,4 +1,5 @@
 import boto3
+import io
 import requests
 import tempfile
 import urllib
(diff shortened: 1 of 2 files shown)


sethstoudenmier (approved):
Stepped through comments on PR related to this change in cpython and everything seems to be in line with what is planned for that. LGTM.

PR #2976, opened 2021-02-18
Staging -> QAT

tony-sappe:
Downmerge of DEFC U and fix for the file retrieval helper class (necessary after the pandas version update)


collinwr (approved):
Approved this down merge

PR #2956, opened 2021-02-11
Mod/dev 6552 add unlinked count to agencies overview

brett-varney:
**Description:**
add unlinked award counts to api/v2/reporting/agencies/overview/ , replacing the "0" placeholders with actual data

**Technical details:**
add the C & D missing values now being stored in ReportingAgencyOverview

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [x] Frontend na-no changes
    - [x] Operations na
    - [x] Domain Expert na
4. [x] Matview impact assessment completed na
5. [x] Frontend impact assessment completed na
6. [x] Data validation completed
7. [x] Appropriate Operations ticket(s) created na
8. [x] Jira Ticket [DEV-6552](https://federal-spending-transparency.atlassian.net/browse/DEV-6552):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected (API | Script | Download) na, simply adding table values to results
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #2992, opened 2021-02-24
Mod/dev 6552 add unlinked count to agencies overview

brett-varney:
**Description:**
add unlinked award counts to api/v2/reporting/agencies/overview/ , replacing the "0" placeholders with actual data
NOTE: this PR simply extends https://github.com/fedspendingtransparency/usaspending-api/pull/2956 by sorting null values to the end in all cases

**Technical details:**
add the C & D missing values now being stored in ReportingAgencyOverview

    - [ ] Backend
    - [x] Frontend na-no changes
    - [x] Operations na
    - [x] Domain Expert na
4. [x] Matview impact assessment completed na
5. [x] Frontend impact assessment completed na
7. [x] Appropriate Operations ticket(s) created na
8. [x] Jira Ticket [DEV-6552](https://federal-spending-transparency.atlassian.net/browse/DEV-6552):
    - [x] Performance evaluation of affected (API | Script | Download) na, simply adding table values to results

```
```

### DIFF
--- a/usaspending_api/reporting/v2/views/agencies/overview.py
+++ b/usaspending_api/reporting/v2/views/agencies/overview.py
@@ -161,7 +161,7 @@ class AgenciesOverview(AgencyBase, PaginationMixin):
         if self.pagination.sort_order == "desc":
             result_list = result_list.order_by(F(self.pagination.sort_key).desc(nulls_last=True))
         else:
-            result_list = result_list.order_by(F(self.pagination.sort_key).asc(nulls_first=True))
+            result_list = result_list.order_by(F(self.pagination.sort_key).asc(nulls_last=True))
 
         return self.format_results(result_list)
 
--- a/usaspending_api/reporting/v2/views/agencies/overview.py
+++ b/usaspending_api/reporting/v2/views/agencies/overview.py
@@ -156,8 +156,13 @@ class AgenciesOverview(AgencyBase, PaginationMixin):
                 "unlinked_contract_award_count",
                 "unlinked_assistance_award_count",
             )
-            .order_by(f"{'-' if self.pagination.sort_order == 'desc' else ''}{self.pagination.sort_key}")
         )
+
+        if self.pagination.sort_order == "desc":
+            result_list = result_list.order_by(F(self.pagination.sort_key).desc(nulls_last=True))
+        else:
+            result_list = result_list.order_by(F(self.pagination.sort_key).asc(nulls_first=True))
+
         return self.format_results(result_list)
 
     def format_results(self, result_list):
--- a/usaspending_api/reporting/tests/integration/test_agencies_overview.py
+++ b/usaspending_api/reporting/tests/integration/test_agencies_overview.py
@@ -446,25 +446,6 @@ def test_pagination(setup_test_data, client):
     assert len(response["results"]) == 3
     expected_results = [
         {
-            "agency_name": "Test Agency",
-            "abbreviation": "ABC",
-            "toptier_code": "123",
-            "agency_id": 1,
-            "current_total_budget_authority_amount": None,
-            "recent_publication_date": None,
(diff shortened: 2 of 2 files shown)


usaspending_api/reporting/v2/views/agencies/overview.py
```diff
@@ -156,8 +156,13 @@ def get_agency_overview(self):
                 "unlinked_contract_award_count",
                 "unlinked_assistance_award_count",
             )
-            .order_by(f"{'-' if self.pagination.sort_order == 'desc' else ''}{self.pagination.sort_key}")
         )
+
+        if self.pagination.sort_order == "desc":
+            result_list = result_list.order_by(F(self.pagination.sort_key).desc(nulls_last=True))
```
sethstoudenmier, line 162: Would recommend COALESCE the `null` values to 0 instead of simply having them be first or last. In column where values can be negative it doesn't make sense. Screenshot of my local on your branch below:

Screen Shot 2021-02-25 at 7 23 35 AM

sethstoudenmier: Would also recommend adding a secondary sort of `toptier_code` (or just some unique column that is non-null) to make sure we get a consistent sort.

sethstoudenmier (changes requested):

usaspending_api/reporting/v2/views/agencies/overview.py
brett-varney: > 
> 
> Would recommend COALESCE the `null` values to 0 instead of simply having them be first or last. In column where values can be negative it doesn't make sense. Screenshot of my local on your branch below:
> 
Makes sense. I tried for over a day to get a COALESCE in here, failed to, and did what Tony suggested. But we hadn't thought of negatives.

sethstoudenmier: Technically speaking your columns in this ticket won't be displaying negatives, but because we hand't thought about it before now your changes will affect those columns.

brett-varney: @sethstoudenmier My confusion: this is sorting amounts; my update only had to do with counts. What's the correlation here? Or is this a different update we need to fix?

sethstoudenmier: > By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise.
(https://www.postgresql.org/docs/10/queries-order.html)

Before your changes the `NULL` values for the spending columns were sorting according to the description above. Now they are sorting out of order in a different way. We can bring up the option to create a different ticket to correct sort for those, but it is going to require undoing your changes to write up a new way to sort that correctly handles nulls for all columns.

tony-sappe: I think simply placing nulls last for both directions will fix the issue you reported @sethstoudenmier

brett-varney: @sethstoudenmier I get it, now I'm resorting nulls in all values, not just the ones I'm fixing.

tony-sappe: @brett-varney your code changes are affecting to sort behavior for every field

> @sethstoudenmier My confusion: this is sorting amounts; my update only had to do with counts. What's the correlation here? Or is this a different update we need to fix?

@brett-varney your code changes affect sorting across all of the fields

sethstoudenmier (approved):
Originally had concerns that the order of the NULL values at the end might change, but after testing I could not get the order to change. Conversation occurred in Slack to confirm that NULL values will be placed at the end of the results and `0` will be replaced with `--` for the frontend when `NULL`.

PR #3029, opened 2021-03-12
[DEV-7012] Change how Awards with negated outlay are filtered out

sethstoudenmier:
**Description:**
Make sure that Awards with negated outlay (e.g. `100` and `-100`) are filtered out taking into account the addition of the new downward adjustment columns that are part of the calculation.

**Technical details:**
After going back and forth multiple times on how to implement I settled on this implementation for performance reasons. A search of Staging showed that the cases where the outlay is zero the negated File C records share the same DEFC (which make sense). I wanted to keep the functionality where a user is able to filter by a single DEFC on the endpoint. The shared DEFC between negated File C means that a user should not be able to filter by a single DEFC and be robbed of data they would otherwise be able to see if the implementation was granular to DEFC instead of the entire Award. 

The bulk of the work is not shown in code, but was in performance testing and trying to test different implementations. If we wanted to filter out by the sum of a single award and group by DEFC then that would require adding another `sum` and `bucket_selector` aggregation at the top level to filter on.

Two two tests are only on Object Class spending since the change is in a central part of the base class where all implementations (such as Object Class spending) will hit. They test to make sure that negated Obligation and Outlay (taking into account the downward adjustment columns) will filter out an Award and that a nonzero Obligation with a negated Outlay will be counted.

    - [ ] Backend
8. [x] Jira Ticket [DEV-7012](https://federal-spending-transparency.atlassian.net/browse/DEV-7012):

```
```

### DIFF
--- a/usaspending_api/common/query_with_filters.py
+++ b/usaspending_api/common/query_with_filters.py
@@ -469,23 +469,6 @@ class _NonzeroFields(_Filter):
         return ES_Q("bool", should=non_zero_queries, minimum_should_match=1)
 
 
-class _NonzeroNestedSumFields(_Filter):
-    """Query for when the sum of nested fields should not be zero"""
-
-    underscore_name = "nonzero_sum_fields"
-
-    @classmethod
-    def generate_elasticsearch_query(
-        cls, filter_values: List[str], query_type: _QueryType, nested_path: str = ""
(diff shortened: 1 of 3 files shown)


tony-sappe (approved):
makes sense

PR #3031, opened 2021-03-12
[DEV-6957] Adding new SQL view for only toptier agencies with DABS

tony-sappe:
**Description:**
Creating a SQL view which can be used for many different endpoints and ETL scripts: `vw_published_dabs_toptier_agency`

**Technical details:**
This appears to be a more correct implementation of the `Toptier.objects.account_agencies()` function and will be replacing it in a following PR

**Requirements for PR merge:**

1. [x] Unit & integration tests updated (N/A)
2. [x] API documentation updated (N/A)
3. [x] Necessary PR reviewers:
    - [x] Backend
4. [x] Matview impact assessment completed
5. [x] Frontend impact assessment completed
6. [x] Data validation completed
7. [x] Appropriate Operations ticket(s) created
8. [x] Jira Ticket [DEV-6957](https://federal-spending-transparency.atlassian.net/browse/DEV-6957):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected "Script": <1s
    - [x] Before / After data comparison: 107 agencies are in the new SQL view

**Area for explaining above N/A when needed:**
```
No impact to the API, materialized views, or tests
```

PR #3032, opened 2021-03-12
[DEV-6957] Leverage new SQL View to return the subset of Agencies with published DABS

tony-sappe:
**Description:**
Most of the changes were to mock data in tests so the new SQL view will contain the expected data

Minor Items:

1. Added a secondary sort to v2/reporting/agencies/overview of the `toptier_code` since it is possible to agencies to have $0 for the submission period
2. `ToptierAgency.objects.account_agencies()` is gone, all instances were replaced with the new view. While some performance tests were performed on Sandbox, it was not comprehensive across all endpoints which were modified (specifically the v2/agency endpoints) which are slated to be used for future website pages and might need to be modified anyway
3. Added defensive programming against divide-by-0/null errors in v2/reporting/agencies/<toptier_code>/overview
4. Cleaned up `populate_reporting_agency_overview` a bit. It might look odd in this PR but I expect to introduce a following PR to modify that ETL, so I left the cleanup changes here in case the follow-up PR doesn't materialize.

**Technical details:**
Building on #3031 

    - [ ] Backend
8. [x] Jira Ticket [DEV-6957](https://federal-spending-transparency.atlassian.net/browse/DEV-6957):
    - [x] Performance evaluation of affected API (all endpoint SQL completed in under 2s)

```
No changes to API definitions or materialized views
```


usaspending_api/reporting/v2/views/agencies/toptier_code/overview.py
```diff
@@ -130,7 +130,7 @@ def format_results(self, result_list):
                 "current_total_budget_authority_amount": result["total_budgetary_resources"],
                 "total_budgetary_resources": result["gtas_total_budgetary_resources"],
                 "percent_of_total_budgetary_resources": round(
-                    result["total_budgetary_resources"] * 100 / result["gtas_total_budgetary_resources"], 2
+                    result["total_budgetary_resources"] * 100 / (result["gtas_total_budgetary_resources"] or 100), 2
```
sethstoudenmier, line 133: If this is suppose to be a percent why are you having the default be to divide by 100?

sethstoudenmier (commented):
One minor question. Aside from that changes look good and seem reasonable.

usaspending_api/reporting/v2/views/agencies/toptier_code/overview.py
tony-sappe: Good question. I put in it to fix divide-by-zero errors and forgot to put in a permanent fix. I think it works because when `gtas_total_budgetary_resources` is null `total_budgetary_resources` is 0.

tony-sappe: pushed change

sethstoudenmier (approved):

sethstoudenmier:
Don't think it was cause any problems, but not merging right now as I would prefer to get the Staging -> QAT down merge in first.

PR #3038, opened 2021-03-15
[DEV-6496] DUNS Endpoint Optimization

collinwr:
**Description:**
This PR improves the performance of the `/api/v2/recipient/duns/` endpoint by working on the following three areas:
- **Calculates the `count` field of the response using a separate, cached endpoint.** This prevents the count from needing to be queried from the db for each request when the `keyword` and `award_type` filters have already been included in a request.
- **Adds indexes to accomodate sorting.** An index was added for each `award_type` field. A second index was added for `recipient_unique_id` and `recipient_name` because their sorts put nulls last, and can't share an index for `ASC` and `DESC` queries
- **Increases the limit of results from 100 to 1000.** 

**Technical details:**
Some things to look out for:
- The `request_count()` method in `list_recipients.py`
- The manually edited migration: `0006_auto_20210315_2028.py`. Two index were added directly to the migration rather than to the model because Django doesn't support `NULLS LAST` in its indexes.

    - [ ] Backend
8. [x] Jira Ticket [DEV-6496](https://federal-spending-transparency.atlassian.net/browse/DEV-6496):

```
```


usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -70,13 +71,59 @@ def get_recipients(filters={}):
...
+    keyword = "-"
+    if "keyword" in validated_payload:
+        keyword = validated_payload["keyword"]
+
+    return ".".join(
+        [
+            award_type,
+            keyword,
+        ]
+    )
```
tony-sappe, line 90: If I understand this correctly, this is creating a separate cache key which is separate from the key generated when the API endpoint is directly requested?

Is this is acceptable it might be helpful to comment on why this is necessary

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -70,13 +71,59 @@ def get_recipients(filters={}):
...
+
+
+class DunsCount(APIView):
+    """
+    This route takes a single keyword filter and agency_type, and returns a count of matching recipients
```
tony-sappe, line 95: agency type?

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -70,13 +71,59 @@ def get_recipients(filters={}):
...
 class ListRecipients(APIView):
     """
     This route takes a single keyword filter (and pagination filters), and returns a list of recipients
     """
 
     endpoint_doc = "usaspending_api/api_contracts/contracts/v2/recipient/duns.md"
 
+    def request_count(self, filters={}):
+        response = DunsCount.as_view()(request=self.request._request).data
+        return response["count"]
```
tony-sappe, line 125: Would it be better design to handle the exceptions in this method instead of requiring a try-except wrapper below?

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -86,10 +133,22 @@ def post(self, request):
...
+
         new_sort = {"type": "enum", "enum_values": ["name", "duns", "amount"], "default": "amount"}
         models = update_model_in_list(models, "sort", new_sort)
         models = update_model_in_list(models, "limit", {"default": 50})
         validated_payload = TinyShield(models).block(request.data)
 
-        results, page_metadata = get_recipients(filters=validated_payload)
+        try:
+            count = self.request_count(validated_payload)
+        except Exception as e:
```
tony-sappe, line 148: How likely is this to fail? What would be the reasons for it to throw an exception?

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -86,10 +133,22 @@ def post(self, request):
         ]
         models.extend(copy.deepcopy(PAGINATION))  # page, limit, sort, order
 
+        # Override pagination default limit of 100
+        for model in models:
+            if model["name"] == "limit":
+                model["max"] = 1000
```
tony-sappe, line 139: Shouldn't the API contract be updated to include this?

tony-sappe: Also was there any experimentation to see if a page size higher than 1000 would be beneficial?

usaspending_api/recipient/models.py
```diff
@@ -96,14 +96,38 @@ class Meta:
         managed = True
         db_table = "recipient_profile"
         unique_together = ("recipient_hash", "recipient_level")
-        # Note:  A custom index was added in the migration because there's
+        # Note:  Two custom indexess were added in migrations because there's
```
tony-sappe, line 99:
```suggestion
        # Note:  Two custom indexes were added in migrations because there's
```

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -70,13 +71,59 @@ def get_recipients(filters={}):
...
+
+    return ".".join(
+        [
+            award_type,
+            keyword,
+        ]
+    )
+
+
+class DunsCount(APIView):
```
tony-sappe, line 93: The think `RecipientCount` is more agreeable. I assume there will be many recipients without an associated DUNS returned in the count

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -86,10 +133,22 @@ def post(self, request):
...
         models = update_model_in_list(models, "sort", new_sort)
         models = update_model_in_list(models, "limit", {"default": 50})
         validated_payload = TinyShield(models).block(request.data)
 
-        results, page_metadata = get_recipients(filters=validated_payload)
+        try:
+            count = self.request_count(validated_payload)
+        except Exception as e:
+            count = None
+            logger.error("Unable to retreive count using endpoint. Falling back on using queryset.")
```
tony-sappe, line 150: Feel free to use something else. I was just thinking:
1. state _which_ endpoint
2. Since both methods are using a queryset, it is misleading to say a queryset will be used if there are problems

```suggestion
            logger.error("Unable to retrieve count using `/api/v2/recipient/duns/count`. Falling back on default behavior")
```

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -1,59 +1,60 @@
...
+
     if filters["order"] == "desc":
-        queryset = queryset.order_by(F(api_to_db_mapper[filters["sort"]]).desc(nulls_last=True))
+        queryset = queryset.order_by(F(api_to_db_mapper[filters["sort"]]).desc(nulls_last=nulls_last))
     else:
-        queryset = queryset.order_by(F(api_to_db_mapper[filters["sort"]]).asc(nulls_last=True))
+        queryset = queryset.order_by(F(api_to_db_mapper[filters["sort"]]).asc(nulls_last=nulls_last))
+
+    if count is None:
+        count = queryset.count()
```
tony-sappe, line 56: It might be more trouble than it is worth. But technically the backup code to obtain the count could be obtained outside of this method by running `DunsCount.get_count()`. By sharing the code there is less chance of discrepancies appearing in the future between the count from `/api/v2/recipient/duns/` and `/api/v2/recipient/duns/count/`

usaspending_api/api_contracts/contracts/v2/recipient/duns/count.md
```diff
@@ -0,0 +1,49 @@
+FORMAT: 1A
+HOST: https://api.usaspending.gov
+
+# Recipient Duns [/api/v2/recipient/duns/count]
```
tony-sappe, line 4:
```suggestion
# Recipient Duns [/api/v2/recipient/duns/count/]
```

tony-sappe (changes requested):
Leaving a number of comments. Most are minor. I do have some questions on the new indexes (I assume for sort order) and if all logic branches were accounted for)

usaspending_api/api_contracts/contracts/v2/recipient/duns/count.md
```diff
@@ -0,0 +1,49 @@
...
+
+These endpoints are used to power USAspending.gov's recipient profile pages. This endpoint is called by the [/api/v2/recipient/duns/count] endpoint, so a cache can be used to prevent recalculating the count for each request. 
+
+## POST
```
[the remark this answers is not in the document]
collinwr: I originally made this a GET request, but had to change it, so I could reuse the original request from the other endpoint.

usaspending_api/recipient/migrations/0006_auto_20210315_2028.py
```diff
@@ -0,0 +1,51 @@
...
+            model_name='recipientprofile',
+            index=models.Index(fields=['last_12_direct_payments'], name='recipient_p_last_12_eef9af_idx'),
+        ),
+        migrations.AddIndex(
+            model_name='recipientprofile',
+            index=models.Index(fields=['last_12_loans'], name='recipient_p_last_12_11e0b3_idx'),
+        ),
+        migrations.AddIndex(
+            model_name='recipientprofile',
+            index=models.Index(fields=['last_12_other'], name='recipient_p_last_12_7a7440_idx'),
```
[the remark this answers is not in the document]
collinwr: Yes, the `last_12_*` indexes are used for sorting on those columns.

usaspending_api/api_contracts/contracts/v2/recipient/duns/count.md
```diff
@@ -0,0 +1,49 @@
+FORMAT: 1A
+HOST: https://api.usaspending.gov
+
+# Recipient Duns [/api/v2/recipient/duns/count]
+
+These endpoints are used to power USAspending.gov's recipient profile pages. This endpoint is called by the [/api/v2/recipient/duns/count] endpoint, so a cache can be used to prevent recalculating the count for each request.
```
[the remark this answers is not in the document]
collinwr: Removed

usaspending_api/api_docs/markdown/endpoints.md
```diff
@@ -117,6 +117,7 @@ The currently available endpoints are listed in the following table.
 |[/api/v2/recipient/children/<DUNS\>/](/api/v2/recipient/children/006928857/)|GET| Returns recipient details based on DUNS number |
 |[/api/v2/recipient/duns/<HASH_VALUE\>/](/api/v2/recipient/duns/99a44eeb-23ef-e7c4-1f84-9a695b6f5d2e-R/)|GET| Returns a high-level overview of a specific recipient, given its id |
 |[/api/v2/recipient/duns/](/api/v2/recipient/duns/)|POST| Returns a list of recipients in USAspending DB |
+|[/api/v2/recipient/duns/count/](/api/v2/recipient/duns/count/)|POST| Returns a list of recipients in USAspending DB |
```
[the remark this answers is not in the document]
collinwr: Updated

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -1,59 +1,60 @@
...
 
     queryset = (
         RecipientProfile.objects.filter(qs_filter)
         .values("recipient_level", "recipient_hash", "recipient_unique_id", "recipient_name", amount_column)
         .exclude(recipient_name__in=SPECIAL_CASES)
     )
-
     api_to_db_mapper = {"amount": amount_column, "duns": "recipient_unique_id", "name": "recipient_name"}
 
+    nulls_last = filters["sort"] in ["name", "duns"]
```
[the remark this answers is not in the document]
collinwr: Commented: 

> Nulls Last isn't enabled for the amount sort because it prevents queries sorted by amount columns DESC from using an index on those columns, even though they do cannot contain nulls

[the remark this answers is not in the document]
collinwr: Commented: 

> Nulls Last isn't enabled for the amount sort because it prevents queries sorted by amount columns DESC from using an index on those columns, even though they do cannot contain nulls

usaspending_api/recipient/v2/views/list_recipients.py
collinwr: Commented:
>  A custom cache key is created for this endpoint to ensure that only 'award_type' and 'keyword' fields are used from the request body. Other fields that may come in when this view is called from the RecipientList endpoint, such as 'page', should be ignored.

usaspending_api/recipient/v2/views/list_recipients.py
collinwr: Fixed

usaspending_api/recipient/v2/views/list_recipients.py
collinwr: I think this is unlikely to fail. The try/except blocks are carry over from when this was making an actual http request, and I didn't think of removing them once I switched to directly calling the view. I think they can be removed.

usaspending_api/recipient/v2/views/list_recipients.py
collinwr: See below

usaspending_api/recipient/models.py
collinwr: Fixed

usaspending_api/recipient/v2/views/list_recipients.py
collinwr: Agreed. I updated this. I left them in the same file though, so they can continue to share methods without circular dependencies

usaspending_api/api_contracts/contracts/v2/recipient/duns/count.md
collinwr: Fixed

usaspending_api/recipient/v2/views/list_recipients.py
collinwr: I couldn't find a correct way in the API doc specification to add a max, so I just added it to the body. I originally tried something like this:
```
+ limit (optional, number)
    The number of results to include per page.
        + Default: 5
        + Max: 1000
```
But got contract errors

tony-sappe: In other API Contracts the maximum value is mentioned as a comment on the field. A field comment is text which is aligned with the attributes (in this example, the `+` character) but does not have any leading symbol in the text

usaspending_api/recipient/v2/views/list_recipients.py
tony-sappe: I'm still a little confused. The justification for the custom key doesn't explain why the function is creating a string of `<field1>.<field2>` (or even `-.-`), it seems like an odd format.

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -1,59 +1,68 @@
...
     queryset = (
         RecipientProfile.objects.filter(qs_filter)
         .values("recipient_level", "recipient_hash", "recipient_unique_id", "recipient_name", amount_column)
         .exclude(recipient_name__in=SPECIAL_CASES)
     )
-
     api_to_db_mapper = {"amount": amount_column, "duns": "recipient_unique_id", "name": "recipient_name"}
 
+    # Nulls Last isn't enabled for the amount sort because it prevents queries sorted by amount columns DESC
+    # from using an index on those columns, even though they do cannot contain nulls
```
tony-sappe, line 55: "even though they do cannot contain nulls"

usaspending_api/api_contracts/contracts/v2/recipient/count.md
```diff
@@ -0,0 +1,49 @@
...
+        + `award_type` (optional, enum[string])
+            The award type results are filtered by.
+            + Default: `all`
+            + Members
+                + `all`
+                + `contracts`
+                + `grants`
+                + `loans`
+                + `direct_payments`
+                + `other_financial_assistance`
```
kwhickey, line 32: Should IDV be added here as an award type, even though we don't show it on the recipients listings page?
That may have been an oversight when introducing IDV, that we didn't add it to the page.

usaspending_api/api_docs/markdown/endpoints.md
```diff
@@ -117,6 +117,7 @@ The currently available endpoints are listed in the following table.
 |[/api/v2/recipient/children/<DUNS\>/](/api/v2/recipient/children/006928857/)|GET| Returns recipient details based on DUNS number |
 |[/api/v2/recipient/duns/<HASH_VALUE\>/](/api/v2/recipient/duns/99a44eeb-23ef-e7c4-1f84-9a695b6f5d2e-R/)|GET| Returns a high-level overview of a specific recipient, given its id |
 |[/api/v2/recipient/duns/](/api/v2/recipient/duns/)|POST| Returns a list of recipients in USAspending DB |
+|[/api/v2/recipient/duns/count/](/api/v2/recipient/duns/count/)|POST| Returns the count of recipents for the given filters |
```
kwhickey, line 120: I think the path for this is off. The API docs have it as `/api/v2/recipient/count/` ... you have `duns` in there.

usaspending_api/recipient/migrations/0006_auto_20210315_2028.py
```diff
@@ -0,0 +1,51 @@
...
+        # Performed with RAW SQL because Django migrations don't support 'NULLS LAST'
+        migrations.RunSQL(
+            "CREATE INDEX recipient_profile_recipient_name_d ON public.recipient_profile USING btree (recipient_name DESC NULLS LAST)",
+            reverse_sql="DROP INDEX recipient_profile_recipient_name_d"
+        ),
+        migrations.RunSQL(
+            "CREATE INDEX recipient_profile_recipient_unique_id_d ON public.recipient_profile USING btree (recipient_unique_id DESC NULLS LAST)",
+            reverse_sql="DROP INDEX recipient_profile_recipient_unique_id_d"
+        ),
+        # Performed with RAW SQL because Django migrations don't support GinIndexes with a specific Postgres operator class:
```
kwhickey, line 46: Q: So this then probably means there was another earlier custom RunSQL migration that applied GinIndex to the `recipient_name` column?

usaspending_api/recipient/v2/lookups.py
```diff
@@ -1,3 +1,12 @@
+# based from all_award_types_mappings in usaspending_api.awards.v2.lookups.lookups
+AWARD_TYPES = {
+    "contracts": {"amount": "last_12_contracts", "filter": "contract"},
+    "grants": {"amount": "last_12_grants", "filter": "grant"},
+    "direct_payments": {"amount": "last_12_direct_payments", "filter": "direct payment"},
+    "loans": {"amount": "last_12_loans", "filter": "loans"},
+    "other_financial_assistance": {"amount": "last_12_other", "filter": "other"},
+}
```
kwhickey, line 8: Not sure where these are used and ripple effect, but I suggest keeping pluralization consistent between col name and filter name. 

e.g. 
- `last_12_contracts` -> `contract` (not pluralized, same for `grant`)
- `last_12_loans` -> `loans` (pluralized here)

usaspending_api/recipient/v2/urls.py
```diff
@@ -2,10 +2,11 @@
 from usaspending_api.recipient.v2.views.states import StateMetaDataViewSet, StateAwardBreakdownViewSet, ListStates
 from usaspending_api.recipient.v2.views.recipients import RecipientOverView
 from usaspending_api.recipient.v2.views.recipients import ChildRecipients
-from usaspending_api.recipient.v2.views.list_recipients import ListRecipients
+from usaspending_api.recipient.v2.views.list_recipients import ListRecipients, RecipientCount
 
 urlpatterns = [
     url(r"^duns/$", ListRecipients.as_view()),
+    url(r"^duns/count/$", RecipientCount.as_view()),
```
kwhickey, line 9: Again some inconsistency as to whether this API endpoint should be off of `recipient/count` or `recipient/duns/count`. If it is in fact just counting recipients and there's not implied filter by "duns", I would suggest it just live off of `recipient/count`. But if the `duns` moniker was added to imply that this endpoint allows you to search by duns ... then keep consistent with that I suppose.

usaspending_api/recipient/v2/lookups.py
kwhickey: If the change were to break existing api contracts though, you may not want to do it (it would require keeping backwards compatibility or going to a new v3 for this endpoint).

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -1,59 +1,68 @@
...
 
+    if filters["award_type"] != "all":
+        qs_filter &= Q(award_types__overlap=[AWARD_TYPES[filters["award_type"]]["filter"]])
+
+    return qs_filter
+
+
+def get_recipients(filters={}, count=None):
+    lower_limit = (filters["page"] - 1) * filters["limit"]
+    upper_limit = filters["page"] * filters["limit"]
```
kwhickey, line 39: Deriving these ... will TinyShield have enforced that the overall results to be returned for this page is less than the total (e.g. 1000)? If not, that may need to be enforced here so someone can't give values in their payload that derive a much wider page of results.

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -70,26 +79,77 @@ def get_recipients(filters={}):
...
 
+def duns_count_key_function(view_instance, view_method, request, args, kwargs):
+    """
+    A custom cache key is created for this endpoint to ensure that only 'award_type'
+    and 'keyword' fields are used from the request body. Other fields that may come in
+    when this view is called from the RecipientList endpoint, such as 'page', should
+    be ignored.
+    """
+
+    validated_payload = TinyShield(RECIPIENT_MODELS).block(request.data)
```
kwhickey, line 90: Will this do sanitization of the payload data, or otherwise validate that it conforms to something like strings and numbers. 
If this is becoming part of the key for a Redis cache entry, there is a risk that an injection attack could be used here. This may need some research into:

1. How django-redis or django-cache uses keys (does it has or scrub them before using them?)
2. If not, if you can borrow code from that framework for how it sanitizes the keys first in the default algorithm

Specifically this is a risk for the `keyword` field if I can assume TinyShield is validating that award_type is one of a known set of enum values.

usaspending_api/recipient/v2/views/list_recipients.py
```diff
@@ -1,59 +1,68 @@
...
 
-def get_recipients(filters={}):
-    lower_limit = (filters["page"] - 1) * filters["limit"]
-    upper_limit = filters["page"] * filters["limit"]
-
+def build_duns_base_query(filters):
     qs_filter = Q()
     if "keyword" in filters:
         qs_filter |= Q(recipient_name__contains=filters["keyword"].upper())
         qs_filter |= Q(recipient_unique_id__contains=filters["keyword"])
```
kwhickey, line 29: A possible optimization here could be to also add in a check for `IS NOT NULL` in djangoese 

iirc, you said this turns into sql that looks like
```SQL
"recipient_profile"."recipient_name"::TEXT LIKE '%DEPARTMENT%'
OR "recipient_profile"."recipient_unique_id"::TEXT LIKE '%department%'
```

Since the vast majority of `recipient_unique_id` values are null, would this go faster if that clause was added?

e.g. to result in

```SQL
WHERE
-- ...
(
    "recipient_profile"."recipient_name" IS NOT NULL
    AND
    "recipient_profile"."recipient_name"::TEXT LIKE '%DEPARTMENT%'
)
OR     
(
    "recipient_profile"."recipient_unique_id" IS NOT NULL
    AND
    "recipient_profile"."recipient_unique_id"::TEXT LIKE '%department%'
)
```

So I guess the question is, when doing a `LIKE` does Postgres already ignore nulls, or if not would dropping rows with null help the `LIKE` run more efficiently. 

I can lookup this question. Unless @tony-sappe has the answer.

```

**Review by kwhickey — APPROVED** [review]
Looks really good and clean. Some questions throughout and some risks to address by answer or by code. Nothing glaring standing out though. 

Approving, given the questions are addressed satisfactorily.

File: usaspending_api/recipient/v2/views/list_recipients.py (hunk 24 above)
Comment (tony-sappe) [reply usaspending_api/recipient/v2/views/list_recipients.py:29]: I don't know if documentation will provide the answer. Only trial-and-error experimentation on the exact data and DB hardware will shed light on if adding the checks will help. My assumption is creating partial indexes on non-null values would be more beneficial. That was something that was extensively done on many materialized view indexes.

File: usaspending_api/api_docs/markdown/endpoints.md (hunk 18 above)
Comment (collinwr) [reply usaspending_api/api_docs/markdown/endpoints.md:120]: I have adjusted all these urls to match this format: `/api/v2/recipient/count/`

File: usaspending_api/recipient/v2/urls.py (hunk 21 above)
Comment (collinwr) [reply usaspending_api/recipient/v2/urls.py:9]: I have adjusted all these urls to match this format: `/api/v2/recipient/count/`

File: usaspending_api/recipient/v2/views/list_recipients.py (hunk 22 above)
Comment (collinwr) [reply usaspending_api/recipient/v2/views/list_recipients.py:39]: Not quite sure what you're asking Keith. TinyShield prevents someone from requesting a limit greater than 1000. These `lower_limit` and `upper_limit` use that validated limit to determine the actual range of records to pull from the db.

File: usaspending_api/recipient/v2/views/list_recipients.py (hunk 23 above)
Comment (collinwr) [reply usaspending_api/recipient/v2/views/list_recipients.py:90]: Keith, this has been reworked to make use of our caching library. The values coming from the user are hashed before stored in the cache

File: usaspending_api/recipient/migrations/0006_auto_20210315_2028.py (hunk 19 above)
Comment (collinwr) [reply usaspending_api/recipient/migrations/0006_auto_20210315_2028.py:46]: Yes, there is a comment in the model calling out specific migrations with manual SQL being applied.

File: usaspending_api/api_contracts/contracts/v2/recipient/count.md (hunk 17 above)
Comment (collinwr) [reply usaspending_api/api_contracts/contracts/v2/recipient/count.md:32]: I don't see any IDV in this table. Adding it may require some changes to the ETL

File: usaspending_api/recipient/v2/lookups.py (hunk 20 above)
Comment (collinwr) [reply usaspending_api/recipient/v2/lookups.py:8]: Making this change would require updates to the frontend as well

File: usaspending_api/recipient/v2/views/list_recipients.py (hunk 22 above)
Comment (kwhickey) [reply usaspending_api/recipient/v2/views/list_recipients.py:39]: > TinyShield prevents someone from requesting a limit greater than 1000.
✅

File: usaspending_api/recipient/v2/views/list_recipients.py (hunk 23 above)
Comment (tony-sappe) [reply usaspending_api/recipient/v2/views/list_recipients.py]: > Will this do sanitization of the payload data, or otherwise validate that it conforms to something like strings and numbers.

Yes, that is one of the primary reasons for using TinyShield

File: usaspending_api/api_contracts/contracts/v2/recipient/duns.md [hunk 25]
```diff
@@ -33,6 +33,7 @@ This endpoint returns a list of recipients, their level, DUNS, and amount.
                 + `amount`
         + `limit` (optional, number)
             The number of results to include per page.
+            Maximum: 1000
```
tony-sappe, line 36: This was better as a single line comment

usaspending_api/api_docs/markdown/endpoints.md
```diff
@@ -117,6 +117,7 @@ The currently available endpoints are listed in the following table.
 |[/api/v2/recipient/children/<DUNS\>/](/api/v2/recipient/children/006928857/)|GET| Returns recipient details based on DUNS number |
 |[/api/v2/recipient/duns/<HASH_VALUE\>/](/api/v2/recipient/duns/99a44eeb-23ef-e7c4-1f84-9a695b6f5d2e-R/)|GET| Returns a high-level overview of a specific recipient, given its id |
 |[/api/v2/recipient/duns/](/api/v2/recipient/duns/)|POST| Returns a list of recipients in USAspending DB |
+|[/api/v2/recipient/count/](/api/v2/recipient/count/)|POST| Returns the count of recipents for the given filters |
```
tony-sappe, line 120: alphabetical order is broken from the url change

usaspending_api/recipient/v2/views/list_recipients.py
kwhickey: We can defer this but may want to experiment with this style of query directly in sql if we find things still slower or a high degree of writes (e.g. in-memory table scans) seeing under concurrent querying.

usaspending_api/recipient/v2/lookups.py
kwhickey: no need then

kwhickey (approved):
changes look good

tony-sappe (approved):

PR #3039, opened 2021-03-15
[DEV-6756] Check es doc duplication management command

kwhickey:
**Description:**
Adding a management command to count doc-duplication statistics in an Elasticsearch index.

NOTE: This is a _utility_ management command to adding in data quality checks or troubleshooting/debugging. It can be used as-is, but has some performance caveats. It works, but could be improved upon if further use and specialization is desired.

**Technical details:**
- Uses one of a scroll or partition strategy to progress through all docs
- The former has client-side burden, while the latter has server-side burden
- These are useful if truly needing to get a count of duplication, but doing so comes at a lot of processing time (several hours likely)
    - Could be used for instance to provide assurance that an error in ES indexing or deletion didn't yield duplicate documents in an index

1. [NA] Unit & integration tests updated
    - [ ] Backend
    - [NA] Frontend <OPTIONAL>
    - [NA] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-6756](https://federal-spending-transparency.atlassian.net/browse/DEV-6756):

```
```


usaspending_api/etl/management/commands/check_es_doc_duplication.py
```diff
@@ -0,0 +1,245 @@
...
+
+from django.core.management.base import BaseCommand
+from elasticsearch import Elasticsearch
+from elasticsearch.helpers import scan
+from math import ceil, floor
+from multiprocessing.pool import ThreadPool
+from threading import Lock, get_ident
+
+
+_log = logging.getLogger(__name__)
```
tony-sappe, line 12: Since this is a management command, it should be "script" instead of the module name. Also the convention is to use `logger` instead of `_log`.


```suggestion
logger = logging.getLogger("script")
```

usaspending_api/etl/management/commands/check_es_doc_duplication.py
```diff
@@ -0,0 +1,245 @@
...
+
+    _es_client_config: dict = None
+    _index: str = None
+    _partition_size: int = None
+    _num_partitions: int = None
+    _lock = Lock()
+    _completed_count: int = 0
+    _percents = {p for p in range(1, 101)}
+    _duplicated_doc_ids: dict = {}
+    _duplicated_query_template = {
```
tony-sappe, line 41: Is it necessary to name everything with a trailing underscore? This is a management command, the likelihood of it being imported by other code is incredibly small and discouraged.

usaspending_api/etl/management/commands/check_es_doc_duplication.py
```diff
@@ -0,0 +1,245 @@
...
+    def handle_with_scrolling(self, **options):
+        self._es_client_config = {"hosts": options["es_hostname"], "timeout": options["es_timeout"]}
+        self._partition_size = options["partition_size"]
+        es = Elasticsearch(**self._es_client_config)
+        self._index = options["index"]
+        doc_count = es.count(index=options["index"])["count"]
+        _log.info(f"Found {doc_count:,} docs in index {self._index}")
+        self._num_partitions = ceil(doc_count / self._partition_size)
+
+        # Instantiate a numpy array with the dumber of docs' _id fields we expect to touch
```
tony-sappe, line 133:
```suggestion
        # Instantiate a numpy array with the number of docs' _id fields we expect to touch
```

usaspending_api/etl/management/commands/check_es_doc_duplication.py
```diff
@@ -0,0 +1,245 @@
...
+            "time. The results are then aggregated in Python on the client-side, so it could be memory intensive"
+            "on the client host on which it is run. partition: will run terms aggregations on disjoint "
+            "partitions of the docs, returning --partition-size agg buckets with each call. The aggs being done "
+            "concurrently on the server can cause server straing on CPU, memory, and induce request timeouts",
+        )
+        parser.add_argument(
+            "--es-hostname",
+            type=str,
+            help="The HTTP(S) URL of the Elasticsearch cluster endpoint",
+        )
```
tony-sappe, line 70: Could use the env var + global variable `ES_HOSTNAME` like the other scripts

usaspending_api/etl/management/commands/check_es_doc_duplication.py
```diff
@@ -0,0 +1,245 @@
...
+        self._partition_size = options["partition_size"]
+        es = Elasticsearch(**self._es_client_config)
+        self._index = options["index"]
+        doc_count = es.count(index=options["index"])["count"]
+        _log.info(f"Found {doc_count:,} docs in index {self._index}")
+        self._num_partitions = ceil(doc_count / self._partition_size)
+
+        # Instantiate a numpy array with the dumber of docs' _id fields we expect to touch
+        all_ids = np.empty(doc_count, dtype=int)
+        _log.info(f"Collecting _ids from all {doc_count:,} in scrolling batch queries of size {self._partition_size}")
```
tony-sappe, line 135:
```suggestion
        _log.info(f"Collecting _ids from all {doc_count:,} in scrolling batch queries of size {self._partition_size:,}")
```

tony-sappe (commented):
Leaving a few comments. Since this is a nice-to-have management command they don't need to be blockers to merging

usaspending_api/etl/management/commands/check_es_doc_duplication.py
kwhickey: It's a good practice we should follow everywhere in our python code. If fields are intended to only be used within a class itself (aka "private" in an object-oriented sense), they should be marked via a naming convention as such.

Also generates IDE warnings (as if the underscore wasn't enough) when you are using a "private" field.

usaspending_api/etl/management/commands/check_es_doc_duplication.py
kwhickey: Left out for flexibility of how this is used - a bit more flexibility by not having to have env vars setup and pulled in via `settings.py` to run this. Could be future improvement if used regularly

usaspending_api/etl/management/commands/check_es_doc_duplication.py
kwhickey: I still don't think this is the best way to do this. 
By hard-coding this to a single logger named "script", there is no way to adjust log level of different modules executed _during_ the script run via the logging config in `settings.py`. 

The way this should be done is to have e.g. this (and other) management command modules all getting associated with the `"script"` handler by way of providing their module name as a logger in the `loggers` section of the logging config dict. If that becomes too verbose, that dict can be modified to pattern match (e.g. all modules that have `*.management.*` in their module namespace) to associate them with the `script` handler.

usaspending_api/etl/management/commands/check_es_doc_duplication.py
kwhickey: Fixed

usaspending_api/etl/management/commands/check_es_doc_duplication.py
kwhickey: Fixed

usaspending_api/etl/management/commands/check_es_doc_duplication.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def handle(self, *args, **options):
+        if options["query_type"] == "scroll":
+            _log.info("Starting to gather doc _ids using 'scroll' query strategy")
+            self.handle_with_scrolling(**options)
+        elif options["query_type"] == "partition":
+            _log.info("Starting to gather doc _ids using 'partition' query strategy")
+            self.handle_with_partitioning(**options)
+        else:
+            raise RuntimeError("Invalid argument provided for required option --query-type.")
```
sethstoudenmier, line 122: Not worth blocking on, but you can add this check in the argument parser using `choices`: https://docs.python.org/3/library/argparse.html#choices

usaspending_api/etl/management/commands/check_es_doc_duplication.py
```diff
@@ -0,0 +1,245 @@
...
+            duped_id_count = len(duplicated_doc_ids)
+            max_dupe = np.amax(duped_id_counts)
+            p75 = np.percentile(duped_id_counts, 75)
+            p95 = np.percentile(duped_id_counts, 95)
+            _log.warning(
+                f"Found {duped_id_count:,} _ids with more than one doc in the index. "
+                f"Max duplication (p100) = {max_dupe}; p95 = {p95}; p75 = {p75}"
+            )
+            header = "_id         | duplicated\n" + "-" * 12 + "|" + "-" * 12 + "\n"
+            rows = "\n".join([str(k).ljust(12) + "|" + str(v).ljust(12) for k, v in duplicated_doc_ids.items()])
```
sethstoudenmier, line 173: This printing out of rows is really nice for the `scroll` approach and would be nice to have for the `partition` approach.

sethstoudenmier (approved):
LGTM and is REALLY nice to have! Tested locally by forcing a duplicate document. Both approaches picked up on the duplicate and reporting accurately. The table printout in `scroll` would be nice to have for `partition`, but won't block on it. Approving and leaving for you to merge if not worth adding that table view.

PR #3045, opened 2021-03-18
Mod/dev 6589 refactor valid time periods

brett-varney:
**Description:**
update calculate_last_completed_fiscal_quarter to use DABSSubmissionWindowSchedule table instead of calculated dates

**Technical details:**
return the highest submission_fiscal_quarter that has been revealed for a given fiscal year

2. [x] API documentation updated - NA
    - [ ] Backend
    - [x] Frontend <OPTIONAL> - NA
    - [x] Operations <OPTIONAL> - NA
    - [x] Domain Expert <OPTIONAL> - NA
4. [x] Matview impact assessment completed - NA
5. [x] Frontend impact assessment completed - NA
6. [x] Data validation completed - NA, confirmed by unit tests still working
7. [x] Appropriate Operations ticket(s) created - NA
8. [x] Jira Ticket [DEV-6589](https://federal-spending-transparency.atlassian.net/browse/DEV-6589):
    - [x] Performance evaluation of affected (API | Script | Download) - NA
    - [x] Before / After data comparison - NA, confirmed by unit tests still working

```
```


usaspending_api/common/helpers/fiscal_year_helpers.py
```diff
@@ -185,38 +184,17 @@ def bolster_missing_time_periods(filter_time_periods, queryset, date_range_type,
...
-        else:
-            fiscal_quarter = current_fiscal_quarter - 1
-    elif fiscal_year < current_fiscal_date_adjusted.fiscal_year:
-        fiscal_quarter = 4
-    else:
-        # The future cannot have completed quarters unless you're into shady accounting.
-        fiscal_quarter = None
-
-    return fiscal_quarter
+    row = DABSSubmissionWindowSchedule.objects.filter(
```
sethstoudenmier, line 194: For previous years it works as expected by providing `fiscal_quarter = 4`. However, if the `fiscal_year = <CURRENT FISCAL YEAR>` then it is returning the last `fiscal_quarter` even if it is not completed. I think there is a bug in the current code that is also not handling the case of returning the last completed quarter. Stepping through the old implementation the `as_of_date` is the `current_fiscal_date` (or today + 3 months). Then the FiscalQuarter inside of FiscalDateTime is also adjusting for 3 months.

This comment is what I am referring to by returning the last completed quarter:
```
Returns either the most recently completed fiscal quarter or None if it's too early in the
fiscal year for the first quarter to be considered "completed".
```

SQL for current implementation: 
```
SELECT 
	MAX("dabs_submission_window_schedule"."submission_fiscal_quarter") AS "submission_fiscal_quarter__max"
FROM 
	"dabs_submission_window_schedule"
WHERE 
	"dabs_submission_window_schedule"."submission_reveal_date" <= '2021-03-19T13:57:27.109797+00:00'::timestamptz 
	AND "dabs_submission_window_schedule"."submission_fiscal_year" = 2021;
```

usaspending_api/common/tests/unit/test_fiscal_year_helpers.py
```diff
@@ -69,41 +72,48 @@ def test_dates_are_fiscal_year_bookends():
...
-    assert fyh.calculate_last_completed_fiscal_quarter(2000) == 4
+    now = datetime.now(timezone.utc)
+    yesterday = now + timedelta(days=-1)
+    tomorrow = now + timedelta(days=1)
+    current_fy = fyh.generate_fiscal_year(now)
+
+    mommy.make(
+        "submissions.DABSSubmissionWindowSchedule",
+        submission_fiscal_year=2000,
+        submission_reveal_date=now,
```
sethstoudenmier, line 85: Any reason the fiscal_year is not the same as the year of the reveal date?

usaspending_api/common/tests/unit/test_fiscal_year_helpers.py
```diff
@@ -69,41 +72,48 @@ def test_dates_are_fiscal_year_bookends():
...
+        submission_fiscal_quarter=4,
+    )
+    mommy.make(
+        "submissions.DABSSubmissionWindowSchedule",
+        submission_fiscal_year=current_fy,
+        submission_reveal_date=tomorrow,
+        submission_fiscal_quarter=5,
+    )
+
+    assert fyh.calculate_last_completed_fiscal_quarter(2000) == 1
```
sethstoudenmier, line 113: Would add another test case here to make sure that you get back Q1 through Q3 if they are completed fiscal quarters (kinda similar to the tests that are being removed).

usaspending_api/reporting/tests/integration/test_agencies_overview.py
```diff
@@ -3,18 +3,13 @@
...
-)
+from datetime import datetime, timezone
 
 url = "/api/v2/reporting/agencies/overview/"
 
-CURRENT_FISCAL_YEAR = current_fiscal_year()
-CURRENT_LAST_QUARTER = calculate_last_completed_fiscal_quarter(CURRENT_FISCAL_YEAR) or 1
-CURRENT_LAST_PERIOD = get_final_period_of_quarter(CURRENT_LAST_QUARTER) or 3
+CURRENT_FISCAL_YEAR = 2021
+CURRENT_LAST_QUARTER = 2
```
sethstoudenmier, line 11: Any reason these were updated to be constant values? Very different from calculating the final period.

sethstoudenmier (changes requested):
Few questions and two things that I think should be fixed.

I don't think the method is working as intended and the test cases for that method should cover more Fiscal Quarters (which should help to verify that it is working as intended).

usaspending_api/agency/tests/integration/test_agency_budgetary_resources.py
```diff
@@ -120,6 +121,20 @@ def data_fixture():
         submission=sa2_12,
     )
 
+    mommy.make(
```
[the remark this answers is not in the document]
brett-varney: get_prior_year_agency_budgetary_resources uses calculate_last_completed_fiscal_quarter, so mock data needed to be added for test to run

usaspending_api/common/tests/unit/test_fiscal_year_helpers.py
brett-varney: No, but they don't have to be either. As long as the requested year matches and reveal_date is "now or earlier", it produces expected results.

usaspending_api/common/tests/unit/test_fiscal_year_helpers.py
brett-varney: @sethstoudenmier I don't follow your thinking. The function only returns a single value for the latest quarter "revealed" (if any); you mean to add other test cases simulating different "now" dates? I think I have those covered below (last value is 4, not 3 or 5 though they are all in the same year).

usaspending_api/reporting/tests/integration/test_agencies_overview.py
brett-varney: Wil & I realized this was effectively testing functions that are already tested in their own files. Now that calculate_last_completed_fiscal_quarter is database-dependent, it would require moving all these constants including the assurance_statements below into each test case so it can be marked @pytest.mark.django_db, without adding any value to these tests.

usaspending_api/common/helpers/fiscal_year_helpers.py
brett-varney: @sethstoudenmier So if nothing matches the two filters below, it returns None as expected (confirmed in tests). Maybe we should talk? I'm not understanding the problem you're pointing out.

usaspending_api/common/tests/unit/test_fiscal_year_helpers.py
sethstoudenmier: Followed up in Slack pertaining to this and the larger comment above. Outcome is to look at using the `is_quarter` column to avoid returning a quarter that has only a single period revealed.

usaspending_api/common/tests/unit/test_fiscal_year_helpers.py
```diff
@@ -69,41 +72,60 @@ def test_dates_are_fiscal_year_bookends():
...
+        submission_fiscal_year=current_fy,
+        submission_reveal_date=tomorrow,
+        submission_fiscal_quarter=5,
+        is_quarter=True,
+    )
+
+    assert fyh.calculate_last_completed_fiscal_quarter(2000) == 1 # not 2, since is_quarter=False
+    assert fyh.calculate_last_completed_fiscal_quarter(2001) is None # no row in table for 2001
+    assert fyh.calculate_last_completed_fiscal_quarter(2010) is None # not revealed yet
+    assert fyh.calculate_last_completed_fiscal_quarter(current_fy) == 4 # not 3, since both are revealed & quarters; 5 not revealed yet
```
sethstoudenmier, line 128: > 5 not revealed yet

Didn't notice before that this test case had a fiscal_quarter set to 5. I think you can remove the bit of test data that has `submission_fiscal_quarter=5` since that will never occur realistically.

And that last assertion can be updated to just remove everything after the semicolon because it is still a good test.

usaspending_api/common/helpers/fiscal_year_helpers.py
sethstoudenmier: Followed up in Slack pertaining to this. Outcome is to look at using the is_quarter column to avoid returning a quarter that has only a single period revealed.

sethstoudenmier (approved):

PR #3050, opened 2021-03-18
[DEV-6567] Add TinyShield for Award downloads

sethstoudenmier:
**Description:**
Make sure that the Award downloads are validating filters the same way as the endpoints that return results based on those filters.

**Technical details:**
Added TinyShield to validate the Award downloads. These downloads include:
* `/api/v2/download/awards` (Advanced Search - Award)
* `/api/v2/download/transactions` (Advanced Search - Transaction)
* `/api/v2/bulk_download/awards` (Custom Award Download and Keyword Search)

The implementation will support future downloads that are implemented using TinyShield. The case of the Award downloads is a unique one since all of the endpoints listed above used the same validation function and in the case of Custom Award Downloads it even normalizes the data before passing it over to validation. Ideally validation for future downloads will be much simpler to implement than the few in this PR.

Small changes were made to TinyShield:
* Allowing a list of length == 0
* Using the default value in an object_key

Changes to the Monthly files command are due to no longer needing the `process_filters()` method under `YearLimitedDownloadViewSet`. The overall outcome should be the same, but the normalization and validation was moved. Ideally the Custom Award Download endpoint would be refactored to use the same filters as other endpoints, but I attempted to not change any logic for a download, only how it performed validation.

    - [ ] Backend
8. [x] Jira Ticket [DEV-6567](https://federal-spending-transparency.atlassian.net/browse/DEV-6567):

```
```


usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -58,9 +58,14 @@ def post(
...
-        json_request = request_type.value["validate_func"](request.data)
-        json_request["request_type"] = request_type.value["name"]
-        ordered_json_request = json.dumps(order_nested_object(json_request))
+        if validator_type is not None:
+            validator = validator_type(request.data)
+            json_request = validator.json_request
+        else:
+            json_request = request_type.value["validate_func"](request.data)
+            json_request["request_type"] = request_type.value["name"]
+        json_request = order_nested_object(json_request)
```
collinwr, line 67: What happens if neither a `validator_type` or `json_request` is passed in? Should `json_request` still be initialized?

usaspending_api/common/validator/helpers.py
```diff
@@ -87,7 +87,7 @@ def _verify_float_value(value):
 
 
 def validate_array(rule):
-    rule["min"] = rule.get("min") or 1
+    rule["min"] = 1 if rule.get("min") is None else rule.get("min")  # Allows empty arrays
```
tony-sappe, line 90: Do we want empty arrays? I think the original intention was to not allow a key with an empty array. simply block the request and remove the key & empty array

usaspending_api/download/v2/base_download_viewset.py
sethstoudenmier: Every download should have at least one of those. They are not user provided, but provided when we define the endpoint.

usaspending_api/common/validator/helpers.py
sethstoudenmier: We have a few places in the frontend that are supplying empty arrays as filters to the downloads. Can handle those in the request if we don't want to make this change.

collinwr (approved):
I like the reorganization. Everything seems to be in order. Approving

usaspending_api/common/validator/helpers.py
tony-sappe: Discussed in slack. Plan to resolve in the future

PR #3060, opened 2021-03-23
[OPS-1686] create server Dockerfile with dependencies needed to run the backend and set up and run django management jobs on the CI/CD server 

bminahan-kc:
**Description:**
create `Dockerfile-Server` with dependencies needed to run the backend and set up and run django management jobs on the CI/CD server 

**Technical details:**
There are additional dependencies needed for the docker container to successfully be used to run USASpending jobs on the CI/CD server.
The purpose of this new dockerfile is to be able to run USAspending tasks inside a container with all the appropriate config and access to all the external resources to perform the tasks in AWS-hosted environments. The additional dependencies are:

- the server requirements file
- ansible + awscli to adjust config files before job execution

    - [x] Backend
    - [x] Operations
6. [N/A] Data validation completed
8. [x] Jira Ticket [OPS-1686](https://federal-spending-transparency.atlassian.net/browse/OPS-1686):
    - [N/A] Performance evaluation of affected (API | Script | Download)
    - [N/A] Before / After data comparison

```
this is adding an additional Dockerfile that will only be used by the CI/CD server and is not a change to application code.
```


gpontejos (approved):

Dockerfile-Server
```diff
@@ -0,0 +1,49 @@
...
+    gcc-c++ \
+    postgresql-devel \
+    pcre \
+    pcre-devel \
+    mlocate \
+    wget \
+    unzip \
+    bzip2-devel \
+    memcached \
+    libmemcached \
```
kwhickey, line 40: I'm still not sure where and why `memcached` becomes a dependency to usaspending. This may be a stale holdover from early days _before_ switching to a Redis-based caching system in AWS elasticache.

Dockerfile-Server
```diff
@@ -0,0 +1,49 @@
...
+RUN echo "$(python3 --version)"
+
+##### Copy python packaged
+WORKDIR /dockermount
+COPY requirements/ /dockermount/requirements/
+RUN python3 -m pip install -r requirements/requirements.txt
+
+##### Additional server requirements
+RUN yum install -y \
+    git \
```
kwhickey, line 30: Do you know what we end up using git for _during_ the launch or runtime of the container? I think it should mostly be a packer-time thing with ansible provisioner.

Dockerfile-Server
```diff
@@ -0,0 +1,49 @@
+FROM centos:7
```
kwhickey, line 1: In general, should this mimic how an EC2 is built out? 
If so -- sorry if this blows up lots of things -- but I really liked the idea of using Packer with a Docker builder, and allowing us to still employ our EC2-ansible provisioners with that docker builder. I think there were some comments to that effect in this ticket: (~~can't seem to find it, thought it was under OPS-200~~) this comment:  https://federal-spending-transparency.atlassian.net/browse/OPS-254?focusedCommentId=34453

Perhaps we'd need to refactor our ansible provisioners so we could isolate _just_ the layers needed in this build? If it's not that easy to separate all that would go into an EC2 and the subset that would go into here, maybe just keep them split ... but that does allow drift between how we do it in an EC2 and how we do it in this container. 

But ... that whole concept may lead the goal of this astray. If the intent is to make this _very close_ to the other `Dockerfile` we have, can we get better reuse out of 1) creating a base image the two share? 2) or using the non-Server file _as_ the base image for the server 1, and just adding the extra dependencies, 3) or is there a way with docker multi-stage builds to pull in another docker file as the base series of commands to run (i.e. they both share an included dockerfile?).

Dockerfile-Server
```diff
@@ -0,0 +1,49 @@
...
+    pcre-devel \
+    mlocate \
+    wget \
+    unzip \
+    bzip2-devel \
+    memcached \
+    libmemcached \
+    http://mirror.centos.org/centos/7/os/x86_64/Packages/libmemcached-devel-1.0.16-5.el7.x86_64.rpm
+RUN python3 -m pip install -r requirements/requirements-server.txt
+RUN python3 -m pip install ansible==2.8.3 awscli redis
```
kwhickey, line 43: Are these used on the server in prod or just for devops jobs? If so, probably best to make them part of `requirements-server.txt`

If not, append commands with an `&&` so as to not add additional layers to the docker image.

kwhickey, line 43: Is there a specific redis client lib we can rely on based on the server side versions?

Dockerfile-Server
```diff
@@ -0,0 +1,49 @@
...
+    pcre \
+    pcre-devel \
+    mlocate \
+    wget \
+    unzip \
+    bzip2-devel \
+    memcached \
+    libmemcached \
+    http://mirror.centos.org/centos/7/os/x86_64/Packages/libmemcached-devel-1.0.16-5.el7.x86_64.rpm
+RUN python3 -m pip install -r requirements/requirements-server.txt
```
kwhickey, line 42: `pip install` can take multiple `-r` args for multiple requirements files. I think it may be better to just expand line 26 to take an additional file rather than running pip another time and adding another layer to the docker image.

kwhickey (commented):
Some questions and suggestions inline

Dockerfile-Server
bminahan-kc: these are dependencies to being able to install `pylibmc` pip package which is in the `requirements-server.txt` file.
If we can safely remove this application code dependency, we can also remove these two yum packages.

Dockerfile-Server
bminahan-kc: the ultimate goal of this image is to be able to run our various management commands in an isolated docker environment. Not necessarily run the application. the main docker image built with the regular `Dockerfile` would be responsible for that. Based on _how_ we currently run our management commands is where we get all these additional dependencies (example: pulling a file from S3 using the CLI before running a job)

bminahan-kc: I think it does make sense though to start with the existing image built using `Dockerfile` and then just add stuff for the purposes of running all our management commands in the way we _currently_ run them (don't want to out scope this ticket to refactoring how we run all our jobs)

Dockerfile-Server
bminahan-kc: used for jobs only, not by a running api server.

Dockerfile-Server
bminahan-kc: the dependency on memcached and libmemcached being installed before `requirements-server.txt` kind of throws this out the window.

Dockerfile-Server
bminahan-kc: we sometimes use it to get the current code branch name or SHA

bminahan-kc: packer is not involved in building this container.

Dockerfile-Server
bminahan-kc: resolved as part of latest patch

Dockerfile-Server
bminahan-kc: resolved by latest patch

bminahan-kc: we actually don't have to install redis, it is already a dependency of django-redis in app requirements. resolved in latest patch

Dockerfile-Server
bminahan-kc: resolved in latest patch

PR #3063, opened 2021-03-26
[DEV-6866] Add TinyShield to Account Download and more cleanup

sethstoudenmier:
Currently has "Do Not Merge" label since #3055 should be merged first. Will be updated to DEV branch following that PR merging into DEV.

**Description:**
Add TinyShield to Account Download for consistent request validation. 

**Technical details:**
Added TinyShield to Account Download following the new pattern for validation. Also made a minor change to how TinyShield handles validation of dictionaries. This shows up when the key is something like `filters|agency` and TinyShield attempts to run `filters.get("agency")` making the assumption that "filters" is a dictionary. This is causing endpoints to blow up when something like `{"filters": "this will fail"}` is passed to the endpoint since a string doesn't have a `.get()` method.

    - [ ] Backend
8. [x] Jira Ticket [DEV-6866](https://federal-spending-transparency.atlassian.net/browse/DEV-6866):

```
```


usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -50,13 +41,8 @@ def post(
...
             return self.get_download_response(file_name=filename)
 
-        if validator_type is not None:
-            validator = validator_type(request.data)
-            json_request = validator.json_request
-        else:
-            json_request = request_type.value["validate_func"](request.data)
-            json_request["request_type"] = request_type.value["name"]
-        json_request = order_nested_object(json_request)
+        validator = validator_type(request.data)
```
collinwr, line 44: `validator_type` is an optional parameter, will removing the check that it exists cause problems if it's not provided.

sethstoudenmier: Realistically it should not because we are deciding what validator type to pass back not the user, but I can add a check and throw an error to make sure it is clear. It is only optional if you provide a "special" request type where we pre-generate it (right now only disaster download).

collinwr: I think add some error handling here would be useful for future developers

collinwr:
Other than my comment above, everything looked good. I checked to see that any validation logic cut out was added back somewhere else.

collinwr (approved):

PR #3067, opened 2021-03-29
[WARMFIX] Fix calculation of total_budgetary_resources for publish_dates

emily-brents:
**Description:**
Fixes an issue with the publish_dates endpoint on the about the data page where the total_budgetary_resources amount was being summed per fiscal year when it should have been taking the latest entry for that fiscal year

The actual percentages on the website page will also require a frontend fix, since the total federal amount is also being summed per fiscal year, which is incorrect

    - [ ] Backend

```
```

### DIFF
--- a/usaspending_api/reporting/tests/integration/test_agencies_publish_dates.py
+++ b/usaspending_api/reporting/tests/integration/test_agencies_publish_dates.py
@@ -13,20 +13,24 @@ def publish_dates_data(db):
         submission_reveal_date="2020-01-01 00:00:00.000000+00",
         submission_fiscal_year=2020,
         submission_fiscal_month=7,
-        is_quarter=True,
     )
     dabs2 = mommy.make(
         "submissions.DABSSubmissionWindowSchedule",
         submission_reveal_date="2020-01-02 00:00:00.000000+00",
         submission_fiscal_year=2019,
         submission_fiscal_month=12,
-        is_quarter=True,
     )
     dabs3 = mommy.make(
-        "submissions.DABSSubmissionWindowSchedule", submission_reveal_date="2019-01-01 00:00:00.000000+00"
+        "submissions.DABSSubmissionWindowSchedule",
+        submission_reveal_date="2019-01-01 00:00:00.000000+00",
+        submission_fiscal_year=2019,
+        submission_fiscal_month=12,
     )
     dabs4 = mommy.make(
-        "submissions.DABSSubmissionWindowSchedule", submission_reveal_date="2019-01-02 00:00:00.000000+00"
+        "submissions.DABSSubmissionWindowSchedule",
+        submission_reveal_date="2019-01-02 00:00:00.000000+00",
+        submission_fiscal_year=2019,
+        submission_fiscal_month=12,
     )
     tas1 = mommy.make("accounts.TreasuryAppropriationAccount")
     tas2 = mommy.make("accounts.TreasuryAppropriationAccount")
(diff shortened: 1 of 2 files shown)


usaspending_api/reporting/v2/views/agencies/publish_dates.py
```diff
@@ -38,16 +38,18 @@ def get_agency_data(self):
         agency_filters = []
         if self.filter is not None:
             agency_filters.append(Q(name__icontains=self.filter) | Q(abbreviation__icontains=self.filter))
+        submission_window = DABSSubmissionWindowSchedule.objects.filter(
```
sethstoudenmier, line 41: Do you foresee any issues with this query not taking into account Quarter vs Monthly submission? Only ask since it is possible for them to have different reveal dates. For example, in Prod FY2021 Q3 and FY 2021 Q3 P9 have two different reveal dates even though they both have a FY of 2021 and Fiscal Period of 9.

usaspending_api/reporting/tests/integration/test_agencies_publish_dates.py
```diff
@@ -573,7 +579,7 @@ def test_publication_date_sort(client, publish_dates_data):
     )
     mommy.make(
         "reporting.ReportingAgencyOverview",
-        toptier_code="001",
+        toptier_code="001" "",
```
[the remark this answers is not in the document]
emily-brents: beats me

usaspending_api/reporting/v2/views/agencies/publish_dates.py
emily-brents: That is a good point, do we default to one over the other in cases like this? iirc we generally go with the quarter?

sethstoudenmier: I think quarter is a safer bet since it is typically later than its period alternative. Would need to double check in other places something like this occurs to be certain.

sethstoudenmier: Looking closer I think this needs similar logic to the `periods` subquery below. Two examples I used:
* SBA --> `toptier_code = 073`
* CFTC --> `toptier_code = 339`

For FY21, SBA has a value for `total_budgetary_resources` that is in Period 4 (which is past the reveal date and not quarterly). CFTC has a value for `total_budgetary_resources` that is in Period 3 and submits quarterly so no value for Period 4 right now. 

Previous approach returned `Max("submission_fiscal_month") == 4` which works for SBA but returns `0` for CFTC. The new approach returns `Max("submission_fiscal_month") == 3` which works for CFTC but not SBA because it isn't the most recent closed period for SBA.

Current SQL generated by Django for that one part:
```
SELECT U0."total_budgetary_resources" 
FROM "reporting_agency_overview" U0 
WHERE 
     (U0."fiscal_period" = 4 AND U0."fiscal_year" = 2021 
     AND U0."toptier_code" = ("vw_published_dabs_toptier_agency"."toptier_code"))
```

I think we want something like this:
```
SELECT RAO.total_budgetary_resources
FROM reporting_agency_overview RAO
WHERE RAO.fiscal_period = (
	SELECT MAX(DSWS.submission_fiscal_month) 
	FROM submission_attributes SA 
	INNER JOIN dabs_submission_window_schedule DSWS ON (SA.submission_window_id = DSWS.id)
	WHERE SA.reporting_fiscal_year = 2021 AND RAO.toptier_code = SA.toptier_code AND DSWS.submission_reveal_date <= '2021-03-29T16:01:00.816980+00:00'::timestamptz
) AND RAO.fiscal_year = 2021 AND RAO.toptier_code = vw_published_dabs_toptier_agency.toptier_code
```

Thoughts? I have a concern around the performance of this, but I think we need to break apart Quarter and Period submissions when retrieving a single value otherwise you could end up with the wrong values using the example above.

sethstoudenmier (approved):
Query is still under 1 second for performance and handles quarterly vs period submissions. LGTM.

PR #3083, opened 2021-04-05
Staging -> Master

collinwr:
DEV-6606 - Write Obligations by Award Type API contract
DEV-6496 - Profile and optimize api/v2/recipient/duns/ endpoint DB queries which can slow down the prod DB
DEV-6612 - Implement Budgetary Resources bar chart API endpoint changes
DEV-6589 - Revisit how valid time periods are determined by the API business logic
DEV-6902 - Update api/v2/federal_accounts// with agency detail
DEV-6567 - Downloads from Advanced Search page results lead to empty downloads if filters strings lead with whitespace characters
DEV-6872 - Spending By Geography returns HTTP 500 if there is no value for name


kwhickey (approved):
Approving for merge into master in prep for prod deploy.

PR #3085, opened 2021-04-07
[DEV-6435] /api/v2/agency/{toptier_code}/budgetary_resources/ Contract Update

MaxwellKendall:
**Description:**
adds `"agency_obligation_by_period"` property to response for route `/api/v2/agency/{toptier_code}/budgetary_resources/`

**Technical details:**
N/A

**Requirements for PR merge:**

1. `N/A` Unit & integration tests updated
2. [x] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - `N/A` Operations <OPTIONAL>
    - [`N/A`Domain Expert <OPTIONAL>
4. `N/A` Matview impact assessment completed
5. [x] Frontend impact assessment completed
6. `N/A`Data validation completed
7. `N/A` Appropriate Operations ticket(s) created
8. [ ] Jira Ticket [DEV-123](https://federal-spending-transparency.atlassian.net/browse/DEV-7131):
    - [ ] Link to this Pull-Request
    - `N/A` Performance evaluation of affected (API | Script | Download)
    - `N/A` Before / After data comparison

**Area for explaining above N/A when needed:**
```
contract update only
```

PR #3091, opened 2021-04-15
[DEV-7131] Update obligations on bugetary_resources endpoint

emily-brents:
**Description:**
Adds a breakdown of obligations by period to the agency `budgetary_resources` endpoint. API contract was updated in https://github.com/fedspendingtransparency/usaspending-api/pull/3085

**Technical details:**
Pulls the `obligations_incurred_total_by_tas_cpe` per period from File A / appropriation_account_balances table.

    - [ ] Backend
8. [X] Jira Ticket [DEV-7131](https://federal-spending-transparency.atlassian.net/browse/DEV-7131):

```
```


usaspending_api/agency/v2/views/budgetary_resources.py
```diff
@@ -66,6 +85,7 @@ def get_agency_budgetary_resources(self):
                 "agency_budgetary_resources": x["agency_budgetary_resources"],
                 "agency_total_obligated": x["agency_total_obligated"],
                 "total_budgetary_resources": resources.get(x["submission__reporting_fiscal_year"]),
+                "agency_obligation_by_period": self.get_periods_by_year(x["submission__reporting_fiscal_year"]),
```
sethstoudenmier, line 88: This line is resulting in a DB query to get fired off for each FY in addition to the query at the beginning of this method. This endpoint in QAT right now completes in about .5 seconds. My test against QAT had this going up to 2.5 seconds. That isn't terrible, but is a noticeable jump. Is it possible to change this to minimize queries against the DB?

emily-brents: I've pushed some changes that make the query run roughly the same when testing against the QAT db locally (3.45 s v. 3.54 s), but please double check since those are both worse times than you saw while testing

sethstoudenmier (approved):
LGTM. Still a couple second jump for performance, but not terrible and there are now much fewer DB calls.

PR #3096, opened 2021-04-20
Removes extra option on mkdir in README

jspeis:
**Description:**
In this context I don't think you would need -p option on `mkdir` since a nested folder is not being created in the example.

**Technical details:**

N/A

Simple change in README, does not change any code/data/configuration.


kwhickey:
@jspeis thanks for the thought on simplifying this, however the `-p` wasn't so much there for nested, but more if someone had already created a `usaspending` folder or perhaps was running through the steps a second time. The `-p` would allow the step to succeed if the folder was already there; and it doesn't hurt to have it if it's not there.

PR #3115, opened 2021-05-03
[DEV-7191] Sort with nulls at the bottom of results

sethstoudenmier:
**Description:**
Better handling of NULLs in DEV-7191 caused a slight issue with how we were trying to sort results with the Django ORM but then manipulated some of the values to be NULL depending on the value of the publication date.

**Technical details:**
Removed the Django ORM Order By since it is no longer effective at sorting NULL values for our use case in the endpoint. Also adjust the sorting for both Overview endpoints to be consistent with NULLs sorted last. 

The approach to handle the NULLs sorted last is to create a tuple where the first value is True or False depending on the combination of selected sort order and if a field is NULL. The remaining fields of the tuple will sort the non-NULL results correctly.

    - [ ] Backend
8. [x] Jira Ticket [DEV-7191](https://federal-spending-transparency.atlassian.net/browse/DEV-7191):

```
```

### DIFF
--- a/usaspending_api/reporting/tests/integration/test_agency_code_overview.py
+++ b/usaspending_api/reporting/tests/integration/test_agency_code_overview.py
@@ -303,25 +303,6 @@ def test_pagination(setup_test_data, client):
     assert len(response["results"]) == 3
     expected_results = [
         {
-            "fiscal_year": 2019,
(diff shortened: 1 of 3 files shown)


collinwr (approved):
Approved. Cool tuple logic in the sort

PR #3120, opened 2021-05-10
[DEV-7281] Makes FK on historical_appropriation_account_balances casc…

collinwr:
**Description:**
DEV-7189 involves updating toptier_codes for agencies. Because `historical_appropriation_account_balances` has a foreign key to `toptier_agency`, errors will occur if a previously existing toptier_code is removed. This PR updates the `on_delete` behavior of the FK to 'CASCADE` so this migration can take place error free. 

**Technical details:**
`historical_appropriation_account_balances` does not look like it is currently being used in the codebase. 

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-7281](https://federal-spending-transparency.atlassian.net/browse/DEV-7281):

```
```


usaspending_api/accounts/models/historical_appropriation_account_balances.py
```diff
@@ -55,7 +55,7 @@ class HistoricalAppropriationAccountBalances(models.Model):
     reporting_fiscal_quarter = models.IntegerField(blank=True, null=True)
     reporting_fiscal_period = models.IntegerField(blank=True, null=True)
 
-    owning_toptier_agency = models.ForeignKey("references.ToptierAgency", models.DO_NOTHING, null=True)
+    owning_toptier_agency = models.ForeignKey("references.ToptierAgency", models.CASCADE, null=True)
```
sethstoudenmier, line 58: Concerned that making this change will result in data being silently deleted from this table in the future, should we decide to make use of it. My understanding is that this case does not occur very often and could be handled manually each time or via something such as `models.SET` (https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.SET).

sethstoudenmier: Additionally, this table contains only data from FY14 through FY17. This means it is even more likely to be silently deleted since we would not have a step in our nightly pipeline to refresh it each night.

PR #3122, opened 2021-05-12
[DEV-7197] Spending by Award DEFC filter updates

emily-brents:
**Description:**
Adjusts the values returned by the `spending_by_award` endpoint to filter the Covid-19 related outlays and obligations by the provided DEFC filter.

**Technical details:**
If the COVID-19 Obligations or COVID-19 Outlays values are requested in a call to `spending_by_award`, the values returned will be filtered by the `def_codes` filter (if present). So if Award 1 has $100 in obligations under DEFC "L" and 
 $20 in obligations under DEFC "P", if the def_codes filter contains only "L", the value returned for Award 1's obligations will be $100, not $20.

    - [ ] Backend
8. [X] Jira Ticket [DEV-7197](https://federal-spending-transparency.atlassian.net/browse/DEV-7197):

```
```

### DIFF
--- a/usaspending_api/search/v2/views/spending_by_award.py
+++ b/usaspending_api/search/v2/views/spending_by_award.py
@@ -365,7 +365,7 @@ class SpendingByAwardVisualizationViewSet(APIView):
                 )
             if row.get("def_codes"):
                 if self.filters.get("def_codes"):
-                    row["def_codes"] = filter(lambda x: x in self.filters.get("def_codes"), row["def_codes"])
+                    row["def_codes"] = list(filter(lambda x: x in self.filters.get("def_codes"), row["def_codes"]))
             row["generated_internal_id"] = hit["generated_unique_award_id"]
             row["recipient_id"] = hit.get("recipient_unique_id")
             row["parent_recipient_unique_id"] = hit.get("parent_recipient_unique_id")
--- a/usaspending_api/search/v2/views/spending_by_award.py
+++ b/usaspending_api/search/v2/views/spending_by_award.py
@@ -363,6 +363,9 @@ class SpendingByAwardVisualizationViewSet(APIView):
                         for x in row.get("COVID-19 Outlays")
                     ]
                 )
+            if row.get("def_codes"):
+                if self.filters.get("def_codes"):
+                    row["def_codes"] = filter(lambda x: x in self.filters.get("def_codes"), row["def_codes"])
             row["generated_internal_id"] = hit["generated_unique_award_id"]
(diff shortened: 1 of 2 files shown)


usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -343,6 +343,29 @@ def construct_es_response_for_prime_awards(self, response) -> dict:
...
+                        x["outlay"]
+                        if (self.filters.get("def_codes") is not None and x["defc"] in self.filters["def_codes"])
+                        or self.filters.get("def_codes") is None
+                        else 0
+                        for x in row.get("COVID-19 Outlays")
+                    ]
+                )
+            if row.get("def_codes"):
+                if self.filters.get("def_codes"):
+                    row["def_codes"] = filter(lambda x: x in self.filters.get("def_codes"), row["def_codes"])
```
sethstoudenmier, line 368: You might need to wrap the `filter()` in `list()` because testing locally it appears that DRF is representing a filter object as an empty list.

sethstoudenmier: Also, do we want to limit the DEFC returned to only those that are filtered on? Realized while working on the download counterpart that it wasn't mentioned in any AC and keep going back and forth on it. Originally was for limiting the DEFC shown to only those that are filtered to, but could come across as hiding information. My thinking is that Outlay and Obligation are different since on that page we don't list everything out, only a single number (unlike the list of DEFC next to it).

sethstoudenmier (changes requested):
Obligations and Outlays appear to work as expected. The changes to `def_codes` in the response are always returning an empty list. I do wonder if we even want the changes to the `def_codes` returned though.

usaspending_api/search/v2/views/spending_by_award.py
emily-brents: Initially I was not going. to filter the DEFC array, but based on [this](https://data-transparency-bfs.slack.com/archives/C0122HUCD17/p1620831604086100) conversation with Beth, it seems that the PO wants the array to be filtered as well. I also am not seeing the bug you're talking about with the `def_codes` array, but I'll try and look into it some more.

sethstoudenmier: >Initially I was not going. to filter the DEFC array, but based on this conversation with Beth, it seems that the PO wants the array to be filtered as well

Gotcha. I will make the change to the download as well to reflect this change.

sethstoudenmier (approved):

PR #3143, opened 2021-05-26
[DEV-7118] Use external load dates to determine cutoff for download cache

sethstoudenmier:
**Description:**
Download "cache" is actually a reference to the DownloadJob table and the download .zip folder in S3. Previously the download was using a cutoff date of midnight every night, meaning any download after midnight would be a new download. Anything prior to midnight would first look to see if it had been previously run.

**Technical details:**
Use a combination of FPDS, FABS, ES Transactions, and ES Awards external load dates paired with the latest submission's reveal date. The goal is to have data as fresh as possible while still keeping the cache functionality. The download cache could be more granular but that would require adding a lot more external load dates to map each stage of the pipeline with a download.

The approach taken will compare the `update_date` of a DownloadJob to the most recent and relevant External Load Date. The "relevant" part comes into play only for Elasticsearch based downloads (Advanced Search Awards and Transactions) because we do not update the Elasticsearch Indexes until the end of the pipeline. 

This approach does mean the cache will be "cleared" multiple times through the pipeline, but it will help to guarantee the most fresh data possible and also help to prevent someone from downloading data that does not match the site. 

    - [ ] Backend
8. [x] Jira Ticket [DEV-7118](https://federal-spending-transparency.atlassian.net/browse/DEV-7118):

```
```

### DIFF
--- a/usaspending_api/download/v2/base_download_viewset.py
+++ b/usaspending_api/download/v2/base_download_viewset.py
@@ -35,7 +35,6 @@ class BaseDownloadViewSet(APIView):
         special_request_type: Optional[str] = None,
         validator_type: Optional[Type[DownloadValidatorBase]] = None,
     ):
-        # commit to refresh codeclimate
         if special_request_type == "disaster":
             filename = (
                 DownloadJob.objects.filter(
--- a/usaspending_api/download/v2/base_download_viewset.py
+++ b/usaspending_api/download/v2/base_download_viewset.py
@@ -147,11 +147,12 @@ class BaseDownloadViewSet(APIView):
         ).aggregate(Max("last_load_date"))["last_load_date__max"]
 
         # Conditional put in place for local development where the external dates may not be defined
+        cached_download = None
         if updated_date_timestamp:
             recent_submission_window_date = DABSSubmissionWindowSchedule.objects.filter(
                 submission_reveal_date__lt=datetime.max.replace(tzinfo=timezone.utc)
             ).aggregate(Max("submission_reveal_date"))["submission_reveal_date__max"]
-            return (
+            cached_download = (
                 DownloadJob.objects.filter(
                     json_request=ordered_json_request,
                     update_date__gte=max(updated_date_timestamp, recent_submission_window_date),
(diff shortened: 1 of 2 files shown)


usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -53,14 +57,7 @@ def post(
         ordered_json_request = json.dumps(json_request)
 
         # Check if the same request has been called today
-        # TODO!!! Use external_data_load_date to determine data freshness
-        updated_date_timestamp = datetime.strftime(datetime.now(timezone.utc), "%Y-%m-%d")
-        cached_download = (
-            DownloadJob.objects.filter(json_request=ordered_json_request, update_date__gte=updated_date_timestamp)
```
collinwr, line 59: How was this ever true? When would the last update_date ever be greater than the current time?

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -125,6 +122,47 @@ def _get_status_url(self, file_name: str) -> str:
...
+            id for name, id in EXTERNAL_DATA_TYPE_DICT.items() if name in external_data_type_name_list
+        ]
+
+        # Clear the download "cache" based on most recent date in the list of relevant ExternalDataLoadDate objects
+        updated_date_timestamp = ExternalDataLoadDate.objects.filter(
+            external_data_type_id__in=external_data_type_id_list
+        ).aggregate(Max("last_load_date"))["last_load_date__max"]
+
+        # Conditional put in place for local development where the external dates may not be defined
+        if updated_date_timestamp:
```
collinwr, line 150: Minor request: Can you take the return out of this condition by setting a default value then updating it in the condition?

collinwr:
Cool approach here Seth. I don't see any problems!

usaspending_api/download/v2/base_download_viewset.py
sethstoudenmier: `updated_date_timestamp` would have had the value of today at midnight. If a download was downloaded after midnight then the `update_date` would be greater than midnight since it is a DateTimeField.

PR #3158, opened 2021-06-09
[DEV-7339] Improve performance of Custom Downloads

sethstoudenmier:
**Description:**
Custom Downloads have grown a lot over the years and the resulting Django ORM built query is not performant. This PR leverages a combination of SQL Views and the Django ORM to help areas that are affecting performance the most without having to do a complete overhaul of the Django ORM driven approach.

**Technical details:**
Currently the Custom Account Downloads are based off of the File A, B, and C tables. Many of the fields that we need for the download are able to be easily acquired through Django ORM query paths and annotations. However, there are some areas that Django is not handling very well. In an attempt to increase performance of current state of the downloads and also provide freedom to use more SQL as needed in the future the Custom Account Downloads are now based off of SQL Views. 

These Views inherit all properties of the tables that were previously used to allow direct support of previous functionality, while making changes for two fields (CGAC related fields) that were previously giving the downloads an issue. On average the EXPLAIN PLAN for the New vs Old is about 1/3rd the cost. 

TODO:
* Test in Sandbox to make sure no functionality was affected
* ~Look at applying this pattern to the Custom Award Download since it has potential to have worse performance than Custom Account~ Going to handle this in another PR

**Unfortunate Oddities:**

I noticed while testing that a `SELECT *` in a SQL View will not actually store the query as a `SELECT *` but instead as the fields on the table at the time of the query. For production use this is not a concern since we will re-build the views as need be. However, for tests this can be a problem since you could be missing fields from the views if the `0004...` Download migration is run prior to other migrations that would add fields to the base tables of the different views.

As of right now I am unsure of the best way around this other than updating the dependencies of the migration as we add more fields.

    - [ ] Backend
8. [ ] Jira Ticket [DEV-7339](https://federal-spending-transparency.atlassian.net/browse/DEV-7339):

```
```


usaspending_api/accounts/models/appropriation_account_balances.py
```diff
@@ -56,6 +41,33 @@ class AppropriationAccountBalances(DataSourceTrackedModel):
...
     final_of_fy = models.BooleanField(blank=False, null=False, default=False, db_index=True)
 
+    class Meta:
+        abstract = True
+
+
+class AppropriationAccountBalancesManager(models.Manager):
+    def get_queryset(self):
+        """ Get only records from the last submission per TAS per fiscal year. """
+        return super(AppropriationAccountBalancesManager, self).get_queryset().filter(final_of_fy=True)
```
kwhickey, line 51: This may look like new code because of refactoring ... but, did we always have this Manager class, and was it always defaulting to have a filter that forces `final_of_fy=True`. I thought we had plans to actually remove `final_of_fy` or stop using it on any A/B/C files. 

I'd be interested in what code was using this Model manager. Perhaps _that_ code is very aware of the filter and taking advantage of it.

usaspending_api/accounts/v2/filters/account_download.py
```diff
@@ -187,10 +186,6 @@ def generate_gross_outlay_amount_derived_field(filters, account_type):
 def generate_treasury_account_query(queryset, account_type, tas_id, filters):
     """ Derive necessary fields for a treasury account-grouped query """
     derived_fields = {
-        "allocation_transfer_agency_identifier_name": get_agency_name_annotation(
```
kwhickey, line 190: Looks like removal of `get_agency_name_annotation` what was the reason and benefit for this? And what goes in its place if those fields are still needed?

kwhickey: I think i see now -- this was an artifact of needing to do a subquery in django to join `cgac` table?

usaspending_api/download/migrations/0004_download_views.py
```diff
@@ -0,0 +1,53 @@
...
+            name="AccountBalancesView",
+            fields=[],
+            options={
+                "db_table": "vw_account_balances",
+                "managed": False
+            }
+        ),
+        migrations.RunSQL(
+            sql=[f"{Path('usaspending_api/download/sql/account_balances.sql').read_text()}"],
+            reverse_sql=["DROP VIEW IF EXISTS vw_account_balances;"],
```
kwhickey, line 27: Oooo. I really like this. I didn't know you could just point django migrations at a sql file. Clean way to version control stuff.

kwhickey: Oh, you can't do that, but are just reading a file. Clever.

usaspending_api/download/sql/account_balances.sql
```diff
@@ -0,0 +1,12 @@
+DROP VIEW IF EXISTS vw_account_balances;
+CREATE VIEW vw_account_balances AS
+SELECT
+    "appropriation_account_balances".*,
```
kwhickey, line 4: I'm sure you probably copy-pasted from django output for getting this up quickly, but it'd be clean and nice to get rid of the double quotes. Optional to alias the tables if you'd like. This and the other views.

usaspending_api/financial_activities/models.py
```diff
@@ -78,6 +65,27 @@ class FinancialAccountsByProgramActivityObjectClass(DataSourceTrackedModel):
...
+        abstract = True
+
+
+class FinancialAccountsByProgramActivityObjectClassManager(models.Manager):
+    def get_queryset(self):
+        """
+        Get only records from the last submission per TAS per fiscal year.
+        """
+
+        return super(FinancialAccountsByProgramActivityObjectClassManager, self).get_queryset().filter(final_of_fy=True)
```
kwhickey, line 78: same comment as with the AppropriationsAccountBalances manager.

usaspending_api/download/models/award_financial.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
+
+from usaspending_api.awards.models import AbstractFinancialAccountsByAwards
+
+
+class AwardFinancialView(AbstractFinancialAccountsByAwards):
```
kwhickey, line 6: One question is the names of the views.
I think i remember seeing a comment somewhere saying "one day, we'll finally get all these dabs tables/models named with the same convention" and i think it was referring to the `FinancialAccountsBy` convention `financial_accounts_by_`
And maybe the holdover to fix was file A.
So maybe think about changing the view model and view names to all start with `FinancialAccountsBy...` and `vw_financial_accounts_by_` resp.

And then for the views, keeping convention of the vw_ prefix. But differentiating the fact that these are download-specific views with a suffix
`vw_financial_accounts_by_awards_downloads` or `vw_financial_accounts_by_awards_dl`

(if there's a strong argument to lean back towards the broker naming convention for these tables -- i'm good to hear that too)

usaspending_api/download/sql/award_financial.sql
```diff
@@ -0,0 +1,12 @@
...
+CREATE VIEW vw_award_financial AS
+SELECT
+    "financial_accounts_by_awards".*,
+	"cgac_aid"."agency_name" AS "agency_identifier_name",
+	"cgac_ata"."agency_name" AS "allocation_transfer_agency_identifier_name"
+FROM "financial_accounts_by_awards"
+INNER JOIN "submission_attributes" ON ("financial_accounts_by_awards"."submission_id" = "submission_attributes"."submission_id")
+LEFT OUTER JOIN "treasury_appropriation_account" ON ("financial_accounts_by_awards"."treasury_account_id" = "treasury_appropriation_account"."treasury_account_identifier")
+LEFT OUTER JOIN "cgac" AS "cgac_aid" on ("treasury_appropriation_account"."agency_id" = "cgac_aid"."cgac_code")
+LEFT OUTER JOIN "cgac" AS "cgac_ata" on ("treasury_appropriation_account"."allocation_transfer_agency_id" = "cgac_ata"."cgac_code")
```
kwhickey, line 11: I guess i'm thinking -- could there be more tucked into this sql to take some of the work off the Django ORM and make the download queries even faster? 

I see now that this code was all aimed at fixing that `cgac` join, but this does provide a good clean foundation for continuing to chip away at the ugly ORM-produced-sql, and instead put cleaner and more efficient starter-sql in here.

kwhickey: After seeing the SQL though, i see these don't go too deep into download-specific sql (e.g. joining awards and such). Should they? 
If they will head there, i'd say suffix with `downloads` or `dl`. If they will stay a minimally augmented version of the file A / B / C tables, then maybe leave them named as is.

kwhickey: Another thought that comes to mind if it was just to augment the data with `cgac` fields, could we have just done that at ETL-time, when pulling this data over from broker (join in cgac where it exists, and update these tables as that data changes?)

kwhickey (approved):
Looks really good and clean as-is. A smart mix of orm for usability and sql for directness and efficiency. 

A handful of questions embedded about whether the sql goes far enough, and the choices of names used.

usaspending_api/accounts/models/appropriation_account_balances.py
sethstoudenmier: This is old code. I only see the `final_objects` referenced in one place (outside of definition). It could probably be removed, but I tried to limit changes in this PR.

usaspending_api/accounts/v2/filters/account_download.py
sethstoudenmier: > this was an artifact of needing to do a subquery in django to join cgac table?

Correct. It was performing the CGAC subquery and is no longer needed now that we handle that in SQL.

usaspending_api/download/sql/account_balances.sql
sethstoudenmier: Oh 100% a copy-paste and cleanup. Will get this cleaned up.

usaspending_api/download/models/award_financial.py
sethstoudenmier: Can get those changed over. Wanted the views to match the table they are based off of, but if we are planning to eventually change those tables then I am happy to make this change now for the new Views..

usaspending_api/download/sql/award_financial.sql
sethstoudenmier: > I guess i'm thinking -- could there be more tucked into this sql to take some of the work off the Django ORM and make the download queries even faster?

Was hoping to add more as well. Minimal "extra" fields on the View will help to reduce data changes for downloads in the future (except in terrible performance scenarios). Additionally, the other parts that aren't a straight `SELECT` (e.g. `CASE` statements) are easily done with Django or they need to be dynamically added based on filters (e.g. the `gross_outlay_amount_by_award_cpe` field).

> If they will head there, i'd say suffix with downloads or dl

Good call, will add that to the naming convention. 

>Another thought that comes to mind if it was just to augment the data with cgac fields, could we have just done that at ETL-time, when pulling this data over from broker (join in cgac where it exists, and update these tables as that data changes?)

Another possible route. Personally more of a fan for trying to keep some USAspending tables closer to their Broker counter part.

usaspending_api/submissions/helpers.py
```diff
@@ -132,8 +131,12 @@ def get_last_closed_periods_per_year() -> List[ClosedPeriod]:
                             submission_reveal_date <= now()
                     order   by submission_fiscal_year, -submission_fiscal_month
                 ) as m on m.submission_fiscal_year = q.submission_fiscal_year
+        where   q.submission_fiscal_year = %(fiscal_year)s
```
kwhickey, line 134: I'm concerned about this query a bit. It seems a bit too smart for its own good. 

In the case where it is just monthly period 2 or monthly periods 2 and 3 that have closed, but the first quarer, Q1, has not yet closed, this query returns no results. 
However it should somehow get the latest monthly period that has closed, yes?

Run these statements in sequence in sandbox to see what I mean (don't forget to rollback!)
```sql
begin;
update dabs_submission_window_schedule set submission_reveal_date = '9999-12-31 23:59:59.999999+00' where id > 2021030;
select * from dabs_submission_window_schedule order by submission_reveal_date desc, id desc;
select  coalesce(q.submission_fiscal_year, m.submission_fiscal_year) as fiscal_year,
        q.submission_fiscal_quarter as fiscal_quarter,
        m.submission_fiscal_month as fiscal_month
from    (
            select  distinct on (submission_fiscal_year)
                    submission_fiscal_year, submission_fiscal_quarter
            from    dabs_submission_window_schedule
            where   is_quarter is true and
                    submission_reveal_date <= now()
            order   by submission_fiscal_year, -submission_fiscal_quarter
        ) as q
        full outer join (
            select  distinct on (submission_fiscal_year)
                    submission_fiscal_year, submission_fiscal_month
            from    dabs_submission_window_schedule
            where   is_quarter is false and
                    submission_reveal_date <= now()
            order   by submission_fiscal_year, -submission_fiscal_month
        ) as m on m.submission_fiscal_year = q.submission_fiscal_year
where   q.submission_fiscal_year = 2021
;
rollback;
```

What it used to return before the FY where clause added
image

But what it returns in the case of the above data and query:
image

I think the `WHERE` needs to be moved outside the full outer join, as applied on the full results of the query as if it were a subquery. 

But then I worry about other times of the year and if that join on `m.submission_fiscal_year = q.submission_fiscal_year` might be incorrectly joining quarterly and monthly data. That may be ok though -- needs further tests at various times during the year (like when P12 has closed but no yet Q4...).

kwhickey: Seems to work okay as long as you apply the year filtering as an outer-query. Alternatively, you could just get all years, and then filter the handful of results in python, to pick out the year you're interested in.

kwhickey (changes requested):
One major issue to address with filtering the `get_last_closed_periods_for_year(fiscal_year: int)`. 

Then just a thought on whether to default to CASCADE in the Abstract, and override in the view.

usaspending_api/awards/models/financial_accounts_by_awards.py
```diff
@@ -129,6 +129,17 @@ class FinancialAccountsByAwards(DataSourceTrackedModel):
...
+    class Meta:
+        abstract = True
+
+
+class FinancialAccountsByAwards(AbstractFinancialAccountsByAwards):
+    # Overriding attributes from the Abstract Fields;
+    # This needs to occur primarily for the values of "on_delete" and "related_name"
+    award = models.ForeignKey("awards.Award", models.CASCADE, null=True, related_name="financial_set")
+    treasury_account = models.ForeignKey("accounts.TreasuryAppropriationAccount", models.CASCADE, null=True)
+    submission = models.ForeignKey("submissions.SubmissionAttributes", models.CASCADE)
```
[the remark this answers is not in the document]
sethstoudenmier: Good call, will make the change.

usaspending_api/submissions/helpers.py
sethstoudenmier: After looking closer I don't think that SQL is needed at all. New change removes that function and creates a ClosedPeriod object with the filters provided. This is closer to the previous functionality but was overlooked during initial changes. The main difference between old vs new is that the new code is not going through all closed periods since they really don't matter, only the period we are searching for.

usaspending_api/accounts/v2/filters/account_download.py
```diff
@@ -95,29 +97,35 @@ def account_download_filter(account_type, download_table, filters, account_level
...
-    return queryset.filter(submission_filter, **query_filters)
+    return queryset.filter(submission_filter, nonzero_filter, **query_filters)
 
 
 def get_submission_filter(account_type, filters):
-
+    """
+    Limits the overall File A, B, and C submissions that are looked at. For File A and B we only
+    look at the most recent submissions for the provided filters. For File C we expand this to
+    include all submissions up to the provided filters.
```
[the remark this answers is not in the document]
sethstoudenmier: Will change this out

kwhickey (approved):
Approved! Looks good! Ready for Testing

PR #3173, opened 2021-06-30
[DEV-7480] Increase performance of Custom Award Download

sethstoudenmier:
**Description:**
Improve Custom Award Download by reducing the number of expensive, static subqueries that are performed for every row. This creates a larger overhead, but overall a lower total cost and runtime.

**Technical details:**

Top line from `EXPLAIN ANALYZE`
```
OLD --> Gather  (cost=1,002.27..4,657,195,474.66 rows=17511656 width=1348) (actual time=77.034..5,685,887.869 rows=25,057,142 loops=1)
NEW --> Hash Left Join  (cost=39,045,969.43..4,346,625,773.95 rows=41312092 width=1348) (actual time=843,092.686..3,976,798.546 rows=25,057,142 loops=1)
```

Subawards were not adjusted as part of this work. The subaward download is efficient enough that the added overhead would  put it past the worst case scenario that was found.

Some of the changes to remove the `# Award Level` download path were cleanup as that path is no longer valid following the change a while back to use Elasticsearch for Advanced Search downloads. 

    - [ ] Backend
8. [x] Jira Ticket [DEV-7480](https://federal-spending-transparency.atlassian.net/browse/DEV-7480):

```
```


usaspending_api/conftest.py
```diff
@@ -67,6 +72,23 @@ def add_view_protection():
...
+    """
+    In some cases the download views are based off of either Tables or Materialized Views that are converted to
+    Views for ease of use with PyTest. In these cases the download view is dropped prior to creation of the base
+    Table/Materialized View/View. This function ensures that all download Views exist in case they were dropped.
+    """
+    try:
+        download_view_dir = Path("usaspending_api/download/sql")
+        download_view_files = [download_view_dir / file_name for file_name in os.listdir(download_view_dir)]
+        sql = ""
+        for view in download_view_files:
```
collinwr, line 85: Should the exception handling take place within the for loop here, so if one view already exists and throws and exception, later views will still be created?

sethstoudenmier: It certainly could, but it could also be removed in hindsight. The SQL that it is running `DROP IF EXISTS` the view and then `CREATE`.

usaspending_api/download/migrations/0004_download_views.py
```diff
@@ -12,7 +12,9 @@ class Migration(migrations.Migration):
         ("accounts", "0002_auto_20200320_1307"),
         ("awards", "0074_financialaccountsbyawards_distinct_award_key"),
         ("download", "0003_auto_20180306_1726"),
-        ("financial_activities", "0003_financialaccountsbyprogramactivityobjectclass_disaster_emergency_fund")
+        ("financial_activities", "0003_financialaccountsbyprogramactivityobjectclass_disaster_emergency_fund"),
```
collinwr, line 15: Why is this migration being updated instead of adding a new one? Won't this migration already be marked as having run?

sethstoudenmier: Correct. Realistically the migration is put in place for the sake of tests and the model. The actual SQL run to create the views can be run manually. For example, if in the future we need to add another field to the SQL View then it wouldn't make sense to add another migration. Instead we would add a field to the model, the View, and then re-run the SQL that DROPs and CREATEs the View.

Benefit of this approach is that all Download SQL View migrations are in one place.

usaspending_api/download/tests/unit/test_file_generation.py
```diff
@@ -5,18 +5,18 @@
...
+def test_get_elasticsearch_awards_csv_sources(db):
+    original = VALUE_MAPPINGS["elasticsearch_awards"]["filter_function"]
+    VALUE_MAPPINGS["elasticsearch_awards"]["filter_function"] = MagicMock(returned_value="")
     csv_sources = download_generation.get_download_sources(
-        {"download_types": ["awards"], "filters": {"award_type_codes": list(award_type_mapping.keys())}}
+        {"download_types": ["elasticsearch_awards"], "filters": {"award_type_codes": list(award_type_mapping.keys())}}
     )
     assert len(csv_sources) == 2
-    VALUE_MAPPINGS["awards"]["filter_function"] = original
+    VALUE_MAPPINGS["elasticsearch_awards"]["filter_function"] = original
```
collinwr, line 15: Do you know why these need to be set back to the original value? For both this method and the one below.

sethstoudenmier: I removed the `awards` path because it was no longer in use. And when doing so I realized we didn't have a test for the `elasticsearch_transactions` path.

usaspending_api/search/models/transaction_search.py
```diff
@@ -2,18 +2,10 @@
...
-    Fields in this model have all, with the exception of primary/foreign keys, been made nullable because it
-    is directly populated by the contents of a materialized view. The fields used to create the materialized view
-    may or may not be nullable, but those constraints are not enforced in this table.
-    """
-
-    transaction = models.OneToOneField(TransactionNormalized, on_delete=models.DO_NOTHING, primary_key=True)
-    award = models.ForeignKey(Award, on_delete=models.DO_NOTHING, null=True)
+class AbstractTransactionSearch(models.Model):
+    transaction = models.OneToOneField("awards.TransactionNormalized", on_delete=models.DO_NOTHING, primary_key=True)
+    award = models.ForeignKey("awards.Award", on_delete=models.DO_NOTHING, null=True)
```
collinwr, line 8: What does changing the syntax for referencing another model do here?

sethstoudenmier: It is called a "lazy reference" for Django. The exact same functionality, but cleans up imports and prevents any cases of circular imports in models.

usaspending_api/conftest.py
collinwr: I'm good with either approach

usaspending_api/download/migrations/0004_download_views.py
collinwr: So will this ticket include a Data Change ticket to run that SQL?

sethstoudenmier: Sure will. Still need to make that, but will include it as a step.

collinwr (approved):
Approved

PR #3177, opened 2021-07-06
DEV-7438: Object Class Updates

dpb-bah:
**Description:**
Updating Object Class loader and File B/C loaders to account for the object class format.

**Technical details:**
* We will be back-cleaning the older Object Class codes in the data, so there's no need to support both the old and new formats moving on.
* New format includes dropping the 4 digit codes starting with 1/2 and adding 4 digit codes that simply add a `0` to the 3 digit equivalents.

    - [x] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
7. [x] Data validation completed
8. [x] Appropriate Operations ticket(s) created
9. [x] Jira Ticket [DEV-7438](https://federal-spending-transparency.atlassian.net/browse/DEV-7438):

```
```


usaspending_api/etl/submission_loader_helpers/object_class.py
```diff
@@ -41,20 +41,19 @@ def get_object_class_row(row):
...
-        # the object class field is the 3 digit version, so grab direct/reimbursable information from a separate field
-        try:
-            direct_reimbursable = ocdr.BY_DIRECT_REIMBURSABLE_FUN_MAPPING[row.by_direct_reimbursable_fun]
-        except KeyError:
-            # So Broker sort of validates this data, but not really.  It warns submitters that their data
-            # is bad but doesn't force them to actually fix it.  As such, we are going to just ignore
-            # anything we do not recognize.  Terrible solution, but it's what we've been doing to date
-            # and I don't have a better one.
-            direct_reimbursable = None
+    if len(object_class) == 4 and object_class[3] == "0":
```
sethstoudenmier, line 44: Should any additional handling be put in place in case there is an Object Class of length 4 that doesn't end with "0"? Right now if that case ever happened to occur it seems it would make it through without an issue.

sethstoudenmier (commented):
One minor question, aside from that code looks good and object_class loader appears to work as expected with the new file format. 

After Travis tests are happy I think this is good to go

usaspending_api/etl/submission_loader_helpers/object_class.py
dpb-bah: That makes sense. I can account for that.

usaspending_api/references/management/commands/load_object_classes.py
```diff
@@ -14,7 +14,7 @@
 from usaspending_api.references.models import ObjectClass
 
 
-OBJECT_CLASS_PATTERN = re.compile("[12]?[0-9]{3}")
+OBJECT_CLASS_PATTERN = re.compile("^[0-9]{3}0?$")
```
sethstoudenmier, line 17: Good idea to add the starting and ending regex

sethstoudenmier (approved):
LGTM. Good to merge assuming Travis passes

sethstoudenmier (approved):

PR #3182, opened 2021-07-09
[DEV-7480] Ensure transaction download view is updated with table

sethstoudenmier:
**Description:**
Need to ensure that the new download view for Transaction based downloads is up-to-date after matview recreate.

**Technical details:**
An oversight on my part resulted in the download view for transactions being dropped during a matview recreate since we swap in a new `transaction_search` table. This change will recreate the `vw_transaction_seach_download` View whenever the `transaction_search` table is swapped out to make sure it is up to date.

    - [ ] Backend
8. [x] Jira Ticket [DEV-7480](https://federal-spending-transparency.atlassian.net/browse/DEV-7480):

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/combine_transaction_search_chunks.py
+++ b/usaspending_api/etl/management/commands/combine_transaction_search_chunks.py
@@ -195,6 +195,7 @@ class Command(BaseCommand):
     def swap_tables(self, rename_indexes, rename_constraints):
 
         swap_sql = (self.matview_dir / "componentized" / f"{TABLE_NAME}__renames.sql").read_text()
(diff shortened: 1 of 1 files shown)


collinwr:
I think this new SQL is only necessary when new migrations/changes are applied to the Django model. I can’t think of a reason why it would hurt though

collinwr:
> I think this new SQL is only necessary when new migrations/changes are applied to the Django model. I can’t think of a reason why it would hurt though

After speaking with Seth, I realize that it is necessary because the table the view is based off of is being recreated.

collinwr (approved):

PR #3186, opened 2021-07-13
Update gitignore to include matview folders

sethstoudenmier:
**Description:**
Updated the `.gitignore` to include the matview directories that are generated as part of the `matvier_runner`.


emily-brents (approved):
A good idea

PR #3187, opened 2021-07-15
[OPS-1881] Set the work-mem for download

sethstoudenmier:
**Description:**
Set the `work_mem` so that more memory is used by the downloader instead of relying so much on writing to disk.

**Technical details:**
Set the `work_mem` for each download DB call so that we can leverage more memory. Currently the downloader does not use much memory compared to what the RDS instance has.

    - [ ] Backend

```
```


emily-brents (approved):
LGTM

usaspending_api/settings.py
```diff
@@ -26,6 +26,9 @@
 DOWNLOAD_DB_TIMEOUT_IN_HOURS = 4
 CONNECTION_MAX_SECONDS = 10
 
+# Value is set per environment using formula: ( TOTAL_MEMORY * 0.25 ) / MAX_CONNECTIONS
```
kwhickey, line 29: Could you add an extra line that states that `MAX_CONNECTIONS` here is the max connections that the DB serving downloads is expected to see (which may be different than the DB serving api requests) @sethstoudenmier

PR #3188, opened 2021-07-16
[DEV-7579] Adds column definition to TS inserts

collinwr:
**Description:**
This PR updates the SQL statements used to insert records from `transaction_search_*` matviews into the `transaction_search` table. It uses the column definition from the `TransactionSearch` model to specify a list of columns to insert.

**Technical details:**
This PR breaks the pattern of generating all SQL for chunked matviews in the `chunked_matview_sql_generator.py` script. This is because there isn't a great way to get a list of columns from that script. Here are the ideas that I had to do so along with why they won't work. 
- **Get the list of columns from `transaction_search.json`** - Contrary to what I originally remembered, the JSON doesn't have a good list of columns. The file contains a list of lines of the table DDL, but this would need complicated parsing logic in order to get the list of columns.
- **Get the list of columns from the TransactionSearch model (liked implemented in the combine command)** - This script is separate from our Django project, so models cannot be accessed. This script would need to be converted to a Django management command to access the models, which would require major a refactor.
- **Query the columns from the Database** - This method faces a similar issue to the last. Because the script is not part of the Django project, creating a connection to the database would not be easy.

**NOTE: A change `daims_map.py` was included to fix a codeclimate error.**

    - [ ] Backend
8. [x] Jira Ticket [DEV-7579](https://federal-spending-transparency.atlassian.net/browse/DEV-7579):

```
```

### DIFF
--- a/usaspending_api/data/daims_maps.py
+++ b/usaspending_api/data/daims_maps.py
@@ -24,6 +24,29 @@ cases. These must be numbered from 1, and store conditions in the "case_<INT>" a
 For an example, see TransactionNormalized.action_type_map
 """
 
+change_reasons = {
+    "A": "Additional Work (new agreement, FAR part 6 applies)",
+    "B": "Supplemental Agreement for work within scope",
+    "C": "Funding Only Action",
+    "D": "Change Order",
+    "E": "Terminate for Default (complete or partial)",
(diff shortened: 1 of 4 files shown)


sethstoudenmier (approved):
Run on QAT looks good and I don't see any issues with code.

PR #3190, opened 2021-07-19
[DEV-7480] Revert changes

sethstoudenmier:
**Description:**
This reverts changes from DEV-7480. Testing in Staging and NonProd environments showed that this could help performance, but after living in Production for about a week it has not helped as expected and possibly hurt performance. Best guess as of right now is that the initial overhead of the new approach hurts when a large amount of downloads are queued up vs a few for testing. 

**Technical details:**
Reverting all changes from DEV-7480 except for tests that were added since some of the tests were simply added for better coverage of downloads, not necessarily pertaining to these changes.

    - [ ] Backend
8. [ ] Jira Ticket [DEV-7480](https://federal-spending-transparency.atlassian.net/browse/DEV-7480):

```
```


usaspending_api/download/lookups.py
```diff
@@ -77,19 +76,19 @@
     # Transaction Level
```
collinwr, line 76: I noticed that "awards" is not being added back to the `VALUE_MAPPINGS` was it originally removed because it was no longer being used even before the download changes?

usaspending_api/download/migrations/0004_download_views.py
```diff
@@ -54,16 +54,4 @@ class Migration(migrations.Migration):
             sql=[f"{Path('usaspending_api/download/sql/vw_financial_accounts_by_program_activity_object_class_download.sql').read_text()}"],
```
collinwr, line 54: Do the new dependencies added by the previous PR need to stay? (I couldn't comment on unchanged lines)

usaspending_api/download/lookups.py
collinwr: I just remembered the conversation we had about this in the last PR. I think we're good here.

collinwr:
Looks good to me. I will approve once some additional testing has been done tomorrow.

usaspending_api/download/lookups.py
sethstoudenmier: Yeh it was no longer being used.

usaspending_api/download/migrations/0004_download_views.py
sethstoudenmier: Can get that updated as well. Doesn't case any harm to leave it, but should probably remove it since the dependencies aren't actually needed.

collinwr (approved):

PR #3193, opened 2021-07-20
[DEV-7271] add duns uei

emily-brents:
**Description:**
Adds the `uei` and `ultimate_parent_uei` to the `duns` table.

**Technical details:**
Also adjusts the `recipient_hash` generation to default to the uei field (if present). 

    - [ ] Backend
8. [X] Jira Ticket [DEV-7271](https://federal-spending-transparency.atlassian.net/browse/DEV-7271):

```
```


usaspending_api/recipient/management/sql/restock_summary_award_recipient.sql
```diff
@@ -11,7 +11,10 @@ UPDATE public.summary_award_recipient AS sar SET
   action_date = a.date_signed,
   recipient_hash = MD5(UPPER(
     CASE
-      WHEN COALESCE(fpds.awardee_or_recipient_uniqu, fabs.awardee_or_recipient_uniqu) IS NOT NULL THEN CONCAT('duns-', COALESCE(fpds.awardee_or_recipient_uniqu, fabs.awardee_or_recipient_uniqu))
+      WHEN COALESCE(fpds.awardee_or_recipient_uei, fabs.awardee_or_recipient_uei) IS NOT NULL
+        THEN CONCAT('uei-', COALESCE(fpds.awardee_or_recipient_uei, awardee_or_recipient_uei))
```
collinwr, line 15: Should `awardee_or_recipient_uei` be `fabs.awardee_or_recipient_uei`?

collinwr:
Looks like there are some `recipient_hash` calculations being done in the matviews. These should be updated as well.

usaspending_api/recipient/management/sql/recipient_lookup/050_sam_missing_name.sql
```diff
@@ -19,7 +19,7 @@ INSERT INTO public.temporary_restock_recipient_lookup (
 )
 SELECT
   DISTINCT ON (awardee_or_recipient_uniqu, legal_business_name)
-  MD5(UPPER(CONCAT('duns-', awardee_or_recipient_uniqu)))::uuid AS recipient_hash,
+  MD5(COALESCE(UPPER(CONCAT('uei-', uei)), UPPER(CONCAT('duns-', awardee_or_recipient_uniqu))))::uuid AS recipient_hash,
```
kwhickey, line 22: Use of `CONCAT` with a text and null value will always result in non-null, so you'll get `uei-` as the value and it will never use the `awardee_or_recipient_uniqu` in case of a null `uei`. See if you can figure out how to move the coalesce inward, to work on the individual fields and then do the prefix and MD5 on the result of coalesce.

image

usaspending_api/recipient/management/sql/recipient_lookup/030_sam_parents.sql
```diff
@@ -10,7 +10,7 @@ INSERT INTO public.temporary_restock_recipient_lookup (
 )
 SELECT
   DISTINCT ON (ultimate_parent_unique_ide, ultimate_parent_legal_enti)
-  MD5(UPPER(CONCAT('duns-', ultimate_parent_unique_ide)))::uuid AS recipient_hash,
+  MD5(COALESCE(UPPER(CONCAT('uei-', ultimate_parent_uei)), UPPER(CONCAT('duns-', ultimate_parent_unique_ide))))::uuid AS recipient_hash,
```
kwhickey, line 13: Same comment as below (above?) about `CONCAT` always yielding a non-null result.

usaspending_api/recipient/management/sql/recipient_lookup/030_sam_parents.sql
```diff
@@ -10,7 +10,7 @@ INSERT INTO public.temporary_restock_recipient_lookup (
 )
 SELECT
   DISTINCT ON (ultimate_parent_unique_ide, ultimate_parent_legal_enti)
```
kwhickey, line 12: Can you add a bugfix here while you are changing things:
```
      -- BUG fixed here: 
      --   - parent DUNS ID (value of ultimate_parent_unique_ide) can show up with the same value on many rows
      --   - on those rows, it could have different (i.e. older) names for itself
      --   - originally this was partitioning on BOTH the parent DUNS and parent name. 
      --   - That may introduce the same DUNS multiple times with different names, which would yield entering the same recipient_hash value into the destination table more than once.
      --   - FIXED: by partitioning only on the parent DUNS. It will take the parent Name from the most recent row
```

i.e. it should be
```
SELECT
  DISTINCT ON (ultimate_parent_unique_ide)
```

Note that this was discovered when prototyping this ETL using `MERGE` and `PARTITION BY`. Replace "partition by" above with "distinct on".

usaspending_api/recipient/management/sql/recipient_lookup/030_sam_parents.sql
```diff
@@ -10,7 +10,7 @@ INSERT INTO public.temporary_restock_recipient_lookup (
...
-  MD5(UPPER(CONCAT('duns-', ultimate_parent_unique_ide)))::uuid AS recipient_hash,
+  MD5(COALESCE(UPPER(CONCAT('uei-', ultimate_parent_uei)), UPPER(CONCAT('duns-', ultimate_parent_unique_ide))))::uuid AS recipient_hash,
   UPPER(ultimate_parent_legal_enti) AS legal_business_name,
   ultimate_parent_unique_ide AS duns,
   'sam-parent' as source,
```
kwhickey, line 16: Also, it's not a bug in this `030` step, but there's a related bug in the `070` step which would be good to fix here if possible:

In `070 Adding Recipient records from SAM parent data without a name`, basically need to add the `AND ultimate_parent_legal_enti IS NULL` to the end of the WHERE clause. 
```
    -- BUG fixed here: originally wasn't pinning ultimate_parent_legal_enti to NULL, which is reintroducing data where both are provided
    WHERE ultimate_parent_unique_ide IS NOT NULL AND ultimate_parent_legal_enti IS NULL
```

usaspending_api/recipient/management/sql/recipient_lookup/030_sam_parents.sql
kwhickey, line 12: Can you also add a bug fix for the `010 Adding Recipient records from SAM` step?

In that .sql script, instead of: 
```
SELECT
  DISTINCT ON (awardee_or_recipient_uniqu, legal_business_name)
```

It should be

```
SELECT
  DISTINCT ON (awardee_or_recipient_uniqu, UPPER(legal_business_name))
```

If not done it will collect the duns number more than once with different casings of the legal_business_name.

usaspending_api/recipient/management/sql/recipient_lookup/030_sam_parents.sql
kwhickey, line 13: I think the `010` step, which build the hash on the DUNS number, and `070` step, which builds the hash on the parent DUNS number are also missing from this PR. 
- To follow this pattern, wouldn't those build it on `uei` if available (for 010), and `ultimate_parent_uei` if available (for `070`)

In general though, i'm not sure if this is the pattern to follow. We may just need to add more steps for uei fields in this layered build-up of recipient records from available hash-worthy fields in an order-of-precedence. -- More in the review comment.

kwhickey (changes requested):
Overall good coverage and pattern following for the most part. But I do see some things missing in the pattern and some fundamental questions to the approach here. 

(1) I think COALESCEing `uei` and `ultimate_parent_uei` where present should be done to follow the patterns in the `010` step and `070` step

(2) In general though, i'm not sure if this is the pattern to follow. We may just need to add more steps for uei fields in this layered build-up of recipient records from available hash-worthy fields in an order-of-precedence. -- More in the review comment. 
- For instance, in the future, when a new recipient comes along, they will have a `NULL` value for the duns number (`awardee_or_recipient_uniqu`) and for the parent duns number. But they may have a non-null UEI value.
- In those cases the windowed "partitions" made by any `DISTINCT ON (... [duns or parent duns]...)` statement is not going to pick out unique records form which to build a `recipient_hash` field... since it'll grab ALL new recipients with a NULL duns. 
- So I think we need to add _NEW_ steps in this order-of-precedence and ensure we account for cases where
    - uei is not null and duns is null
    - uei is not null and duns is not null
    - uei is null and duns is not null
    - the above, plus ... parent uei is [not] null and parent duns is [not] null
- the goal is to make sure if a UEI is there or a parent UEI is there, it takes precedence as input for the recipient_hash over the duns or parent duns, resp.

(3) Then I have a question on what this means for the hash _changing_ for a recipient that didn't change at all except for getting a new uei. I think @collinwr mentioned trying to keep the recipient hash the same, so as not to affect saved/bookmarked URLs? ... at least for those recipients that had a duns and get a new UEI. 
- How is that maintained with this approach, since the coalesce will take the uei to build the `recipient_hash` field rather than the duns?
- Or does the magic happen when the `recipient_profile` record is built?
- This may have been figured out and I just missed the nuance of how it works (again)

emily-brents:
Closing due to UEI being delayed

PR #3220, opened 2021-08-25
[DEV-7773] Fix award summary page download to avoid empty columns

sethstoudenmier:
**Description:**
The Award Summary download is populated outlay fields with `NULL` because there is no FY filter. This fixes that and also makes a correction to Treasury Account downloads.

**Technical details:**
Previously any download without a FY filter was setting outlay fields to be `NULL` not taking into consideration that there is no FY filter for Award Summary page. Adding back logic to handle the absence of a FY filter corrects that issue. Part of that fix is adding back some logic that was originally removed in #3158. However, the new implementation is using Django ORM instead of raw SQL.

An additional adjustment was made for Treasury Account downloads so that the CSV produced will match the `Federal Accounts Funding` section of the Award Summary page. Since there is no `GROUP BY` present for Treasury Account level downloads there isn't a need to have additional CASE statement around which outlay values to use. As a result, that logic was removed for Treasury Account downloads and left in place for Federal Account since it rolls up all TAS into their Federal Account.

    - [ ] Backend
8. [x] Jira Ticket [DEV-7773](https://federal-spending-transparency.atlassian.net/browse/DEV-7773):

```
```


usaspending_api/accounts/v2/filters/account_download.py
```diff
@@ -159,64 +160,75 @@ def get_nonzero_filter():
...
 
-def generate_treasury_account_query(queryset, account_type, filters):
+def generate_treasury_account_query(queryset, account_type):
     """ Derive necessary fields for a treasury account-grouped query """
     derived_fields = {
         "submission_period": get_fyp_or_q_notation("submission"),
-        "gross_outlay_amount": generate_gross_outlay_amount_derived_field(filters, account_type),
-        "gross_outlay_amount_fyb_to_period_end": generate_gross_outlay_amount_derived_field(filters, account_type),
+        "gross_outlay_amount": generate_gross_outlay_amount_derived_field(account_type),
+        "gross_outlay_amount_fyb_to_period_end": generate_gross_outlay_amount_derived_field(account_type),
```
collinwr, line 222: Will `gross_outlay_amount` and `gross_outlay_amount_fyb_to_period_end` always be the same because `generate_gross_outlay_amount_derived_field` is called for both with the same parameter? Same for `generate_federal_account_query`

sethstoudenmier: The `account_balances` download refers to `gross_outlay_amount` and the other two refer to `gross_outlay_amount_fyb_to_period_end`. Both are annotated and then the correct columns are used for each download.

collinwr: Oh, I see what you mean. Gotcha

collinwr (approved):
Code looks good, and a test Award Summary download included correct outlay results. Approved

PR #3222, opened 2021-08-31
[DEV-7831] Refactor /award/amount to use Elasticsearch

sethstoudenmier:
**Description:**
The `/api/v2/disaster/award/amount/` consistently runs over 1 min and occasionally over 5 minutes. This means that it cannot cache and will stay in a failing state with each call. By switching to Elasticsearch this consistently takes under 1 minute.

**Technical details:**
Previously we have always tried to get a unique count with Elasticsearch `cardinality` aggregation. This does not work in many cases since it can only guarantee accuracy up to 10k. However, this new `scripted_metric` approach that has been used in a few different places has proven to be accurate (when testing against Staging and local) while completing in enough time to be cached.

The only test case that was updated is `def test_award_amount_success()`. This test case looked for results with Award Type Codes `["A", "09", "10"]` and DEF Codes `["L"]`. The issue is that it was getting the sum of Obligation and Outlay for all File C records under the Award as long as it had a single File C with DEF Code instead of only those values for File C with DEF Codes = "L". The change is that the returned Obligation and Outlay sum will only be for those File C records with the matching DEF Code to the filter that belong to an Award with with the correct Award Type Code.

Some additional notes around MATERIALIZED VIEWS and Table Partitions are added to the JIRA ticket as a comment.

    - [ ] Backend
8. [x] Jira Ticket [DEV-7831](https://federal-spending-transparency.atlassian.net/browse/DEV-7831):

```
```

### DIFF
--- a/usaspending_api/disaster/tests/integration/test_disaster_award_amount.py
+++ b/usaspending_api/disaster/tests/integration/test_disaster_award_amount.py
@@ -2,44 +2,50 @@ import pytest
 
 from rest_framework import status
 
+from usaspending_api.search.tests.data.utilities import setup_elasticsearch_test
+
 url = "/api/v2/disaster/award/amount/"
 
 
 @pytest.mark.django_db
-def test_award_amount_success(client, monkeypatch, generic_account_data, unlinked_faba_account_data, helpers):
+def test_award_amount_success(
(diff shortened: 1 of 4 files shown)


kwhickey (approved):
Looks good. Clean transition to ES (if a complicated ES query).

PR #3324, opened 2022-01-10
[DEV-8199] Update Source Procurement fields to be Boolean instead of Text

sethstoudenmier:
**Description:**
Broker is updating a lot of their TextFields that contain boolean style data to be a nullable Boolean field. This is the migration to our source table to match their changes.

**Technical details:**
Changing about 87 fields to be a NullBooleanField from TextField. These fields contain boolean style data (e.g. "Yes", "No", 1, 0, "true", "false") which convert easily from Text to Boolean. Since this table is only used in the pipeline a migration is used to handle the update. The pattern for the update is Stage and Swap:

* Create new NullBooleanField with `_temp` in the name
* Copy over data from current TextField to the NullBooleanField
* Rename the TextField to have `_old` in the name
* Rename the NullBooleanField to no longer have `_temp` in the name
* Remove the TextFields

In addition to this migration, there are a few changes to support reading in Boolean fields into TransactionFPDS table. 

    - [ ] Backend
8. [x] Jira Ticket [DEV-8199](https://federal-spending-transparency.atlassian.net/browse/DEV-8199):

```
```


emily-brents (approved):
Haven't checked every single field, but generally LGTM

PR #3334, opened 2022-01-14
[WARMFIX] Fix "other" award category grouping

emily-brents:
**Description:**
In the `obligations_by_award_category` endpoint, the "other" category both accounts for the defined award category "other" as well as anything that doesn't match one of the existing award categories. The endpoint was overriding the defined "other" category amount with the sum of non-matching categories, resulting in the "other" category to be 0 when it should not be.

**Technical details:**
Switched a `=` to a `+=` sign when formatting the "other" results. 

    - [ ] Backend
8. [N/A] Jira Ticket [DEV-123](https://federal-spending-transparency.atlassian.net/browse/DEV-123):

```
```

### DIFF
--- a/usaspending_api/agency/v2/views/obligations_by_award_category.py
+++ b/usaspending_api/agency/v2/views/obligations_by_award_category.py
@@ -69,8 +69,8 @@ class ObligationsByAwardCategory(AgencyBase):
                 results_map[key] = round(obligation_value, 2)
             else:
                 other_total += obligation_value
+        results_map["other"] += round(other_total)
 
-        results_map["other"] = round(other_total)
 
         # Convert Map back to list with extra categories
         for category, category_mapping in category_map.items():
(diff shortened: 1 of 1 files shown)


usaspending_api/agency/v2/views/obligations_by_award_category.py
```diff
@@ -69,8 +69,8 @@ def format_results(self, results):
                 results_map[key] = round(obligation_value, 2)
             else:
                 other_total += obligation_value
+        results_map["other"] += round(other_total)
```
sethstoudenmier, line 72: Could we just remove this and the else above? Cannot remember if this was put in place to cover a specific reason or if the `else` is redundant.

emily-brents: i'm pretty sure the idea was that if there was an award with a category that we don't have in the mapping it would go in that "other" sum, but I don't know if we have any data like that.

sethstoudenmier: Makes sense. Code LGTM then, but Travis has a few complaints.

sethstoudenmier (approved):

sethstoudenmier (approved):

PR #3342, opened 2022-01-26
[DEV-8370] UEI recipient hash

emily-brents:
**Description:**
Moves UEI to be the main source of the recipient_hash field
**Technical details:**
Had to change the way the recipient_hash is grouped for the recipient_lookup table to reduce the number of duplicate entries that are the same DUNS but are missing a UEI.

    - [ ] Backend
    - [ ] Frontend 
8. [X] Jira Ticket [DEV-8370](https://federal-spending-transparency.atlassian.net/browse/DEV-8370):

```
```


usaspending_api/recipient/management/sql/recipient_lookup/100_finalize.sql
```diff
@@ -35,22 +35,24 @@ UPDATE public.recipient_lookup rl SET
...
+        OR rl.parent_legal_business_name  IS DISTINCT FROM tem.parent_legal_business_name
+        OR rl.recipient_hash              IS DISTINCT FROM tem.recipient_hash
+        OR rl.source                      IS DISTINCT FROM tem.source
+        OR rl.state                       IS DISTINCT FROM tem.state
+        OR rl.zip4                        IS DISTINCT FROM tem.zip4
+        OR rl.zip5                        IS DISTINCT FROM tem.zip5
+  )
+  AND (
+        rl.duns                           IS DISTINCT FROM tem.duns
+        OR rl.uei                         IS DISTINCT FROM tem.uei
```
collinwr, line 55: Why is this moved out into a separate `AND` block? Shouldn't we update the record whether the duns/uei changes or one of the other fields?

usaspending_api/recipient/models.py
```diff
@@ -137,7 +137,7 @@ class RecipientLookup(models.Model):
     id = models.BigAutoField(primary_key=True)
     recipient_hash = models.UUIDField(unique=True, null=True)
     legal_business_name = models.TextField(null=True, db_index=True)
-    duns = models.TextField(unique=True, null=True)
+    duns = models.TextField(null=True)
```
sethstoudenmier, line 140: Curious why you removed `unique=True`? Would think if we might want to actually add it to `uei`.

sethstoudenmier (changes requested):
The committed code overall LGTM. Should add `uei` to `obtain_recipient_uri()` conditionals and then left a question for the model update. 

Additionally, see a couple other places we would need to switch the UEI around:
* restock_summary_award_recipient.sql
* recipient_agency.sql

usaspending_api/recipient/models.py
emily-brents: Because recipient_lookup uses the hash as the primary key, if we have records that have the same DUNS but different UEI (or one is null and one is not), we can have different records in recipient_lookup with the same DUNS. This happens a lot with transaction records that aren't backfilled vs / SAM records that were updated iirc.

usaspending_api/common/recipient_lookups.py
```diff
@@ -19,7 +25,9 @@ def obtain_recipient_uri(recipient_name, recipient_unique_id, parent_recipient_u
     Return example string: 11fcdf15-3490-cdad-3df4-3b410f3d9b20-C
 
     """
-    if (is_parent_recipient and not recipient_unique_id) or not (recipient_unique_id or recipient_name):
+    if (is_parent_recipient and not recipient_unique_id) or not (
+        recipient_unique_id or recipient_name or recipient_uei
+    ):
         return None
 
     if recipient_unique_id:
```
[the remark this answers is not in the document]
emily-brents: yeah that's fair

sethstoudenmier (approved):
LGTM. Just make sure to update deploy steps.

usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -198,7 +198,7 @@ def create_recipient_object(db_row_dict: dict) -> OrderedDict:
                     db_row_dict["_parent_recipient_unique_id"],
                     None,  # parent_recipient_unique_id
                     True,  # is_parent_recipient
-                    db_row_dict["_recipient_uei"],
+                    db_row_dict["_parent_recipient_uei"],
```
sethstoudenmier, line 201: Good catch! Completely missed this in review

sethstoudenmier (approved):

sethstoudenmier (approved):

PR #3348, opened 2022-01-27
[Prod] Fix travis by removing newer ES

sethstoudenmier:
**Description:**
Travis recently came default with ES 7.16 instead of the previous version of 5.5. For some reason the process of downgrading from 7.16 causes the install of our currently used version (7.1) to fail when starting. The fix created in #3343 is to remove the previously installed version of Elasticsearch prior to installing 7.1.

The PR mentioned above was created for testing purposes and branched off QAT. As a result, this new branch was created from `master` in order to get the Travis fix to all environments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ python:
 
 services:
   - postgresql
+  # NOTE: This is added, but re-installed at a different version below
+  - elasticsearch
 
 addons:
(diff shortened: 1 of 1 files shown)


collinwr (approved):
I can see successful builds using this new config. Approving.

PR #3352, opened 2022-01-28
[DEV-8388] Have ES based downloads check that replication has lookup ids

sethstoudenmier:
**Description:**
Help to resolve issues with downloads having incomplete data.

**Technical details:**
It appears that replication might be causing issues with our downloads in environments that utilize a Primary/Replica structure for the API/Downloads. Since the replica is readonly the lookup IDs for the Elasticsearch based downloads are inserted into the Primary and then it would appear that they are not being replicated over to the replica in time for the first CSV downloaded. 

This PR adds a wait following the step that inserts lookup IDs to make sure they are on the DB designated for downloads prior to proceeding with the download.

    - [ ] Backend
8. [ ] Jira Ticket [DEV-8338](https://federal-spending-transparency.atlassian.net/browse/DEV-8338):

```
```


usaspending_api/download/helpers/elasticsearch_download_functions.py
```diff
@@ -86,11 +88,39 @@ def _populate_download_lookups(cls, filters: dict, download_job: DownloadJob, si
         )
 
         created_obj_list = DownloadJobLookup.objects.bulk_create(download_lookup_obj_list, batch_size=10000)
```
kwhickey, line 90: Realizing this when I was chatting with you earlier, but how does this code (which is running on the BD instance) have a connection to the the master PG db? Does it?
I guess, is it's connection _always_ to the master DB... and it's only the `psql` command that gets handed the read-only download replica connection?

Just asking: does this bulk create insert into master? or try to insert into the read-only replica?

usaspending_api/settings.py
```diff
@@ -279,17 +279,26 @@ def _configure_database_connection(environment_variable):
...
+# must also be provided.
 if os.environ.get("DB_SOURCE"):
     if not os.environ.get("DB_R1"):
         raise EnvironmentError("DB_SOURCE environment variable defined without DB_R1")
+    if not os.environ.get("DOWNLOAD_DATABASE_URL"):
+        raise EnvironmentError("DB_SOURCE environment variable defined without DOWNLOAD_DATABASE_URL")
     DATABASES = {
         DEFAULT_DB_ALIAS: _configure_database_connection("DB_SOURCE"),
         "db_r1": _configure_database_connection("DB_R1"),
+        "db_download": _configure_database_connection("DOWNLOAD_DATABASE_URL"),
```
kwhickey, line 292: This generally looks okay, but it'd be good to talk through the need for why this all the sudden has to be in settings and we got by without it before. i.e. what changed? Is it that the python code running in BD needs to distinguish databases?

Also, as it's written here, will all envs and env-component-configs now need to provide the `DOWNLOAD_DATABASE_URL`? even deployed API components?

kwhickey (dismissed):
Approving as it looks, but have some questions to discuss, mostly around handling of connection strings.

Also, it occurs to me that this is all introduced by managing this intermediate download ID data in the database. Something I suggested we do. 
An alternative approach I'm not sure if we considered is to join a "virtual table" inline of the query, rather than physically inserting it, waiting for replication, and then joining. It regresses a bit back towards the complicated `IN` clause we had, but is a bit different. Postgres could do this with what it calls "constant tables`, which is just use of the `VALUES (..), (..)` statement. see: https://www.enterprisedb.com/postgres-tutorials/how-create-constant-table-postgresql

the question would remain, like with the IN clause, does putting 500k IDs inline of the query in a `VALUES` statement eat up memory/CPU and cause postgres to choke.

usaspending_api/settings.py
sethstoudenmier: Added this in order to use `.using()` in the Django ORM to specify which DB to hit for the Queryset. Probably better to conditionally add this if `DOWNLOAD_DATABASE_URL` is provided. 

It isn't that the value is all of sudden needed in `settings.py` but that it is needed in the `DATABASES` setting. Currently `DOWNLOAD_DATABASE_URL` is used directly as a connection string in the PSQL command.

kwhickey (approved):
Looks good 👍

PR #3357, opened 2022-02-02
[DEV-8368] Fix subcomponent bug

emily-brents:
**Description:**
Fixes two bugs - one that was causing a 500 error due to null values in the sort field, the other due to the incorrect fiscal period being chosen for FY2022.

**Technical details:**
The base class used for all agency pages defaults to the last period of the quarter - when we use that with Files A/B and we don't have submissions for that period but do have them for other periods in the quarter-i.e., no submissions for period 3 but they exist for period 2, no data was being returned. This fix changes the code to pick the most recent period, not based on the most recent quarter

    - [ ] Backend
8. [X] Jira Ticket [DEV-8368](https://federal-spending-transparency.atlassian.net/browse/DEV-8368):

```
```


usaspending_api/agency/v2/views/subcomponents.py
```diff
@@ -69,7 +71,7 @@ def format_results(self, file_a_response, file_b_response):
                 }
                 for x in combined_response
             ],
-            key=lambda x: x.get(self.pagination.sort_key),
+            key=lambda x: (x.get(self.pagination.sort_key) is not None, x.get(self.pagination.sort_key)),
```
sethstoudenmier, line 74: Would recommend using something like `sort_with_null_last()` here so that we can start to be more consistent. I know in some other places we have added Null values to the end of results.

usaspending_api/agency/v2/views/subcomponents.py
```diff
@@ -110,11 +112,18 @@ def get_file_b_queryset(self):
...
+            )
+            .values("reporting_fiscal_year")
+            .annotate(max_fiscal_period=Max(F("reporting_fiscal_period")))
+            .values("max_fiscal_period")
+        )
         filters = [
             Q(**{f"{treasury_account_keyword}__federal_account__parent_toptier_agency": self.toptier_agency}),
             Q(submission__reporting_fiscal_year=self.fiscal_year),
-            Q(submission__reporting_fiscal_period=self.fiscal_period),
+            Q(submission__reporting_fiscal_period=latest[0]["max_fiscal_period"]),
```
sethstoudenmier, line 126: Should the `fiscal_period` property in AgencyBase be changed instead? If this issue is coming up with the switch to File A/B then won't this potentially affect anywhere using that property?

emily-brents: Yeah, theoretically it would, but I was under the impression that this was intentionally based on the comment under the `fiscal_period()` property in agency_base.py:
```
        """
        This is the fiscal period we want to limit our queries to when querying CPE values for
        self.fiscal_year.  If it's prior to Q1 submission window close date, we will return
        quarter 1 anyhow and just show what we have (which will likely be incomplete).
        """
```
Now I don't know if that's still what we want to do, given the situation, but it does seem like we did it on purpose. I also don't want to randomly switch every use of `fiscal_period` on the agency page to a new value without more testing.

sethstoudenmier: Screenshot of what I am seeing with your branch against QAT DB
Screen Shot 2022-02-02 at 10 30 46 AM

emily-brents: I'll update the bureau federal account endpoint, since that was also changed to file a/b, but the rest of the endpoints on the agency page I believe were intended to operate as-is

sethstoudenmier (approved):
LGTM. Checked out changes on local frontend against QAT data.

PR #3366, opened 2022-02-11
[DEV-8477] Allow filtering on non-COVID DEFC

sethstoudenmier:
**Description:**
Pull in all DEFC to the Elasticsearch Award Index. Additionally, keep in place filtering limits for COVID DEFC, but remove them for non-COVID DEFC.

**Technical details:**
Update the Award materialized views to pull in all DEFC, not just those that are COVID DEFC. The JOIN where these values are now sourced from is already a part of the materialized views and will minimally affect performance to pull in all DEFC instead of only COVID DEFC. Test cases were updated to contain necessary data for this JOIN to populate the DEFC. From a business logic perspective these fields will always be populated, but our test data in a lot of cases did not specify item such as a Treasury Appropriation Account, Federal Account, or Toptier Agency because they simply weren't needed for the current test.

Changes were made to the DEFC filter for Elasticsearch to re-structure how the COVID limits on DEFC filter are applied and also removing those limits in cases where we aren't searching for COVID DEFC.

NOTE: Added a `do not merge` while these changes are run in lower environments for larger scale testing prior to merge.
UPDATE: Removed `do not merge` since changes look good in QAT following new Award matviews and ES index.

    - [ ] Backend
8. [x] Jira Ticket [DEV-8477](https://federal-spending-transparency.atlassian.net/browse/DEV-8477):

```
```


usaspending_api/common/query_with_filters.py
```diff
@@ -422,16 +423,47 @@ def generate_elasticsearch_query(cls, filter_values: List[str], query_type: _Que
...
+            else:
+                query = ES_Q("bool", should=covid_queries, minimum_should_match=1)
+
+            def_codes_query.append(query)
+
+        if other_queries:
+            query = ES_Q("bool", should=other_queries, minimum_should_match=1)
+            def_codes_query.append(query)
+
+        if len(def_codes_query) > 1:
```
collinwr, line 458: Since the final else takes the same action as the first, any reason not to condense this down to:
```
if len(def_codes_query) != 1:
    ...
else:
    ...
```

usaspending_api/common/query_with_filters.py
```diff
@@ -422,16 +423,47 @@ def generate_elasticsearch_query(cls, filter_values: List[str], query_type: _Que
...
+
+        if covid_queries:
+            if query_type == _QueryType.TRANSACTIONS:
+                query = ES_Q(
+                    "bool",
+                    must=ES_Q("range", action_date={"gte": "2020-04-01"}),
+                    should=covid_queries,
+                    minimum_should_match=1,
+                )
+            elif query_type == _QueryType.AWARDS:
```
collinwr, line 444: Why add a special case for Awards requiring non-zeroes?

sethstoudenmier: Special case already exists, but is in the `_handle_defc_query` function down below. Instead of sharing similar logic in two places around DEFC I combined it into one location.

collinwr: Oh I see that now

collinwr (approved):
Approved

PR #3377, opened 2022-02-17
[DEV-8370] UEI Based hash redo

emily-brents:
**Description:**
Changes the generation of the `recipient_hash` field to use the UEI field instead of the DUNS field by default. 

**Technical details:**
Changing the recipient_hash to be UEI based causes several issues - we have to remove the requirement for unique DUNS on the recipient_lookup table, since UEI -> DUNS is not a one-to-one relation. There is also the case where a DUNS might be listed once with a UEI and once with a NULL UEI. In order to remove the listing with a NULL UEI, we calculate a second DUNS-based hash for the row that does have a UEI and compare that to the hash of the row without one (without UEI means that the hash will either be DUNS or name based), and then delete that record and replace it with the new UEI version.
This also means that in addition to all the recipient_hash calculations being rewritten in all the matviews, we also have to adjust any joins that are dependent on DUNS and switch those to use the recipient_hash to avoid any duplicate records.

    - [ ] Backend
8. [X] Jira Ticket [DEV-8370](https://federal-spending-transparency.atlassian.net/browse/DEV-8370):


usaspending_api/database_scripts/matview_generator/subaward_view.json
```diff
@@ -93,15 +93,17 @@
     "LEFT OUTER JOIN",
     "  (SELECT",
     "    legal_business_name AS recipient_name,",
-    "    duns",
+    "    duns,",
+    "    ROW_NUMBER() OVER(PARTITION BY duns) AS row",
     "  FROM recipient_lookup AS rlv",
-    "  ) recipient_lookup ON recipient_lookup.duns = recipient_unique_id AND recipient_unique_id IS NOT NULL",
+    "  ) recipient_lookup ON recipient_lookup.duns = recipient_unique_id AND recipient_unique_id IS NOT NULL AND row = 1",
```
collinwr, line 99: Is it safe to continue to join to `recipient_lookup` based on `duns`? When we stop receiving transactions with duns, we'll need something else. `uei` or `recipient_hash`.

emily-brents: We currently do not have UEI data for subawards in the usaspending databases, when we get that data we can update the subaward view to use those, but for now this is what it is.

collinwr: I'll create a ticket to load `uei` from the broker `subaward` table into USAS. We can update this at that time.

usaspending_api/recipient/management/sql/recipient_lookup/000_temp_relations.sql
```diff
@@ -20,10 +20,22 @@ CREATE MATERIALIZED VIEW public.temporary_transaction_recipients_view AS (
     ))::uuid AS recipient_hash,
```
collinwr, line 20: UEI is not given priority when generating `recipient_hash` in the above case statement

collinwr (approved):
Approved

PR #3353, opened 2022-01-31
[DEV-7590] Proving out the ability for USAspending to use more than one database schema / database

todddunagan-bah:
**Description:**
PoC to demonstrate how we might support database tables and other artifacts across schemas and/or physical databases.

**Technical details:**
* Manual copied db table design for agency, toptier_agency, and subtier_agency
* Manual export import above table data
* Config changes: 
  * Added ability to supply `search_path` during database bootstrap
  * Updated the base database router to be aware of `secondary` models in order to know how to route to a non-primary db
* Migration considerations:
  * Migrations must be run to include db specificity: `python3 manage.py migrate references  --database db_secondary`
  * Had much trouble trying to train the app to know how to run migrations correctly to tables it previously knew to be in a different location. Ended up having to run: `python3 manage.py migrate references  --fake --database db_secondary` in order to get migrations to a state to accept a new migration to the `references` app.

PR #3418, opened 2022-04-07
[PIPE-192] Demonstrate migration and code changes for new schemas

sethstoudenmier:
# Approach
This Draft PR demonstrates the first option below. Django does not natively support schemas and so a workaround has to be used.

* Options:
   * Create minimum two Users on each Database (one for ETL and one for API / Downloads) that have different `search_path` 
      * PRO:
         * Minimal code changes with only a single migration and some ETL changes (specifically where a table, view, or materialized view is created)
         * As long as tables are queried without the schema name (e.g. `broker_subaward` instead of `raw.broker_subaward`) each user will only query the schemas in their `search_path`
      * CON:
         * Any time a new table is created an additional step will be required to change it to the correct schema since `public` will be the default (temp tables would not be an issue since they would be created in `public` and then removed)
   * Define additional Django connections that leverage different `search_paths` (example found in #3353)
      * PRO:
         * More clearly defined change since the router would be updated to say specifically which actions are taken against which connection (read, write, and migrations)
      * CON:
         * Would require that the router is kept up to date as new models are created otherwise this would need to be paired with the third option below where the schema name is combined with the table name
   * Define each Django model with the schema in the table name (e.g. `'"raw"."broker_subaward"'` instead of `"broker_subaward"` and set User permissions on what schema they are able to access
      * PRO:
         * Django migrations would automatically create new tables in the correct schema without any additional SQL
         
      * CON:
         * Would require some updates to ETL code where the schema name is added to the table name (e.g. would generate `public.raw.broker_subaward` without a change in `load_subawards`)
         * Would have to leverage Django's [SeparateDatabaseAndState](https://docs.djangoproject.com/en/3.2/ref/migration-operations/#separatedatabaseandstate) migration in order to prevent a table name change and instead move to another schema
         * This would allow a DB User to query against all tables regardless of `search_path` and thus a permission level change would be required to make sure that the API / Download User is only able to see `public` and `rpt` schemas

# Demonstration
In this Travis build (https://app.travis-ci.com/github/fedspendingtransparency/usaspending-api/builds/249099648) you will see all tests fail because in the migrations the Awards table is not defined.
```
E               psycopg2.errors.UndefinedTable: relation "awards" does not exist
E               LINE 5:     from awards 
E 
```
However, in this Travis build (https://app.travis-ci.com/github/fedspendingtransparency/usaspending-api/builds/249101392) all tests pass after the change was pushed to update the Travis DB user to contain the correct `search_path`.

The specific line added to the Travis build:
```
- psql -c "ALTER USER ${USASPENDING_DB_USER} SET search_path TO public,raw,int,temp,rpt"
```

# TODO
* Determine path forward for local development; should a developer be required to manually adjust their DB user's search_path or should something be changed in the Dockerfile to make that update?
* Break this up to match the JIRA tickets currently created for an easier review.
* Verify if any tables in the as-is plan for each schema have the same name. If the answer is yes, determine a new name to prevent conflict. Tables will the same name will be an issue when querying across different schemas with the use of `search_path`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,7 @@ before_script:
   - ./cc-test-reporter before-build
   # Our Postgres DB provided by Travis needs to have the (super) users specified by our env var DB URLs used
   - psql -c "CREATE USER ${USASPENDING_DB_USER} PASSWORD '${USASPENDING_DB_PASSWORD}' SUPERUSER"
+  - psql -c "ALTER USER ${USASPENDING_DB_USER} SET search_path TO public,raw,int,temp,rpt"
   - psql -c "CREATE USER ${BROKER_DB_USER} PASSWORD '${BROKER_DB_PASSWORD}' SUPERUSER"
   - docker build -t ${BROKER_DOCKER_IMAGE} ${BROKER_REPO_FOLDER}  # Build image from which to call Broker scripts
   - until curl --silent -XGET --fail http://localhost:9200; do printf '.'; sleep 1; done
--- a/usaspending_api/awards/management/commands/load_subawards.py
+++ b/usaspending_api/awards/management/commands/load_subawards.py
@@ -48,8 +48,8 @@ class Command(mixins.ETLMixin, BaseCommand):
     def _perform_load(self):
         """ Grab the Broker subaward table and use it to update ours. """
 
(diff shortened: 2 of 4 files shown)


sethstoudenmier:
New Pull Requests have been opened to implement this new schema functionality.

PR #3428, opened 2022-04-19
[PIPE-195/196/197/201/204] Move transaction related tables into different schemas

sethstoudenmier:
**Description:**
Creating new schemas and moving Transaction related tables.

**Technical details:**
Creating a migration inside of the `etl` app that has some dependencies on other apps to make sure we move the most recent state of the transaction tables.

    - [ ] Backend
8. [x] Jira Ticket [PIPE-192](https://federal-spending-transparency.atlassian.net/browse/PIPE-192):

```
```

### DIFF
--- a/usaspending_api/database_scripts/matview_generator/chunked_matview_sql_generator.py
+++ b/usaspending_api/database_scripts/matview_generator/chunked_matview_sql_generator.py
@@ -30,7 +30,8 @@ POSTGRES INDEX FORMAT
 
 EXAMPLE SQL DESCRIPTION JSON FILE:
 
-{   "final_name": "example_matview",
+{   "schema_name": "public",
+    "final_name": "example_matview",
     "matview_sql": [
     "SELECT",
     "  action_date,",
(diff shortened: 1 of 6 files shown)


usaspending_api/database_scripts/matview_generator/chunked_matview_sql_generator.py
```diff
@@ -159,6 +162,7 @@ def create_componentized_files(sql_json):
 
 def create_chunked_componentized_files(sql_json):
     table_name = sql_json["final_name"]
+    matview_schema_name = sql_json.get("chunked_schema_name", "public")
```
kwhickey, line 165: I can have a verbal convo to settle this:

I see this with a default of "public". Is that convention everywhere -- fall back to `public` when not found to be set?

And then I see `"chunked_schema"` set like:
```
{
  "schema_name": "rpt",
  "chunked_schema_name": "temp",
```

So the net effect is that the chunked matviews will be created in the `temp` table schema, yes?

kwhickey (approved):
Code looks good. 
One comment that we can cover verbally, but I think it should be fine -- approving contingent on all of this being tested/exercised through the pipeline and jobs. 

One other thing that needs to happen or get resolved is the monthly deltas and database archive getting figured out with these changes.

usaspending_api/database_scripts/matview_generator/chunked_matview_sql_generator.py
sethstoudenmier: > Is that convention everywhere -- fall back to public when not found to be set?
The matviews all had that as a default previously so that is why I carried it forward.

> So the net effect is that the chunked matviews will be created in the temp table schema, yes?
That is correct.

PR #3431, opened 2022-04-20
Ops 2229 add postgres version local env

gpontejos:
**Description:**

**Technical details:**

This just adds the ability to use a different version of PostgreSQL in your local environment. The version that was previously pinned was several minor versions behind the currently deployed version. Also, we're moving to the PostgreSQL 13 so we'll need to accommodate for that in the local env. 

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [OPS-2229](https://federal-spending-transparency.atlassian.net/browse/OPS-2229):

```
```

### DIFF
--- a/.env
+++ b/.env
@@ -2,6 +2,8 @@
 ##  run docker-compose config to see the templated variables ##
 POSTGRES_USER=usaspending
 POSTGRES_PASSWORD=usaspender
+POSTGRES_CLIENT_VERSION=10
+POSTGRES_DB_VERSION=10.18
 
 ## Change to host.docker.internal if you are running a local Postgres. Otherwise leave as-is, so
 ## Docker will use the Postgres created by Compose.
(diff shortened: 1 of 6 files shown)


.env
```diff
@@ -2,6 +2,8 @@
 ##  run docker-compose config to see the templated variables ##
 POSTGRES_USER=usaspending
 POSTGRES_PASSWORD=usaspender
+POSTGRES_CLIENT_VERSION=10
+POSTGRES_DB_VERSION=10.18
```
sethstoudenmier, line 6: I believe this file was removed from version control in some work that Keith did. @gpontejos a new `.env.template` file exists for people to create their own `.env` file with. Could you move the change to there and also have it default to Postgres 13 since that is what we are upgrading to.

PR #3440, opened 2022-05-03
[DEV-7792] Upgrade to Django 3.2 from 2.2

sethstoudenmier:
**Description:**
Upgrade to Django 3.2 (current LTS version) from Django 2.2 (previous LTS version).1

**Technical details:**
The work was committed in three stages:

> Get all tests to pass by upgrading to Django 3.2 and upgrading libraries that also needed to be upgraded to support Django 3.2

* Upgrades to libraries that rely on Django
* All usages of `AppConfig` need to now have the `AppConfig.name` defined as a dot notation path
* Any annotation where you could have the potential for a mixed type (e.g. `Text and Char` or `Int and Float`) the `output_field=` has to be a defined argument
* The `StringAgg` does not have the `output_field` set and can create an issue depending on the delimiter; this was resolved in Django's development branch, but not currently in a release
* With Django 3.2 we are now able to filter on Exists() instead of requiring an annotation; places that were previously performing an Exists in an annotation were refactored to filter on Exists for two reasons:
   * More performant to simply filter on EXISTS instead of an annotation
   * There are some issues with using Django CTE and a subquery (e.g EXISTS in an annotation) when subquery models are not defined with `objects = CTEManager()`
* Similar to the sub-bullet above, some models had to have `objects = CTEManager()` set in order to allow their use as a subquery of a queryset that contains a CTE; this is a known issue with a PR open in django_cte, but this workaround also works fine
* Similar to the new `StringAggWithDefault` used in place of `StringAgg` the `output_field` was also added to `ConcatAll` to handle mixed types
* `response` objects passed to `process_response` no longer have `_headers` as a private field but instead use `headers`
* Django in 3.1.1 decided that in order to fix an issue with custom json_decoders they would break functionality for raw queries that use a Django connection to query a JSONB field (all Django JSON fields are JSONB). A workaround for this is to cast raw queries of JSONB types to be a JSON types through the use of `::JSON` in the raw query, while this is quite annoying it is one that it does not seem the Django community cares to fix (https://code.djangoproject.com/ticket/31991#comment:1)
   * The bug is that all raw queries using a Django connection directly (instead of a psycopg2 connection which Django uses under the hood) any JSONField (JSONB data type) will be returned as a string, but any query to JSONField through the ORM will return a dictionary as expected
* `partial_index` was previously used since Django did not support such indexes. This library was not updated to support Django 3.2, but partial indexes were added to Django through the use of the `condition=` argument.
   * In order for migrations to run with Django 3.2 all legacy migrations that included `partial_indexes` were refactored to not use `partial_indexes` but instead the Django approach with `condition=`. Unfortunately, unique indexes are still not support via Django even though they are supported in Postgres. The end result was to add migrations using RawSQL that create a unique index and map to a typical pattern used to create an index (e.g. https://github.com/fedspendingtransparency/usaspending-api/commit/78da1c6fc2d07389c15149c69c4d38799dad8ca0#diff-b62d15a74328635bbe0166019a7e2b3d713354501edee910ec9f4dba6128a3aeR13).
 * Django has marked the `django.contrib.postgres.fields.JSONField` as deprecated in favor of the regular `models.JSONField`
 * Lastly, libraries that we did not use and also did not support Django 3.2 were removed from requirements and INSTALLED_APP

> Make the change from `model_mommy` to `model_bakery` since `model_mommy` is no longer supported and `model_bakery` is the version which has support for Django 3.2

* All just renames to use `bakery`

> Remaining version / code changes needed to silence some of the warnings

* Warnings primarily about deprecation:
   * Pandas 
   * Django Debug Toolbar
   * DRF API Tracking
   * Django
      * `NullBooleanField` deprecated in favor of `BooleanField(null=True, blank=True)`
      * `get_random_string()` is deprecated in favor of the same method where `length=` is a defined argumen
* Some libraries complained that `DEFAULT_AUTO_FIELD` was not defined so this was set in the `settings.py` with the value matching Django's own default value

    - [ ] Backend
8. [ ] Jira Ticket [DEV-7792](https://federal-spending-transparency.atlassian.net/browse/DEV-7792):

```
```


kwhickey (commented):
First commit reviewed ✅ 
Questions: ❓ 
- how many places did you need to change `select *` to `select fields, field:JSON` to cast?
- why removing the index creations in migrations and making them RunSQL? What lib were we using before that supported unique=True on the indices. Seems a big oversight for them to no longer support this kind of index.
- does the datadog middleware integration still work with this version of django? we should test this when deploying to staging.

requirements/requirements-app.txt
```diff
@@ -25,7 +25,7 @@ Markdown<3.0
 marshmallow==3.14.1
 numpy==1.19.*
 openpyxl==2.4.7
-pandas==1.1.*
+pandas==1.3.*
```
kwhickey, line 28: Why change the pandas version? We should look at specifically regression testing code that uses pandas.

sethstoudenmier: Was changed to silence more warnings that appear during tests. Not directly related to Django 3.2 so happy to revert this back if wanted.

sethstoudenmier:
> * how many places did you need to change `select *` to `select fields, field:JSON` to cast?
The only place I can remember making that change was around the submission loader when it creates the publish_history JSON object. And the change from `SELECT *` was only in the test case IIRC, the change that affected the loader was only adding `::JSON` to a single field.

> * why removing the index creations in migrations and making them RunSQL? What lib were we using before that supported unique=True on the indices. Seems a big oversight for them to no longer support this kind of index.
Previously we used `django-partial-index` to create the unique partial indexes. `django-partial-index` does not support Django 3.2 and Django supports partial indexes, but not unique partial indexes without the creation of a separate Index and Unique Constraint. 

> * does the datadog middleware integration still work with this version of django? we should test this when deploying to staging.
Plan to poke around more as we get this out to Staging.

@kwhickey Answers to other questions above

requirements/requirements-app.txt
kwhickey: Looked at versions and some deprecations but mostly advanced features. Going forward with upgrade.

kwhickey (approved):
Reviewed all major commits and discussed verbally. All looks good.

kwhickey (approved):
APPROVING! ... after minor updates from recent changes on qat.

PR #3245, opened 2021-09-27
[DEV-7790] combine_transaction_search_chunks AccessLock Error Mitigation

collinwr:
**Description:**
This PR updates the `combine_transaction_search_chunks` command to reduce the risk of failures due to AccessLocks. This is accomplished in two main ways.
1. Instead of renaming the constraints on the `transaction_search_old` table, the constraints are removed
2. Retry logic has been added to the step that removes the `transaction_search_old`

Additionally, a warning is returned from the command if the `transaction_search_old` table is unable to be dropped. This warning will be reported in message sent from Jenkins to Slack after a pipeline finishes. 

**Technical Details**
Two additional PRs will be created to update the staging and production pipelines. These PRs will include a change to send the warning to Slack.
1. Staging Pipeline - https://github.com/fedspendingtransparency/usaspending-config/pull/1096
2. Production Pipeline - https://github.com/fedspendingtransparency/usaspending-config/pull/1097 

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
4. [x] Matview impact assessment completed
5. [x] Frontend impact assessment completed
6. [x] Data validation completed
7. [ ] Appropriate Operations ticket(s) created
8. [x] Jira Ticket [DEV-7790](https://federal-spending-transparency.atlassian.net/browse/DEV-7790):
    - [x] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #3442, opened 2022-05-09
[PIPE-231] Remove constraints from Transaction Search to improve locking during swap

sethstoudenmier:
**Description:**
Improve locks that are taken during the combine transactions stage of the Nightly Pipeline to prevent deadlock scenarios.

**Technical details:**
Currently the Nightly Pipeline runs into a deadlock issue due to conflicting `AccessShareLock` and `AccessExclusiveLock`. The `AccessExclusiveLock` occurs when the combine transactions stage attempts to `DROP CONSTRAINT` for the two different FOREIGN KEYs (`awards` and `transaction_normalized`). One example of the `AccessShareLock` that has been problematic for this is a simple SELECT that joins `awards` and `transaction_normalized` since it takes an `AccessShareLock` on both of those tables. The deadlock scenario occurs when:
1. DROP CONSTRAINT for FOREIGN KEY to `awards` takes an `AccessExclusiveLock`
2. SELECT ... FROM ... takes an `AccessShareLock` on `awards` and `transaction_normalized`
3. DROP CONSTRAINT for FOREIGN KEY to `transaction_normalized` takes an `AccessExclusiveLock`

In the scenario above, Lock 1 will not let up until Lock 3 does since they are in a transaction. And Lock 2 will not let up until Lock 1 does since `AccessExclusiveLock` blocks all operations to the table under lock.

The reason this lock is taken on these tables is because of how Postgres handles FOREIGN KEY CONSTRAINTS. When a FOREIGN KEY CONSTRAINT is created it also creates a trigger on the underlying table (e.g. `transaction_search -> awards` and `transaction_search -> transaction_normalized`). When the CONSTRAINT is dropped it also needs to drop the trigger, which results in the `AccessExclusiveLock` on both Primary and Secondary table. This is also why attempts to resolve this in #3245 did not work. That PR added error handling on the DROP TABLE section fo the job, but also moved the DROP CONSTRAINTS to the Swap / Rename section. So while it appears that a new problem was created, it actually just moved to problem since the DROP TABLE would have also performed DROP CONSTRAINT via CASCADE. 

This leaves two solutions:
* Simply remove the DROP CONSTRAINT statements out of the DB transaction
   * Appears that having these statements in a DB transaction is actually bad practice for this deadlock reason
* OR, remove the CONSTRAINTS entirely from `transaction_search`
   * This PR takes this approach because really this table is a report built on top of tables that already have constraints
   * Also, the constraints are kinda needed for the Django ORM, but you can trick the Django ORM into thinking that the fields are Foreign Key or One-To-One fields even though they are simple data types AND NOT keys

Lastly, this PR removes changes around error handling since that is actually not a problem.

5. [x] API documentation updated
6. [ ] Necessary PR reviewers:
    - [ ] Backend
7. [x] Matview impact assessment completed
8. [x] Frontend impact assessment completed
9. [x] Data validation completed
10. [ ] Appropriate Operations ticket(s) created
11. [x] Jira Ticket [PIPE-231](https://federal-spending-transparency.atlassian.net/browse/PIPE-231):

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/combine_transaction_search_chunks.py
+++ b/usaspending_api/etl/management/commands/combine_transaction_search_chunks.py
@@ -5,8 +5,6 @@ from pathlib import Path
 from django.db import connection, transaction
 from django.core.management.base import BaseCommand
 
-from psycopg2 import OperationalError
-
 from usaspending_api.common.data_connectors.async_sql_query import async_run_creates
 from usaspending_api.common.helpers.sql_helpers import execute_sql_simple
 from usaspending_api.common.helpers.timing_helpers import ConsoleTimer as Timer
@@ -60,8 +58,6 @@ class Command(BaseCommand):
 
         logger.info(f"Chunk Count: {chunk_count}")
 
-        # Constraints must be read first, so indexes created by pk/fk are not repeated
-        create_temp_constraints, rename_constraints, delete_constraints = self.read_constraint_definitions()
         create_temp_indexes, rename_indexes = self.read_index_definitions()
 
         with Timer("Recreating table"):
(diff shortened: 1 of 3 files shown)


kwhickey (commented):
Well deduced to trap this scenario. I'll reach out to chat verbally to clarify the deadlocking scenario -- it would seem to me that Postgres should not let itself get into this scenario: either all locks in the transaction are take or none are, and/or while the exclusive lock exists, no reads (share lock) should be allowed to be taken. 
But maybe the order of operations within the DB transaction lent itself to this vulnerable-to-interleaving state.

I like the approach of just ridding ourselves of the overly-complex (unneeded) DB constraints. But I wonder if we were to have forced locks early in the transaction by taking the locks explicitly, whether it would have made the transaction less prone to deadlocks.

Another thing that should be considered is that in ridding this code of the functions to handle error retries and constraint reading/drops, might we need them again for other gold tables? Or are we deciding that none of them should have (or can have) FK constraints in the future?

kwhickey (approved):
After conversation, this change can go through as it nicely circumvents the deadlock vulnerability. But please record what we learned (In this PR most of the info is there already), as this will likely come up again as we do more table swaps like this

PR #3447, opened 2022-05-12
[Pipe 274] PySpark Functions to Identify differences in data frames

bkschmidt:
**Description:**
Pr adds Diff function for comparing Spark data frames with shared schema. 

**Technical details:**
Diff function takes two data frames with matching schema and lists of columns to compare. Returns a changes by category
.

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-123](https://federal-spending-transparency.atlassian.net/browse/PIPE-274):

```
```


usaspending_api/etl/data_diff_tool.py
```diff
@@ -0,0 +1,42 @@
+from gresearch.spark.diff import *
+
+def data_diff_function(df1, df2):
+    #assumes dataframes as a delta table
+    table1 = spark.read.load(df1)
+    table2 = spark.read.load(df2)
+    return (table1.diff(table2, table1.columns).show)
```
collinwr, line 7: I'm guessing that this method makes use of the gresearch library? What does this method provide that the `diff` method below does not? Is it necessary?

usaspending_api/etl/data_diff_tool.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def diff(left: DataFrame, right: DataFrame, unique_key_col="id", compare_cols=[], collect=False):
```
collinwr, line 10: Can you add a comment to describe what this method and each of its parameters does? Check out some examples from `usaspending_api/common/helpers/sql_helpers.py`

usaspending_api/etl/data_diff_tool.py
```diff
@@ -0,0 +1,42 @@
...
+    # unique_key_col = "id"
+
+    # If compare_cols is left as an empty list, ALL columns will be compared to each other
+    # (for rows where unique_key_col values match).
+    # Otherwise, specify the columns that determine "sameness" of rows
+    # compare_cols = []
+    # compare_cols = ["value", "v2"]
```
collinwr, line 17: Looks like there are some commented out debug statements here. Can you clean lines 11-17 up to only include documentation comments?

collinwr (changes requested):
Left a few comments. Looks like this is also failing the `black` check

usaspending_api/etl/data_diff_tool.py
```diff
@@ -0,0 +1,42 @@
+from gresearch.spark.diff import *
```
kwhickey, line 1: I don't think this is needed @bkschmidt -- this code that was created to diff two dataframes using a full outer join is completely standalone, yes?

usaspending_api/etl/data_diff_tool.py
kwhickey: Concur @bkschmidt I think we remove this function. Folks can use it ad-hoc in notebooks if needed, but I'd rather not have our code having a dependency on this library being pip installed. As it stands right now, it would fail since there is no update to requirements.txt that would bring the dependent library into a development env.

usaspending_api/etl/data_diff_tool.py
kwhickey: Concur - we use the Google Python Style Guide generally. Take a look at how docstrings are used for functions and methods: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods, esp _`Args`_

usaspending_api/etl/data_diff_tool.py
```diff
@@ -0,0 +1,42 @@
...
+
```
kwhickey, line 2: RE: the python module (aka the .py file name). I don't think this needs to stand alone in its own python module (file). Can you move this to the `usaspending_api/common/etl/spark.py` utilities module, as a utility that could be used in a data processing workflow?

kwhickey (changes requested):
Some changes desired:
1. Remove dependency on external library
1. Move functions into existing spark.py module
1. Use google python style docstrings
1. Also would be good to have a sample unit test that showcases use of this. Two dummy dataframes of in-memory dictionary or array data being compared and asserting the output is as expected. Could add in a new test module `usaspending_api/common/tests/unit/test_spark.py`

usaspending_api/etl/data_diff_tool.py
bkschmidt: Correct, I included the other function that uses gresearch here but given the later discussion, both that and this line should be removed.

usaspending_api/common/etl/spark.py
```diff
@@ -261,3 +261,45 @@ def merge_delta_table(spark: SparkSession, source_df: DataFrame, delta_table_nam
...
+        .join(right.withColumn("exists", lit(1)).alias("r"), left[unique_key_col] == right[unique_key_col], "fullouter")
+        .withColumn("diff", expr(compare_expr))
+    )
+    cols_to_show = ["diff"] + [f"l.{c}" for c in compare_cols] + [f"r.{c}" for c in compare_cols]
+    differences = differences.select(*cols_to_show)
+    # NOTE: .show() shows only a few rows. Call with collect=True to get ALL results, then show
+    if collect:
+        return differences.collect()
+    else:
+        differences.show(truncate=False)
```
collinwr, line 305: The return type should be consistent whether `collect` is true/false. In the current version, there is only a return when collect is true. They should either both return or show the differences. Make sure to update the documentation at the top to reflect the change.

bkschmidt: Updated to clarify

kwhickey (changes requested):
Noticed that the PR details are not filled out (High Level Description, Technical Details). Please fill those out so if this PR is ever referred back to, we can gather than context. 

image

usaspending_api/common/etl/spark.py
```diff
@@ -261,3 +261,47 @@ def merge_delta_table(spark: SparkSession, source_df: DataFrame, delta_table_nam
         WHEN NOT MATCHED THEN INSERT *
     """
     )
+
+
+def diff(left: DataFrame, right: DataFrame, unique_key_col="id", compare_cols=[], collect=False):
+    """
+    Compares two dataframes that share a schema and returns difference at a row level. 
+    Meant to be run in a databricks notebook.
```
kwhickey, line 269: Please remove this line in the comment. While it _can_ be run there, it doesn't have to be, and would actually require a fair bit of knowledge to get our code installed in a platform like that to get it to be imported and run (i.e. it's more likely someone would just copy-paste it where they want it). Also we don't want to provide code that is platforms specific if we can avoid it or suggest that the code is platform specific (this code currently is not).

usaspending_api/common/etl/spark.py
```diff
@@ -261,3 +261,47 @@ def merge_delta_table(spark: SparkSession, source_df: DataFrame, delta_table_nam
...
+    Args:
+        left: assumes a DataFrame wtih a schema that matches the right DataFrame
+        right: assumes a DataFrame wtih a schema that matches the right DataFrame
+        unique_key_col: indidcates a unique indentification column for comparison
+        compare_cols: list of columns to be compared, must exist in both left and right dataframe
+        collect: spark function to limit number of rows returned
```
kwhickey, line 275: While collect is a function on its own, here the parameter is a boolean flag indicating whether to collect the results and return those or not. Should read something more like:
```
collect (bool): flag indicating whether to simply show and print the results (if False) or to instead run the Spark ``.collect()`` function on the diff results DataFrame and return that list of ``Row`` objects
```

usaspending_api/common/etl/spark.py
kwhickey: The return type was always a bit confused on this. Maybe we just get it merged as is, and then extend/overload it. I'm thinking there should be a base `diff(...) -> DataFrame:` function, and then maybe a `diff_show(...) -> None:` and `diff_collect(...) -> List[Row]:` wrappers that "overload" the base function. 

That would allow folks that want to do further filtering on the differences to get the DataFrame and do so.

kwhickey (changes requested):
- Some docstring format/cleanup. 
- Suggestion on return type and overloading later

usaspending_api/common/etl/spark.py
bkschmidt: I'll remove the comment.

usaspending_api/common/etl/spark.py
bkschmidt: That clarifies that arg, I'll use similar language.

usaspending_api/common/etl/spark.py
```diff
@@ -261,3 +261,47 @@ def merge_delta_table(spark: SparkSession, source_df: DataFrame, delta_table_nam
...
+        If compare_cols is left as an empty list, ALL columns will be compared to each other
+        (for rows where unique_key_col values match).
+        Otherwise, specify the columns that determine "sameness" of rows
+    Returns:Row level differences based on the dataframes. If collect is called, output is returned
```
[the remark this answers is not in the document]
bkschmidt: Updated the syntax

usaspending_api/common/etl/spark.py
```diff
@@ -261,3 +261,47 @@ def merge_delta_table(spark: SparkSession, source_df: DataFrame, delta_table_nam
...
+    Meant to be run in a databricks notebook.
+    Args:
+        left: assumes a DataFrame wtih a schema that matches the right DataFrame
+        right: assumes a DataFrame wtih a schema that matches the right DataFrame
+        unique_key_col: indidcates a unique indentification column for comparison
+        compare_cols: list of columns to be compared, must exist in both left and right dataframe
+        collect: spark function to limit number of rows returned
+        If compare_cols is left as an empty list, ALL columns will be compared to each other
+        (for rows where unique_key_col values match).
+        Otherwise, specify the columns that determine "sameness" of rows
```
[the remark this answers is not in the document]
bkschmidt: Moved to the appropriate location

kwhickey (approved):
Approving after changes.

collinwr (dismissed):

collinwr (dismissed):

PR #3469, opened 2022-05-31
[PIPE-284] Updates settings_databricks.py to match Postgres upgrade version

collinwr:
**Description:**
This PR updates the `settings_databricks.py` to match the changes made to `settings.py` by https://github.com/fedspendingtransparency/usaspending-api/pull/3440/files#diff-6f927744266f66a4e5bb69251ac908ade18155d8a14a62d6320b92ef94063906

### DIFF
--- a/usaspending_api/etl/management/commands/create_delta_table.py
+++ b/usaspending_api/etl/management/commands/create_delta_table.py
@@ -34,7 +34,7 @@ TABLE_SPEC = {
     },
     "sam_recipient": {
         "schema_sql_string": sam_recipient_sql_string,
-        "source_table": "duns",
+        "source_table": "raw.duns",
         "destination_database": "raw",
         "partition_column": "broker_duns_id",
         "partition_column_type": "numeric",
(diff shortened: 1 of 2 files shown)


usaspending_api/etl/management/commands/create_delta_table.py
```diff
@@ -34,7 +34,7 @@
     },
     "sam_recipient": {
         "schema_sql_string": sam_recipient_sql_string,
-        "source_table": "duns",
+        "source_table": "raw.duns",
```
sethstoudenmier, line 37: @collinwr did you find that you had to add this? Should be fine to not use this as long as you are using the API or ETL users.

sethstoudenmier (dismissed):
Approving with the assumption that you don't see any future headaches with having the `<schema>.<table>` pattern used for the "source_table". If you think this could cause headache in the future then I would push to have the connection updated to use the right user and removed this change.

Leave merging up to you.

sethstoudenmier (approved):

PR #3472, opened 2022-06-01
[PIPE-279/280] Load Table to Delta Tests

dpb-bah:
**Description:**
Adding an integration test for loading data from the postgres database to delta spark tables.

**Technical details:**
* Includes a conftest addition that updates and resets the `POSTGRES_DSN` for test databases
* Updated `load_table_to_delta` to use the `get_jdbc_url()` command
* Updated `generate_test_db_connection_string` -> `get_database_dsn_string` throughout the code.
* Updated FABS SQL scripts `indirect_federal_sharing` portion to specify a precision that works with Spark's schema.
* Updated `TABLE_SPEC` with `model` fields for use among the tests.

2. API documentation updated
    - [ ] Backend
    - Frontend <OPTIONAL>
    - Operations <OPTIONAL>
    - Domain Expert <OPTIONAL>
8. [x] Jira Ticket [PIPE-279](https://federal-spending-transparency.atlassian.net/browse/PIPE-279) [PIPE-280](https://federal-spending-transparency.atlassian.net/browse/PIPE-280):

```
```


dpb-bah:
Added note: `Update dummy data to use factories with fuzzy data instead of blank values for a more accurate comparison`. This is something we see on the Broker that I feel should be imported to the USAS code but I feel is out of scope with this ticket.

usaspending_api/etl/tests/integration/test_load_table_to_delta.py
```diff
@@ -0,0 +1,87 @@
...
+
+def equal_datasets(ds1, ds2):
+    """Helper function to compare the two datasets. Note the column types of ds1 will be used to cast columns in ds2."""
+    datasets_match = True
+    for i, ds1_row in enumerate(ds1):
+        for k, val_1 in ds1_row.items():
+            val_1_type = type(val_1)
+            val_2 = ds2[i][k]
+            if not isinstance(val_1, type(None)) and not isinstance(val_1, type(val_2)):
+                val_2 = val_1_type(val_2)
```
collinwr, line 23: Why perform a cast here? Don't we want to catch types that don't match?

usaspending_api/etl/tests/integration/test_load_table_to_delta.py
```diff
@@ -0,0 +1,87 @@
...
+def _verify_delta_table_loaded(spark: SparkSession, delta_table_name: str, s3_bucket: str):
+    """Generic function that uses the create_delta_table command to create the given table and assert it was created
+    as expected
+    """
+    # make the table and load it
+    call_command("create_delta_table", f"--destination-table={delta_table_name}", f"--spark-s3-bucket={s3_bucket}")
+    call_command("load_table_to_delta", f"--destination-table={delta_table_name}")
+
+    # get the postgres data to compare
+    model = TABLE_SPEC[delta_table_name]["model"]
```
collinwr, line 41: Nice use of adding `model` to the `TABLE_SPEC` here

collinwr: If this is to combat casting that is done during the ETL. Would it make sense to use the `custom_schema` property from the `TABLE_SPEC` to target specific casts.

usaspending_api/conftest.py
```diff
@@ -123,6 +123,15 @@ def django_db_setup(
             ensure_business_categories_functions_exist()
             call_command("load_broker_static_data")
 
+            # temporarily updating the environment so commands in tests will point to the test database
+            old_environ = os.environ.copy()
+            os.environ["DATABASE_URL"] = generate_test_db_connection_string()
+
+    def reset_environ():
```
kwhickey, line 130: Maybe add clarifying comments that this is called by the finalizer below, and use of a finalizer is basically like auto-baking in a "tear-down" behavior to every test (if that is true).

usaspending_api/etl/tests/integration/test_load_table_to_delta.py
```diff
@@ -0,0 +1,87 @@
...
+    partition_col = TABLE_SPEC[delta_table_name]["partition_column"]
+    dummy_data = list(model.objects.order_by(partition_col).all().values())
+
+    # get the spark data to compare
+    received_data = spark.sql(f"select * from {delta_table_name} order by {partition_col}")
```
kwhickey, line 46: Are they always going to have a partition_col? Maybe have a guard here that only adds the order by of that col if it exists. Can it fall back to numeric (does `order by 1` work?) in the case where partition col does not exist?

usaspending_api/etl/management/commands/create_delta_table.py
```diff
@@ -31,43 +34,48 @@
         "custom_schema": "",
     },
     "recipient_lookup": {
-        "schema_sql_string": recipient_lookup_sql_string,
+        "model": RecipientLookup,
         "source_table": "recipient_lookup",
         "destination_database": "raw",
         "partition_column": "id",
         "partition_column_type": "numeric",
+        "schema_sql_string": recipient_lookup_sql_string,
```
kwhickey, line 42: It may be too much of a refactor for this scope, but do you and @collinwr also think that the term `schema_sql_string` can be misleading? Should it be more explicit like `table_create_sql` or something along those lines. "Schema" is getting overloaded to mean a lot of things currently in this context.

usaspending_api/conftest.py
```diff
@@ -123,6 +123,15 @@ def django_db_setup(
             ensure_business_categories_functions_exist()
             call_command("load_broker_static_data")
 
+            # temporarily updating the environment so commands in tests will point to the test database
+            old_environ = os.environ.copy()
+            os.environ["DATABASE_URL"] = generate_test_db_connection_string()
```
kwhickey, line 128: I think maybe we should be more precise here, and only store the single values we changed, and then only restore those single values. While the env _probably shouldn't_ be changing permanently during the course of a test run ... who knows. 

So could this be a dict that captures any environ value we change, and then in the finalizer, the dict's values are reset into the os.environ?

kwhickey (changes requested):
Good and concise accomplishment of the tests. Some suggestions inline for things we can do to minimize code changes, increase safety, and amplify comments.

usaspending_api/etl/tests/integration/test_load_table_to_delta.py
kwhickey: Yeah we need to think about what scenarios may cause a failure. Maybe there are a handful of types we treat as "synonymous" and tolerate casting, but others we don't? ... e.g. numeric and decimal ... but not decimal and int for instance?

kwhickey (changes requested):
Just a few more on the data comparer

usaspending_api/etl/tests/integration/test_load_table_to_delta.py
dpb-bah: No the `model` is for database querying for testing.
`dummy_data = list(model.objects.order_by(partition_col).all().values())` 
and can be used to also retreive the meta table name if we wanted to remove the `source_table` attribute.

usaspending_api/common/helpers/generic_helper.py
```diff
@@ -310,7 +310,7 @@ def get_account_data_time_period_message():
 
 def generate_test_db_connection_string():
     db = connection.cursor().db.settings_dict
-    return "postgres://{}:{}@{}:5432/{}".format(db["USER"], db["PASSWORD"], db["HOST"], db["NAME"])
+    return "postgres://{}:{}@{}:{}/{}".format(db["USER"], db["PASSWORD"], db["HOST"], db["PORT"], db["NAME"])
```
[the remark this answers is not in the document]
dpb-bah: Yeah I just found this trying to not remake the wheel. I can refactor to use that instead throughout the code.

usaspending_api/etl/tests/integration/test_load_table_to_delta.py
collinwr: What type of mismatches are you seeing here? 
If these mismatches are caused by the casting in the ETL done by `custom_schema`, would it make sense to use the `custom_schema` property from the `TABLE_SPEC` to target specific casts?

usaspending_api/etl/tests/integration/test_load_table_to_delta.py
dpb-bah: I'd be worried if a table doesn't have a `partition_column` or a `pk`. I can make these changes and account for that regardless though.

usaspending_api/etl/management/commands/create_delta_table.py
dpb-bah: I generally agree, as that's the primary purpose of it and the schema is more laid out in the model/db. @collinwr mind if I change it to `table_create_sql`?

usaspending_api/conftest.py
```diff
@@ -123,12 +124,21 @@ def django_db_setup(
...
 
-            # temporarily updating the environment so commands in tests will point to the test database
-            old_environ = os.environ.copy()
-            os.environ["DATABASE_URL"] = generate_test_db_connection_string()
+            # The Django function "call_command", which runs a separate process of a django management command, will
+            # use the same settings as calling it normally via commandline. This is fine except when calling it in a
+            # Django test that usually involves the creation/teardown of a test database, and the new management command
+            # will work off the normal database (not the newly created test database). This resolves the issue by
+            # temporarily mocking the DATABASE_URL environment value (in python storage, not actually modifying your
+            # local environment) which will then allow the new python process to accurately point to the test database.
```
kwhickey, line 132: Are you sure this is the case? I recently looked into this and the `call_command`, called from code, _simulates_ a CLI-call, but actually doesn't start a new process (or at least it didn't look like it did). If it where an actual call to `manage.py cmd_name`, yes that does start a new process. 

If that's the case, it may actually recycle the already parsed settings module, which in this case, would include any substitutions for settings that pytest-django has already done.

kwhickey (changes requested):
A couple more questions that we can discuss offline/verbally

usaspending_api/etl/management/commands/load_table_to_delta.py
```diff
@@ -75,14 +73,13 @@ def handle(self, *args, **options):
...
-        jdbc_url = get_jdbc_url_from_pg_uri(jdbc_url)
+        jdbc_url = get_jdbc_url()
         if not jdbc_url:
-            raise RuntimeError(f"Looking for JDBC URL passed to env var '{JDBC_URL_KEY}', but not set.")
+            raise RuntimeError(f"Could'nt find JDBC url, please properly configure your CONFIG.")
         if not jdbc_url.startswith("jdbc:postgresql://"):
             raise ValueError("JDBC URL given is not in postgres JDBC URL format (e.g. jdbc:postgresql://...")
 
-        if partition_column_type == "numeric":
+        if partition_column_type in ("numeric", "long"):
```
sethstoudenmier, line 82: @collinwr is correcting the types that caused the need for this in his PR #3478. I incorrectly used `long` originally instead of `numeric` not realizing that this isn't talking about the actual type. Should be able to remove this change following that PR's merge into QAT.

kwhickey (approved):
All changes look good

PR #3481, opened 2022-06-06
[PIPE-10] Standard Postgres DB connection config var names

kwhickey:
**Description:**
Standardizing names of DB url config vars and env vars. Then getting rid of POSTGRES_DSN, and ensuring DATABASE_URL and `USASPENDING_DB_*` vars are complete and consistent

**Technical details:**
Use of `pydantic.PostgresDsn` caused confusion
- you can create it with just the url
- if you do, the "parts" are left None
- you can also create it with the url AND parts
- if you do, it will unfortunately allow the parts and url to differ
- and then you could create it with just parts ... in which case both url and parts are populated, and consistent with each other. 

So really only the last case is beneficial... but also the least likely to be used.

There really isn't much value in keeping a `CONFIG` var around of type `PostgresDsn` if it is more likely than not that it will either have `None` parts, or that its parts will differ than what's shown in the URL. So refactoring to:
1. just have the parts as config vars
1. and the DATABASE_URL as a config var
1. and NOT have POSTGRES_DSN as a config var
1. and add a factory method for DATABASE_URL that 
1. will use DATABASE_URL if provided (as an env var or .env value) as priority
1. if not provided, attempt to use the parts (if provided) to build it, and then set DATABASE_URL to the result
1. AND THEN finally validate that the parts match the URL (in case both are provided, and it would lead to an inconsistent state). Throw an error if the user accidentally provided both / ALL and they differ. 

I think this is a safer and more concise approach. More defensive.

    - [x] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [PIPE-10](https://federal-spending-transparency.atlassian.net/browse/PIPE-10):

```
```


usaspending_api/config/envs/default.py
```diff
@@ -60,57 +67,126 @@ def __new__(cls, *args, **kwargs):
...
+                    cls, values, "USASPENDING_DB_PORT", lambda: str(url_parts.port)
+                )
+            if "USASPENDING_DB_NAME" not in values or values["USASPENDING_DB_NAME"] is None:
+                values = eval_default_factory_from_root_validator(
+                    cls, values, "USASPENDING_DB_NAME", lambda: url_parts.path.lstrip("/")
+                )
+            if "USASPENDING_DB_USER" not in values or values["USASPENDING_DB_USER"] is None:
+                values = eval_default_factory_from_root_validator(cls, values, "USASPENDING_DB_USER", lambda: username)
+            if "USASPENDING_DB_PASSWORD" not in values or values["USASPENDING_DB_PASSWORD"] is None:
+                values = eval_default_factory_from_root_validator(
```
dpb-bah, line 113: Is it possible to replace these with python dict `.get()` coalesce equivalents? Ex:
`"USASPENDING_DB_HOST" not in values or values["USASPENDING_DB_HOST"] is None`
=>
`values.get("USASPENDING_DB_HOST", None) is None`

dpb-bah (changes requested):

usaspending_api/config/envs/default.py
kwhickey: Good idea! changing ...

dpb-bah (approved):

PR #3486, opened 2022-06-09
[PIPE-257] Create transaction_search delta table

sethstoudenmier:
**Description:**
Create the `transaction_search` delta table to match the current process of creating the `transaction_search` Postgres table via the matview pipeline.

**Technical details:**
Two new values added to TABLE_SPEC:
* `transaction_search`
   * creation of `transaction_search` using a query that pulls from delta tables and views
* `transaction_search_testing`
   * creation of `transaction_search_testing` as a direct copy of `transaction_search` to aid testing

Few things to note:
* a UDF was created to make sure that any recipient_hash created ad hoc is converted to a UUID formatted string
* after fighting to have `federal_accounts` value represented as a Struct in PySpark I finally realized (and verified) that dropping a JSON formatted string into a Postgres JSON or JSONB column correctly converts to JSON or JSONB
* the `award_id` value for FABA was changed to be a LONG because that is the value used by the `awards` table for its PK
* test data was more detailed for `transaction_search` and `transaction_search_testing` to make sure that all appropriate FK relationships were created and nothing was omitted with a NULL since the query uses primarily LEFT JOINs

Local testing was done for the performance of loading using the UDF.

Timing with use of a UDF for formatting recipient_hash: 
* 4 m 30 s
* 4 m 13 s
* 4 m 11 s
* 4 m  3 s

Timing with direct copy of recipient_hash:
* 3 m 49 s
* 5 m 47 s (outlier but kept to demonstrate that local testing has some variation)
* 3 m 55 s

Timing with use of UDF and INSERT OVERWRITE approach:
* 3 m 43 s
* 3 m 48 s
* 4 m 21 s

Overall there is a performance hit, but does not appear to be too much. Worth keeping a closer eye on during testing in Databricks.

    - [ ] Backend
8. [x] Jira Ticket [PIPE-257](https://federal-spending-transparency.atlassian.net/browse/PIPE-257):

```
```


usaspending_api/etl/tests/integration/test_load_to_delta.py
```diff
@@ -60,6 +277,20 @@ def equal_datasets(psql_data: List[Dict[str, Any]], spark_data: List[Dict[str, A
...
+
+            if isinstance(spark_val, list):
+                spark_val = sorted(spark_val)
+
+            # Handle JSON formatted strings where the only difference is spacing between keys
+            if isinstance(psql_val, str):
+                psql_val = psql_val.replace(" ", "")
+
+            if isinstance(spark_val, str):
+                spark_val = spark_val.replace(" ", "")
```
kwhickey, line 292: isn't this removing all spaces in ALL string fields? How does this ensure it's just for JSON strings?
And what about spaces within the value of JSON fields?

usaspending_api/etl/tests/integration/test_load_to_delta.py
```diff
@@ -94,6 +327,18 @@ def _verify_delta_table_loaded(spark: SparkSession, delta_table_name: str, s3_bu
...
+def create_and_load_all_delta_tables(spark: SparkSession, s3_bucket: str):
+    non_aggregate_table_specs = {
+        key: val
+        for key, val in TABLE_SPEC.items()
+        if val.get("source_table") is not None and key == val.get("source_table")
+    }
+    for dest_table in non_aggregate_table_specs:
+        call_command("create_delta_table", f"--destination-table={dest_table}", f"--spark-s3-bucket={s3_bucket}")
+        call_command("load_table_to_delta", f"--destination-table={dest_table}")
+    create_ref_temp_views(spark)
```
kwhickey, line 339: Might need to / want to see how we can incorporate `--alt-db` at least, and maybe also `--alt-table` in this. It could be difficult though. If this is just used by unit tests, may be unnecessary.

usaspending_api/transactions/delta_models/transaction_search.py
```diff
@@ -1,104 +1,491 @@
...
+        {", ".join([f'{key} {val}' for key, val in _transaction_search_types.items()])}
     )
     USING DELTA
-    LOCATION 's3a://{SPARK_S3_BUCKET}/{DELTA_LAKE_S3_PATH}/{DESTINATION_DATABASE}/{DESTINATION_TABLE}'
-    """
+    LOCATION 's3a://{{SPARK_S3_BUCKET}}/{{DELTA_LAKE_S3_PATH}}/{{DESTINATION_DATABASE}}/{{DESTINATION_TABLE}}'
+"""
+
+transaction_search_load_sql_string = fr"""
+    INSERT OVERWRITE rpt.transaction_search
```
kwhickey, line 113: So does the `OVERWRITE` keyword here effectively do the same as `saveMode("overwrite")` in the pyspark syntax? ... 
And ... does that mean it marks all prior parquet files for that table as inactive (to be vacuumed) and everything that is part of _this_ INSERT becomes the _only_ data in the table?

usaspending_api/transactions/delta_models/transaction_search_testing.py
```diff
@@ -0,0 +1,104 @@
+transaction_search_testing_sql_string = r"""
+    CREATE OR REPLACE TABLE {DESTINATION_TABLE} (
+        transaction_id LONG NOT NULL,
```
kwhickey, line 3: Could we also have this use the fields dictionary to maintain the same order?

usaspending_api/transactions/delta_models/transaction_search.py
```diff
@@ -1,104 +1,491 @@
...
+    (
+        {",".join([col for col in _transaction_search_types])}
+    )
+    SELECT
```
kwhickey, line 117: I'll want to walk this through with you verbally, since there's so much packed into this matview-now-spark-sql statement. Things like:

- how `REGEXP_EXTRACT` worked out
- any array functions used
- e.g. whether `COLLECT_SET` is working okay as an array_agg substitute. Did we need to sort any of the agg'd results?
- UDFs working ok
- did we fix the `LATERAL JOIN` here
- if this was taken from the most recent transaction_search SQL, which incorporated all the UEI adjustments
- whether there's been any performance testing?
- any data validation testing thus far?

kwhickey (changes requested):
Looks really really good. Everything comes together in this one. 

Just a handful of questions and potential risks, as well as walking through some of this verbally.

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -0,0 +1,91 @@
...
+        "partition_column": "transaction_id",
+        "delta_table_create_sql": transaction_search_create_sql_string,
+        "custom_schema": "recipient_hash STRING, federal_accounts STRING",
+        "user_defined_functions": [
+            {
+                "name": "format_as_uuid",
+                "f": lambda val: str(uuid.UUID(val)) if val else None,
+                "returnType": StringType(),
+            },
+        ],
```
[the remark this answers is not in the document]
sethstoudenmier: There is a level of assumption with this function. It will handle the casing in regards to the output of MD5 / format of the UUID because that will always be lowercase. It does not handle the casing of values provided. It is assumed that the values reaching this are either upper case or lower case prior to reaching MD5 (our current approach for use of MD5 in Postgres).

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+
+class Command(BaseCommand):
+
+    help = """
+    This command reads data via a Spark SQL query that relies on delta tables that have already been loaded paired
+    with temporary views of tables in a Postgres database. As of now, it only supports a full reload of a table.
+    All existing data will be deleted before new data is written.
```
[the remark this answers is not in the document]
sethstoudenmier: Admittedly this was a copy and paste from the other load command. Happy to change the verbiage.

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Resolve Parameters
+        destination_table = options["destination_table"]
+
+        table_spec = TABLE_SPEC[destination_table]
+        destination_database = table_spec["destination_database"]
+
+        # Set the database that will be interacted with for all Delta Lake table Spark-based activity
+        logger.info(f"Using Spark Database: {destination_database}")
+        spark.sql(f"use {destination_database};")
```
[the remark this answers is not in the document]
sethstoudenmier: Will do

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Create User Defined Functions if needed
+        if TABLE_SPEC[destination_table].get("user_defined_functions"):
+            for udf_args in TABLE_SPEC[destination_table]["user_defined_functions"]:
+                spark.udf.register(**udf_args)
```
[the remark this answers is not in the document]
sethstoudenmier: Hadn't thought about that, but will verify / make necessary change.

usaspending_api/etl/management/commands/load_table_to_delta.py
```diff
@@ -1,20 +1,134 @@
...
+    # Additional definitions for use in testing
+    "transaction_search_testing": {
+        "model": TransactionSearch,
+        "source_table": "transaction_search",
+        "destination_database": "raw",
+        "partition_column": "transaction_id",
+        "partition_column_type": "numeric",
+        "delta_table_create_sql": transaction_search_create_sql_string,
+        "custom_schema": "recipient_hash STRING, federal_accounts STRING",
+    },
```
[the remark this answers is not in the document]
sethstoudenmier: Added this to copy over the TransactionSearch table directly from Postgres for comparison in Databricks with what we get from the load_query_to_delta job. Happy to remove this if you don't see much value.

usaspending_api/etl/tests/integration/test_load_to_delta.py
sethstoudenmier: It does, but given it is just a test case I was trying to think of the best way to verify data matches without making it too complicated. Can take another pass thinking about it.

usaspending_api/etl/tests/integration/test_load_to_delta.py
sethstoudenmier: Will do

usaspending_api/etl/management/commands/load_query_to_delta.py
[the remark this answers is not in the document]
kwhickey: ok. A quick test though looks like if `val` is an uppercase UUID, this will normalize it to lowercase, which is good --> everything coming out of this function is lowercased.

usaspending_api/etl/tests/integration/test_load_to_delta.py
kwhickey: I may have missed something in the way github redacts code... if this is just a function for _your_ test case, then it's all good. But if this is the shared function used by many test cases, then all the string fields are going to get transformed for every table's test case that uses that function.

usaspending_api/transactions/delta_models/transaction_search.py
kwhickey: Ok, i think those docs make it clear to me. But we ought to be careful about writing with partitions. Maybe it just works out though, that we'd only want to truncate data for changed partitions. But perhaps a row's partition column changed ; we'd need to make sure everything was wiped, not just the new partition (a lot like the ES requirement to delete-before-adding-in-order-to-update ... due to data landing on different shards).

So all that to mean, we may want to see if there's a `DELETE FROM` or `TRUNCATE` command we _explicitly_ run ahead of the `INSERT` on the _WHOLE_ table if we truly want to blow away the whole table.

kwhickey: Talked about it and it seems that there's no _implied_ partition (i.e. looking up the partition(s) affected and just replacing those based on the partition_col value of the rows inserted) ... you have to _explicitly_ tell it what partition you want to affect, if you want to affect just 1 partition. 

Still worth a test to confirm that assumption about the docs but code should be safe as is.

usaspending_api/transactions/delta_models/transaction_search.py
kwhickey: Talked all these through. 
Actions:
- see if we can order the COLLECT_SET when aggregating a list of items, for future comparisons in MERGE's and the like

usaspending_api/etl/management/commands/load_query_to_delta.py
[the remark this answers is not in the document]
sethstoudenmier: Was able to verify that this shouldn't be a problem. PySpark is creating temporary UDF in this case (https://github.com/apache/spark/blob/v3.2.1/sql/core/src/main/scala/org/apache/spark/sql/UDFRegistration.scala#L69). A temporary function in Spark (https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-ddl-create-function.html):

> Indicates the scope of function being created. When TEMPORARY is specified, the created function is valid and visible in the current session. No persistent entry is made in the catalog for these kind of functions.

That would mean that the only way two functions could interfere is if a developer were to register two of the same functions in this particular spot of the code, which at that point I would hope someone would realize that have two things named the same.

kwhickey (approved):
All changes look good!

PR #3494, opened 2022-06-22
Define schema columns to have consistency in created and loaded tables

sethstoudenmier:
**Description:**
Defined column order when loading data from Postgres to make sure they align with the defined schema.

**Technical details:**
Each delta_model's schema definition was moved into a dictionary to be used by both the schema and the load so that column order is guaranteed to match between the created table and the loaded data.

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>

```
```


usaspending_api/etl/management/commands/load_table_to_delta.py
```diff
@@ -127,6 +127,9 @@ def handle(self, *args, **options):
                 properties=get_jdbc_connection_properties(),
             )
 
+        if table_spec.get("column_names"):
+            df = df.select(table_spec.get("column_names"))
+
```
kwhickey, line 132: Would be good to add a comment here saying why we're doing this (e.g. to indicate preserving a consistent column ordering)

kwhickey (changes requested):
Looks good, just two minor suggestions
1. add a comment when ordering by cols
2. Maybe call the dictionaries `<TABLE_NAME>_COLUMNS` (or `_COLS`) rather than just `_TYPES`? I think it clarifies what the usage of that dict is moreso. e.g `AWARDS_TYPES` on its own makes me think of the different award types (contracts, loans, etc.)

usaspending_api/etl/management/commands/load_table_to_delta.py
sethstoudenmier: >AWARDS_TYPES on its own makes me think of the different award types (contracts, loans, etc.

Very good point, will make that change and add comment.

kwhickey (approved):
Looks good

PR #3497, opened 2022-06-24
PIPE-289: Delta -> Postgres command

dpb-bah:
**Description:**
Command to pull a delta table to postgres as a temp table.

**Technical details:**
Includes helper functions to copy constraints and indexes preserving their names.

2. API documentation updated
    - [ ] Backend
    - Frontend <OPTIONAL>
    - Operations <OPTIONAL>
    - Domain Expert <OPTIONAL>
8. [x] Jira Ticket [PIPE-289](https://federal-spending-transparency.atlassian.net/browse/PIPE-289):

```
```


usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+    As of now, it only supports a full reload of a table. All existing data will be deleted before new data is written.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--delta-table",
+            type=str,
+            required=True,
+            help="The source Delta Table to read the data",
+            choices=list(TABLE_SPEC.keys()),
```
sethstoudenmier, line 41: Not worth blocking over, but for a dictionary `list(<DICTIONARY>)` is the exact same thing as `list(<DICTIONARY_KEYS>)`.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+                source_columns = [(column.name, type(column)) for column in source_model._meta.get_fields()]
+            else:
+                source_columns = list(source_schema.items())
+
+            for column_name, column_type in source_columns:
+                if column_type in SPECIAL_TYPES_MAPPING:
+                    special_columns[column_name] = column_type
+                    delta_type = SPECIAL_TYPES_MAPPING[column_type]["delta"].format(column_name=column_name)
+                    alter_column_sql.append(
+                        f"ALTER TABLE {temp_destination_table}" f" ALTER COLUMN {column_name} TYPE {delta_type}"
```
kwhickey, line 181: Will the table be empty when making this alteration? If not, it could take a long long time to execute.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+        logger.info(f"LOAD (FINISH): Loaded data from Delta table {delta_table} to {temp_destination_table}")
+
+        if source_table or source_schema:
+            # Recast the special case columns to their postgres equivalents
+            alter_column_sql = []
+            for column_name, column_type in special_columns.items():
+                delta_type = SPECIAL_TYPES_MAPPING[column_type]["postgres"].format(column_name=column_name)
+                alter_column_sql.append(
+                    f"ALTER TABLE {temp_destination_table}" f" ALTER COLUMN {column_name} TYPE {delta_type}"
```
kwhickey, line 209: So in this case, they definitely will have a lot of rows in the table when the column data types are altered. How long does a query like this take with 100M+ records?

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+        logger.info(f"LOAD (START): Loading data from Delta table {delta_table} to {temp_destination_table}")
+        df.write.options(customSchema=custom_schema, truncate=True).jdbc(
+            url=jdbc_url, table=temp_destination_table, mode="overwrite", properties=get_jdbc_connection_properties()
+        )
+        logger.info(f"LOAD (FINISH): Loaded data from Delta table {delta_table} to {temp_destination_table}")
+
+        if source_table or source_schema:
+            # Recast the special case columns to their postgres equivalents
+            alter_column_sql = []
+            for column_name, column_type in special_columns.items():
```
kwhickey, line 206: Since this may have the potential of a long wait while it executes, it'd be good to put a log statement in here, before the for loop:
e.g. 
f"Altering columns back to their postgres-side types for columns: {special_columns}"

kwhickey, line 206: Since this may have the potential of a long wait while it executes, it'd be good to put a log statement in here, before the for loop:
e.g. 
f"Altering columns back to their postgres-side types for columns: {special_columns}"

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+            "--alt-db",
+            type=str,
+            required=False,
+            help="An alternate database (aka schema) in which to create this table, overriding the TABLE_SPEC db",
+        )
+        parser.add_argument(
+            "--alt-name",
+            type=str,
+            required=False,
+            help="An alternate delta table name for the created table, overriding the TABLE_SPEC destination_table "
```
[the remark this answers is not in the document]
kwhickey: Looks like this is for the delta side. So maybe clarify that in the help comments, or we could just rename them to be `--alt-delta-db` and `--alt-delta-table`

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+            # Only applicable if we're copying down a table that was already in postgres
+            alter_column_sql = []
+            source_columns = []
+            if source_model:
+                source_columns = [(column.name, type(column)) for column in source_model._meta.get_fields()]
+            else:
+                source_columns = list(source_schema.items())
+
+            for column_name, column_type in source_columns:
+                if column_type in SPECIAL_TYPES_MAPPING:
```
kwhickey, line 177: Similar to a comment below:

> Since this may have the potential of a long wait while it executes, it'd be good to put a log statement in here, before the for loop:
> e.g.
> f"Altering columns back to their postgres-side types for columns: {special_columns}"

usaspending_api/etl/tests/integration/test_load_to_from_delta.py
```diff
@@ -233,7 +234,7 @@ def _handle_string_cast(val: str) -> Union[str, dict]:
     """
     try:
         return json.loads(val)
-    except TypeError:
+    except (TypeError, json.decoder.JSONDecodeError):
```
kwhickey, line 237: We definitely wan to swallow this error? If so you should put a code comment in explaining why this is ok.

usaspending_api/transactions/delta_models/transaction_search.py
```diff
@@ -101,9 +101,110 @@
     "funding_office_name": "STRING",
 }
 
+TRANSACTION_SEARCH_POSTGRES_COLUMNS = {
```
kwhickey, line 104: I wonder if there's a way to not have to redefine ALL the columns again if most of them are not changing, but only transform one set of columns into the other (and still allow it to be referred to as a dict constant like this).

It also helps answer the question that anyone would ask when they see these two dicts: _What's the different between these two?_

kwhickey (changes requested):
A lot of good work in here. Some questions to walk through with you before approval.

usaspending_api/etl/management/commands/load_table_from_delta.py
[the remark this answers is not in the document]
collinwr: Looks like the `TABLE_SPEC` imported here from `create_delta_table.py` is the unioned version.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+                """
+            with db.connection.cursor() as cursor:
+                logger.info(f"Creating {temp_destination_table}")
+                cursor.execute(create_temp_sql)
+                logger.info(f"{temp_destination_table} created.")
+
+                # Copy over the constraints before indexes
+                # Note: we could of included indexes above (`INCLUDING CONSTRAINTS`) but we need to drop the
+                #       foreign key ones
+                if source_table:
```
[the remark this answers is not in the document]
collinwr: I was able to find a scenario where `source_table` is `None`. In the `TABLE_SPEC` for `award_search` we don't specify a `source_table` to read from because we're not writing that data from Postgres to Delta Lake, but we want the ability to write it back to Postgres.
I'm not sure we're going to get the expected behavior here though because we'd like to copy constraints from `award_search` to `award_search_temp`, which won't happen based on this if statement.

usaspending_api/etl/management/commands/load_table_from_delta.py
kwhickey: FYI ... first test writing this data to postgres, saw this statement running for over 1 hour, and it was still running as I wrote this.

kwhickey: Seeing how this runs, and anticipating the runtime of index creation, I think we need to separate spark work and non-spark work from this job and others like it. 

Reason is: An 8-node cluster will keep all 8 nodes up for maybe 1-3 hours extra while those nodes sit idle, just to issue a SQL statement on the DB and wait for it to complete. The way this kind of statement works is that it will run from the _driver_ node, and not create distributable tasks that are sent out to the worker nodes. So cost-wise, we don't want to keep a cluster up if the cluster is not doing work. 

We may be able to run the spark part in one job, and then continue the remaining parts in Jenkins job steps. It's not great that the job then isn't "complete", but we'll have to figure it out. The alternative would be an auto-scaling cluster, that could scale workers down if they're idle ... but I've heard Python jobs are not eligible on autoscaling clusters. We'll need to fact-check that.

kwhickey (changes requested):
Issue not getting indexes on `transaction_search`

usaspending_api/etl/management/commands/load_table_from_delta.py
dpb-bah: Good call. I think I copied that from another command but I don't see it in them anymore. Fixing.

usaspending_api/etl/management/commands/load_table_from_delta.py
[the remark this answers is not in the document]
dpb-bah: Yeah this is a kind of rub I need to square with other devs. If we want to copy indexes, constraints, etc. from SQL (not already a table in postgres), we need to update the TABLE_SPEC to include something that specifies these. I want to avoid having to regex SQL scripts to find said metadata.

usaspending_api/etl/management/commands/load_table_from_delta.py
dpb-bah: Table should be empty at this point. If it already existed, we've cleared it out and rebuilt it without any data.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+        custom_schema = table_spec["custom_schema"]
+        special_columns = {}
+
+        # Delta side
+        destination_database = options["alt_db"] or table_spec["destination_database"]
+        delta_table_name = options["alt_name"] or delta_table
+        delta_table = f"{destination_database}.{delta_table_name}" if destination_database else delta_table_name
+
+        # Postgres side
+        source_table = None
```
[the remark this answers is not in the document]
dpb-bah: I couldn't guarantee that with some cases (see `transaction_search`)

usaspending_api/etl/tests/integration/test_load_to_from_delta.py
dpb-bah: Thought I needed it until solving the issue of the delta <-> postgres column mismatch. Fixed

usaspending_api/transactions/delta_models/transaction_search.py
dpb-bah: Yeah I agree this was thrown together hastily. I didn't want to get involved with updating the other scripts/tests that used those columns. I could standardize it if the backend devs didn't mind me updating it everywhere.

@sethstoudenmier @collinwr @emily-brents @bkschmidt

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+    make_copy_constraints,
+)
+
+SPECIAL_TYPES_MAPPING = {
+    db.models.UUIDField: {"postgres": "UUID USING {column_name}::UUID", "delta": "TEXT"},
+    "UUID": {"postgres": "UUID USING {column_name}::UUID", "delta": "TEXT"},
+    db.models.JSONField: {"postgres": "JSONB using {column_name}::JSON", "delta": "TEXT"},
+    "JSONB": {"postgres": "JSONB using {column_name}::JSON", "delta": "TEXT"},
+    db.models.DateTimeField: {"postgres": "TIMESTAMP WITH TIME ZONE", "delta": "TIMESTAMP WITHOUT TIME ZONE"},
+    "TIMESTAMP": {"postgres": "TIMESTAMP WITH TIME ZONE", "delta": "TIMESTAMP WITHOUT TIME ZONE"},
```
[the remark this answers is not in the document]
dpb-bah: Added `USING {column_name} AT TIME ZONE 'UTC'` to the datetime updates. @sethstoudenmier confirmed when loaded normally, all the datetimes are without time zone, so this step is necessary to update them.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,227 @@
...
+                delta_type = SPECIAL_TYPES_MAPPING[column_type]["postgres"].format(column_name=column_name)
+                alter_column_sql.append(
+                    f"ALTER TABLE {temp_destination_table}" f" ALTER COLUMN {column_name} TYPE {delta_type}"
+                )
+            if alter_column_sql:
+                with db.connection.cursor() as cursor:
+                    cursor.execute(";".join(alter_column_sql))
+
+            if source_table:
+                # Load indexes if applicable
```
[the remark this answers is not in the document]
dpb-bah: I can give that a shot. We definitely need to refactor/optimize `TABLE_SPEC` (with this or next sprint) to be explicit about each of the fields we're using cause I guarantee `source_table` and `swap_table` will get mixed up.

usaspending_api/database_scripts/matview_generator/chunked_matview_sql_generator.py
```diff
@@ -16,6 +18,10 @@
     make_matview_refresh,
     TEMPLATE,
 )
+from usaspending_api.etl.broker_etl_helpers import dictfetchall
```
collinwr, line 21: This line causes this script to fail because it tries to reference Django settings.
Usage in pipeline:
```
python3 usaspending_api/database_scripts/matview_generator/chunked_matview_sql_generator.py --file usaspending_api/database_scripts/matview_generator/transaction_search.json --chunk-count 10 --dest chunked_matviews
```

usaspending_api/etl/management/commands/copy_table_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+                if max_parallel_maintenance_workers:
+                    logger.info(f"Setting max_parallel_maintenance_workers to {max_parallel_maintenance_workers}")
+                    cursor.execute(f"SET max_parallel_maintenance_workers = {max_parallel_maintenance_workers}")
+                if maintenance_work_mem:
+                    logger.info(f"Setting maintenance_work_mem to '{maintenance_work_mem}GB'")
+                    cursor.execute(f"SET maintenance_work_mem = '{maintenance_work_mem}GB'")
+                copy_index_sql = make_copy_indexes(cursor, source_table, dest_table)
+        if copy_index_sql:
+            logger.info(f"Copying indexes over from {source_table}.")
+            CTSC_Command().create_indexes(copy_index_sql, index_concurrency)
```
collinwr, line 145: The `combine_transaction_search_chunks` command will eventually be removed. Can you move the `create_indexes` method to `copy_table_metadata` and have `combine_transaction_search_chunks` reference it from there instead?

dpb-bah: I'll give it a go. I'm a little worried it may run into a circular dependency.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,288 @@
...
+from pyspark.sql import SparkSession
+
+from usaspending_api.common.helpers.spark_helpers import (
+    configure_spark_session,
+    get_active_spark_session,
+    get_jdbc_connection_properties,
+    get_jdbc_url,
+    get_jvm_logger,
+)
+from usaspending_api.etl.management.commands.create_delta_table import TABLE_SPEC
```
collinwr, line 15: I'm thinking this should `TABLE_SPEC` options from `load_query_to_delta` instead of `create_delta_table`. `create_delta_table` doesn't have the new, necessary options like `swap_table` and `swap_schema`

dpb-bah: In that case, I'm more tempted to add `swap_table` to those and set them to `None`s for consistency across all of them.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -0,0 +1,288 @@
...
+                temp_table,
+            ]
+            if not copy_constraints:
+                copy_table_metadata_args.append("--skip-constraints")
+            if not copy_indexes:
+                copy_table_metadata_args.append("--skip-indexes")
+            for option in ["max_parallel_maintenance_workers", "maintenance_work_mem", "index_concurrency"]:
+                if options[option]:
+                    copy_table_metadata_args.extend([f'--{option.replace("_", "-")}', options[option]])
+            call_command("copy_table_metadata", *copy_table_metadata_args)
```
collinwr, line 288: Aren't we trying to avoid this because it will be using cluster resources for database intensive tasks? Any reason to keep this logic here if we're not planning on using it in the pipeline?

dpb-bah: Generally yes. I kept it in as a kind of reminder that it's necessary that these should be done in succession and an option if we didn't run this in a pipeline. By default it won't but it could be used if someone wanted to kick it off and not have to come back to run the next command. I can take it out and leave a note if we don't think we'll ever need that option.

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -39,10 +46,15 @@
     "award_search": {
         "model": AwardSearchView,
         "source_query": award_search_load_sql_string,
+        "source_database": None,
+        "source_table": None,
         "destination_database": "rpt",
+        "swap_table": "vw_award_search",
```
collinwr, line 52: I think this should just be `award_search`. [PIPE-291](https://federal-spending-transparency.atlassian.net/browse/PIPE-291) will involve refactoring some code to use the `award_search` table instead of `vw_award_search` view.

dpb-bah: Yeah I couldn't find an `award_search` postgres table to test with and the model was `AwardSearchView`, so I thought it'd make sense to update `swap_table` alongside the model in PIPE-291. If you still want me to update this value now I could do it.

collinwr: Updating it later as a part of 291 makes sense!

usaspending_api/etl/management/commands/load_table_from_delta.py
collinwr: Just checked, looks like you did add them for those tables. I missed those earlier. I don't see why we'd want to add configuration for copies that we're never going to use though

dpb-bah: I was going for consistency across all the options in `TABLE_SPEC` and allowing to pull from other tables like `sam_recipient` or other tables which aren't available in the `TABLE_SPEC` in `load_query_to_delta`

usaspending_api/database_scripts/matview_generator/chunked_matview_sql_generator.py
collinwr: Cool! The command works with the pipeline usage now.

usaspending_api/etl/management/commands/copy_table_metadata.py
collinwr: This looks good!

usaspending_api/etl/management/commands/load_table_from_delta.py
dpb-bah: Took it out and included an additional log.

collinwr: Logging looks good. Nice touch to add the necessary parameters for the next step

kwhickey (approved):
Approved - 1 minor comment re code-comments, but perhaps you can apply this as this code is reviewed for award_search

PR #3506, opened 2022-07-11
Ftr/pipe 311 sam recipient load table from brokerdb

bkschmidt:
**Description:**
Add support to load the sam_recipient table from broker to load_query_to_delta.py

**Technical details:**
The high level tech approach for this story is to create a jdbc proxy view on the broker database sam_recipient table. Once this view has been created, SQL very similar to that in the existing restock_duns.sql file can be used to transform the data into the required schema.

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [x] Jira Ticket [PIPE-311](https://federal-spending-transparency.atlassian.net/browse/PIPE-311):

```
```


usaspending_api/common/helpers/spark_helpers.py
```diff
@@ -59,8 +60,20 @@
     DisasterEmergencyFundCode,
     SubmissionAttributes,
     DABSSubmissionWindowSchedule,
+    sam_recipient,
```
dpb-bah, line 63: `from usaspending_api.recipient.models import StateData, DUNS`

and use the `DUNS` model? Feels weird mixing delta models with the postgres django models...

usaspending_api/common/helpers/spark_helpers.py
```diff
@@ -401,6 +414,19 @@ def get_jdbc_url():
     return get_jdbc_url_from_pg_uri(CONFIG.DATABASE_URL)
 
 
+def get_broker_jdbc_url(databroker_uri: str)->str:
+    """Getting a JDBC-compliant Broker Postgres DB connection string hard-wired to the POSTGRES vars set in CONFIG"""
+    if not CONFIG.DATA_BROKER_DATABASE_URL:
+        raise ValueError("DATABASE_URL config val must provided")
```
dpb-bah, line 420: `DATA_BROKER_DATABASE_URL` config val must be provided

usaspending_api/common/helpers/spark_helpers.py
```diff
@@ -401,6 +414,19 @@ def get_jdbc_url():
     return get_jdbc_url_from_pg_uri(CONFIG.DATABASE_URL)
 
 
+def get_broker_jdbc_url(databroker_uri: str)->str:
```
dpb-bah, line 417: Any reason why this isn't following the other methods? Feels like it makes more sense to just do `get_broker_jdbc_url()` that's just like `get_jdbc_url` except it's `DATA_BROKER_DATABASE_URL` and that it reuses `get_jdbc_url_from_pg_uri` as both will be postgres connection strings.

usaspending_api/common/helpers/spark_helpers.py
```diff
@@ -59,8 +60,20 @@
...
 
+BROKER_PROXY_TABLES =[
+    sam_recipient,
+    subaward,
+    published_fabs,
+    detached_award_procurement,
+    submissions,
+    appropriation,
+    object_class_program_activity,
+    award_financial,
```
dpb-bah, line 75: May want to comment these out as they're a little big to store as temp tables potentially in memory.

dpb-bah (changes requested):

usaspending_api/common/helpers/spark_helpers.py
bkschmidt: Updated, thank you!

usaspending_api/common/helpers/spark_helpers.py
bkschmidt: Refactored to follow other methods

usaspending_api/common/helpers/spark_helpers.py
bkschmidt: Commented out; they are out of scope of this ticket and present a broader design question for bringing in broker tables.

usaspending_api/common/helpers/spark_helpers.py
bkschmidt: Let me know if this change makes sense; not sure if the change I just pushed is what you mean.

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -48,6 +51,16 @@
         "custom_schema": "recipient_hash STRING, federal_accounts STRING, cfdas ARRAY<STRING>,"
         " tas_components ARRAY<STRING>",
     },
+    "sam_recipient": {
+        "model": DUNS,
+        "source_table": "duns",
+        "destination_database": "raw",
```
collinwr, line 57: This should be written to the `int` schema

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -48,6 +51,16 @@
         "custom_schema": "recipient_hash STRING, federal_accounts STRING, cfdas ARRAY<STRING>,"
         " tas_components ARRAY<STRING>",
     },
+    "sam_recipient": {
+        "model": DUNS,
+        "source_table": "duns",
```
collinwr, line 56: A `source_query` is necessary to run the `load_query_to_delta` command. This should be the query that transforms the table from it's raw broker form to the intermediary needed by later ETLs. It should apply the same transformations as the query in `restock_duns.sql`

bkschmidt: Added and called sam_recipient_sql_string

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -54,6 +54,17 @@
         " tas_components ARRAY<STRING>",
         "column_names": list(AWARD_SEARCH_COLUMNS),
     },
+    "sam_recipient": {
```
collinwr, line 57: Ben, can you make the key for this table `int.sam_recipient`, so we won't have a key conflict with the TABLE_SPEC in `load_table_to_delta`?

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,116 @@
...
     "zip4": "STRING",
-    "business_types_codes": "ARRAY<STRING>",
-    "dba_name": "STRING",
-    "entity_structure": "STRING",
+    "zip5": "STRING",
+    "alternate_names": "ARRAY<STRING>",
+    "source": "STRING NOT NULL",
+    "update_date": "TIMESTAMP NOT NULL",
     "uei": "STRING",
-    "ultimate_parent_uei": "STRING",
```
collinwr, line 20: Looks like these columns were updated to match the source `sam_recipient` table from broker, however these should match the target delta model. Can you change these back?

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,116 @@
...
+        sam_recipient.country_code,
+        sam_recipient.congressional_district,
+        COALESCE(sam_recipient.business_types_codes, text[]) AS business_types_codes,
+        sam_recipient.entity_structure,
+        sam_recipient.sam_recipient_id,
+        COALESCE(sam_recipient.activation_date, sam_recipient.deactivation_date) as record_date,
+        sam_recipient.uei,
+        sam_recipient.ultimate_parent_uei
+      FROM
+        sam_recipient
```
collinwr, line 80: You need to prefix this table name with the database used for reference tables `global_temp`. I'd also give this table an alias, maybe something like `AS broker_sam_recipient`, so it's clear that this data is coming from the broker in the query.

collinwr (changes requested):
Left a few more comments. 

Also looks like there's no change to add broker reference tables. I know @dpb-bah is working on this as a part of his `subaward_search` story. If you want to test your new command, you may want to include his changes to `usaspending-api/usaspending_api/common/etl/spark.py`

usaspending_api/etl/management/commands/load_query_to_delta.py
bkschmidt: updated key to avoid conflict

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: I have made the prefix broker_SR and the alias refers to global_temp.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,107 @@
...
         {", ".join([f'{key} {val}' for key, val in SAM_RECIPIENT_COLUMNS.items()])}
     )
     USING DELTA
     LOCATION 's3a://{{SPARK_S3_BUCKET}}/{{DELTA_LAKE_S3_PATH}}/{{DESTINATION_DATABASE}}/{{DESTINATION_TABLE}}'
     """
+sam_recipient_load_sql_string = fr"""
+-- Create temp table to reload into the duns table without dropping the destination table
+DROP TABLE IF EXISTS int.sam_recipient;
+
+CREATE TABLE int.sam_recipient AS (
```
collinwr, line 35: Here we should try to avoid recreating the table. Check out the pattern of `INSERT OVERWRITE` used by `award_search_load_sql_string` and `transaction_search_load_sql_string`

kwhickey: [INSERT OVERWRITE docs](https://docs.databricks.com/sql/language-manual/sql-ref-syntax-dml-insert-into.html)

> If you specify OVERWRITE the following applies:
> - Without a partition_spec the table is truncated before inserting the first row.
> - Otherwise all partitions matching the partition_spec are truncated before inserting the first row.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,107 @@
...
-sam_recipient_sql_string = rf"""
+sam_recipient_create_sql_string = fr"""
     CREATE OR REPLACE TABLE {{DESTINATION_TABLE}} (
         {", ".join([f'{key} {val}' for key, val in SAM_RECIPIENT_COLUMNS.items()])}
     )
     USING DELTA
     LOCATION 's3a://{{SPARK_S3_BUCKET}}/{{DELTA_LAKE_S3_PATH}}/{{DESTINATION_DATABASE}}/{{DESTINATION_TABLE}}'
     """
+sam_recipient_load_sql_string = fr"""
+-- Create temp table to reload into the duns table without dropping the destination table
```
kwhickey, line 32: Fix comment here, referencing `duns` table, which is not the target.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,107 @@
...
+    broker_SR.state AS state,
+    broker_SR.zip AS zip,
+    broker_SR.zip4 AS zip4,
+    broker_SR.business_types_codes AS business_types_codes,
+    broker_SR.dba_name as dba_name,
+    broker_SR.entity_structure as entity_structure,
+    broker_SR.uei AS uei,
+    broker_SR.ultimate_parent_uei AS ultimate_parent_uei
+  FROM
+    global_temp.broker_SR as broker_SR
```
kwhickey, line 57: Where did `broker_SR` get added as a `global_temp` (global temporary view) view? I don't see that in the PR code. Make sure to bring in latest changes from `qat` branch, as the ref data code moved to `spark.py`.
It should have a more clear name, like `broker_sam_recipient` rather than an abbreviation.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,107 @@
...
+      SELECT
+        DISTINCT ON (broker_SR.awardee_or_recipient_uniqu)
```
kwhickey, line 59: I'm pretty sure `DISTINCT ON` is not supported by Spark SQL. Did you find it to be? 
In the recipient_lookup Spark SQL, we had to change this to windowing functions in CTEs like
`ROW_NUMBER() OVER (PARTITION BY uei, awardee_or_recipient_uniqu ORDER BY update_date DESC NULLS LAST) AS row_num`

And then you select form the CTE `WHERE row_num = 1` to get the top item of the ordered partition.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,107 @@
...
+        broker_SR.entity_structure,
+        broker_SR.sam_recipient_id,
+        COALESCE(broker_SR.activation_date, broker_SR.deactivation_date) as record_date,
+        broker_SR.uei,
+        broker_SR.ultimate_parent_uei
+      FROM
+        global_temp.sam_recipient as broker_SR
+      ORDER BY
+        broker_SR.awardee_or_recipient_uniqu,
+        broker_SR.activation_date DESC NULLS LAST)') AS broker_duns
```
kwhickey, line 83: Typo here with parens and single quotes?

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,107 @@
...
+            zip4 text,
+            country_code text,
+            congressional_district text,
+            business_types_codes text[],
+            entity_structure text,
+            sam_recipient_id text,
+            record_date date,
+            uei text,
+            ultimate_parent_uei text
+          )
```
kwhickey, line 104: Is this the leftover postgres `dblink` style query? I'm pretty sure we're getting away from that completely here, and querying the broker DB directly, rather than querying it through the usaspending db over `dblink`. 

It looks like maybe this SQL is just a placeholder for the ultimate Spark SQL? 
You'll want some unit tests that exercise whatever SQL you make to ensure it does in fact work. That will be a challenge to get the broker<->usaspending integration tests working just right. 
I'd say
1. Fix the sql
2. Point your local env at a remote broker DB (e.g. sandbox) while on VPN
3. Test the `load_query_to_delta` locally with something like: `make localhost-spark-submit django_command="load_query_to_delta --destination-table int.sam_recipient"`
4. Ensure you see the data you want in your local delta tables by opening up `make pyspark-shell` (you'll need to pull qat branch to get this) and then run `spark.sql("select * from int.sam_recipient").show()`
5. Then work on getting your own local broker DB setup in docker (it doesn't need data)
6. Then ensure you have the right local config to run broker integration tests (Daniel or I could help with this)
7. Then write some unit tests that inject some `data_broker.public.sam_recipient` data into the postgres DB, and then run your command to ensure it is working across DBs.

kwhickey (changes requested):
Preliminary review with some needed changes.

bkschmidt:
DO NOT MERGE: SEE https://github.com/fedspendingtransparency/usaspending-api/pull/3562 for current codebase

PR #3525, opened 2022-07-21
[PIPE-289] Drop Unique Award ID Constraint in Award Search

dpb-bah:
**Description:**
Simply drops the unique constraint on `award_id` in `rpt.award_search`

**Technical details:**
Note: this unique index is not managed by the model. 

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [PIPE-289](https://federal-spending-transparency.atlassian.net/browse/PIPE-289):

```
```

### DIFF
--- a/usaspending_api/search/migrations/0008_awardsearch_table.py
+++ b/usaspending_api/search/migrations/0008_awardsearch_table.py
@@ -5,10 +5,6 @@ from django.db import migrations, models
 import django.db.models.deletion
 import django.db.models.expressions
 
-additional_sql = [
-    'ALTER TABLE rpt.award_search DROP COLUMN id',
-    'CREATE UNIQUE INDEX as_idx_award_id ON rpt.award_search(award_id)'
-]
 
 class Migration(migrations.Migration):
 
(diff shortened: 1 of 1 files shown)


usaspending_api/search/migrations/0008_awardsearch_table.py
```diff
@@ -115,7 +119,7 @@ class Migration(migrations.Migration):
         # Trick Django into believing this is a foreign primary key for purposes of using the ORM,
         # but avoid the headache that comes with foreign keys and the primary key constraint
         migrations.RunSQL(
-            sql="ALTER TABLE rpt.award_search DROP COLUMN id",
+            sql=";".join(additional_sql),
```
sethstoudenmier, line 122: If we aren't using that SQL anywhere else I would say put it here instead of splitting it between two places.

usaspending_api/search/migrations/0008_awardsearch_table.py
```diff
@@ -115,7 +119,7 @@ class Migration(migrations.Migration):
...
             reverse_sql="",
```
sethstoudenmier, line 123: Can you think of anytime where not having reverse_sql specified could be an issue? Think it would need to be added to allow this migration to be reversed (should that need ever come up).

sethstoudenmier: Unapplying migrations runs `DROP TABLE "rpt"."award_search" CASCADE;` so can disregard.

sethstoudenmier (approved):

PR #3532, opened 2022-07-25
Adds production config for databricks

collinwr:
**Description:**
Adds production config for databricks

    - [ ] Backend
    - [ ] Operations

```
```

### DIFF
--- a/usaspending_api/config/envs/production.py
+++ b/usaspending_api/config/envs/production.py
@@ -1,0 +1,29 @@
+########################################################################################################################
+# Production runtime env configuration
+# - Inherits all defaults from DefaultConfig in default.py
+# - Overrides any default config variables with local runtime env specific values
+# - Users may override local config vars values declared here by putting values in a git-ignored .env file alongside
+#   docker-compose.yml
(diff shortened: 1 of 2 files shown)


usaspending_api/config/envs/production.py
```diff
@@ -0,0 +1,29 @@
...
+        Attributes inherited from or overridden from ``DefaultConfig``
+
+    """
+
+    # ==== [Global] ====
+    ENV_CODE: ClassVar[str] = "prd"
+
+    # ==== [AWS] ====
+    AWS_PROFILE: str = None
+    AWS_REGION = ""
```
kwhickey, line 28: Only thing I'm not sure of is why we have
```
AWS_REGION = ""
```
I would think we'd want that set, and it could be set in the `DefaultConfig` and not needing overriding in lower envs

And if we remove it, probably remove it from each qat.py, staging.py, production.py
If we keep it, I think it needs to be typed like it is in `DefaultConfig`

```
AWS_REGION: str = ""
```

kwhickey (approved):
Looks fine. Plan to remove override of aws region later.

PR #3540, opened 2022-08-04
[PIPE-380] SQL bulk COPY of CSV to write Delta tables back to Postgres

kwhickey:
**Description:**
Change the `load_table_from_delta.py` Django command to use a SQL bulk `COPY` command strategy in writing Delta tables to Postgres tables. 

**Technical details:**
- Timings show the using `COPY` maximizes DB resources, such as write throughput and IOPS, and leads to performance of about one-half the time to write a large table from Delta Lake to Postgres. 
- Unfortunately, Spark's `DataFrameWriter` for JDBC does not support the bulk `COPY` command, and simply does an huge array of `INSERT` statements. 
- Our approach to leveraging the cluster to do `COPY` was to write a custom `COPY` function that streams a gzipped CSV file from S3 to the `psycopg2` `copy_expert` command, which writes it to the temp postgres table.
- We `parallelized` the S3 file paths across an `RDD` and used the `rdd.mapPartitionWithIndex` to apply this function on the workers
- By default, the SaveMode is `"overwrite"` in `DataFrame.csv.write.options(...).mode("overwrite").csv(...)`, which will nuke and pave the .csv.gz files in the target directory on each run.

**TODO**
- [x] Regression fixes of any failing unit tests
- [x] New unit tests if there are gaps to cover
- [x] Build into the job or as a follow-on step away to "vacuum" or "garbage collect" or clean-up the temporary CSV files stored in S3 that are only used as an intermediary format for the `COPY` command
- [ ] Test that the resultant data is still valid and matches the target tables data (before swapping)

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [PIPE-380](https://federal-spending-transparency.atlassian.net/browse/PIPE-380):

```
```


usaspending_api/config/envs/default.py
```diff
@@ -212,18 +214,44 @@ def _validate_database_url(cls, values, db_url_conf_name, db_conf_prefix, requir
...
+        # noinspection PyArgumentList
+        cls._validate_database_url(
+            cls=cls, values=values, db_url_conf_name="DATABASE_URL", db_conf_prefix="USASPENDING_DB", required=True
+        )
+        return values
+
+    # noinspection PyMethodParameters
+    # Pydantic returns a classmethod for its validators, so the cls param is correct
+    @root_validator
+    def _DATA_BROKER_DATABASE_URL_and_parts_factory(cls, values):
```
dpb-bah, line 238: I'm fine with keeping this split. I think I merged them at one point worried that values was being populated going into the Broker call but that was an issue with testing on my part. I think this is fine.

usaspending_api/search/delta_models/award_search.py
```diff
@@ -76,7 +76,7 @@
     "pop_congressional_population": {"delta": "INTEGER", "postgres": "INTEGER"},
     "cfda_program_title": {"delta": "STRING", "postgres": "TEXT"},
     "cfda_number": {"delta": "STRING", "postgres": "TEXT"},
-    "cfdas": {"delta": "ARRAY<STRING>", "postgres": "[TEXT]"},
+    "cfdas": {"delta": "ARRAY<STRING>", "postgres": "[TEXT]"},  # TODO: <---- these pg text array types valid? = TEXT[]
```
dpb-bah, line 79: Yeah transaction_search seems to have it as `TEXT[]`. I thought I saw `[TEXT]` as valid and working but regardless, this is a good note to check/test soon.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -247,3 +293,219 @@ def handle(self, *args, **options):
...
+            csv_path = (
+                f"{CONFIG.SPARK_CSV_S3_PATH}/temp/{delta_db}/{delta_table_name}/"
+                f"{datetime.strftime(datetime.utcnow(), '%Y%m%d%H%M%S')}/"
+            )
+        s3_bucket_with_csv_path = f"s3a://{spark_s3_bucket_name}/{csv_path}"
+
+        logger.info(f"LOAD: Starting dump of Delta table to temp gzipped CSV files in {s3_bucket_with_csv_path}")
+        df_no_arrays = convert_array_cols_to_string(df, is_postgres_array_format=True, is_for_csv_export=True)
+        df_no_arrays.write.options(
+            maxRecordsPerFile=CONFIG.SPARK_PARTITION_ROWS * 3,  # e.g. 25k - 50k seems a good sweet spot
```
dpb-bah, line 360: Given this seems like a number based from our experience and less with the logic, I'd suggest moving this to the top as a constant if the sweet spot moves around depending on circumstances.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -247,3 +293,219 @@ def handle(self, *args, **options):
...
+        with psycopg2.connect(dsn=db_dsn) as connection:
+            with connection.cursor() as cursor:
+                cursor.execute("SHOW max_parallel_workers")
+                max_parallel_workers = int(cursor.fetchone()[0])
+                # For whatever reason, one-half of the max parallel workers seemed to be the sweet spot, but probably
+                # don't want to go below 8
+                partitions = max(ceil(max_parallel_workers / 2), partitions)
+
+        # Give more memory to each connection during the COPY operation of large files to avoid spillage to disk
+        work_mem_for_large_csv_copy = 256 * 1024  # MiB of work_mem * KiBs in 1 MiB
```
dpb-bah, line 406: Same suggestion as above, moving these to constants for ease of updating based on environments, etc.

usaspending_api/common/csv_stream_s3_to_pg.py
```diff
@@ -0,0 +1,182 @@
...
+
+The main functions: copy_csv_from_s3_to_pg and copy_csvs_from_s3_to_pg
+are used in distributed/parallel/multiprocess execution (by Spark) and is
+pickled via cloudpickle. As such it must not have any presumed setup code that would have run (like Django setup,
+logging configuration, etc.) and must encapsulate all of those dependencies (like logging config) on its own.
+
+Adding new imports to this module may inadvertenly introduce a dependency that can't be pickled.
+
+As it stands, even if new imports are added to the modules it already imports, it could lead to a problem.
+"""
```
dpb-bah, line 14: Just noting how fragile and worrisome this comment/description reads. That may be intentional and get the point across, just may be a TODO or tech debt to make this less fragile if this is the case.

dpb-bah (approved):
Approving with a minor suggestion of moving some values to be constants since they are related to performance which can vary over time (or configs if based on the environment).

usaspending_api/config/envs/default.py
kwhickey: Yeah, this is me uncertain of how the pydantic framework discovers and processes validators (and the `values` return value from them). I thought if 2 values are set in `values`, to do it in two steps so each value being set is deliberate.

usaspending_api/common/csv_stream_s3_to_pg.py
kwhickey: I think I got it as "stable" as I can. It's a matter of what other code gets latched on to this module or added to things it depends on. The unit tests that exercise this should blow up if any of those occurrences happen.

usaspending_api/search/delta_models/award_search.py
kwhickey: I think I might just change it in this PR.

dpb-bah (dismissed):

dpb-bah (dismissed):

dpb-bah (approved):

PR #3539, opened 2022-08-03
[PIPE-316] Recipient Profile Databricks

emily-brents:
**Description:**
Adapts the `restock_recipient_profile` sql script to wirk in databricks.

**Requirements for PR merge:**

1. [X] Unit & integration tests updated
2. [X] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [ ] Frontend
4. [X] Matview impact assessment completed
5. [X] Frontend impact assessment completed
6. [ ] Data validation completed
7. [ ] Appropriate Operations ticket(s) created
8. [X] Jira Ticket [PIPE-316](https://federal-spending-transparency.atlassian.net/browse/PIPE-316):
    - [X] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #3541, opened 2022-08-09
[PIPE-315] Implement Recipient Lookup ETL with Spark

sethstoudenmier:
**Description:**
Implemented Recipient Lookup with Spark and made updates to test cases.

**Technical details:**
* Implemented the Recipient Lookup ETL which required increasing the allocated memory in the MAKE command used to run the ETL locally
* Implemented test cases for Recipient Lookup `load_query_to_delta` command; first step in the test is to make sure initial load matches and then validate the MERGE INTO picks up new changes correctly
* Updated test cases to include the `hive_unittest_metastore_db` fixture; without this fixture I was running into issues where tests would fail locally if not run in a specific order
* Transaction and Award Search were updated temporary to pull from `test.recipient_lookup_testing` since that version is a direct copy of the Recipient Lookup table; if PIPE-373 is not deployed in the same sprint then the Nightly Pipeline will need to be updated to create `test.recipient_lookup_testing` temporarily

    - [ ] Backend
8. [ ] Jira Ticket [PIPE-315](https://federal-spending-transparency.atlassian.net/browse/PIPE-315):

```
```


sethstoudenmier:
Adding Do Not Merge briefly while I try to figure out a data difference discovered when running in QAT

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+    # -----
+    # Delete any instances of a NULL recipient_hash and handle cleanup following alternate names generation
+    # -----
+    r"""
+    DELETE FROM temp.temporary_restock_recipient_lookup
+    WHERE recipient_hash IS NULL OR row_num_union != 1
+    """,
+    # -----
+    # Merge the temporary_restock_recipient_lookup table into recipient_lookup
+    # -----
```
collinwr, line 544: Similar comment to one I made on Emily's PR:
>What's the purpose of populating a temporary table, temporary_restock_recipient_profile, then performing a series of merges to make recipient_profile match?
I believe this was done in Postgres to prevent a live table from being updated by a series of steps (1-9) that could fail, but because this DBR ETL is updating a table that will later be nuke/paved, would it make sense to clear recipient_profile at the start of the script and update directly? 

Any reason not to be populating Recipient Lookup directly instead of the temporary table?

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -61,6 +86,13 @@ class Command(BaseCommand):
     All existing data will be deleted before new data is written.
     """
 
+    # Values defined in the handler
+    auto_increment_max_id: Optional[int]
+    destination_database: str
+    destination_table_name: str
+    spark: SparkSession
+    spark_s3_buket: str
```
collinwr, line 94: Typo here: `spark_s3_buket`

collinwr: Although, if that temp table is no longer being used after updates based on my other comment, this parameter may no longer be necessary

usaspending_api/recipient/delta_models/recipient_lookup.py
sethstoudenmier: Following up on discussions from Slack.

Current thinking is that utilizing a separate temp table allows for slightly different columns between recipient_lookup and the temp_recipient_lookup without persisting those columns into DBR. The approach was changed however to use `INSERT OVERWRITE` instead of `MERGE INTO` in hopes that blowing away the old table and then inserting will be more performant than trying to go row by row.

usaspending_api/etl/management/commands/load_query_to_delta.py
sethstoudenmier: Good catch. Autocomplete definitely helped me to persist that throughout that page.

usaspending_api/recipient/delta_models/recipient_lookup.py
kwhickey: Ditto. 

basically could the temp table be the final table? It would exist in a half-baked state during the pipeline, but perhaps we have error handling that rolls it back to the previous version if a problem occurred

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+    # Populate recipient_lookup with incremented Primary Key fields
+    # -----
+    r"""
+    CREATE OR REPLACE TEMPORARY VIEW temp_recipient_lookup_view AS (
+        SELECT
+            {AUTO_INCREMENT_MAX_ID} + ROW_NUMBER() OVER (ORDER BY recipient_hash) AS new_id,
+            rl.*
+        FROM {DESTINATION_DATABASE}.{DESTINATION_TABLE} AS rl
+        WHERE rl.id = 0
+    )
```
kwhickey: Could we continue to leave the ID column off in delta, and just let the auto-incrementing ID col in the target postgres table create the PK ids? Are they meaningful to keep in delta?

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+            zip5,
+            action_date AS update_date,
+            NULL AS alternate_names,
+            ROW_NUMBER() OVER (PARTITION BY recipient_hash ORDER BY action_date DESC, is_fpds, transaction_unique_id) AS row_num
+        FROM temporary_transaction_recipients_view
+        WHERE COALESCE(uei, awardee_or_recipient_uniqu) IS NULL
+    ),
+    union_all AS (
+        SELECT * FROM latest_duns_sam WHERE row_num = 1
+        UNION ALL
```
kwhickey, line 420: Is UNION ALL what we want here? It collects all the rows from each step and doesn't preserve priority (i.e. if the SAME row is surfaced by priority 1 step and priority 2 step, it's not just priority 1 that remains, but both of them).

I guess a straight UNION, even if supported by Spark (not sure if they do both UNION and UNION ALL differently) wouldn't work if we're adding the priority in though. So maybe there's no way around it.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+        parent_legal_business_name,
+        parent_uei,
+        state,
+        zip4,
+        zip5,
+        alternate_names,
+        source,
+        update_date,
+        row_num_union
+    FROM union_all_priority
```
kwhickey, line 488: Should there be something that narrows down to top-priority items?
e.g. 
```
WHERE row_num_union = 1
```

kwhickey (commented):
No major blockers, but a few suggestions and some questions around optimizations/correctness.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,583 @@
...
+            NULL AS country_code,
+            ultimate_parent_unique_ide AS parent_duns,
+            UPPER(ultimate_parent_legal_enti) AS parent_legal_business_name,
+            ultimate_parent_uei AS parent_uei,
+            NULL AS state,
+            NULL AS zip4,
+            NULL AS zip5,
+            update_date,
+            NULL AS alternate_names,
+            ROW_NUMBER() OVER (PARTITION BY ultimate_parent_uei, ultimate_parent_unique_ide ORDER BY update_date DESC NULLS LAST) AS row_num
```
kwhickey, line 230: Similar to the `010` step, this should probably coalesce the two hash inputs, one of which could be null, on the partition:
`PARTITION BY COALESCE(ultimate_parent_uei, ultimate_parent_unique_ide) ORDER BY update_date DESC NULLS LAST) AS row_num`

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,583 @@
...
+            country_code,
+            ultimate_parent_unique_ide AS parent_duns,
+            UPPER(ultimate_parent_legal_enti) AS parent_legal_business_name,
+            ultimate_parent_uei AS parent_uei,
+            state,
+            zip4,
+            zip AS zip5,
+            update_date,
+            NULL AS alternate_names,
+            ROW_NUMBER() OVER (PARTITION BY uei, awardee_or_recipient_uniqu ORDER BY update_date DESC NULLS LAST) AS row_num
```
kwhickey, line 295: Same thing here, since one could be null, probably want to partition by a coalesce of both:

`PARTITION BY COALESCE(uei, awardee_or_recipient_uniqu) ORDER BY update_date DESC NULLS LAST) AS row_num`

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,583 @@
...
+            NULL AS country_code,
+            ultimate_parent_unique_ide AS parent_duns,
+            NULL AS parent_legal_business_name,
+            ultimate_parent_uei AS parent_uei,
+            NULL AS state,
+            NULL AS zip4,
+            NULL AS zip5,
+            update_date,
+            NULL AS alternate_names,
+            ROW_NUMBER() OVER (PARTITION BY ultimate_parent_uei, ultimate_parent_unique_ide ORDER BY update_date DESC NULLS LAST) AS row_num
```
kwhickey, line 360: Same thing here, since one could be null, `PARTITION BY COALESCE(ultimate_parent_uei, ultimate_parent_unique_ide) ORDER BY update_date DESC NULLS LAST) AS row_num`

usaspending_api/recipient/delta_models/recipient_lookup.py
kwhickey: The thought I had, not completely thought-out, was to `UNION` the datasets in reverse priority order, so last-one-wins, replacing the same rows from earlier datasets. Or something like that. But I guess ALL rows would have to be equal to overwrite, and it's only some of them that are. 
So maybe first collecting them in 1 `UNION ALL` dataset, and then using `PARTITION` to get the _1st_ row (based on e.g. the `ORDER BY priority` column) ... would achieve the same effect. --> which you have below in the `union_all_priority`. 

And PARTITIONing on `recipient_hash` (which I _think_ will _always_ have a value in the above CTEs) is the same as sequential `MERGE` statements with `recipient_hash` as the merge-matching-key. So I think all is good here 👍

usaspending_api/recipient/delta_models/recipient_lookup.py
kwhickey: Or ... is this deferred, so you can generate alternate names from the full/redundant dataset? I see below you later delete where `row_num_union != 1`

kwhickey (changes requested):
Round 2 review: 
I think there may need to be some coalesces when partitioning by the inputs to the hash (uei and duns numbers). Not sure if it will make a difference on the result of the partition if coalesced vs. kept separate. 

Some further questions too.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+            city,
+            congressional_district,
+            country_code,
+            ultimate_parent_unique_ide AS parent_duns,
+            ultimate_parent_legal_enti AS parent_legal_business_name,
+            ultimate_parent_uei AS parent_uei,
+            state,
+            zip4,
+            zip5,
+            action_date AS update_date,
```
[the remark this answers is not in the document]
kwhickey: Talked this through and its the most logical value for this -- as it is the same and first value use in ORDER BY of partition entires below.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+            NULL AS city,
+            NULL AS congressional_district,
+            NULL AS country_code,
+            ultimate_parent_unique_ide AS parent_duns,
+            ultimate_parent_legal_enti AS parent_legal_business_name,
+            ultimate_parent_uei AS parent_uei,
+            NULL AS state,
+            NULL AS zip4,
+            NULL AS zip5,
+            action_date AS update_date,
```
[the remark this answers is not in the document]
kwhickey: Talked this through and its the most logical value for this -- as it is the same and first value use in ORDER BY of partition entires below.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+            city,
+            congressional_district,
+            country_code,
+            ultimate_parent_unique_ide AS parent_duns,
+            ultimate_parent_legal_enti AS parent_legal_business_name,
+            ultimate_parent_uei AS parent_uei,
+            state,
+            zip4,
+            zip5,
+            action_date AS update_date,
```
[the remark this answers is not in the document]
kwhickey: Talked this through and its the most logical value for this -- as it is the same and first value use in ORDER BY of partition entires below.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+            NULL AS city,
+            NULL AS congressional_district,
+            NULL AS country_code,
+            ultimate_parent_unique_ide AS parent_duns,
+            NULL AS parent_legal_business_name,
+            ultimate_parent_uei AS parent_uei,
+            NULL AS state,
+            NULL AS zip4,
+            NULL AS zip5,
+            action_date AS update_date,
```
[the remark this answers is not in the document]
kwhickey: Talked this through and its the most logical value for this -- as it is the same and first value use in ORDER BY of partition entires below.

usaspending_api/recipient/delta_models/recipient_lookup.py
sethstoudenmier: Addressed this following example that Keith had shared to utilize a series of VIEWs.

usaspending_api/etl/management/commands/load_query_to_delta.py
sethstoudenmier: Removed this following changes mentioned above.

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -94,37 +133,65 @@ def handle(self, *args, **options):
...
+        create_ref_temp_views(self.spark)
+
+        # Make sure that the "temp" database exists since this is used by some
+        # load queries to store temporary tables
+        self.spark.sql(f"CREATE DATABASE IF NOT EXISTS temp;")
+
+        load_query = table_spec["source_query"]
+        if isinstance(load_query, list):
+            for query in load_query:
+                self.run_spark_sql(query)
```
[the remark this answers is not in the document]
sethstoudenmier: To avoid adding more clutter to the recipient_lookup file I changed this to log the index of the query and the first 100 characters.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+            tn.is_fpds,
+            REGEXP_REPLACE(
+                MD5(UPPER(
+                    CASE
+                        WHEN COALESCE(fpds.awardee_or_recipient_uei, fabs.uei) IS NOT NULL THEN CONCAT('uei-', COALESCE(fpds.awardee_or_recipient_uei, fabs.uei))
+                        WHEN COALESCE(fpds.awardee_or_recipient_uniqu, fabs.awardee_or_recipient_uniqu) IS NOT NULL THEN CONCAT('duns-', COALESCE(fpds.awardee_or_recipient_uniqu, fabs.awardee_or_recipient_uniqu))
+                        ELSE CONCAT('name-', COALESCE(fpds.awardee_or_recipient_legal, fabs.awardee_or_recipient_legal)) END
+                )),
+                '^(\.{{8}})(\.{{4}})(\.{{4}})(\.{{4}})(\.{{12}})$',
+                '\$1-\$2-\$3-\$4-\$5'
```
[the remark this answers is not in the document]
sethstoudenmier: Added a comment to clarify. Essentially this is how we can format MD5 to be a "UUID" as what would be expected in Postgres without the need for a UDF since all UUIDs should have the same formet.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,686 @@
...
+            tn.action_date,
+            CASE
+                WHEN tn.is_fpds = TRUE THEN CAST('fpds' AS STRING)
+                ELSE CAST('fabs' AS STRING)
+            END AS source
+        FROM raw.transaction_normalized AS tn
+        LEFT OUTER JOIN raw.transaction_fpds AS fpds ON (tn.id = fpds.transaction_id)
+        LEFT OUTER JOIN raw.transaction_fabs AS fabs ON (tn.id = fabs.transaction_id)
+        WHERE tn.action_date >= '2007-10-01'
+        ORDER BY tn.action_date DESC
```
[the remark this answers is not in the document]
sethstoudenmier: Shouldn't be needed and thus was removed. Most likely there from me copying it over and not paying attention to it.

usaspending_api/recipient/delta_models/recipient_lookup.py
sethstoudenmier: ID col was removed from delta following conversations outside of GitHub.

usaspending_api/recipient/delta_models/recipient_lookup.py
sethstoudenmier: For now this has been updated to limit with `WHERE row_num_union = 1`, however, I plan to create a bug ticket since this functionality is actually a bug since it limits the recipients that we use to generate alternate_names and also remove old recipients that are still aligned to their DUNS and not UEI.

usaspending_api/recipient/delta_models/recipient_lookup.py
sethstoudenmier: Mentioned this in the response above.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,583 @@
...
+            country_code,
+            ultimate_parent_unique_ide AS parent_duns,
+            UPPER(ultimate_parent_legal_enti) AS parent_legal_business_name,
+            ultimate_parent_uei AS parent_uei,
+            state,
+            zip4,
+            zip AS zip5,
+            update_date,
+            NULL AS alternate_names,
+            ROW_NUMBER() OVER (PARTITION BY uei, awardee_or_recipient_uniqu ORDER BY update_date DESC NULLS LAST) AS row_num
```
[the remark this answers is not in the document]
sethstoudenmier: In this implementation and the prototype code there is a CTE below
```
union_all_priority AS (
        SELECT
            *,
            ROW_NUMBER() OVER (PARTITION BY recipient_hash ORDER BY priority ASC) as row_num_union
        FROM union_all
    )
```

This would limit the value based on the hash which in the case of sam does not take into account the `legal_business_name`. Since it doesn't take into account the name it seems that it is wasted effort to try and partition by the name as well since two recipients with the same UEI and different names would get condensed into a single recipient anyways.

As for the COALESCE on UEI and DUNS I don't think it makes any difference. I have not tried with the COALESCE, but I have seen the data match without it.

usaspending_api/etl/tests/integration/test_load_to_from_delta.py
```diff
@@ -803,9 +1057,19 @@ def test_load_table_to_from_delta_for_award_search_testing(
...
 @mark.django_db(transaction=True)
 def test_load_table_to_from_delta_for_award_search_alt_db_and_name(
-    spark, s3_unittest_data_bucket, populate_data_for_transaction_search
+    spark, s3_unittest_data_bucket, populate_data_for_transaction_search, hive_unittest_metastore_db
 ):
-    create_and_load_all_delta_tables(spark, s3_unittest_data_bucket)
+    tables_to_load = [
+        "awards",
+        "financial_accounts_by_awards",
+        "recipient_lookup",
```
kwhickey, line 1065: For this, and other places where `create_and_load_all_delta_tables` is called, just take a look at probably next-sprint refactoring considering this comment: https://github.com/fedspendingtransparency/usaspending-api/pull/3539#discussion_r944792655

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,552 @@
...
+    (
+        {",".join([col for col in RECIPIENT_LOOKUP_COLUMNS_WITHOUT_ID])}
+    )
+    SELECT
+        {",".join([col for col in RECIPIENT_LOOKUP_COLUMNS_WITHOUT_ID])}
+    FROM
+        temp_recipients_with_alt_names_view
+    """,
+    # -----
+    # Cleanup the temporary table and views
```
kwhickey, line 546: Minor thing here to remove "table" from the comment, as I think all that's left is views.

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -1,30 +1,552 @@
...
+            FULL OUTER JOIN alt_parent_names AS apn ON an.recipient_hash = apn.recipient_hash
+        )
+        SELECT
+            {
+                ", ".join(
+                    [
+                        "temp_stale_recipients_removed_view." + val
+                        for val
+                        in set(RPT_RECIPIENT_LOOKUP_DELTA_COLUMNS) - {"alternate_names"}
+                    ]
```
kwhickey, line 519: Clever work here and elsewhere to cut down on column-copy-paste 👍

kwhickey (changes requested):
One thing for awareness, and one place to add table_spec entries (which may come in on a merge), and one comment to fix.

kwhickey (approved):
All looks good! Approved

PR #3543, opened 2022-08-10
[OPS-2361] Implement delta for uwsgi max lifetime restart

jasvatos:
**Description:**
We are periodically seeing instances in prod get replaced by the load balancer due to failing health checks on random intervals that sometimes round to 29 hours.
Uwsgi workers are currently set to restart after a lifetime of 29 hours. This causes all of the workers to go down at the same time if their lifetime is the same. This addresses the issue by implementing a delta to spread the restart times out.
Also removing an unsupported value for the max-requests-delta.

**Technical details:**

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [ ] Jira Ticket [OPS-2361](https://federal-spending-transparency.atlassian.net/browse/OPS-2361):

```
```

### DIFF
--- a/config/uwsgi.ini
+++ b/config/uwsgi.ini
@@ -40,5 +40,5 @@ cheaper-rss-limit-hard = 7730941133
 # Configuration to limit the lifespan of a worker to mitigate the
 #  risk of memory leaks in the application or dependencies
 max-requests = 1000
-max-requests-delta = 1
 max-worker-lifetime = 104400
+max-worker-lifetime-delta = 200
--- a/config/uwsgi.ini
+++ b/config/uwsgi.ini
@@ -41,4 +41,4 @@ cheaper-rss-limit-hard = 7730941133
 #  risk of memory leaks in the application or dependencies
 max-requests = 1000
 max-worker-lifetime = 104400
-max-worker-lifetime-delta = 200
+max-worker-lifetime-delta = 20


config/uwsgi.ini
```diff
@@ -40,5 +40,5 @@ cheaper-rss-limit-hard = 7730941133
 # Configuration to limit the lifespan of a worker to mitigate the
 #  risk of memory leaks in the application or dependencies
 max-requests = 1000
-max-requests-delta = 1
 max-worker-lifetime = 104400
+max-worker-lifetime-delta = 20
```
kwhickey, line 44: @jasvatos - I'm assuming this is in seconds. Is 20 seconds enough time for an instance to go down and come back up operational before the next instance goes down?

kwhickey: Or ... this is the worker process... right not a full instance. 
So --> enough time for a single worker process to be replaced and report healthy before its sibling gets recycled?

jasvatos: It is in seconds and doesn't wait for another instance to report healthy. However, the delta is a calculation of `worker_id * delta`. 
With a default of 50 workers, the first worker restarts 20 seconds after the max lifetime is reached. 
The 50th worker will restart 1000 seconds after the max lifetime is reached.

kwhickey (approved):
looks good. Approving.

PR #3547, opened 2022-08-12
[PIPE-284] Narrowing `extract_db_data_frame` to get partition bounds using unique partition cols

kwhickey:
**Description:**
When running SQL against Postgres tables to get min, max, count in order to build out partition bounds for the JDBC read, do so with SQL that can take advantage of the fact that the partition col has no duplicate values in the table. 

**Technical details:**
The `get_partition_bounds_sql` function used by `extract_db_data_frame` has always had a bool parameter for whether the col used for partition bounds was unique or not. We had been cautiously setting this always to `False`, just to get things up and running, but never tuned it to each target table col's uniqueness. Fixing that and adding partition col uniqueness as a metadata entry in the `TABLE_SPEC` dictionaries. 

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [x] Jira Ticket [PIPE-284](https://federal-spending-transparency.atlassian.net/browse/PIPE-284):

```
```


usaspending_api/etl/management/commands/load_table_to_delta.py
```diff
@@ -313,18 +325,18 @@ def handle(self, *args, **options):
...
+                jdbc_url=jdbc_url,
+                partition_rows=SPARK_PARTITION_ROWS,
+                min_max_sql=get_partition_bounds_sql(
+                    table_name=source_table,
+                    partitioning_col_name=partition_column,
+                    partitioning_col_alias=partition_column,
+                    is_partitioning_col_unique=is_partition_column_unique,
                 ),
-                source_table,
-                partition_column,
```
collinwr, line 327: Nice changes on this call to include named parameters 👍

usaspending_api/etl/management/commands/load_table_to_delta.py
```diff
@@ -136,6 +141,7 @@
         "swap_schema": None,
         "partition_column": "published_fabs_id",
         "partition_column_type": "numeric",
+        "is_partition_column_unique": True,
```
collinwr, line 144: There isn't actually a unique constraint on the partition columns for `transaction_fabs` and `transaction_fpds` (`published_fabs_id` and `detached_award_procurement_id` respectively). Looking in the code to see if there is another way we can guarantee these are unique.

kwhickey: Good point. Should we really be using those columns then? Why not just use the guaranteed unique (and indexed) `transaction_id` col for each one?

collinwr: Looks like those fields are PKs/unique on `source_assistance_transaction` and `source_procurement_transaction`, which should be 1 to 1 with fabs/fpds tables, so we're good here.

collinwr: I think you're right that `transaction_id` could be a better column for partitioning. Want to include that change? We can do a test run in staging over the weekend.

collinwr (approved):
Verified the accuracy of each `is_partition_column_unique` field. Approving

PR #3555, opened 2022-08-22
[PIPE-322] Few changes to allow copy back of Recipient Lookup

sethstoudenmier:
**Description:**
Few changes that allow the copy back of Recipient Lookup to 

**Technical details:**
* docker-compose.yaml was updated to point to Postgres 13 image to allow use of `INCLUDING GENERATED` in the loader
* `INCLUDING GENERATED` was added to copy back references to the sequence
* `load_table_from_delta` was updated to optionally reset a sequence; resets the sequence back to previous value if the job fails
* updated `travis.yml` to use Postgres 13 as well

    - [ ] Backend
8. [ ] Jira Ticket [PIPE-322](https://federal-spending-transparency.atlassian.net/browse/PIPE-322):

```
```


usaspending_api/etl/management/commands/copy_table_metadata.py
```diff
@@ -14,7 +14,7 @@
 # since that script uses relative imports which do not work well with the rest of the repo code
 TEMPLATE = {
     "read_indexes": "SELECT indexname, indexdef FROM pg_indexes WHERE schemaname = '{}' AND tablename = '{}';",
-    "read_constraints": "select conname, pg_get_constraintdef(oid) from pg_constraint where contype IN ('p', 'f') and conrelid = '{}'::regclass;",
+    "read_constraints": "select conname, pg_get_constraintdef(oid) from pg_constraint where contype IN ('p', 'f', 'u') and conrelid = '{}'::regclass;",
```
kwhickey, line 17: I felt like we went through a discussion on this with transaction_search and award_search, and decided to NOT copy unique constraints.

Or ... did we just decide to not copy them for _those_ tables, because they weren't actually necessary on those tables. So we just kept on _not copying them_? 

If the latter, what is it about the unique constraints you're seeing that makes you feel like we ought to keep them, rather than discard them like we did for award_search and transaction_search?

usaspending_api/recipient/delta_models/recipient_lookup.py
```diff
@@ -26,7 +26,7 @@
 
 RECIPIENT_LOOKUP_DELTA_COLUMNS = {k: v["delta"] for k, v in RECIPIENT_LOOKUP_COLUMNS.items()}
 RPT_RECIPIENT_LOOKUP_DELTA_COLUMNS = {k: v["delta"] for k, v in RECIPIENT_LOOKUP_COLUMNS_WITHOUT_ID.items()}
-RECIPIENT_LOOKUP_POSTGRES_COLUMNS = {k: v["postgres"] for k, v in RECIPIENT_LOOKUP_COLUMNS.items()}
+RECIPIENT_LOOKUP_POSTGRES_COLUMNS = {k: v["postgres"] for k, v in RECIPIENT_LOOKUP_COLUMNS_WITHOUT_ID.items()}
```
kwhickey, line 29: When did the postgres columns transition over to the column set that was _without ID_? I thought the idea was we were (maybe not permanently) keeping ID in there and letting it auto-populate with the sequence?

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -100,6 +104,12 @@ def add_arguments(self, parser):
             default=CONFIG.SPARK_S3_BUCKET,
             help="The destination bucket in S3 to write the data",
         )
+        parser.add_argument(
+            "--reset-sequence",
+            action="store_true",
+            help="In the case of a Postgres sequence for the provided 'delta-table' the sequence will be reset to 1. "
+            "If the job fails for some unexpected reason then the sequence will be reset to the previous value.",
```
collinwr, line 111: Is this second line of the parameter help info directly related to the parameter? Looks like the sequence value is reset to the previous value whether this argument is used or not.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -294,15 +318,31 @@ def handle(self, *args, **options):
...
+
+        Returns the previous value use by the sequence.
+        """
+        new_seq_val = val if val else 1
+        self.logger.info(f"Setting the Postgres sequence to {new_seq_val} for: {seq_name}")
+        with db.connection.cursor() as cursor:
+            cursor.execute(f"SELECT last_value FROM {seq_name}")
+            last_value = cursor.fetchone()[0]
+            cursor.execute(f"ALTER SEQUENCE IF EXISTS {seq_name} RESTART WITH {new_seq_val}")
+        return last_value
```
kwhickey, line 342: Is the plan still to see if we can drop this as technical debt, by dropping the PK column from the table so that no sequence needs to be maintained? 
What are the remaining challenges to doing so? Django having a PK?

kwhickey (commented):
Some questions for clarification came up in the review.

usaspending_api/recipient/delta_models/recipient_lookup.py
sethstoudenmier: The table is created via the `LIKE` command which created the `id` column that references the sequence. This statement is all of the columns we are looking to copy back from Delta to Postgres. Not having `id` specified means that it will be auto populated by the sequence.

usaspending_api/etl/management/commands/load_table_from_delta.py
sethstoudenmier: >Is the plan still to see if we can drop this as technical debt, by dropping the PK column from the table so that no sequence needs to be maintained?
Not entirely sure if that is planned technical debt or not.

>What are the remaining challenges to doing so? Django having a PK?
Most likely it is just that, but we could trick Django into thinking a PK column exists. The problem is that I have no idea what issues that could cause if Django expects something like a PK but doesn't have one.

usaspending_api/etl/management/commands/load_table_from_delta.py
sethstoudenmier: Sequence value is only reset if there is a value for `postgres_seq_last_value`; which would only exist if it was previously changed.

collinwr: Okay gotcha. I'm following

kwhickey (dismissed):
Questions answered, and approving for now and can again once caught up with merges.

collinwr (approved):

PR #3544, opened 2022-08-10
[PIPE-327] Delta raw.broker_subaward -> rpt.subaward_search

dpb-bah:
**Description:**
Initial delta model for rpt.subaward_search and configuration to populate it via `load_query_to_delta` 

**Technical details:**
* Split out subawards from `test_load_to_from_delta`

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. API documentation updated
3. [x] Necessary PR reviewers:
    - [x] Backend
    - Frontend <OPTIONAL>
    - Operations <OPTIONAL>
    - Domain Expert <OPTIONAL>
4. Matview impact assessment completed
5. Frontend impact assessment completed
6. [x] Data validation completed
7. Appropriate Operations ticket(s) created
8. [x] Jira Ticket [PIPE-327](https://federal-spending-transparency.atlassian.net/browse/PIPE-327):
    - [x] Link to this Pull-Request
    - [x] Performance evaluation of affected (API | Script | Download) (10min to load)
    - [x] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #3556, opened 2022-08-23
[PIPE-327] Delta rpt.subaward_search -> Postgres

dpb-bah:
**Description:**
SubawardSearch model and configuration to load it from databricks. Note: Not including the cutover code which will be a separate PR.

**Technical details:**
Includes alembic migration for SubawardSearch.

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
7.  Appropriate Operations ticket(s) created
    - [x] Performance evaluation of affected (API | Script | Download) (around 6-7 min)
9. [ ] Merged after https://github.com/fedspendingtransparency/usaspending-api/pull/3544

```
```


usaspending_api/database_scripts/matviews/functions_and_enums.sql
```diff
@@ -1,8 +1,5 @@
 -- Postgres extensions, functions, and enums necessary for some views
 
-CREATE EXTENSION IF NOT EXISTS intarray;
-
-
```
kwhickey, line 5: Did you intend to remove this? Are you confirming that no other SQL code somehow needed this postgres extension? Or maybe it's native now in our version of Postgres?

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -57,6 +58,7 @@
         "custom_schema": "recipient_hash STRING, federal_accounts STRING, cfdas ARRAY<STRING>,"
         " tas_components ARRAY<STRING>",
         "column_names": list(AWARD_SEARCH_COLUMNS),
+        "tsvectors": None,
```
kwhickey, line 61: We're having a lot of dict keys pop up. A good unit test would be to fetch the `TABLE_SPEC` (combined), and throw an error if any key is in one entry that's missing from any other entry.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -25,6 +25,8 @@
 
 from usaspending_api.etl.management.commands.create_delta_table import TABLE_SPEC
 
+DEFAULT_TEXT_SEARCH_CONFIG = "pg_catalog.simple"
```
kwhickey, line 28: While this is used here, it feels like it should get brought up to a higher level. Like `settings.py` ... or if it's only used in spark stuff for now, possibly in `usaspending_api.config.envs.default.py`

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -239,6 +242,19 @@ def handle(self, *args, **options):
                     cursor.execute(create_temp_sql)
                     logger.info(f"{temp_table} created.")
 
+                    # If there are vectors, add the triggers that will populate them based on other calls
+                    logger.info(f"Adding tsvector triggers for the following columns: {list(tsvectors.keys())}")
+                    for tsvector_name, derived_from_cols in tsvectors.items():
```
kwhickey, line 247: Does this need to be guarded by a check that `tsvectors` is not `None`. The `.get(...)` above could allow it to be. And if we instead default it to empty list, the log statement should be moved into the for loop or otherwise dealt with so that we don't log about an empty dict of tsvectors

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -239,6 +242,19 @@ def handle(self, *args, **options):
...
+                        derived_from_cols_str = ", ".join(derived_from_cols)
+                        tsvector_trigger_sql = f"""
+                            CREATE TRIGGER tsvector_update_{tsvector_name} BEFORE INSERT OR UPDATE
```
kwhickey, line 250: Any way to do "if not exists" for triggers? Given kerfuffle and error states with swaping, and copying-back, we may find some db objects already existing.

usaspending_api/search/migrations/0010_subaward_search.py
```diff
@@ -0,0 +1,487 @@
...
+                ('keyword_ts_vector', django.contrib.postgres.search.SearchVectorField()),
+                ('award_ts_vector', django.contrib.postgres.search.SearchVectorField()),
+                ('recipient_name_ts_vector', django.contrib.postgres.search.SearchVectorField()),
+                ('award', models.BigIntegerField(null=True, db_column="award_id")),
+                ('awarding_agency', models.IntegerField(null=True, db_column="awarding_agency_id")),
+                ('cfda', models.IntegerField(null=True, db_column="cfda_id")),
+                ('funding_agency', models.IntegerField(null=True, db_column="funding_agency_id")),
+            ],
+            options={
+                'db_table': 'rpt"."subaward_search',
```
kwhickey, line 217: Is this a typo, or am I just having a hard time parsing ...?
`'db_table': 'rpt"."subaward_search'`

usaspending_api/search/models/subaward_search.py
```diff
@@ -0,0 +1,450 @@
...
+    sub_place_of_perform_city_code = models.TextField(null=True, blank=True)
+    sub_place_of_perform_congressio = models.TextField(null=True, blank=True)
+
+    # USAS Vectors
+    keyword_ts_vector = SearchVectorField()
+    award_ts_vector = SearchVectorField()
+    recipient_name_ts_vector = SearchVectorField()
+
+    class Meta:
+        db_table = 'rpt"."subaward_search'
```
kwhickey, line 223: Same comment as before, and this is where it came from in the migration. Is this the right quoting syntax?

kwhickey (changes requested):
A few minor changes and other places needing clarification.

usaspending_api/database_scripts/matviews/functions_and_enums.sql
dpb-bah: It's included in the migration for `subaward_search` as its needed for the vectors.

usaspending_api/etl/management/commands/load_query_to_delta.py
dpb-bah: Added a quick unit test for this.

usaspending_api/etl/management/commands/load_table_from_delta.py
dpb-bah: It's just the default setting of our Postgres DBs - so ill just move it to settings.py.

usaspending_api/etl/management/commands/load_table_from_delta.py
dpb-bah: Updating it to `tsvectors = table_spec.get("tsvectors") or []` which should account for if it's not included *and* if they do include it in the TABLE_SPEC but set it to `None`.

usaspending_api/etl/management/commands/load_table_from_delta.py
dpb-bah: Postgres seems to have something like that in Postgres 14. In the meantime, I'll just drop it if it does exist to be safe.

https://stackoverflow.com/questions/35927365/create-or-replace-trigger-postgres

usaspending_api/search/models/subaward_search.py
dpb-bah: (same response as the other)

kwhickey (dismissed):
Changes look good. Approving.

kwhickey (approved):
Approving after qat merge and CI pass.

PR #3562, opened 2022-08-26
[Pipe 311] Load Sam_recipient Table from Broker

bkschmidt:
**Description:**
Add support to load the sam_recipient table from broker to load_query_to_delta.py

**Technical details:**
The high level tech approach for this story is to create a jdbc proxy view on the broker database sam_recipient table. Once this view has been created, SQL very similar to that in the existing restock_duns.sql file can be used to transform the data into the required schema.

FOR ADDITIONAL CONVERSATION, SEE THIS PR: https://github.com/fedspendingtransparency/usaspending-api/pull/3506

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [DEV-123](https://federal-spending-transparency.atlassian.net/browse/PIPE-311):

```
```


.pre-commit-config.yaml
```diff
@@ -9,4 +9,5 @@ repos:
     rev: 20.8b1
     hooks:
     -   id: black
+        additional_dependencies: [ 'click==8.0.4' ]
```
sethstoudenmier, line 12: Would remove this since it isn't needed and could cause confusion by having this library set in two places.

usaspending_api/common/etl/spark.py
```diff
@@ -31,10 +31,13 @@
...
     DisasterEmergencyFundCode,
 )
+from usaspending_api.recipient.delta_models import (
+    sam_recipient,
+)
 from usaspending_api.submissions.models import SubmissionAttributes, DABSSubmissionWindowSchedule
 
 MAX_PARTITIONS = CONFIG.SPARK_MAX_PARTITIONS
-_RDS_REF_TABLES = [
+usas_RDS_REF_TABLES = [
```
sethstoudenmier, line 40: Constants have a convention of all upper case names. Please change to `USAS_RDS_REF_TABLES`.

usaspending_api/common/etl/spark.py
```diff
@@ -54,6 +57,10 @@
     DABSSubmissionWindowSchedule,
 ]
 
+_Broker_Proxy_tables = [
```
sethstoudenmier, line 60: Similar to the above, Python convention is that constants should be all uppercase. If not a constant then make all lowercase.

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -85,6 +89,23 @@
         "custom_schema": "recipient_hash STRING",
         "column_names": [x for x in list(RECIPIENT_PROFILE_DELTA_COLUMNS) if x != "id"],
     },
+    "int.sam_recipient": {
+        "model": None,
+        "is_from_broker": True,
+        "source_query": sam_recipient_load_sql_string,
+        "source_database": None,
+        "source_table": None,
+        "destination_database": "int",
```
sethstoudenmier, line 98: Shouldn't this be going into `raw`?

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -85,6 +89,23 @@
         "custom_schema": "recipient_hash STRING",
         "column_names": [x for x in list(RECIPIENT_PROFILE_DELTA_COLUMNS) if x != "id"],
     },
+    "int.sam_recipient": {
```
sethstoudenmier, line 92: Since you removed the other `sam_recipient` from `load_table_to_delta` there shouldn't be a need to have `int.` in the name.

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -85,6 +89,23 @@
...
+        "model": None,
+        "is_from_broker": True,
+        "source_query": sam_recipient_load_sql_string,
+        "source_database": None,
+        "source_table": None,
+        "destination_database": "int",
+        "swap_table": "sam_recipient",
+        "swap_schema": "int",
+        "partition_column": "awardee_or_recipient_uniqu",
```
sethstoudenmier, line 101: Shouldn't this be the `broker_duns_id`? That is the PK on `sam_recipient` for Broker.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,90 @@
...
     CREATE OR REPLACE TABLE {{DESTINATION_TABLE}} (
-        {", ".join([f'{key} {val}' for key, val in SAM_RECIPIENT_COLUMNS.items()])}
+        {", ".join([f'{key} {val}' for key, val in SAM_RECIPIENT_DELTA_COLUMNS.items()])}
     )
     USING DELTA
     LOCATION 's3a://{{SPARK_S3_BUCKET}}/{{DELTA_LAKE_S3_PATH}}/{{DESTINATION_DATABASE}}/{{DESTINATION_TABLE}}'
     """
+sam_recipient_load_sql_string = [
+    rf"""
+CREATE OR REPLACE TEMPORARY VIEW Source_data AS (
```
sethstoudenmier, line 34: Would lowercase this name from `Source_data` to `source_data`

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,90 @@
...
+    broker_sam_recipient.zip AS zip,
+    broker_sam_recipient.zip4 AS zip4,
+    broker_sam_recipient.business_types_codes AS business_types_codes,
+    broker_sam_recipient.dba_name as dba_name,
+    broker_sam_recipient.entity_structure as entity_structure,
+    broker_sam_recipient.uei AS uei,
+    broker_sam_recipient.ultimate_parent_uei AS ultimate_parent_uei
+  FROM
+    Source_data
+    Where Source_data.row_num !=0
```
sethstoudenmier, line 86: `If you are going to uppercase other SQL terms then do the same here with `Where`.  Also adjust the indentation to match the formatting that you have.

sethstoudenmier: `ROW_NUMBER` starts at the value 1 and so this condition will match all data. This should instead look only for those rows where `row_num = 1` since that will be the first item from the "distinct" group; matching the current Postgres `DISTINCT ON` implementation.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,90 @@
...
+        broker_sam_recipient.sam_recipient_id,
+        COALESCE(broker_sam_recipient.activation_date, broker_sam_recipient.deactivation_date) as record_date,
+        broker_sam_recipient.uei,
+        broker_sam_recipient.ultimate_parent_uei
+      FROM
+        global_temp.sam_recipient as broker_sam_recipient
+);""",
+    rf"""INSERT OVERWRITE {{DESTINATION_DATABASE}}.{{DESTINATION_TABLE}} AS (
+  SELECT
+    broker_sam_recipient.awardee_or_recipient_uniqu AS awardee_or_recipient_uniqu,
```
sethstoudenmier, line 65: All of these selects reference `broker_sam_recipient` but that doesn't exist in this scope.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,90 @@
...
-    "business_types_codes": "ARRAY<STRING>",
-    "dba_name": "STRING",
-    "entity_structure": "STRING",
-    "uei": "STRING",
-    "ultimate_parent_uei": "STRING",
+    "awardee_or_recipient_uniqu": {"delta": "STRING", "postgres": "TEXT"},
+    "legal_business_name": {"delta": "STRING", "postgres": "TEXT"},
+    "ultimate_parent_unique_ide": {"delta": "STRING", "postgres": "TEXT"},
+    "broker_duns_id": {"delta": "STRING", "postgres": "TEXT"},
+    "update_date": {"delta": "STRING", "postgres": "TEXT"},
```
sethstoudenmier, line 6: This should be a `DATE`

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,90 @@
...
-    "zip4": "STRING",
-    "business_types_codes": "ARRAY<STRING>",
-    "dba_name": "STRING",
-    "entity_structure": "STRING",
-    "uei": "STRING",
-    "ultimate_parent_uei": "STRING",
+    "awardee_or_recipient_uniqu": {"delta": "STRING", "postgres": "TEXT"},
+    "legal_business_name": {"delta": "STRING", "postgres": "TEXT"},
+    "ultimate_parent_unique_ide": {"delta": "STRING", "postgres": "TEXT"},
+    "broker_duns_id": {"delta": "STRING", "postgres": "TEXT"},
```
sethstoudenmier, line 5: This should be an `INTEGER` I believe

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,90 @@
...
+        COALESCE(broker_sam_recipient.business_types_codes, text[]) AS business_types_codes,
+        broker_sam_recipient.entity_structure,
+        broker_sam_recipient.sam_recipient_id,
+        COALESCE(broker_sam_recipient.activation_date, broker_sam_recipient.deactivation_date) as record_date,
+        broker_sam_recipient.uei,
+        broker_sam_recipient.ultimate_parent_uei
+      FROM
+        global_temp.sam_recipient as broker_sam_recipient
+);""",
+    rf"""INSERT OVERWRITE {{DESTINATION_DATABASE}}.{{DESTINATION_TABLE}} AS (
```
sethstoudenmier, line 63: Go look at `transaction_search_load_sql_string` and follow a similar pattern to how `INSERT OVERWRITE` is used.

sethstoudenmier (changes requested):
Initial pass with some requested changes.

usaspending_api/common/etl/spark.py
bkschmidt: Renamed and updated.

usaspending_api/common/etl/spark.py
bkschmidt: Renamed and updated

usaspending_api/etl/management/commands/load_query_to_delta.py
bkschmidt: changed it to raw; if needs to be changed to int, can be done.

usaspending_api/etl/management/commands/load_query_to_delta.py
bkschmidt: naming convention simplified.

usaspending_api/etl/management/commands/load_query_to_delta.py
bkschmidt: Corrected1

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: renamed view to source_data.

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: Thank you for clarifying the DISTINCT ON implementation.

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: references to broker_sam_recipient renamed to source_data

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: Changed back to Date

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: Changed to Integer

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: changed to follow transaction_search_load_sql_string pattern.

.pre-commit-config.yaml
bkschmidt: Removed

usaspending_api/recipient/delta_models/sam_recipient.py
sethstoudenmier: Still shows `STRING`

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,94 @@
...
+    source_data.zip AS zip,
+    source_data.zip4 AS zip4,
+    source_data.business_types_codes AS business_types_codes,
+    source_data.dba_name as dba_name,
+    source_data.entity_structure as entity_structure,
+    source_data.uei AS uei,
+    source_data.ultimate_parent_uei AS ultimate_parent_uei
+  FROM
+    source_data
+  WHERE Source_data.row_num =1
```
sethstoudenmier, line 90: Typo here; it is `source_data` not `Source_data`

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,94 @@
...
+      FROM
+        global_temp.sam_recipient as broker_sam_recipient
+);""",
+    rf"""
+  INSERT OVERWRITE {{DESTINATION_DATABASE}}.{{DESTINATION_TABLE}}
+    (
+        {",".join([col for col in SAM_RECIPIENT_DELTA_COLUMNS])}
+    )
+  SELECT
+    source_data.awardee_or_recipient_uniqu AS awardee_or_recipient_uniqu,
```
sethstoudenmier, line 69: Not worth blocking over, but the use of `AS` in most of these is redundant.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,94 @@
...
+        broker_sam_recipient.ultimate_parent_legal_enti,
+        broker_sam_recipient.address_line_1,
+        broker_sam_recipient.address_line_2,
+        broker_sam_recipient.city,
+        broker_sam_recipient.state,
+        broker_sam_recipient.zip,
+        broker_sam_recipient.zip4,
+        broker_sam_recipient.country_code,
+        broker_sam_recipient.congressional_district,
+        COALESCE(broker_sam_recipient.business_types_codes, text[]) AS business_types_codes,
```
sethstoudenmier, line 54: You cannot coalesce to a type: `text[]`

usaspending_api/common/etl/spark.py
```diff
@@ -54,6 +57,10 @@
     DABSSubmissionWindowSchedule,
 ]
 
+_BROKER_PROXY_TABLES = [
```
sethstoudenmier, line 60: The `_` in at the beginning of the name is used to denote that something is private since Python doesn't have private types. Can remove that since you import that below.

usaspending_api/common/etl/spark.py
```diff
@@ -519,3 +526,19 @@ def create_ref_temp_views(spark: SparkSession):
...
+
+    logger.info(f"Creating the following tables under the global_temp database: {broker_ref_tables}")
+    for broker_rdf_table in broker_ref_tables:
+        spark_sql = f"""
+         CREATE OR REPLACE GLOBAL TEMPORARY VIEW {broker_rdf_table}
+         USING JDBC
+         OPTIONS (
+           driver '{jdbc_conn_props["driver"]}',
+           fetchsize '{jdbc_conn_props["fetchsize"]}',
+           url '{get_usas_jdbc_url()}',
```
sethstoudenmier, line 539: You are using the USAspending JDBC connection here to reference data that will live on Broker.

sethstoudenmier (changes requested):

usaspending_api/common/etl/spark.py
bkschmidt: changed to get_broke_jdbc

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: I think something like this should work? CAST(COALESCE(broker_sam_recipient.business_types_codes, 0), AS TEXT)

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: Worth Changing though.

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: changed

usaspending_api/common/etl/spark.py
bkschmidt: changed out.

usaspending_api/etl/tests/integration/test_spark_app.py
```diff
@@ -250,6 +250,20 @@ def test_create_ref_temp_views(spark: SparkSession):
...
-    for rds_ref_table in _RDS_REF_TABLES:
+    for rds_ref_table in USAS_RDS_REF_TABLES:
+        spark_count = spark.sql(f"select count(*) from global_temp.{rds_ref_table._meta.db_table}").collect()[0][0]
+        assert rds_ref_table.objects.count() == spark_count
+
+
+def test_create_ref_broker_views(spark: SparkSession):
+    # Add dummy data to each test views
+    for rds_ref_table in _BROKER_PROXY_TABLES:
+        baker.make(rds_ref_table)
```
sethstoudenmier, line 261: Not going to be able to use model_bakery to create Broker tables.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,94 @@
...
+      FROM
+        global_temp.sam_recipient as broker_sam_recipient
+);""",
+    rf"""
+  INSERT OVERWRITE {{DESTINATION_DATABASE}}.{{DESTINATION_TABLE}}
+    (
+        {",".join([col for col in SAM_RECIPIENT_DELTA_COLUMNS])}
+    )
+  SELECT
+    source_data.awardee_or_recipient_uniqu,
```
sethstoudenmier, line 69: Since you have all of the defined columns above I would recommend replacing the SELECT to be more like https://github.com/fedspendingtransparency/usaspending-api/blob/qat/usaspending_api/recipient/delta_models/recipient_lookup.py#L597.

That way the order of the columns isn't quite as fragile.

sethstoudenmier (changes requested):
Couple more small comments. Also, I don't see any tests to actually test the load of sam_recipient to delta.

usaspending_api/recipient/delta_models/sam_recipient.py
bkschmidt: Good call, changed to follow the stable pattern.

usaspending_api/etl/tests/integration/test_load_to_from_delta.py
```diff
@@ -682,6 +701,68 @@ def verify_delta_table_loaded_to_delta(
     assert equal_datasets(dummy_data, received_data, TABLE_SPEC[delta_table_name]["custom_schema"], ignore_fields)
 
 
+def verify_delta_query_loaded_to_delta(
```
kwhickey, line 704: Maybe there's a weird merge history. This isn't all new here is it? Doesn't this python module already have all this code. Why would it be newly added in this PR.

kwhickey: Or was it just moved around in the module?

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,76 @@
...
+        broker_sam_recipient.country_code,
+        broker_sam_recipient.congressional_district,
+        CAST(COALESCE(broker_sam_recipient.business_types_codes, 0), AS TEXT) AS business_types_codes,
+        broker_sam_recipient.entity_structure,
+        broker_sam_recipient.sam_recipient_id,
+        COALESCE(broker_sam_recipient.activation_date, broker_sam_recipient.deactivation_date) as record_date,
+        broker_sam_recipient.uei,
+        broker_sam_recipient.ultimate_parent_uei
+      FROM
+        global_temp.sam_recipient as broker_sam_recipient
```
kwhickey, line 61: I guess I see why a broker proxy table was added now. 

Since the table is only used for the scope of connecting to and loading from the Broker DB over JDBC ... I feel like perhaps creating a DataFame for that table, and then `df.createOrReplaceTempView("broker_sam_recipient")` or something like that might be a better way to only make that remote table available over JDBC when needed for the load.

usaspending_api/recipient/delta_models/sam_recipient.py
```diff
@@ -1,29 +1,76 @@
...
+    rf"""
+  INSERT OVERWRITE {{DESTINATION_DATABASE}}.{{DESTINATION_TABLE}}
+    (
+        {",".join([col for col in SAM_RECIPIENT_DELTA_COLUMNS])}
+    )
+  SELECT
+        {",".join([col for col in SAM_RECIPIENT_DELTA_COLUMNS])}
+  FROM
+    source_data
+  WHERE source_data.row_num =1
```
kwhickey, line 72: Minor spacing issue here on `=1`

kwhickey (changes requested):
A few questions on the approach. I think there's a better way to get at the broker DB table then setting up a global temp view as if it were reference data. Worth discussing.

usaspending_api/recipient/delta_models/sam_recipient.py
kwhickey: In talking with Seth, I think given we're going to make this `int.sam_recipient`, and copy the ETL logic/procedure as it exists now in the pipeline to render USAS PG `raw.duns`, this could be left as `STRING` (since that's what it is in the USAS db). Eventually this delta data will be used to replace data in that pg table, so to minimize refactoring, I think keeping it as a STRING initially is correct. But we should take on tech-debt to do a deliberate refactor to `INTEGER`

usaspending_api/etl/tests/integration/test_load_to_from_delta.py
kwhickey: Was not in the latest PR.

usaspending_api/recipient/delta_models/sam_recipient.py
kwhickey: Changed to inline temp view.

usaspending_api/common/etl/spark.py
```diff
@@ -519,3 +527,19 @@ def create_ref_temp_views(spark: SparkSession):
         """
         spark.sql(spark_sql)
     logger.info(f"Created the reference views in the global_temp database")
+    broker_ref_tables = [rds_ref_table._meta.db_table for rds_ref_table in BROKER_PROXY_TABLES]
```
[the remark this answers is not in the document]
kwhickey: Removed, and changed to spark sql inline temp view.

kwhickey (approved):
Approving review for merge. ✅

PR #3346, opened 2022-01-27
[DEV-8368] Update Status of Fund Endpoint to use File A/B

collinwr:
**Description:**
This PR updates the following two endpoints to use Files A and B as the primary data source rather than GTAS.
- `/v2/agency/{toptier_code}/subcomponents`
- `/v2/agency/toptier_code/sub_components/bureau_slug`

These are the formulas used to calculate Obligations, Outlays, and Total Budgetary Resources:
- **Obligations** - File B - `total_budgetary_resources_amount_cpe`
- **Outlays** - File B - `gross_outlay_amount_by_program_object_class_cpe`
- **Total Budgetary Resources** - File A - `obligations_incurred_by_program_object_class_cpe`

**Technical details:**
Because these fields are coming from different files, I aggregated both files separately with different queries then combined the results before formatting them. 

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [ ] Domain Expert <OPTIONAL>
4. [x] Matview impact assessment completed
5. [x] Frontend impact assessment completed
6. [ ] Data validation completed
7. [x] Appropriate Operations ticket(s) created
8. [x] Jira Ticket [DEV-8368](https://federal-spending-transparency.atlassian.net/browse/DEV-8368):
    - [ ] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #3594, opened 2022-09-23
[DEV 6968] Update Federal Accounts endpoint for agency profile page

aguest-kc:
**Description:**
Outlays, Obligations, and TBR are updated to use File A/B instead of GTAS.

**Technical details:**
Added `bureau_slug` and `total_budgetary_resources` fields to federal account objects and added an optional `bureau_slug` query parameter for filtering results. Removed pagination of results. 

    - [x] Backend
    - [x] Frontend
7. [x] Appropriate Operations ticket(s) created ( N/A ) 
8. [ ] Jira Ticket: [DEV-6968](https://federal-spending-transparency.atlassian.net/browse/DEV-6968)

```
```


usaspending_api/agency/v2/views/federal_account_list.py
```diff
@@ -20,15 +23,20 @@ class FederalAccountList(PaginationMixin, AgencyBase):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.params_to_validate = ["fiscal_year", "filter"]
+        self.params_to_validate = ["fiscal_year", "filter", "bureau_slug"]
+
+    def format_results(self, file_a_response, file_b_response):
+        for row_a, row_b in itertools.product(file_a_response, file_b_response):
```
sethstoudenmier, line 29: Why are you doing a cartesian product here? It seems like overkill when you only care about where rows match.

usaspending_api/agency/v2/views/federal_account_list.py
```diff
@@ -58,20 +68,42 @@ def format_results(self, rows):
     def get(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         self.sortable_columns = ["name", "obligated_amount", "gross_outlay_amount"]
         self.default_sort_column = "obligated_amount"
-        results = self.format_results(self.get_federal_account_list())
-        page_metadata = get_pagination_metadata(len(results), self.pagination.limit, self.pagination.page)
+        results = self.format_results(self.get_tbr_from_file_a_queryset(), self.get_federal_account_list_from_file_b())
+        page_metadata = get_pagination_metadata(len(results), len(results), self.pagination.page)
+        print(page_metadata)  # debug
```
sethstoudenmier, line 73: Should remove this `print()`.

usaspending_api/agency/v2/views/federal_account_list.py
```diff
@@ -58,20 +68,42 @@ def format_results(self, rows):
     def get(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         self.sortable_columns = ["name", "obligated_amount", "gross_outlay_amount"]
         self.default_sort_column = "obligated_amount"
-        results = self.format_results(self.get_federal_account_list())
-        page_metadata = get_pagination_metadata(len(results), self.pagination.limit, self.pagination.page)
+        results = self.format_results(self.get_tbr_from_file_a_queryset(), self.get_federal_account_list_from_file_b())
+        page_metadata = get_pagination_metadata(len(results), len(results), self.pagination.page)
```
sethstoudenmier, line 72: Would suggest defining a variable for the length instead of calling `len()` twice. Might not be the most expensive call, but no point having to get that result twice.

usaspending_api/agency/v2/views/federal_account_list.py
```diff
@@ -58,20 +68,42 @@ def format_results(self, rows):
...
 
-    def get_federal_account_list(self) -> List[dict]:
+    def get_tbr_from_file_a_queryset(self):
+        """
+        Query Total Budgetary Resources from File A
+        """
+        filters, bureau_info_subquery = self.get_common_query_objects("treasury_account_identifier")
+
+        results = (
+            (AppropriationAccountBalances.objects.filter(*filters))
```
sethstoudenmier, line 91: Will also need to filter by the `is_final_balalance_of_fy` to make sure that you are not getting all of the TBR values associated.

Or better yet, follow the example used in the `get_federal_account_list_from_file_b` below where the latest submissions_ids are retrieved and then used to determine which records to view.

The problem here is that the `_cpe` value stands for Current Period Ending. Since this value is a sum up to that point it will vary period to period for an agency across a single Fiscal Year.

usaspending_api/api_contracts/contracts/v2/agency/toptier_code/federal_account.md
```diff
@@ -106,7 +109,7 @@ Returns a list of Federal Accounts and Treasury Accounts in the Agency's appropr
 + `obligated_amount` (required, number)
 + `gross_outlay_amount` (required, number)
 
-## TreasuryAccount (object
+## TreasuryAccount (object)
```
sethstoudenmier, line 112: Good catch on this missing parenthesis.

sethstoudenmier (changes requested):
I would highly suggest moving away from that cartesian product since in Prod size data it could be problematic. Also, I believe you need to limit the TBR to only be that of the last period associated with the FY.

.gitignore
```diff
@@ -59,6 +59,7 @@ __pycache__
 *.pyc
 *.swp
 \.cache
+.devcontainer/
```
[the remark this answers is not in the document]
aguest-kc: VS Code has an option to do your actual development inside of a Docker container so that you have a repeatable dev environment. The `.devcontainer/` directory is where VS Code stores the files related to building and customizing your dev container.

There is no privacy or security issue with adding this directory to source control, I added it to .gitignore because I didn't think anyone else develops inside of a Docker container.

usaspending_api/agency/v2/views/agency_base.py
```diff
@@ -149,6 +158,10 @@ def agency_type(self):
     def award_type_codes(self):
         return self._query_params.get("award_type_codes")
 
+    @property
+    def bureau_slug(self):
+        return self._query_params.get("bureau_slug")
```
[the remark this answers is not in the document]
aguest-kc: I've moved **bureau_slug** to be a cached property and added validation to check that, if provided, the given slug exists.

Updated in commit 19f9a668c8f8f70a85398079b42206a8dfe4f12e

usaspending_api/agency/v2/views/federal_account_list.py
aguest-kc: Removed print statement in commit 4f77912ecff45e5bb3ba1c9c4960e7bc575de59d

usaspending_api/agency/v2/views/federal_account_list.py
aguest-kc: Created a `results_length` variable in commit d776dd0bbdded2dd8057f2e6b9f56a08d08b1fd6

usaspending_api/agency/v2/views/federal_account_list.py
aguest-kc: I added a filter on the `submission_ids` that are returned in the File A results in commit d145e39d6de95e5714301d53ccb7f0693773da12. Also, I moved the `get_latest_submission_ids_for_fiscal_year()` function call so that it's only run once per request.

usaspending_api/agency/v2/views/federal_account_list.py
```diff
@@ -44,37 +52,64 @@ def format_results(self, rows):
...
+                "treasury_account_identifier__tas_rendering_label",
+            )
+            .annotate(
+                bureau_info=bureau_info_subquery
+            )
+        )
+        return results
+
+    def get_federal_account_list_from_file_b(self) -> List[dict]:
+        _, bureau_info_subquery = self.get_common_query_objects("treasury_account")
```
sethstoudenmier, line 108: Any reason you have the `Q(treasury_account__funding_toptier_agency=self.toptier_agency),` filter defined in this method, but then in the TBR method you are using `Q(**{f"{treasury_account_keyword}__federal_account__parent_toptier_agency": self.toptier_agency}),`?

sethstoudenmier (changes requested):
Overall I think you have the right idea on where the values come from, but still some areas for improvement. Building out the tests more will help to validate your implementation.

* Current changes tot he test cases are only adding data, but you actually added to the response of this endpoint so your test updates should include more than just changes to test data
* Still would like you to take another look at the use of `itertools.product()` to hopefully change the approach away from a cartesian product
* Travis tests and code formatting need attention

usaspending_api/agency/v2/views/federal_account_list.py
kwhickey: @sethstoudenmier Maybe this can be disregarded, or reframed based on the below:

Each queryset function (FA list from B, TBR from A) is calling `get_common_query_objects` and it's the `get_common_query_objects` function that uses
`Q(**{f"{treasury_account_keyword}__federal_account__parent_toptier_agency": self.toptier_agency})`. 


With a different field name for the`TreasuryAppropriationAccount` model FK in each of the File A and File B model objects. 

`get_federal_account_list_from_file_b` uses `get_common_query_objects` I guess to get the bureau info subqery. But then also does its own filtering by agency via `FinancialAccountsByProgramActivityObjectClass.treasury_account__funding_toptier_agency=self.toptier_agency`

So maybe it begs the questions: _What is the **right** way to filter File A and/or File B objects by Agency? Why is it being done differently in different places here (e.g. via the TAS's funding toptier agency, or via the TAS'S FA'S parent toptier agency)? And do we need to be consistent in how this filtering is applied to File A or File B records?_

usaspending_api/agency/v2/views/federal_account_list.py
kwhickey: Diving into this question ... 
@aguest-kc what determines the granularity of results getting returned by this API endpoint? 

- Is each record built and returned at the granularity of 1 record per: Agency? Federal Account? TAS? Bureau? Or is it variable -- and if so how?
- Should a result be returned for every File A record that matches the API filter params, and then augmented with totals from File B for the same criteria?
- Or is a result returned for every File B record that matches the API filter params, and then augmented with budget info from File A for the same criteria? 
- Or is a result returned for records EITHER File A OR File B records that match API filter params, and then the result is filled in with File A or File B data where it can be found?

usaspending_api/agency/v2/views/federal_account_list.py
```diff
@@ -20,15 +23,20 @@ class FederalAccountList(PaginationMixin, AgencyBase):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.params_to_validate = ["fiscal_year", "filter"]
+        self.params_to_validate = ["fiscal_year", "filter", "bureau_slug"]
+
+    def format_results(self, file_a_response, file_b_response):
```
kwhickey, line 28: I know this isn't originally your code, but many of the steps in this final `format_results` function could use some inline comments walking the code-reader through the logic and flow of what's happening with the data-reshaping and accumulation.

usaspending_api/agency/v2/views/federal_account_list.py
sethstoudenmier: Here is an old PR from Wil that did something similar without the need to loop over every possible combination of results: https://github.com/fedspendingtransparency/usaspending-api/pull/3346/files#diff-87b27ed6231b4e97bfa0370021d6677ba778af56e5f1d0022ec1e9d7abe34f27R78

sethstoudenmier: Using SQL as an example, your `itertools.product()` approach is a CROSS JOIN which is typically not very performant or at least drops in performance very quickly as data grows. Wil instead uses an approach more inline with a FULL OUTER JOIN. Screenshot is added here for a visual representation of the two. 

Screen Shot 2022-10-11 at 1 13 01 PM

aguest-kc: I pushed an update that replaces `itertools.product` with an approach similar to the one from Wil's PR:

```python
# Combine File A and B responses
        combined_list_dict = {}

        for row in file_b_response:
            combined_list_dict[row['treasury_account__tas_rendering_label']] = row

        for row in file_a_response:
            # Append the total_budgetary_resources_amount_cpe value to the dict with the appropriate tas_rendering_label for this row
            if row['treasury_account_identifier__tas_rendering_label'] in combined_list_dict:
                combined_list_dict[row['treasury_account_identifier__tas_rendering_label']].update(
                    {'total_budgetary_resources': row['total_budgetary_resources_amount_cpe']}
                )
```

usaspending_api/agency/v2/views/agency_base.py
```diff
@@ -120,6 +129,18 @@ def toptier_agency(self):
             raise NotFound(f"Agency with a toptier code of '{self.toptier_code}' does not exist")
         return toptier_agency
 
+    @cached_property
+    def bureau_slug(self):
+        input_bureau_slug = self._query_params.get("bureau_slug")
+
+        if input_bureau_slug:
+            bureau_slug = BureauTitleLookup.objects.filter(
```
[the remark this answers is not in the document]
aguest-kc: Added a `.lower()` call on any bureau_slug input.

PR #3609, opened 2022-10-12
[PIPE-341] Transaction ETL in Spark

sethstoudenmier:
**Description:**
Create the Transaction ETL in Spark as a Django command.

**Technical details:**
Create the Transaction ETL in Spark as a Django command.

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [PIPE-341](https://federal-spending-transparency.atlassian.net/browse/PIPE-341):

```
```


usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -0,0 +1,416 @@
...
+        set_cols = [f"silver_table.{col.silver_name} = source_view.{col.silver_name}" for col in col_info]
+        silver_table_cols = ", ".join([col.silver_name for col in col_info])
+
+        sql = f"""
+            MERGE INTO int.transaction_{transaction_type} AS silver_table
+            USING vw_{transaction_type}_from_source AS source_view
+            ON silver_table.transaction_id = source_view.transaction_id
+            WHEN MATCHED
+                THEN UPDATE SET
+                    {", ".join(set_cols)}
```
kwhickey, line 308: Will `updated_at` (or whatever the update field is) get a new date or the "now" date when doing this?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -0,0 +1,416 @@
...
+                    )
+                    OR
+                    (
+                        transaction_id_lookup.published_fabs_id IS NOT NULL
+                        AND NOT EXISTS (
+                            SELECT *
+                            FROM raw.published_fabs
+                            WHERE transaction_id_lookup.published_fabs_id = published_fabs.published_fabs_id
+                        )
+                    )
```
kwhickey, line 339: Sometimes a `UNION` (or `UNION ALL`) will have better performance in the query optimizer (not sure about in spark) than a complex `OR`. Did you try both ways?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -0,0 +1,416 @@
...
+                THEN UPDATE SET
+                    {", ".join(set_cols)}
+            WHEN NOT MATCHED
+                THEN INSERT
+                    ({silver_table_cols})
+                    VALUES ({silver_table_cols})
+        """
+        return sql
+
+    def delete_records_sql(self):
```
kwhickey, line 316: So does the delete strategy just use the broker data as the set to compare against to determine deletes? Vs. using the deleted transaction logs csv file in S3? Does the current transaction loader do the same or does it rely on the logs to pick out what should be deleted? If this approach is different than the current, any assessment of whether it's "safe" (yields the same results in all conditions)?

Also, if this is how it deletes, do we need to _create_ new records in the CSV delete logs and is that being done in this code yet?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -0,0 +1,416 @@
...
+
+        # Before running INSERT, make sure expected keys have no NULLs
+        table_names = ("int.transaction_normalized", "raw.detached_award_procurement", "raw.published_fabs")
+        key_names = ("transaction_unique_id", "detached_award_proc_unique", "afa_generated_unique")
+        for table_name, key_name in zip(table_names, key_names):
+            test_results = self.spark.sql(
+                f"SELECT COUNT(*) AS count FROM {table_name} WHERE {key_name} IS NULL").collect()
+
+            assert test_results[0]['count'] == 0, \
+                f"Found {test_results[0]['count']} NULLs in {key_name} in table {table_name}!"
```
kwhickey, line 400: `assert` works sometimes, but there's always the risk that where this python code runs, assertions don't raise errors. So best to raise an error here instead.

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -0,0 +1,416 @@
...
+
+        # Insert existing transactions into the lookup table
```
kwhickey, line 402: I'll need to talk through what this is doing. Not following at a glance.

usaspending_api/etl/migrations/0004_create_transaction_id_seq.py
```diff
@@ -0,0 +1,21 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("etl", "0003_duns_schema_change"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="CREATE SEQUENCE IF NOT EXISTS transaction_id_seq AS BIGINT",
+            reverse_sql="DROP SEQUENCE IF EXISTS transaction_id_seq"
```
kwhickey, line 18: Do we need to "seed" the sequence with the value it starts from or is that handled in the code?

usaspending_api/broker/helpers/get_business_categories.py
```diff
@@ -416,3 +416,429 @@ def get_business_categories(row, data_type):
             "Invalid object type provided to update_business_categories. "
             "Must be one of the following types: TransactionFPDS, TransactionFABS"
         )
+
+
+def get_business_categories_fabs(business_types):
```
kwhickey, line 421: Wow. 
How did this used to be done? Postgres function? Or done just in Python because it was RBAR'ing 1 row at a time (possibly from a chunked row-set)?

kwhickey (commented):
Initial pass through. Really clean and concise! Does this template out all the flow that needs to take place or is there parts still missing/unknown?

Also curious about if and how any field-level manipulations/transformations are being handled. Maybe it's just mostly upper'ing and casting. But is there any whitespace manipulation we need to account for that's done in Python currently?

Bunch more questions inline that we can discuss verbally.

usaspending_api/etl/management/commands/load_transactions_in_delta.py
kwhickey: TODO: Need to investigate broker data and see if it is always UTC or if it could hold another timezone, as the current code would shift the time to align to UTC in the `cast_datetime_to_utc` function being used.

PR #3620, opened 2022-10-24
[HOTFIX] [PIPE-446] Further transaction block around CSV copy from delta

kwhickey:
**Description:**
Guarding the CSV write step with a transaction so if it fails and is retried, the previous data attempting to be written will roll back.

**Technical details:**
Note: Further consideration needed for the bounds of the RETRY. Due to using (rdd.parallelize) one full Task which spark is handling (in our case there's only 8 for the entire 1921 CSV files) may encompass many files to write in sequence. If any one of the files fails to write, the whole task (e.g. maybe 240 individual files) will be re-attempted. So the ones that DID succeed and were committed, will be written again along with the ones that did not succeed the first time around. 
In other words, the bounds of the retry are not aligned to the bounds of the transaction.

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>

```
```

### DIFF
--- a/usaspending_api/common/csv_stream_s3_to_pg.py
+++ b/usaspending_api/common/csv_stream_s3_to_pg.py
@@ -160,17 +160,21 @@ def copy_csvs_from_s3_to_pg(
                     cursor.execute("SET work_mem TO %s", (work_mem_override,))
                 s3_client = _get_boto3_s3_client()
                 for s3_obj_key in s3_obj_keys:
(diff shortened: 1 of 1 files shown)


kwhickey:
Closing this as that type of transaction management is only available in psycopg3. Best we can do is autocommit, which it already has. This will be addressed initially with setting `spark.task.maxFailures=1`

PR #3646, opened 2022-11-21
[PIPE-461] Transaction Search Agg Key Rework

dpb-bah:
**Description:**
Updating the ES transaction index to store only the codes in the agg_keys, and then using said code along with the database to surface their current values.

**Technical details:**
Updated the following endpoints
    * `spending_by_category/*`
    * `spending_by_geography/*`

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [PIPE-461](https://federal-spending-transparency.atlassian.net/browse/PIPE-461):
9. [x] Merged after https://github.com/fedspendingtransparency/usaspending-api/pull/3648

```
```


usaspending_api/search/v2/views/spending_by_category_views/spending_by_recipient.py
```diff
@@ -64,18 +63,35 @@ def _get_recipient_id(row: list) -> str:
...
+        for recipient_info in recipient_info_query.all():
+            current_recipient_info[str(recipient_info["recipient_hash"])] = recipient_info
 
+        # Build out the results
         results = []
-        location_info_buckets = response.get("group_by_agg_key", {}).get("buckets", [])
-        for bucket in location_info_buckets:
-            recipient_info = json.loads(bucket.get("key"))
+        for bucket in recipient_info_buckets:
+            result_hash, result_level = tuple(bucket.get("key").split("/")) if bucket.get("key") else (None, None)
```
emily-brents, line 81: Just double checking, but would this accurately catch the condition where the key has a hash but not a recipient_level? I believe this situation can happen with some of the "special" recipients since the levels come from a join that excludes them and the hash does not.

usaspending_api/search/v2/views/spending_by_category_views/spending_by_recipient.py
```diff
@@ -64,18 +63,35 @@ def _get_recipient_id(row: list) -> str:
...
 
             results.append(
                 {
                     "amount": int(bucket.get("sum_field", {"value": 0})["value"]) / Decimal("100"),
-                    "recipient_id": recipient_info["hash_with_level"] or None,
-                    "name": recipient_info["name"] or None,
-                    "code": recipient_info["duns"] or "Recipient not provided",
+                    "recipient_id": result_hash_with_level,
+                    "name": (
+                        recipient_info["legal_business_name"].upper()
```
emily-brents, line 90: This should already be uppercased if you are pulling the name from the recipient_lookup table.

usaspending_api/search/v2/views/spending_by_geography.py
```diff
@@ -381,40 +380,83 @@ def build_elasticsearch_search_with_aggregation(self, filter_query: ES_Q) -> Opt
...
+
+        # Get the codes
         geo_info_buckets = response.get("group_by_agg_key", {}).get("buckets", [])
+        geo_codes = [_key_to_geo_code(bucket["key"]) for bucket in geo_info_buckets if bucket.get("key")]
+
+        # Get the current geo info
+        current_geo_info = {}
+        if self.geo_layer == GeoLayer.STATE:
+            geo_info_query = (
+                PopCounty.objects.filter(state_code__in=geo_codes, county_number="000")
```
emily-brents, line 394: These queries look pretty good, but double check that none of the data comes from CityCountyStateCode or StateData, sometimes the names or other data can differ between the tables, so just check which table transaction_search is joining to and make sure they line up.

usaspending_api/search/v2/views/spending_by_category_views/spending_by_recipient.py
dpb-bah: It's a good call to double-check as it can get funky. I just reviewed the `test_spending_by_recipients.py` which includes scenarios of `MULTIPLE RECIPIENTS` without levels in the transaction search data. The recipient agg key stores them as `[hash]/` in the index and then this would split them into the variables (quick local run of that line):
```
>>> bucket = {'key': 'example_hash/'}
>>> result_hash, result_level = tuple(bucket.get("key").split("/")) if bucket.get("key") else (None, None)
>>> result_hash
'example_hash'
>>> result_level
''
```
And then the reponse has something like:
```
{"amount": 5000.0, "code": "096354360", "name": "MULTIPLE RECIPIENTS", "recipient_id": None}
or
{
     "amount": 5000000.0,
     "code": "Recipient not provided",
    "name": "MULTIPLE RECIPIENTS",
    "recipient_id": None,
},
```
I think this should cover most of those weird cases.

usaspending_api/search/v2/views/spending_by_geography.py
dpb-bah: Yeah I was actually making my own `code_to_state` via StateData till I saw the import and we were using the one from `abbreviations.py` for the query django equivalent. I'm a little concerned that's just a dictionary that could be out of sync with the database but it's just states and their codes which rarely change.

As for the queries, I definitely use `StateData` in `spending_by_locations.py` for `self.location_type == LocationType.STATE_TERRITORY`, though it seems like it's used in the transaction search query? 
```
    LEFT OUTER JOIN (
        SELECT code, name, fips, MAX(id)
        FROM global_temp.state_data
        GROUP BY code, name, fips
    ) POP_STATE_LOOKUP ON (
        POP_STATE_LOOKUP.code = COALESCE(transaction_fpds.place_of_performance_state, transaction_fabs.place_of_perfor_state_code)
    )
    ...
    LEFT OUTER JOIN (
        SELECT code, name, fips, MAX(id)
        FROM global_temp.state_data
        GROUP BY code, name, fips
    ) RL_STATE_LOOKUP ON (
        RL_STATE_LOOKUP.code = COALESCE(transaction_fpds.legal_entity_state_code, transaction_fabs.legal_entity_state_code)
    )
```
Doesn't looks like I'm using `CityCountyStateCode` anywhere though.

emily-brents: okay sounds good, I ran into an issue when I was testing where the county names varied slightly, but hopefully that was just for award search and the transactions won't have that issue.

emily-brents (approved):

emily-brents (dismissed):

emily-brents (approved):

PR #3647, opened 2022-11-21
Initial work for transaction_f[ab|pd]s handlers for delta transaction loader code

DManowitz-BAH:
**Description:**
Next step in moving the transaction loaders to Spark, after [PIPE-341](https://federal-spending-transparency.atlassian.net/browse/PIPE-341)

**Technical details:**
Took work done by @sethstoudenmier in PR #3609 for handling transaction_f[ab|pd]s ETL levels, added it to the code from PR #3625, added a few initial tests, and modified this code to streamline it a bit and to handle a problem encountered with tearing down views in the testing framework.

Still need to write additional tests and to complete other steps below.

**Requirements for PR merge:**

1. [ ] Unit & integration tests updated
2. [ ] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
    - [ ] Domain Expert <OPTIONAL>
4. [ ] Matview impact assessment completed
5. [ ] Frontend impact assessment completed
6. [ ] Data validation completed
7. [ ] Appropriate Operations ticket(s) created
8. [ ] Jira Ticket [PIPE-340](https://federal-spending-transparency.atlassian.net/browse/PIPE-340):
    - [ ] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #3657, opened 2022-12-06
[PIPE-339] Handle transaction_normalized ETL level in Spark transaction loader

DManowitz-BAH:
**Description:**
Next step in moving the transaction loaders to Spark, after [PIPE-340](https://federal-spending-transparency.atlassian.net/browse/PIPE-340)

**Technical details:**
Took work done by @sethstoudenmier in PR https://github.com/fedspendingtransparency/usaspending-api/pull/3609 for handling transaction_normalized ETL level, added it to the code from PR https://github.com/fedspendingtransparency/usaspending-api/pull/3647, and modified it to fully handle the updating the int.transaction_normalized table 

Still need to write tests and to complete other steps below.

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>

```
```


usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -184,24 +223,133 @@ def source_subquery_sql(self, transaction_type):
...
+                        CAST(COALESCE({bronze_table_name}.federal_action_obligation, 0)
+                                + COALESCE({bronze_table_name}.non_federal_funding_amount, 0)
+                             AS NUMERIC(23, 2)) AS funding_amount
+                        """,
+                    ]
                 )
+                map_col_info = copy.copy(FABS_TO_NORMALIZED_COLUMN_INFO)
             else:
-                select_cols.append(f"{bronze_table_name}.{col.bronze_name} AS {col.silver_name}")
+                fpds_business_category_columns = copy.copy(fpds_boolean_columns)
```
collinwr, line 269: Can you make the `else` above this line into an `elif` then add an `else` to catch bad `transaction_types` at the bottom, similar to what you have on line 224:
>  raise ValueError("Invalid value for 'transaction_type'; must select either: 'fabs' or 'fpds'")

collinwr: I see that you're already doing this check in `transaction_normalized_merge_into_sql()`, so this is optional. I still like to be explicit about what condition leads to this code path for the ease of readability though. Maybe no need for the final `else`, but keep the `elif`

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -184,24 +223,133 @@ def source_subquery_sql(self, transaction_type):
...
+                fpds_business_category_columns.extend(["contracting_officers_deter", "domestic_or_foreign_entity"])
```
collinwr, line 270: Why are these being added here rather than existing in the list from `usaspending_api/broker/helpers/build_business_categories_boolean_dict.py`?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
DManowitz-BAH: ok

usaspending_api/etl/management/commands/load_transactions_in_delta.py
DManowitz-BAH: These are non-boolean columns that are used in the business category logic, so they are not in that list.

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -54,24 +80,37 @@ def add_arguments(self, parser):
...
 
     def handle(self, *args, **options):
         with self.prepare_spark():
             self.etl_level = options["etl_level"]
             self.spark_s3_bucket = options["spark_s3_bucket"]
+            self.no_initial_copy = options["no_initial_copy"]
+
+            # Capture minimum last load date of the source tables to update the "last_load_date" after completion
+            last_procure_load = get_last_load_date("source_procurement_transaction")
+            last_assist_load = get_last_load_date("source_assistance_transaction")
```
collinwr, line 98: We have a method called `get_earliest_load_date()` in `last_load_date.py` that accepts a list of keys and returns the earliest load date. This can simplify this block of code.

collinwr (commented):
Other changes to last load dates look good 👍

usaspending_api/etl/management/commands/load_transactions_in_delta.py
DManowitz-BAH: I took a look at that code.  Unfortunately, I think there is a bug in that code.  For example, if I were to call `get_earliest_load_date(("source_procurement_transaction", "source_assistance_transaction"), datetime(1970, 1, 1))`, then it appears that the code would only return a date corresponding to the earliest last load date of `source_procurement_transaction` or `source_assistance_transaction` if that date were earlier than 1970-01-01, but I don't think that's what we want that call to do.

DManowitz-BAH: Assuming you agree with my assessment, do you want me to fix the code so that it only returns the default value if no last load date is found for any of the keys?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -254,26 +238,40 @@ def handle_column(col, bronze_table_name):
                 regexp_YYYYmmdd = (
                     r"(\\d{4})(?<sep>[-/]?)(\\d{2})(\\k<sep>)(\\d{2})(.\\d{2}:\\d{2}:\\d{2}([+-]\\d{2}:\\d{2})?)?"
                 )
```
collinwr, line 239: I'd say move these regex strings outside the case statements, so `parse_string_date` and `truncate_string_date` can be top level cases. This will prevent the need for nested `if` statements

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -254,26 +238,40 @@ def handle_column(col, bronze_table_name):
...
+                                        regexp_extract({bronze_table_name}.{col.source}, '{regexp_YYYYmmdd}', 3)
+                                        || '-' ||
+                                        regexp_extract({bronze_table_name}.{col.source}, '{regexp_YYYYmmdd}', 5))
+                                       AS DATE)
+                        END AS {col.dest_name}
+                    """
+                else:
+                    # These are string fields that actually hold DATES/TIMESTAMPS, but need the non-DATE part discarded,
+                    # even though they remain as strings
+                    return f"""
```
collinwr, line 260: What do you think about `parse_string_date_truncate_time` as an alternative to `truncate_string_date`?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -244,8 +229,7 @@ def handle_column(col, bronze_table_name):
             elif col.handling == "literal":
                 # Use col.source directly as the value
                 return f"{col.source} AS {col.dest_name}"
-            elif col.handling == "parse_string_date":
-                # return f"parse_string_date({bronze_table_name}.{col.source}) AS {col.dest_name}"
+            elif col.handling in ("parse_string_date", "truncate_string_date"):
                 # Require a separator for mmddYYYY, but not for YYYYmmdd, or there is no way to tell them apart
                 # from just regexp.
```
collinwr, line 233: Can you include a note in this comment about whether a timestamp is allowed and or expected in these regular expressions?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
collinwr: Yes, there is a bug there. Good catch. I would say yes, that would be a good fix

DManowitz-BAH: Done

usaspending_api/etl/management/commands/load_transactions_in_delta.py
DManowitz-BAH: Done

collinwr (approved):

PR #3663, opened 2022-12-22
[PIPE-465] Preserve Views when Swapping Tables

dpb-bah:
**Description:**
Simply detects dependent views off the table being swapped. For each dependent view, it creates a new temp view over that, swaps the two, and deletes the old view if the old table is to be removed.

**Technical details:**
Drops the hardcoded Award_Search view SQL and `VIEWS_TO_MAKE` constant in that script.

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [PIPE-465](https://federal-spending-transparency.atlassian.net/browse/PIPE-465):

```
```


usaspending_api/etl/management/commands/swap_in_new_table.py
```diff
@@ -294,6 +306,43 @@ def validate_indexes(self, cursor):
...
+        dep_views = ordered_dictionary_fetcher(cursor)
+        return dep_views
+
+    def create_new_views(self, cursor):
+        for dep_view in self.dep_views:
+            # Note: Despite pointing at the new temp table at first,
+            #       the recreated views will automatically update and stay pointed to it in the swap via postgres
+            logger.info(f"Recreating dependent view: {dep_view['dep_view_fullname']}_temp")
+            dep_view_sql = dep_view["dep_view_sql"].replace(self.curr_table_name, self.temp_table_name)
+            cursor.execute(f"CREATE VIEW {dep_view['dep_view_fullname']}_temp AS ({dep_view_sql});")
```
collinwr, line 344: Would it be possible to create the temp view in the `temp` schema? I think this would prevent the `api_user` DB user from accessing the view. The schema would need to be updated to swapped in `swap_table_view_sql` as well

dpb-bah: It can be done. Right now it's keeping the temp views in the same schema as their main views, so if the user has access to the main view, it would have access to the temp view as well. Visa versa if they don't have access to the main view in the first place. I personally like it this way as the script is already agnostic to the `temp` schema, but I can hardcode it in if that's what we want.

collinwr (approved):

PR #3674, opened 2023-01-04
Possible SSRF in code. No fix suggested.

openrefactory:
In file: experimental_api_flags.py, there is a method call that fetches a remote resource based on a URL that may come from user-provided, untrusted data. This may allow an attacker to send a crafted request to an unexpected destination. 

The untrusted data in both cases originate from the same source. In experimental_api_flags.py file, line 35, we have

    url = request.build_absolute_uri()

Here variable url is assigned a tainted value from an external source. This is then used in an HTTP method call.

We suggested proper sanitization before user-provided data can be used to fetch remote resources. The actual fix is domain specific; so an actual fix is not provided.

### DIFF
--- a/usaspending_api/common/experimental_api_flags.py
+++ b/usaspending_api/common/experimental_api_flags.py
@@ -46,30 +46,10 @@ def mirror_request_to_elasticsearch(request: Union[HttpRequest, Request]):
     # without disrupting the primary request
     try:
         if request.method == "GET":
-            """
-            ***************** Warning *****************
-            Possible Server-Side Request Forgery (SSRF) attack!
-            Path:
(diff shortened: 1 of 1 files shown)


kwhickey:
Thank you for investigating this. The `mirror_request_to_elasticsearch` function that makes use of `request.build_absolute_uri()` is not in use for the web application. Further, when the Django HTTP Request library executes `build_absolute_uri()` it does sanitize (encode/escape) the received original request's location, path, and query string using `escape_uri_path(path)` `iri_to_uri(iri)` in accordance with RFC 2396 and RFC 3987 to only preserve safe characters and URL encode others.

PR #3681, opened 2023-01-06
[PIPE-347] reset es deletes load date at delta transaction loader initial_run

kwhickey:
**Description:**
Make sure that the `es_deletes` load date is also reset when the new delta load dates are _initially_ set

**Technical details:**
- Set at the `initial_run` of the transaction delta tables **only**

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/load_transactions_in_delta.py
+++ b/usaspending_api/etl/management/commands/load_transactions_in_delta.py
@@ -17,7 +17,11 @@ from usaspending_api.broker.helpers.get_business_categories import (
     get_business_categories_fabs,
     get_business_categories_fpds,
 )
-from usaspending_api.broker.helpers.last_load_date import get_last_load_date, update_last_load_date
+from usaspending_api.broker.helpers.last_load_date import (
+    get_last_load_date,
+    update_last_load_date,
+    get_earliest_load_date,
+)
 from usaspending_api.common.etl.spark import create_ref_temp_views
 from usaspending_api.common.helpers.spark_helpers import (
     get_active_spark_session,
@@ -109,13 +113,9 @@ class Command(BaseCommand):
             self.no_initial_copy = options["no_initial_copy"]
 
             # Capture minimum last load date of the source tables to update the "last_load_date" after completion
-            last_procure_load = get_last_load_date("source_procurement_transaction")
-            last_assist_load = get_last_load_date("source_assistance_transaction")
-            if last_procure_load is None:
-                last_procure_load = datetime.utcfromtimestamp(0)
-            if last_assist_load is None:
-                last_assist_load = datetime.utcfromtimestamp(0)
-            next_last_load = min(last_assist_load, last_procure_load)
+            next_last_load = get_earliest_load_date("source_procurement_transaction", "source_assistance_transaction")
+            if not next_last_load:
+                next_last_load = datetime.utcfromtimestamp(0)
 
             if self.etl_level == "initial_run":
                 self.logger.info("Running initial setup for transaction_id_lookup and award_id_lookup tables")
@@ -1004,6 +1004,8 @@ class Command(BaseCommand):
 
         if set_last_load_date:
             update_last_load_date(destination_table, next_last_load)
+            # es_deletes should remain in lockstep with transaction load dates, so if they are reset, it should be reset
+            update_last_load_date("es_deletes", next_last_load)
 
         # award_id_lookup
         destination_table = "award_id_lookup"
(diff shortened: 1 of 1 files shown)


usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -109,13 +113,9 @@ def handle(self, *args, **options):
...
-            last_procure_load = get_last_load_date("source_procurement_transaction")
-            last_assist_load = get_last_load_date("source_assistance_transaction")
-            if last_procure_load is None:
-                last_procure_load = datetime.utcfromtimestamp(0)
-            if last_assist_load is None:
-                last_assist_load = datetime.utcfromtimestamp(0)
-            next_last_load = min(last_assist_load, last_procure_load)
+            next_last_load = get_earliest_load_date("source_procurement_transaction", "source_assistance_transaction")
+            if not next_last_load:
+                next_last_load = datetime.utcfromtimestamp(0)
```
DManowitz-BAH, line 119: get_earliest_load_date needs an iterable of keys to check, and it also accepts a default that gets returned if no earliest load date is found.

In my code, I have it changed to:
```
next_last_load = get_earliest_load_date(
                ("source_procurement_transaction", "source_assistance_transaction"), datetime.utcfromtimestamp(0)
)
```

DManowitz-BAH (changes requested):

usaspending_api/etl/management/commands/load_transactions_in_delta.py
kwhickey: Great catch on the `Iterable`!
As for the default though, it looks as if it will end up taking the least of _all three_ dates (or however many you provide plus the default). And if you provide a default of `datetime.utcfromtimestamp(0)`, it will beat ANY dates that may be stored in the system and always end up returning `datetime.utcfromtimestamp(0)`. It's not a very intuitive "default" which I would think would act as the way you described. 
Probably means you should double check whether it makes sense where/how you used it as a default.

kwhickey: Fixed

DManowitz-BAH: From Slack:
David Manowitz
  [3:08 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673294929405739)
So, right, there is currently a bug, but I fixed it in my transaction_normalized branch.
[3:09](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673294977522269)
Wil pointed this function out to me, but I noticed that issue, so he told me to fix it.


Keith Hickey
  [3:09 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673294990921999)
hm. but how do you know that was a bug and not as-designed?
[3:10](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295007618149)
(admittedly the design doesn't make a lot of sense)


David Manowitz
  [3:10 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295029930149)
There is a comment in the function I believe as to its intended design.


Keith Hickey
  [3:11 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295065114649)
you'll want to double check all usages of it are "backwards compatible" with your fix ... since some folks that used it may have taken the code literally in its impl and used it knowing the side-effects


David Manowitz
  [3:11 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295079919629)
ok


Keith Hickey
  [3:13 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295181925419)
this comment
default will be returned if no last_load_date is found for any of the supplied
keys
Still seems to be accurate with the way it was implemented.
i.e. if the supplied keys are all null / not populated, the default will be used.
But the trouble comes when they are not null and populated. At that point default may take precedence over the supplied keys if it is lesser than all. (edited) 


David Manowitz
  [3:13 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295228072269)
Good point


Keith Hickey
  [3:14 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295269288799)
I'll leave it to you. Feel free to alter my code after merging it in to use the default if you go through with the adjustment to the implementation of that function. It should have the iterable part of it fixed at least
:+1:
1



David Manowitz
  [3:14 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295298551489)
I'm double-checking the usage of that function now


Keith Hickey
  [3:15 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295319425549)
(my guess is no one ever used a default arg, which would be safe for you)


David Manowitz
  [3:19 PM](https://data-transparency-bfs.slack.com/archives/D042ZT1CG5P/p1673295599066849)
You're right.  Other than in this function, the only other places I see this function called are in etl/elasticsearch_loader_helpers/controller.py and broker/tests/integration/test_last_load_date.py, and even in the test it doesn't use the default!

DManowitz-BAH (approved):

PR #3687, opened 2023-01-14
[PIPE-465] Update swap code for *materialized* views

dpb-bah:
**Description:**
Simply updating the swap code to switch statements when dealing with views vs. matviews.

**Technical details:**
N/A

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [PIPE-465](https://federal-spending-transparency.atlassian.net/browse/PIPE-465):

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/swap_in_new_table.py
+++ b/usaspending_api/etl/management/commands/swap_in_new_table.py
@@ -189,16 +189,21 @@ class Command(BaseCommand):
             sql_table_template.format(old_table_name=self.temp_table_name, new_table_name=f"{self.curr_table_name}"),
         ]
 
-        sql_view_template = "ALTER VIEW {old_view_name} RENAME TO {new_view_name};"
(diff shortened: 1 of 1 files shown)


collinwr:
I'm a little hesitant about the Matviews being created during the swap. It only took about 10 minutes in the run you sent me, but we want those swaps to happen as quickly as possible, so the new data is all revealed close to the same time. 

I just did a quick test, and it looks like a Matview will throw the same error as a View when trying to delete the underlying data. 

I wonder if at this point, we should go ahead and convert `mv_agency_autocomplete` to a table. If it stays as a Matview that's recreated by the swap command, we'll never actually "refresh" the Matview, so maybe it should just be a table. It can be created after `award_search` lands as a temp table then later swapped with the other tables.

kwhickey (approved):
Good adaptation to matviews. Approved.

PR #3691, opened 2023-01-19
[PIPE-474] Modify initial_run code in Delta loader to handle orphaned transactions & awards

DManowitz-BAH:
**Description:**
This PR addresses the issue of "orphaned" transactions (records in the `transaction_fabs|fpds|normalized` tables that have no associated records in the source (`raw.detached_award_procurement` or `raw.published_fabs`) tables, and any awards that are impacted by orphaned transactions.

Due to timing of reviews, etc, this PR also fixes an issue with the `awards` ETL level.

**Technical details:**
During the `initial_run` ETL level of the `load_transactions_in_delta` command, this code first identifies any orphaned transactions in the `raw.transaction_normalized` table.  It then prevents those transactions from being added to the `int.transaction_id_lookup` or `id.award_id_lookup` tables, and prevents them from being copied from the `raw` to `int` versions of the `transaction_fabs|fpds|normalized` tables.  It also identifies any awards that have some orphaned transactions associated with them and marks them for modification during the next run of the `awards` ETL level.  Lastly, it finds any awards that are themselves orphaned (lacking any transactions) once the orphaned transactions have been excluded, and it excludes them from being copied from the `raw.awards` table to the `int.awards` table.

NOTE: All of the current tests for `load_transactions_to_delta` pass with these changes, but no new unit tests have been added to test out this functionality yet.

For the `awards` ETL level, needed to change an `INNER JOIN` to a `LEFT JOIN` to handle a sparse CTE. 

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [ ] Jira Ticket [PIPE-474](https://federal-spending-transparency.atlassian.net/browse/PIPE-474):

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/load_transactions_in_delta.py
+++ b/usaspending_api/etl/management/commands/load_transactions_in_delta.py
@@ -995,6 +995,7 @@ class Command(BaseCommand):
         Procedure to create & set up transaction_id_lookup and award_id_lookup tables and create other tables in
         int database that will be populated by subsequent calls.
         """
+
         @contextmanager
         def prepare_temp_tables():
             # Since the table to track the orphaned transactions is only needed for this function, just using a
@@ -1155,7 +1156,7 @@ class Command(BaseCommand):
                         award_id LONG NOT NULL
                     )
                     USING DELTA
-                    LOCATION 
+                    LOCATION
                         's3a://{self.spark_s3_bucket}/{delta_lake_s3_path}/{destination_database}/{destination_table}'
                 """
             )
@@ -1202,7 +1203,7 @@ class Command(BaseCommand):
                         generated_unique_award_id STRING NOT NULL
                     )
                     USING DELTA
-                    LOCATION 
+                    LOCATION
                         's3a://{self.spark_s3_bucket}/{delta_lake_s3_path}/{destination_database}/{destination_table}'
                 """
             )
@@ -1269,7 +1270,7 @@ class Command(BaseCommand):
                                     -- Again, want to exclude orphaned transactions, as they will not be copied into the
                                     -- int schema.
                                     WHERE ucase(pfabs.afa_generated_unique) NOT IN (
-                                        SELECT 
+                                        SELECT
                                             transaction_unique_id FROM temp.orphaned_transaction_info WHERE NOT is_fpds
                                     )
                                 )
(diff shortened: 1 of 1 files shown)


usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -1049,6 +1096,8 @@ def initial_run(self, next_last_load):
                     FROM raw.transaction_normalized AS tn INNER JOIN raw.detached_award_procurement AS dap ON (
                         tn.transaction_unique_id = ucase(dap.detached_award_proc_unique)
                     )
+                    -- Want to exclude orphaned transactions, as they will not be copied into the int schema.
+                    WHERE tn.id NOT IN (SELECT transaction_id FROM temp.orphaned_transaction_info WHERE is_fpds)
```
kwhickey, line 1100: Should the `WHERE is_fpds` and `WHERE NOT is_fpds` be comparing to a boolean? Does this work or evaluate correctly if not? Just not sure if it comes up with falsey/truthy logic so long as there is any boolean value and it's not null.

kwhickey (approved):
Only a few comments inside and still approving. 
- Good catch on the LEFT join for sparse exec. comp. data
- The use of is_fpds without comparing to e.g. `is_fpds = TRUE` makes me a bit wary
- maybe try to make the temp tables drop even in case of error
- Will all this orphan code only run for the `initial_run` invocation?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -1276,3 +1377,6 @@ def initial_run(self, next_last_load):
                         # es_deletes should remain in lockstep with transaction load dates, so if they are reset,
                         # it should be reset
                         update_last_load_date("es_deletes", next_last_load)
+
+        self.spark.sql("DROP TABLE temp.orphaned_transaction_info")
+        self.spark.sql("DROP TABLE temp.orphaned_award_info")
```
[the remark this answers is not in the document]
DManowitz-BAH: Good catch on this.  However, I was inspired by Seth's use of a `@contextmanager` function to handle the spark context, so I decided to use that here, too.

usaspending_api/etl/management/commands/load_transactions_in_delta.py
DManowitz-BAH: This column is declared as `NOT NULL`, so no need to worry about NULLs.

kwhickey (changes requested):
1 safeguard requested

kwhickey (approved):
Looks good!

PR #3694, opened 2023-01-20
[PIPE-380] Ensure empty s3 folder for delta table CSV files

kwhickey:
**Description:**
Prevent any old/stale CSV files from past runs getting swept up in `load_table_from_delta` and causing duplicative data.

**Technical details:**
Unsure if this is the culprit for duplicative data, but will be tested shortly. This will ensure each new load from delta has a clean and empty "folder" in S3 in which to dump its CSV files before they are dumped there. 
- Yes, testing seemed to confirm this was a bug: _if a `load_table_from_delta` Spark job is aborted, and some of the .csv.gz part files were created, they will apparently be picked up by the next run_
- This code ensures to clean out any leftover/stale/residual files from previous runs so only the latest data is processed.

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/load_table_from_delta.py
+++ b/usaspending_api/etl/management/commands/load_table_from_delta.py
@@ -427,6 +427,33 @@ class Command(BaseCommand):
             )
         s3_bucket_with_csv_path = f"s3a://{spark_s3_bucket_name}/{csv_path}"
 
+        # Get boto3 s3 resource to interact with bucket where CSV data will land
+        if not CONFIG.USE_AWS:
+            boto3_session = boto3.session.Session(
+                region_name=CONFIG.AWS_REGION,
+                aws_access_key_id=CONFIG.AWS_ACCESS_KEY.get_secret_value(),
+                aws_secret_access_key=CONFIG.AWS_SECRET_KEY.get_secret_value(),
+            )
+            s3_resource = boto3_session.resource(
+                service_name="s3", region_name=CONFIG.AWS_REGION, endpoint_url=f"http://{CONFIG.AWS_S3_ENDPOINT}"
+            )
+        else:
+            s3_resource = boto3.resource(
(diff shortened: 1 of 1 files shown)


usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -427,6 +427,33 @@ def _write_with_sql_bulk_copy_csv(
...
+            )
+        s3_bucket_name = spark_s3_bucket_name
+        s3_bucket = s3_resource.Bucket(s3_bucket_name)
+        objs_collection = s3_bucket.objects.filter(Prefix=csv_path)
+        initial_size = sum(1 for _ in objs_collection)
+
+        if initial_size > 0:
+            self.logger.info(f"LOAD: Starting to delete {initial_size} previous objects in {s3_bucket_with_csv_path}")
+            objs_collection.delete()
+            post_delete_size = sum(1 for _ in objs_collection)
```
collinwr, line 452: Under what circumstances would this be greater than 0?

kwhickey: It usually is `> 0` since the same s3 folder is used and reused as the dumping group for the intermediate csv files used to transport data from delta to postgres. So the csv files from the last run will still be there.

collinwr: Discussed this a little more with Keith. We're expecting `post_delete_size` to always be 0. Keith considered throwing an error if not, but ultimately decided against it:
> i was thinking that, but maybe it's just a timing thing of a few seconds if it ends up being not yet consistent.
And then the savemode=overwrite should also do what it's supposed to afterward.
I think if we find that we get errors with the unique exception, we can go back to this log statement to see if this was the reason. If so, then we need to put in a hard stop

collinwr (approved):

PR #3695, opened 2023-01-20
[PIPE-467] File C to D Linkage in Delta

collinwr:
## Description
This command links File C (financial_accounts_by_awards) records to File D (awards) records. It creates a new copy of the `financial_accounts_by_awards` table in the `int` schema based on the copy in the `raw` schema and applies the updates there. These records are linked to the `awards` table in the `int` schema.

## Technical Details 
This command recreates a `SHALLOW CLONE` of the `raw.financial_accounts_by_awards` table each time it is run. It deletes the previous copy of the of the `int` table. This is because `raw.financial_accounts_by_awards` is fully replaced by the Postgres version each night during the pipeline, so the `int` version also needs to be completely replaced. 

After the clone, views are created based on the new `int` table to determine which FABA records need to be updated. Once these records have been identified, a final table `union_all_priority` is merged into the `int` table and final counts are logged. 

## Requirements for Merge

    - [ ] Backend
8. [ ] Jira Ticket [PIPE-467](https://federal-spending-transparency.atlassian.net/browse/PIPE-467):

```
```


usaspending_api/etl/management/commands/load_table_to_delta.py
```diff
@@ -62,7 +62,7 @@
         "is_from_broker": False,
         "source_table": "vw_awards",
         "source_database": "rpt",
-        "destination_database": "raw",
```
dpb-bah, line 65: Is there a reason why we are updating this and not all the other transaction TABLE_SPECs?

dpb-bah (changes requested):

usaspending_api/etl/management/commands/load_table_to_delta.py
collinwr: Good call! I needed to update it for awards, but it wasn't necessary for Transactions. I suspect that making such an update would impact the Transaction Loader tests. I can give it a shot and see what happens.

collinwr: I just changed it back for now until we decide what we want to do with those table_specs.

usaspending_api/awards/delta_models/financial_accounts_by_awards.py
```diff
@@ -62,3 +62,308 @@
...
+        UNION
+        SELECT * FROM c_to_d_fain
+        UNION
+        SELECT * FROM c_to_d_uri
+        UNION
+        SELECT * FROM c_to_d_fain_and_uri_fain
+        UNION
+        SELECT * FROM c_to_d_fain_and_uri_uri
+        UNION
+        SELECT * FROM c_to_d_deletes
```
kwhickey, line 318: Since each temp view will have a different priority, `UNION` will never collapse down duplicate records. Can a `UNION ALL` then be used, since it typically is more efficient (no record comparisons needed)?

usaspending_api/common/helpers/spark_helpers.py
```diff
@@ -534,6 +548,72 @@ def log_hadoop_config(spark: SparkSession, config_key_contains=""):
     ]
 
 
+def load_dict_to_delta_table(spark, s3_data_bucket, table_schema, table_name, data, overwrite=False):
```
kwhickey, line 551: docstring describing the purpose of this function and when it should/could be used would be helpful.

DManowitz-BAH (commented):
test_load_transactions_in_delta.py and spark_helpers.py look good

usaspending_api/common/helpers/spark_helpers.py
```diff
@@ -534,6 +548,72 @@ def log_hadoop_config(spark: SparkSession, config_key_contains=""):
...
+        "create_delta_table",
+        "--destination-table",
+        table_name,
+        "--alt-db",
+        table_schema,
+        "--spark-s3-bucket",
+        s3_data_bucket,
+    )
+
+    if data:
```
kwhickey, line 578: Would it be helpful to get a list of pyspark `Row` objects?
You can create `Row` objects from dict by [unpacking the dict](https://stackoverflow.com/a/38253641).

```
from pyspark.sql import Row

Row(**data_as_dict)
```

usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
```diff
@@ -0,0 +1,159 @@
...
+        spark_created_by_command = False
+        if not self.spark:
+            spark_created_by_command = True
+            self.spark = configure_spark_session(**extra_conf, spark_context=self.spark)  # type: SparkSession
+
+        # Setup Logger
+        logger = get_jvm_logger(self.spark, __name__)
+
+        # Setup int table. Creates a shallow clone of the `raw` FABA table in the `int` schema.
+        # If the --no-clone option is provided a full table is created instead.
```
kwhickey, line 79: Maybe also clarify under --no-clone that it is a _managed_ table, meaning the data is not stored in a specific S3 bucket
- Do we know where it gets stored when managed?
- i.e. do we at least know that it's not stored on the filesystem of the driver or the cluster nodes, because that data would disappear if so when the cluster is terminated.

dpb-bah (dismissed):

usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
```diff
@@ -0,0 +1,159 @@
...
+        else:
+            self.spark.sql(
+                """
+            CREATE OR REPLACE TABLE int.financial_accounts_by_awards
+            SHALLOW CLONE raw.financial_accounts_by_awards;
+            """
+            )
+
+        # Log initial linkage count
+        unlinked_count = self.get_unlinked_count("raw")
```
kwhickey, line 103: Is this right, that you want to do the initial count from `raw.faba` and not the newly cloned `int.faba`? Maybe it's the same since it just go cloned? Seems it would make more sense / be more readable if the before/after counts were on the same table from the same schema.

Also, what is the provenance of that count?
- at this point, raw.faba is copied up from postgres
- which hopefully had its postgres data c-to-d linked from the previous pipeline run before it was copied up
- I guess this all makes sense if there are no changes to faba in postgres from when it is copied up to delta, and when the PG linkage happens

kwhickey (changes requested):
- Maybe a `UNION ALL`
- some documentation 
- and question about where **managed** table data lives, and if it lives on beyond cluster termination

usaspending_api/common/helpers/spark_helpers.py
collinwr: Are you thinking it might be better to convert the list of dictionaries to a list of Row objects then use a Spark function to insert the rows?

If so, we may still need some logic to cleanup some fields such as converting `date`/`datetime` fields to `isoformat()` or using `col_info` if available to set defaults for empty values.

Lots of the logic for building the INSERT SQL could be removed though

usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
collinwr: You're right that we technically shouldn't see any difference between running this on `raw` and `int`. I can see your point thought that getting the initial unlinked count from `int` might be more readable though. I'll update.

usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
kwhickey: With it now not being managed, you may need to or want to purge the data from the bucket after the table is deleted with the vacuum with retention set to 0 hours. I think the default vacuum is 30 days. So we'd have 30 copies of the data in the S3 bucket before it starts to clean up. 
Actually, does that matter or apply at all with it being a shallow clone? Maybe it just stores _new_ files that are different than the cloned destination? That may be okay to not forcibly vacuum.

usaspending_api/common/helpers/spark_helpers.py
kwhickey: Yeah, fine to leave as is for now if it's working.

usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
kwhickey: Also, in the `--no-clone` scenario, is the `TABLE_SPEC` setup to create the table with the same S3 bucket and path?

```
        # Setup int table. Creates a shallow clone of the `raw` FABA table in the `int` schema.
        # If the --no-clone option is provided a full table is created instead.
        if no_clone:
            self.spark.sql("DROP TABLE IF EXISTS int.financial_accounts_by_awards")
            call_command(
                "create_delta_table",
                "--destination-table",
                "financial_accounts_by_awards",
                "--spark-s3-bucket",
                spark_s3_bucket,
                "--alt-db",
                "int",
            )
```

kwhickey: Resolving. Summary:
- vacuuming needs to be addressed more holistically as this affects all of our tables (and s3 paths)
- creat_delta_table uses s3: Bucket is passed in as a param. Derive path from the schema and table name in the table spec

kwhickey (approved):
Approving latest changes.

dpb-bah (dismissed):

PR #3700, opened 2023-01-25
Update postgresql version and add liblzma headers

aguest-kc:
**Description:**
Updated Dockerfile to add liblzma headers package and to install a newer version of the PostgreSQL client.

**Technical details:**
Added the `xz-devel` package so that Python can build with LZMA compression supported and remove the following warning when launching Python inside a container:

> UserWarning: Could not import the lzma module. Your installed Python is incomplete. Attempting to use lzma compression will result in a RuntimeError.

Updated `postgresql10` to `postgresql13` since the version 10 client is no longer available through **pgdg-redhat-repo-latest.noarch.rpm**.

    - [x] Backend
    - [ ] Ops


kwhickey:
@aguest-kc can you set it to user postgresql13, since that's what we use in our environments.

aguest-kc:
Updated the postgresql version to 13 instead of 14.

dporth-frb (approved):

PR #3707, opened 2023-01-28
[PIPE-466] Usage of INT FABA 

collinwr:
**Descriptions**

Updates a few downstream jobs to use the `int` version of the `financial_accounts_by_awards` table:
- Subaward Search
- Transaction Search
- Award Search
- Touch Covid Awards


kwhickey (approved):
Pretty straightforward. Approved.

PR #3708, opened 2023-01-28
[PIPE-474]  raw fabs to tx norm date parsing fixes

kwhickey:
**Description:**
Adding further date-parsing handling of the published_fabs to transaction_normalized mappings for action_date and the two period of performance columns, since raw fabs records see string dates in a variety of formats"

**Technical details:**
- the `raw.published_fabs.action_date` column was being used in its raw form to derive the fiscal year, and needed to be parsed into a date first
- the `raw.published_fabs.period_of_performance_curr|star` columns were being casted to date, but that will now work with strings in the format of `mm/dd/YYYY`, which this raw table has

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/load_transactions_in_delta.py
+++ b/usaspending_api/etl/management/commands/load_transactions_in_delta.py
@@ -22,6 +22,7 @@ from usaspending_api.broker.helpers.last_load_date import (
     get_last_load_date,
     update_last_load_date,
 )
+from usaspending_api.common.data_classes import TransactionColumn
 from usaspending_api.common.etl.spark import create_ref_temp_views
 from usaspending_api.common.helpers.spark_helpers import (
     get_active_spark_session,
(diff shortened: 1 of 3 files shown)


usaspending_api/transactions/delta_models/transaction_fpds.py
```diff
@@ -363,6 +363,7 @@
     TransactionColumn("modification_number", "award_modification_amendme", "STRING"),
     TransactionColumn("non_federal_funding_amount", "NULL", "NUMERIC(23, 2)", "literal"),
     TransactionColumn("original_loan_subsidy_cost", "NULL", "NUMERIC(23, 2)", "literal"),
+    # All period_of_performance_* fields seen as: YYYY-MM-DD 00:00:00, so cast works
     TransactionColumn("period_of_performance_current_end_date", "period_of_performance_curr", "DATE", "cast"),
```
DManowitz-BAH, line 367: Although we have only seen `period_of_performance_*` fields as YYYY-MM-DD format in fpds, I'd still put the handling as `parse_string_datetime_to_date` to be more robust against the possibility of seeing other formats in the future.

DManowitz-BAH (approved):

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -595,12 +606,19 @@ def select_columns_transaction_fabs_fpds(bronze_table_name):
...
 
         def select_columns_transaction_normalized(bronze_table_name):
+            action_date_col = next(
+                filter(
+                    lambda c: c.dest_name == "action_date" and c.source == "action_date",
+                    FABS_TO_NORMALIZED_COLUMN_INFO if transaction_type == "fabs" else DAP_TO_NORMALIZED_COLUMN_INFO,
+                )
+            )
+            parse_action_date_sql_snippet = handle_column(action_date_col, bronze_table_name, is_result_aliased=False)
             select_cols = [
```
DManowitz-BAH, line 616: Rather than calling `handle_column` here, why not just call `build_date_format_sql`?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -513,67 +514,77 @@ def update_awards(self):
...
+            sql_snippet = f"""
+                CASE WHEN regexp({bronze_table_name}.{col.source}, '{regexp_mmddYYYY}')
+                          THEN {mmddYYYY_fmt}
+                     ELSE {YYYYmmdd_fmt}
+                END{' AS ' + col.dest_name if is_result_aliased else ''}
+            """
+            return sql_snippet
+
+        def handle_column(col, bronze_table_name, is_result_aliased=True):
             if col.handling == "cast":
```
DManowitz-BAH, line 567: If you changed the call to `handle_column` below to `build_date_format_sql`, you don't need to add the `is_result_aliased` argument to this function here, which is never used in its original called location and doesn't really make sense to be allowed for the literal case.

DManowitz-BAH (changes requested):

usaspending_api/transactions/delta_models/transaction_fpds.py
DManowitz-BAH: Had second thoughts after I approved.  See below.

usaspending_api/etl/management/commands/load_transactions_in_delta.py
kwhickey: @DManowitz-BAH the rationale for using the `handle_column` to derive the date-parsing SQL for `action_date` rather than directly calling `build_date_format_sql` was two-fold:
1. to allow that when date-parsing `action_date`, or any other column that needs it, just pass in the `TransactionColumn` object and don't predetermine _what kind_ of date-parsing is needed. Let the logic of `handle_column` read the value of the `TransactionColumn.handling` to determine which of the (currently) two types of date-parsing to do. A bit of defensive programming: If not done this way, and casting to `DATE` is hard coded as `is_casted_to_date=True` or `is_casted_to_date=True` in a call to `build_date_format_sql`, if for some reason the `TranactionColumn.handling` value is changed in the column spec, it will probably lead to a subtle bug because it will be unlikely someone will remember to also go find any call to `build_date_format_sql` for that column and also change `True` to `False` or vice versa.
2. That way a general pattern is established for any other `TransactionColumn` object that needs to have date-parsing SQL built and extracted for use somewhere --> just pass the `TransactionColumn` object to `handle_column` and let it figure out based on the attributes of the object how to build the sql (e.g. cast or not cast)

DManowitz-BAH: OK

DManowitz-BAH (approved):

usaspending_api/transactions/delta_models/transaction_fpds.py
kwhickey: I added the smart-date-parsing for FPDS too and committed

PR #3711, opened 2023-02-06
[PIPE-273] Updates Postgres File to C Linkage with Table

collinwr:
**Description:**
This PR adds a new table `c_to_d_linkage_updates` that contains only two fields `fianancial_accounts_by_awards_id` and `award_id`. This table is written to at the end of the `update_file_c_linkages_in_delta` command. It contains all updates necessary to the table. 

The `update_file_c_linkages` command has also been updated to use this table to determine which File C records need to be updated by default (instead of manually calculating them with the existing SQL files). A flag was added to the command to allow the old process to be used `--recalculate-linkages`

    - [ ] Backend
8. [ ] Jira Ticket [PIPE-273](https://federal-spending-transparency.atlassian.net/browse/PIPE-273):

```
```


usaspending_api/awards/delta_models/financial_accounts_by_awards.py
```diff
@@ -80,7 +80,7 @@
             UPPER(faba.fain) AS fain,
             UPPER(faba.uri) AS uri
         FROM
-            int.financial_accounts_by_awards AS faba
+            raw.financial_accounts_by_awards AS faba
```
collinwr, line 83: Had to change these back to `raw` because the `int` schema already has all the changes (and shows no records pending updates) by the time we write the final temp view to Postgres

collinwr:
> LGTM. Do you know if this still properly manages both the C->D but also the D->C relationships too?
@dpb-bah The process introduced in this command is an alternative to an existing process that only links File C->D. The only updates that it makes to File D are touching award records that were linked to or unlinked from.
In the new approach, there is no need to update awards because they were already updated in Delta before being copied to Postgres.

dpb-bah (approved):
LGTM

usaspending_api/etl/management/commands/update_file_c_linkages.py
```diff
@@ -9,28 +9,75 @@
...
+    UPDATE_LINKAGES_SQL = """
+    UPDATE
+        public.financial_accounts_by_awards AS faba
+    SET
+        award_id = updates.award_id
+    FROM
+        public.c_to_d_linkage_updates AS updates
+    WHERE
+        faba.financial_accounts_by_awards_id = updates.financial_accounts_by_awards_id;
+    """
```
kwhickey, line 29: I think it would be good here to only update the FABA records that need updating, so add an `AND faba.award_id IS DISTINCT FROM updates.award_id` would be helpful for performance. 

I still need to review further though if this sis necessary, or if the updates table only holds the changes in linkages from the last run.

usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
```diff
@@ -150,6 +160,15 @@ def handle(self, *args, **options):
         unlinked_count = self.get_unlinked_count("int")
         logger.info(f"Count of unlinked records after updates: {unlinked_count:,}")
 
+        # Write view back to Postgres for linkages
+        c_to_d_linkage_updates_df = self.spark.sql("SELECT * FROM union_all_priority;")
+        c_to_d_linkage_updates_df.write.jdbc(
+            url=get_usas_jdbc_url(),
+            table="public.c_to_d_linkage_updates",
```
kwhickey, line 167: Since it's only good for 1 day at a time (i.e. updates since last pipeline run) maybe the `temp.` schema is a better choice here.

kwhickey (approved):
Left a few comments for possible improvement, but not holding up approval. I think the first comment is moot as it appears the table only collects the FABA records that need to be updated based on the state of the data for that pipeline run only.

usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
collinwr: I was thinking about that too. I ultimately decided to leave it in `public` because the other tables in the `temp` schema are temporary themselves (swapped, dropped, and recreated), whereas only the data itself is temporary in this table. 
I can definitely see the argument though for the `temp` schema though.

usaspending_api/etl/management/commands/update_file_c_linkages.py
collinwr: I was thinking about something like this, but the table does indeed only hold updates. Shouldn't be any need to add this extra check

PR #3715, opened 2023-02-15
[DEV-9504] Add infrastructure spending amounts to awards ES index

aguest-kc:
**Description:**
Added three new infrastructure spending fields to the Award Search Delta model, the Award Search Postgres model, and the ElasticSearch template. Updated the SQL used to populate the three new Delta table columns.

**Technical details:**
Added `total_iija_outlays`, `total_iija_obligations`, and `iija_spending_by_defc` fields to the Award Search Delta model. Updated the `award_search_load_sql_string` to include a new JOIN to populate the three new infrastructure fields. The new JOIN clause uses the new `infrastructure` DEFC group value. Updated the Award Search Postgres model to include the three new infrastructure spending columns. Added the new fields to the es_award_template.json and award_delta_view.sql files to populate these new values in ElasticSearch.

    - [ ] Backend
3. [ ] Data validation completed
4. [ ] Jira Ticket [DEV-9504](https://federal-spending-transparency.atlassian.net/browse/DEV-9504):


usaspending_api/search/delta_models/award_search.py
```diff
@@ -474,7 +482,64 @@
...
+          GROUPED_BY_DEFC.def_code,
+          'outlay',
+          GROUPED_BY_DEFC.outlay,
+          'obligation',
+          GROUPED_BY_DEFC.obligation
+        )
+      )
+    ) AS iija_spending_by_defc,
+    sum(GROUPED_BY_DEFC.outlay) AS total_iija_outlay,
+    sum(GROUPED_BY_DEFC.obligation) AS total_iija_obligation
```
dpb-bah, line 503: Could these be indented following the other `JOIN`s?

dpb-bah (changes requested):

usaspending_api/search/delta_models/award_search.py
```diff
@@ -474,7 +482,64 @@
         GROUPED_BY_DEFC.award_id IS NOT NULL
     GROUP BY
         GROUPED_BY_DEFC.award_id
-) DEFC on DEFC.award_id = awards.id
+) COVID_DEFC on COVID_DEFC.award_id = awards.id
+-- Infrastructure Investment and Jobs Act (IIJA) spending
```
kwhickey, line 486: Everything looks really good. But I have lots to suggest here. Rather than following suit of the SQL pattern, there is probably an opportunity that needs to be explored on whether we can create e.g. a CTE that is built _ONCE_ that aggregates costs by `defc.code` or by `defc.code, defc.group_name`. Then the aggregations across ALL records only needs to be done once, and you could (maybe?) pick out which group of agg'd values you're interested in by joining/querying the CTE.

Doing in this way will work, but you're paying the cost of looking at and aggregating these records twice, or nearly twice (maybe some efficiencies by limiting the aggregated rows by group -- which you can still do, but it would be limited to all groups you want to agg (currently 2), not just 1 group).

image

Also, thank goodness for Spark. Something like this added to a Postgres materialized view SQL might have added hours to its runtime. You should still run this up in Databricks to get a sense of how much time it adds before/after.

kwhickey (changes requested):
Suggest trying to create a CTE to do the by-disaster-group grouping only once if possible.

usaspending_api/search/delta_models/award_search.py
kwhickey: If this becomes impossible to create a generic grouping a CTE because of too many factors (group_name, date filtering, etc.) maybe just leave it as you have it.

usaspending_api/etl/es_award_template.json
```diff
@@ -1,584 +1,611 @@
...
-        "total_covid_outlay": {
-          "type": "scaled_float",
-          "scaling_factor": 100
-        },
-        "recipient_uei": {
-          "type": "text",
-          "fields": {
-            "keyword": {
-              "type": "keyword"
+    "settings": {
```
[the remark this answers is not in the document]
aguest-kc: Looks like VS Code formats JSON files with 4 spaces and this file was previously indented with 2 spaces per level. I've changed it back to help reduce the diff size.

usaspending_api/search/migrations/0029_add_iija_spending_fields.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.2.18 on 2023-03-02 21:06
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('search', '0028_transaction_search_agency_indexes'),
```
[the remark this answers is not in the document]
aguest-kc: Changed the migration file's name and it's dependencies after merging in the QAT branch.

usaspending_api/etl/es_award_template.json
```diff
@@ -562,6 +562,30 @@
         "total_covid_outlay": {
           "type": "scaled_float",
           "scaling_factor": 100
+        },
+		"iija_spending_by_defc": {
```
dpb-bah, line 566: Could you fix this indent too?

dpb-bah (changes requested):

usaspending_api/etl/es_award_template.json
aguest-kc: The indentation of the iija fields should match the indentation of the covid fields now.

usaspending_api/search/delta_models/award_search.py
aguest-kc: After some testing, converting this to use a CTE instead of the LEFT OUTER JOINs doesn't have much of an affect on performance. Either version of the INSERT SQL will complete in 27 to 25 minutes.

dpb-bah (approved):

kwhickey (approved):
Approving as is

PR #3722, opened 2023-02-27
Fix: Updating `update_award_lookup_ids()` for existing awards

dpb-bah:
This function adds new transaction/award mappings to `award_id_lookup` for incoming transactions. The problem is that it creates new `award_id`s *even if the new transaction points to an existing award*. This updates the new records to reuse the existing `award_id` for already existing awards.

### DIFF
--- a/usaspending_api/etl/management/commands/load_transactions_in_delta.py
+++ b/usaspending_api/etl/management/commands/load_transactions_in_delta.py
@@ -986,7 +986,7 @@ class Command(BaseCommand):
             INSERT INTO int.award_id_lookup
             SELECT
                 COALESCE(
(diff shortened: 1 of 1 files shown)


DManowitz-BAH (commented):
This seems like a reasonable approach.  My only concern is over the COALESCE wrapping the window function.  It sounds like you tested it and it worked, though.  However, it might be leaving holes in the `award_id` sequence where a new transaction that referenced an existing award would have created a new `award_id` previously.  If we don't care that there might be holes in the `award_id` sequence going forward, then go ahead and merge it.  If we do care, then it might be better to either do 2 separate inserts or a single UNION insert where one group is new transactions that reference existing awards, and the other group is new transactions that are also new awards.

PR #3725, opened 2023-02-27
Fix: Updating `update_award_lookup_ids()` for existing awards - QAT

dpb-bah:
This function adds new transaction/award mappings to award_id_lookup for incoming transactions. The problem is that it creates new award_ids even if the new transaction points to an existing award. This updates the new records to reuse the existing award_id for already existing awards.


kwhickey (approved):
This is a good solution to lookup any award that already existed. 
Using the `int.awards` delta table could be an option, but @DManowitz-BAH pointed out that the `award_id_lookup` table is a better option since it will have deleted any award records for deleted awards at this point.

PR #3727, opened 2023-02-28
[PIPE-408] Rename inherited partition indexes and shrink swap transaction boundaries

kwhickey:
**Description:**
1. Keeping inherited indexes named according to convention during migrations
2. Shrinking boundaries around table+view renames during swap to avoid deadlocks

**Technical details:**
1. `copy_table_metadata` does not work if child/inherited indexes are not named the same as parent indexes except for the partition suffix. So migrations (for now) need to rename the auto-named inherited indexes.
2. Deadlocks have occurred if a table rename is happening in the same transaction as the dependent view renames that direct queries to the view
    - The suspicion is that queries can eek in by way of the view, which has not yet been locked for rename, that direct the query to the table, which is already locked for rename, and then the view cannot be locked due to this deadlock scenario
    - So make each rename operation atomic all on its own. Atomic transactions will cover 1) moving to target schema, 2) renaming active to old, 3) renaming temp to active ... but only for 1 table or view at a time, not the whole bunch.

4. [ ] API documentation updated
5. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
6. [ ] Matview impact assessment completed
7. [ ] Frontend impact assessment completed
8. [ ] Data validation completed
9. [ ] Appropriate Operations ticket(s) created
10. [ ] Jira Ticket [DEV-123](https://federal-spending-transparency.atlassian.net/browse/DEV-123):

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/swap_in_new_table.py
+++ b/usaspending_api/etl/management/commands/swap_in_new_table.py
@@ -3,6 +3,7 @@ import logging
 import re
 
 from datetime import datetime
+from pprint import pformat
 from typing import List, OrderedDict, Optional
 from django.core.management import BaseCommand
 from django.db import connection, ProgrammingError, transaction
@@ -179,8 +180,9 @@ class Command(BaseCommand):
             self.analyze_temp_table(cursor)
             self.grant_privs_on_temp_table(cursor)
 
-            # Do the swap via renames
-            self.swap_table_view_sql(cursor)
+            # Do the swaps via renames
+            self.swap_dependent_views(cursor)
+            self.swap_table(cursor)
 
             if not options["keep_old_data"]:
                 self.drop_old_views(cursor)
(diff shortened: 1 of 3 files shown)


usaspending_api/etl/management/commands/swap_in_new_table.py
```diff
@@ -281,39 +283,50 @@ def swap_index_sql(self, cursor):
...
-        rename_sql = [
-            f"ALTER TABLE {self.temp_table_name} SET SCHEMA {self.curr_schema_name};",
-            sql_table_template.format(
-                old_table_name=self.curr_table_name, new_table_name=f"{self.curr_table_name}{self.old_suffix}"
-            ),
-            sql_table_template.format(old_table_name=self.temp_table_name, new_table_name=f"{self.curr_table_name}"),
-        ]
-
-        sql_view_template = "ALTER {mv_s}VIEW {old_view_name} RENAME TO {new_view_name};"
+    def swap_dependent_views(self, cursor):
```
dpb-bah, line 286: is there a reason why we’re not making `swap_dependent_views` atomic like `swap_table`?

dpb-bah (changes requested):

dpb-bah (approved):

usaspending_api/etl/management/commands/swap_in_new_table.py
dpb-bah: Missed the atomic within that function 👍

PR #3739, opened 2023-03-05
[HOTFIX] Using award_search instead of fabs

dpb-bah:


kwhickey (approved):
Tested the query out and it vastly improves or avoids bottlenecks for assistance awards downloads. Merging to prod.

PR #3804, opened 2023-05-04
[DEV-9533] Current Congressional District Derivations

dpb-bah:
**Description:**
Uses the Broker CD derivation tables and the decision trees to derive the current congressional districts for transactions and awards.

**Technical details:**
N/A

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-9533](https://federal-spending-transparency.atlassian.net/browse/DEV-9533):
    - [x] Before / After data comparison (tested on QAT, adds 10-20 minutes per AS/TS generation)

```
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,8 +116,7 @@ script:
   - black --check --diff .
   - python manage.py check_for_endpoint_documentation
   - dredd > dredd-results.txt && echo '! grep -E "^[warn:|error:]" dredd-results.txt' | bash
-#  - pytest --durations 50 --ignore-glob='**/tests/integration/*' --cov=usaspending_api --cov-report= --reuse-db -rsx
-#  - pytest --durations 50 --override-ini=python_files='**/tests/integration/*' --cov=usaspending_api --cov-append --cov-report term --cov-report xml:coverage.xml --reuse-db -rsx
-  - pytest --durations 50 usaspending_api/etl/tests/integration/test_load_to_from_delta_subawards.py --cov-append --cov-report term --cov-report xml:coverage.xml --reuse-db -rsx
+  - pytest --durations 50 --ignore-glob='**/tests/integration/*' --cov=usaspending_api --cov-report= --reuse-db -rsx
+  - pytest --durations 50 --override-ini=python_files='**/tests/integration/*' --cov=usaspending_api --cov-append --cov-report term --cov-report xml:coverage.xml --reuse-db -rsx
 after_script:
   - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT


usaspending_api/search/delta_models/award_search.py
```diff
@@ -291,8 +291,32 @@
...
-  'TEST CUR REC CONGR AS' AS recipient_location_congressional_code_current,
+  (CASE
+      WHEN (
+          COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) <> 'USA'
+      ) THEN NULL
+      WHEN (
+          COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) = 'USA'
+          AND rl_state_cd.congressional_district_no IS NOT NULL
+      ) THEN rl_state_cd.congressional_district_no
+      WHEN (
```
dporth-frb, line 302: There's a lot of code duplication here that I think you can reduce.

You could reduce the number of times this predicate, COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) = 'USA', is duplicated by using a COALESCE.

For example, to lines 302 to 317 can be reduced to only 4 lines.

`
```
WHEN (
          COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) = 'USA'
          AND COALESCE(rl_state_cd.congressional_district_no, rl_zips.congressional_district_no, rl_cd_zips_grouped.congressional_district_no, l_cd_city_grouped.congressional_district_no, rl_cd_county_grouped.congressional_district_no) IS NOT NULL
      ) THEN COALESCE(rl_state_cd.congressional_district_no, rl_zips.congressional_district_no, rl_cd_zips_grouped.congressional_district_no, l_cd_city_grouped.congressional_district_no, rl_cd_county_grouped.congressional_district_no)
```

dporth-frb (changes requested):

usaspending_api/search/delta_models/award_search.py
```diff
@@ -311,8 +335,41 @@
...
+      ) THEN NULL
+      WHEN (
+          transaction_fabs.place_of_performance_scope = 'Multi-State'
+      ) THEN '90'
+      WHEN (
+          (transaction_fabs.place_of_performance_scope = 'State-wide'
+          OR transaction_fpds.place_of_perform_country_c = 'USA')
+          AND pop_state_cd.congressional_district_no IS NOT NULL
+      ) THEN pop_state_cd.congressional_district_no
+      WHEN (
```
dporth-frb, line 351: You can apply the same idea from my previous comment, to lines 351-360

dporth-frb (changes requested):

usaspending_api/transactions/delta_models/transaction_search.py
```diff
@@ -555,8 +555,32 @@
         LPAD(CAST(CAST(REGEXP_EXTRACT(COALESCE(transaction_fpds.legal_entity_congressional, transaction_fabs.legal_entity_congressional), '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 2, '0')
             AS recipient_location_congressional_code,
         RL_DISTRICT_POPULATION.latest_population AS recipient_location_congressional_population,
-        -- TODO: include recipient_location_congressional_code_current derivation
-        'TEST CUR REC CONGR TS' AS recipient_location_congressional_code_current,
+        (CASE
+            WHEN (
+                COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) <> 'USA'
+            ) THEN NULL
+            WHEN (
```
dporth-frb, line 562: I think you can apply the same idea, detailed in my previous comments, here as well.

dporth-frb (changes requested):

usaspending_api/search/delta_models/award_search.py
dpb-bah: Good catch! I had to test it out to make sure everything worked and glad it worked out. Updated accordingly.

dporth-frb (approved):

usaspending_api/search/delta_models/award_search.py
```diff
@@ -291,8 +291,16 @@
...
             AS recipient_location_congressional_code,
-  -- TODO: include recipient_location_congressional_code_current derivation
-  'TEST CUR REC CONGR AS' AS recipient_location_congressional_code_current,
+  (CASE
+      WHEN (
+          COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) <> 'USA'
+      ) THEN NULL
+      WHEN (
+          COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) = 'USA'
+          AND COALESCE(rl_state_cd.congressional_district_no, rl_zips.congressional_district_no, rl_cd_zips_grouped.congressional_district_no, rl_cd_city_grouped.congressional_district_no, rl_cd_county_grouped.congressional_district_no) IS NOT NULL
```
collinwr, line 300: This part of the check might be unnecessary too. Couldn't you use the output value of the `COALESCE` whether it's null or not?

usaspending_api/search/delta_models/award_search.py
```diff
@@ -311,8 +319,36 @@
...
+      WHEN (
+          transaction_fabs.place_of_performance_scope = 'Foreign'
+          OR transaction_fpds.place_of_perform_country_c <> 'USA'
+      ) THEN NULL
+      WHEN (
+          transaction_fabs.place_of_performance_scope = 'Multi-State'
+      ) THEN '90'
+      WHEN (
+          (transaction_fabs.place_of_performance_scope = 'State-wide'
+          OR transaction_fpds.place_of_perform_country_c = 'USA')
```
collinwr, line 332: Similar to my comment about for RL. Can these checks for `OR transaction_fpds.place_of_perform_country_c = 'USA'` (after the first one) be removed to simplify the query?

usaspending_api/search/delta_models/award_search.py
collinwr, line 300: Do we need some kind of check that it's state-wide here? Otherwise, won't this always resolve to ` rl_state_cd.congressional_district_no` because all states exist in this table?

usaspending_api/search/delta_models/award_search.py
collinwr: Same as my comment above about resolving based on the `pop_state_cd` table, won't this always yield a result, essentially setting all records to a state-wide district?

usaspending_api/search/delta_models/award_search.py
```diff
@@ -291,8 +291,16 @@
...
   LPAD(CAST(CAST(REGEXP_EXTRACT(COALESCE(transaction_fpds.legal_entity_congressional, transaction_fabs.legal_entity_congressional), '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 2, '0')
             AS recipient_location_congressional_code,
-  -- TODO: include recipient_location_congressional_code_current derivation
-  'TEST CUR REC CONGR AS' AS recipient_location_congressional_code_current,
+  (CASE
+      WHEN (
+          COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) <> 'USA'
+      ) THEN NULL
+      WHEN (
+          COALESCE(transaction_fpds.legal_entity_country_code, transaction_fabs.legal_entity_country_code) = 'USA'
```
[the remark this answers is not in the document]
dpb-bah: That makes sense. I think it was left over when I was determining the flow and it had more conditions.

usaspending_api/search/delta_models/award_search.py
dpb-bah: Oh I see. Just do `ELSE` with the `COALESCE`. If it's `NULL` it hits the old `ELSE`. If not, it hits the old `WHEN`.

dpb-bah: The `rl_state_cd `/`rl_cd_state_grouped` should only be populated for non `90` due to the `JOIN` definition.

usaspending_api/search/delta_models/award_search.py
dpb-bah: I can see the logic with Recipient Location CDs but in this case we're accounting for two different paths together (one being based on ppop scope for FABS, the other based on county code not being USA for FPDS or FABS). So I'd be worried if we took out the following USA checks for FPDS.

Say a record is FPDS, the ppop country code = 'USA', and it would derive for a state-wide code of '00'. Then when it hits the condition:
```
(
          (transaction_fabs.place_of_performance_scope = 'State-wide')
          AND pop_cd_state_grouped.congressional_district_no IS NOT NULL
      )
      ```
It would move on cause it doesn't meet the ppop scope (since that is null for FPDS) and set the condition to False.

collinwr: Okay, I think I see what you mean. Thanks for the clarification!

collinwr (approved):

PR #3817, opened 2023-05-21
[DEV-9834] Spending over time now supports "New Awards Only"

dporth-frb:
**Description:**
The endpoint `spending_over_time` now consumes the `date_type` field that can appear in the time period filter object. Previously, this endpoint had a bug and did not consume this field. Another consequence, of this bug fix is that this endpoint now supports the "New Awards Only" filter. To get new awards only from this endpoint, include "date_type: date_signed" in the time period filter object of your request to this endpoint.

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-9834](https://federal-spending-transparency.atlassian.net/browse/DEV-9834):

5. The nature of this work does not require frontend impact assessments
7. No operation tickets needed.


usaspending_api/agency/v2/views/new_award_count.py
dporth-frb, line 1: I updated this file with the changes I made because it allowed me to simplify the logic in `_TimePeriods` class. After this change, the logic in that class doesn't need to support 5 different date type keys

usaspending_api/common/query_with_filters.py
```diff
@@ -105,20 +105,39 @@ class _TimePeriods(_Filter):
...
+        if "time_period_obj" not in options or options.get("time_period_obj") is None:
+            return cls._default_elasticsearch_query(filter_values, query_type, **options)
+
         time_period_query = []
-        if options:
-            gte_field = options.get("gte_field", "action_date")
-            lte_field = options.get("lte_field", "date_signed" if query_type == _QueryType.AWARDS else "action_date")
+        for v in filter_values:
+            time_period_obj = options["time_period_obj"]
+            time_period_obj.filter_value = v
```
dpb-bah, line 116: I understand single letter vars for simple looping through items in list comprehensions but here it was a little confusing seeing line 116. Could this var be renamed to `filter_value`? I know this wasn't originally your code but I would still like it fixed for clarity. 

Also applies to the loops starting at 40, 97, 134, and 159.

dpb-bah (changes requested):

usaspending_api/common/query_with_filters.py
dporth-frb: Addressed

dpb-bah (approved):

PR #3823, opened 2023-05-24
[DEV-9834] Spending over time now supports "New Awards Only"

dporth-frb:
**Description:**
The endpoint `spending_over_time` now supports a `new_awards_only` member in the `date_type` field. The previous PR https://github.com/fedspendingtransparency/usaspending-api/pull/3817 was not enough to achieve the desired behavior of new awards only for transactions. Unlike prime awards, transactions require an extra comparison. This PR implements that extra comparison

    - [x] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-9834](https://federal-spending-transparency.atlassian.net/browse/DEV-9834):

5. The nature of this work does not require frontend impact assessments
7. No operation tickets needed.


dpb-bah (dismissed):

dpb-bah (approved):

usaspending_api/api_contracts/contracts/v2/search/spending_over_time.md
```diff
@@ -110,10 +110,16 @@ This endpoint returns a list of aggregated award amounts grouped by time period
...
         + `date_signed`
-          + This date type value is equivalent to `award_date_signed`. Behind the scenes, if you provide this input we map it to `award_date_signed`.
+            This date type value is equivalent to `award_date_signed` for transactions. Behind the scenes, if you provide this input we map it to `award_date_signed`.
+            `date_signed` is not supported when `subawards` field is set to true.
         + `action_date`
-          + This date type value is the default.
+            This date type value is the default.
         + `last_modified_date`
+        + `new_awards_only`
+            Indicates when the results should reflect new awards only. When the `date_type` field equals `new_awards_only` you should expect only
```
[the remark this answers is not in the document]
dporth-frb: Yeah, you're right. I'll fix that.

kwhickey (changes requested):
Some OOP changes desired and thinking about where best to create new packages/namespaces.

usaspending_api/common/filters/time_period/decorators/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .new_awards_only import NewAwardsOnlyTimePeriod
```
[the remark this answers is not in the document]
dporth-frb: > And can time_period be used in a common way, across all django "apps" (i.e. domain areas / bounded contexts ... like awards, accounts, etc.)?

It can be used across all django apps. Anything that uses a time period filter, which include multiple apps.

usaspending_api/common/filters/time_period/time_period.py
```diff
@@ -54,7 +54,27 @@ def lte_date_type(self):
...
+
+    @abstractmethod
+    def gte_date_range(self):
+        """
+        Returns:
+            A nested dictionary indicating a column (the key) which should
+            store a value greater than or equal to (as specified in a key in the nested dictionary) a
+            date (the value in the nested dictionary).
+        """
+        pass
```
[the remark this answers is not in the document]
dporth-frb: Anything that inherits from this class must define this method. A compile time error will occur if a child class inherits from this parent class and doesn't re-declare this method.

usaspending_api/common/filters/time_period/decorators/new_awards_only.py
```diff
@@ -0,0 +1,75 @@
...
+        }
+        self._additional_lte_filters_for_new_awards_only = {
+            _QueryType.TRANSACTIONS: [
+                {"action_date": {"lte": self.end_date()}},
+                {"award_date_signed": {"lte": self.end_date()}},
+            ]
+        }
+
+    @property
+    def filter_value(self):
```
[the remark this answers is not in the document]
dporth-frb: Will do.

usaspending_api/common/filters/time_period/time_period.py
```diff
@@ -54,7 +54,27 @@ def lte_date_type(self):
         """
         Returns:
             A date type that indicates the name of a column in which
-            stores the date values that end date must be less than.
+            stores the date values that end date must be less than or equal to.
             Can return None when lte_date_type doesn't exist.
         """
         pass
+
+    @abstractmethod
```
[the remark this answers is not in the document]
dporth-frb: `I` definitely is a convention of .NET and Java too. However, I included it to imply that this class should not have any implementations only declared behavior. An abstract class can have abstract methods and concrete methods. Does this change your opinion?

usaspending_api/common/filters/time_period/decorators/new_awards_only.py
```diff
@@ -0,0 +1,75 @@
...
+    ON TOP of an existing time period filter.
+    """
+
+    def __init__(self, transaction_search_time_period_obj, query_type, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._query_type = query_type
+        self._transaction_search_time_period_obj = transaction_search_time_period_obj
+        self._additional_gte_filters_for_new_awards_only = {
+            _QueryType.TRANSACTIONS: [
+                {"action_date": {"gte": self.start_date()}},
```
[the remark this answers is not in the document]
dporth-frb: > May be best in this case just to duplicate the logic rather than trying to get maximal reuse. e.g.

Duplicating logic is the whole reason why these bugs exist.

usaspending_api/common/filters/time_period/decorators/__init__.py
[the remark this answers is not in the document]
dporth-frb: I can move it to filters though.

usaspending_api/common/filters/time_period/time_period.py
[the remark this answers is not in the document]
kwhickey: My mistake, all good. For some reason I read one of these functions as not having been declared `@abstractmethod` but I think GitHub was playing tricks on my by hiding lines not affected in the commit:
image

usaspending_api/common/filters/time_period/time_period.py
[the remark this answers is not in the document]
kwhickey: ~~Should all methods be declared abstract then? There were some that were not marked `@abstractmethod` and whose concrete methods were just `pass`.~~  _See [comment below](https://github.com/fedspendingtransparency/usaspending-api/pull/3823#discussion_r1205811384) on me missing this._

Even still, I think we can just leave it declared as an Abstract class (e.g. prefix with `Abstract`) and it just so happens that this abstract class is acting more like an interface in the general programming sense. No need to try to create artificial interfaces in python. You can add in the class comments the usage of this is intended to be like an Interface. in that all methods must be implemented and any new method ought to be marked `@abstractmethod` (I wish there was a way to enforce that in python).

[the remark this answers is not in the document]
dporth-frb: Ok will do.

usaspending_api/common/filters/time_period/decorators/__init__.py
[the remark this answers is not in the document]
kwhickey: Ok, i see it deals with dates generically. Still, may want to do some restructuring of the concrete impls to organize by their implied domain areas. I can take this to an offline conversation.

usaspending_api/common/filters/time_period/transaction_search.py
```diff
@@ -2,7 +2,7 @@
 
 
 class TransactionSearchTimePeriod(ITimePeriod):
```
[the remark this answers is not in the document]
kwhickey: Follow original comment... resolving this one.

usaspending_api/common/filters/time_period/decorators/__init__.py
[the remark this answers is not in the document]
dporth-frb: That'd be great, like maybe put the abstract class in common, then put app specific implementations in their app. So transaction search implementation would go in the search app?

usaspending_api/common/filters/time_period/decorators/new_awards_only.py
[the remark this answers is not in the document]
kwhickey: Yep, I saw the change with taking this out of the constructor to capture them in a implementation-specific property. That'll work.

usaspending_api/common/filters/time_period/decorators/__init__.py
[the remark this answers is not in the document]
dporth-frb: Let me know if you'd like me to do so in this PR. More than happy to do it.

usaspending_api/common/filters/time_period/time_period.py
```diff
@@ -1,28 +1,29 @@
 from abc import ABC, abstractmethod
+from types import list
 
 
-class ITimePeriod(ABC):
+class TimePeriod(ABC):
```
kwhickey: I still think there's value in signifying in the name that this isn't a class that can be instantiated directly. So: `AbstractTimePeriod` or `ATimePeriod` (we've only used `Abstract` as a prefix to my knowledge in the past, when adding that kind of detail to the class name).

usaspending_api/common/filters/time_period/decorators/__init__.py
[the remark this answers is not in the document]
kwhickey: Ideas I'm thinking

option 1:
- move this abstract class up into `__init__.py` as it's truly the base class of a series of concrete classes that may fall into this `time_period` package.
- then within that package, have domain-specific modules that capture the implementation. E.g. in cases where this base class is being used against awards and transactions, create a `...time_period.awards.py` module in the `time_period` package. That module can have both the `NewAwardsOnlyTimePeriod` and `TransactionSearchTimePeriod` classes in it (it's best to store related classes in a module in python, like you would store them in a folder/package in java or .net)

option 2:
- remove the `time_period` package
- have a `...filters.time_period.py` module that has the base Abstract class declared
- move all of the domain-specific impls into e.g. `usaspending_api.awards.v2.filters` 

I tend to like option 1 better if it the impls don't get too entangled with HTTP API-specific features like requests and responses. If the smell like HTTP API handling, move them down to the option 2 level.

usaspending_api/common/filters/time_period/decorators/new_awards_only.py
```diff
@@ -0,0 +1,87 @@
+from usaspending_api.common.filters.time_period.time_period import ITimePeriod
+from usaspending_api.search.filters.elasticsearch.filter import _QueryType
+
+NEW_AWARDS_ONLY_KEYWORD = "new_awards_only"
+
+
+class NewAwardsOnlyTimePeriod(ITimePeriod):
+    """A decorator class that can be used to apply a new awards only filter
```
kwhickey, line 8: Since "Decorators" in python have their own connotation (like attributes in .net and annotations in java), probably change this to say
```
"""A composable class that can be used according to the Decorator software design pattern, to apply ...
```

usaspending_api/common/filters/time_period/time_period.py
```diff
@@ -31,7 +32,7 @@ def start_date(self):
         pass
 
     @abstractmethod
-    def end_date(self):
+    def end_date(self) -> str:
```
kwhickey, line 35: This confuses me that the `end_date` and `start_date` would return a string. It leaves a lot open to interpretation of what date format that string may need to be in, and different filtering implementations may need to actually provide the date filter string in _different_ formats according to the data they're filtering against. So it may be best to make this return the actual `datetime` object, and let it be formatted as needed at the point where the filter is applied. 

If it can't be done as a date object, then the type hints do help a lot, but you should also add some cautions and caveats in the docs for this method as to how and when the date should be converted to a string, and how to determine what string format it should be in.

usaspending_api/common/filters/time_period/time_period.py
```diff
@@ -1,28 +1,29 @@
 from abc import ABC, abstractmethod
+from types import List
 
 
-class ITimePeriod(ABC):
+class TimePeriod(ABC):
     """Time period interface for storing data related to a time period filter."""
 
     def __init__(self, *args, **kwargs):
```
kwhickey, line 8: Per the comment about using dates, is there any way to strongly type the input parameters? It diminishes the usefulness of this if impl can go in many directions and isn't required to comply with some basics like a start_date and end_date in the constructor.

usaspending_api/common/filters/time_period/time_period.py
```diff
@@ -50,11 +51,31 @@ def gte_date_type(self):
...
-    def lte_date_type(self):
+    def lte_date_type(self) -> str:
         """
         Returns:
             A date type that indicates the name of a column in which
-            stores the date values that end date must be less than.
+            stores the date values that end date must be less than or equal to.
             Can return None when lte_date_type doesn't exist.
         """
         pass
```
kwhickey, line 61: If this is returning a column name, `lte_date_type` and `gte_date_type` may not be the best choices for this function. `Type` implies a class type (like `datetime` or `timespan` etc.). 

Can this just say `lte_date_field` and `gte_date_field` or replace field with "col" or "column" or "field_name" or "column_name" etc.

usaspending_api/common/filters/time_period/time_period.py
```diff
@@ -50,11 +51,31 @@ def gte_date_type(self):
...
+
+    @abstractmethod
+    def gte_date_range(self) -> List[dict]:
```
kwhickey, line 64: Get deeper with the type hints here to really hammer home what's expected:

e.g. is it `List[Dict[str, datetime]` or `List[Dict[str, str]`? 
If the latter, see above thoughts around storing string-formatted dates vs. actual date objects. 

Same with `lte_date_range`

kwhickey (changes requested):
Some further thoughts on naming and uses of python types for various functions.

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: Just made that change

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: It doesn't return a datetime. It returns the value in the filter which is a string

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: Got it. Will do.

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: Maybe I need to change the name of this method then?

usaspending_api/common/filters/time_period/decorators/new_awards_only.py
dporth-frb: Ok.

usaspending_api/common/filters/time_period/time_period.py
kwhickey: Right, but my comment was indicating it probably should return a datetime -- or caution in the docstring about how best turn a datetime into a string. 

I think what may be confusing is that we're passing in a filter to this filter's constructor in some of the concrete impls? That seems convoluted, and I was thinking this would be the origin of building-up a filter from primitive values (i.e. datetime objects).

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: Yep, I'll make that change

usaspending_api/common/filters/time_period/decorators/__init__.py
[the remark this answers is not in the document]
dporth-frb: Ok will do.

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: Building up time period filters does not involve any actual date objects. It's all strings, that's what our exiting ES_Q expect. We are passing the time period input from the user into this filter's constructor. I think the name of the methods need to just be updated.

dporth-frb: I don't think we should return a datetime because calling code doesn't need it as a date object. It doesn't need that anywhere. Which is why I suggested changing the method name.

kwhickey: Maybe what we need is a concrete object (not a dictionary) representing the Filter object that is coming in from the API request? 

Or a dataclass. 

Then 
- Api request -> JSON -> <convert from json reqeust filter to python object> = `<<RequestFilterObject>>`
- `ConcreteTimePeriod(request_date_filter: RequestFilterObject)`
- and decide if RequestFilterObject is always used, and put up on AbstractTimePeriod's constructor,
    - Or if TimePeriod has an explicitly constructor with decomposing params into of `TimePeriod(start, end, start_range, end_range)` etc.
    - and concrete impls like `ConcreteTimePeriod(request_date_filter: RequestFilterObject)` then just decompose that RequestFilterObject input param and make a call up to the explicit `TimePeriod(start, end, start_range, end_range)`  constructor with all the parts it extracted out of the  RequestFilterObject

dporth-frb: Wouldn't factory pattern make more sense for choosing which concrete implementation to use? We already know which concrete implementation to use. For example, see usaspending_api/search/v2/views/spending_over_time.py the method 182

dporth-frb: So now instead of having a method that has a bunch of `if` statements directing flow based on whether the query type is awards or transactions, the method works with any subsclass of a `AbstractTimePeriod`. This subclass encapsulates it's custom behavior.

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: Wait, no date type is exactly what we call it in an incoming request to our api. That's why it says date type because the field in the request is `date_type`

kwhickey: But in moving these common filters outside the api realm, it should have no relation to or expectations of knowledge what the API calls things or their request/response payloads. If this is in fact very API specific, it should be moved under API namespaces. 
The way these filters are used is to apply them directly to domain/ORM objects when filtering data, so it shouldn't be coupled to the API request/response life cycle.

dporth-frb: These classes are for our API's filter dictionary.

dporth-frb: So we need to move them, they are not intended for any other filters.

usaspending_api/common/filters/time_period/time_period.py
kwhickey: > I feel like we are over thinking it and a factory may be overkill. Transaction searches use time period consistently and differently than other apps. We know whether we want to use a transaction search time period the moment a request hits our api it doesn't change in the middle of a request.

This may mean that this filter translation is in fact coupled to the API and needs to live closer to the API code rather than a globally reusable and generic time period filter. 

Also, is "TransactionSearch" here indicating
- a) the `TransactionSearch` ORM object which is tied to the `transaction_search` DB table?
- b) or, the transaction data stored in the `*transactions` elasticsearch indexes?
- c) or, the user activity of "using Advanced Search to search across transaction data"

I feel like there's an implication in the name that I may be misinterpreting. It leads me to believe `a`

dporth-frb: These classes are intended to solely represent the time period object that our api accepts. It's coupled to that concept.

dporth-frb: Moving them to awards doesn't make sense, these are not awards specific. Any ideas on where else they could be moved to?

dporth-frb: Moving them to awards doesn't make sense, these are not awards specific. Any ideas on where else they could be moved to?

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: I don't understand how these classes are outside the api realm. They are under `usaspending_api`, how is that not considered the api realm?

usaspending_api/common/filters/time_period/time_period.py
dporth-frb: Going move the classes

usaspending_api/common/query_with_filters.py
```diff
@@ -115,16 +115,18 @@ def generate_elasticsearch_query(cls, filter_values: List[dict], query_type: _Qu
             time_period_obj = options["time_period_obj"]
             time_period_obj.filter_value = filter_value
 
-            start_date = time_period_obj.start_date()
-            end_date = time_period_obj.end_date()
-
             # Keywords must be strings, hence f-strings here
```
collinwr, line 118: This comment is no longer necessary

usaspending_api/search/filters/time_period/decorators/new_awards_only.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def __init__(self, transaction_search_time_period_obj: AbstractTimePeriod, query_type: _QueryType, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._query_type = query_type
+        self._transaction_search_time_period_obj = transaction_search_time_period_obj
+
+    @property
+    def _additional_lte_filters_for_new_awards_only(self):
+        # Making this variable a property to ensure it grabs
+        # end date on the fly in case it wasn't set beofre
```
collinwr, line 23: Typo on `before`

collinwr (approved):
If Keith is good with the changes Desmond and I discussed in response to his feedback, I'm good to approve

usaspending_api/search/filters/time_period/query_types.py
```diff
@@ -1,13 +1,27 @@
...
+        Parameters
+        ----------
+        default_start_date : str
+            This is the start date value to use when start date is not present in the filter value.
+            See start date in [TransactionSearchTimePeriodObject](#usaspending_api/api_contracts/search_filters.md#transaction-search-time-period-object)
+            for definition.
+        default_end_date : str
+            This is the end date value to use when end date is not present in the filter value.
+            See end date in [TransactionSearchTimePeriodObject](#usaspending_api/api_contracts/search_filters.md#transaction-search-time-period-object)
+            for definition.
```
kwhickey, line 23: Minor thing on standard for docstrings -- look to the Google Python Style guide: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

.e.g. `Args` instead of Parameters with lines under

kwhickey (approved):
Approved! (with minor comment about docstring format that could be cleaned up later)

usaspending_api/search/filters/time_period/query_types.py
dporth-frb: Oh gotcha. Ok I'll make that update. I was not aware that was the team standard. I believe I'm using numpy.

collinwr (approved):
Approving latest docstring changes

PR #3835, opened 2023-05-30
DEV-9865: SubawardSearch Current CD Columns

dpb-bah:
**Description:**
Simply includes four new current CD columns in rpt.subaward_search (both in delta and postgres) and includes the derivation logic.

**Technical details:**
N/A

    - [ ] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-9865](https://federal-spending-transparency.atlassian.net/browse/DEV-9865):
    - [x] Performance evaluation of affected (API | Script | Download)  (time to take to generate DBR SubawardSearch: 15min)

```
```


usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -461,22 +465,36 @@
...
+        LATEST_CURRENT_CD.pop_congressional_code_current AS place_of_performance_congressional_current,
         fabs.place_of_performance_scope AS place_of_perform_scope,
         COALESCE(UPPER(bs.sub_place_of_perform_country_co), 'USA') AS sub_place_of_perform_country_co,
         pcc.country_name AS sub_place_of_perform_country_name,
         LPAD(CAST(CAST(REGEXP_EXTRACT(pop.county_numeric, '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 3, '0') AS sub_place_of_perform_county_code,
         pop.county_name AS sub_place_of_perform_county_name,
         LEFT(COALESCE(sub_place_of_performance_zip, ''), 5) AS sub_place_of_perform_zip5,
         pop.census_code AS sub_place_of_perform_city_code,
-        LPAD(CAST(CAST(REGEXP_EXTRACT(UPPER(bs.sub_place_of_perform_congressio), '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 2, '0') AS sub_place_of_perform_congressio
+        LPAD(CAST(CAST(REGEXP_EXTRACT(UPPER(bs.sub_place_of_perform_congressio), '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 2, '0') AS sub_place_of_perform_congressio,
```
dporth-frb, line 491: It may be nice to add a comment explaining what this random regex string is for or what it represents

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -539,6 +560,57 @@
         global_temp.ref_country_code AS rcc
             ON (rcc.country_code = UPPER(bs.sub_legal_entity_country_code)
                 AND bs.sub_legal_entity_country_code IS NOT NULL)
+    LEFT OUTER JOIN
+        global_temp.cd_state_grouped pop_cd_state_grouped ON (
+            pop_cd_state_grouped.state_abbreviation=UPPER(bs.sub_place_of_perform_state_code)
+            AND pop_cd_state_grouped.congressional_district_no <> '90'
```
dporth-frb, line 566: Same with this random 90. Might be helpful for future maintainers to store this in a variable with a descriptive name or leave a comment. Additionally, you may want to go the variable route because 90 is repeated multiple times here.

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -539,6 +560,57 @@
...
+            AND pop_cd_city_grouped.state_abbreviation=UPPER(bs.sub_place_of_perform_state_code)
+        )
+    LEFT OUTER JOIN
+        global_temp.cd_city_grouped rl_cd_city_grouped ON (
+            rl_cd_city_grouped.city_name=UPPER(bs.sub_legal_entity_city_name)
+            AND rl_cd_city_grouped.state_abbreviation=UPPER(bs.sub_legal_entity_state_code)
+        )
+    LEFT OUTER JOIN
+        global_temp.cd_county_grouped pop_cd_county_grouped ON (
+            pop_cd_county_grouped.county_number=LPAD(CAST(CAST(REGEXP_EXTRACT(pop.county_numeric, '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 3, '0')
```
dporth-frb, line 605: This regex has appeared three times now. It may be a good idea to store in a variable so that it does not need to be duplicated.

dporth-frb (changes requested):

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -539,6 +560,57 @@
...
+            AND pop_cd_county_grouped.state_abbreviation=UPPER(bs.sub_place_of_perform_state_code)
+        )
+    LEFT OUTER JOIN
+        global_temp.cd_county_grouped rl_cd_county_grouped ON (
+            rl_cd_county_grouped.county_number=LPAD(CAST(CAST(REGEXP_EXTRACT(rec.county_numeric, '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 3, '0')
```
dporth-frb, line 610: This regex has appeared three times now. It may be a good idea to store in a variable so that it does not need to be duplicated.

dporth-frb (dismissed):

usaspending_api/search/delta_models/subaward_search.py
dpb-bah: Yeah it's everywhere in AwardSearch, TransactionSearch, and SubawardSearch leftover from previous derivations. We could introduce more variables to these functions (like the columns on the `INSERT OVERWRITE`). My hesitation comes from making these queries too dynamic, which results in the labyrinth that is the download code functionality or all string manipulation in  load_transactions_in_delta`. We could go down that route but I would say that's outside the scope of this ticket and should be slated as a whole ticket for tech debt.

usaspending_api/search/delta_models/subaward_search.py
dpb-bah: I can include a comment to explain the `90` but don't want to go down the variable route as explained in the other comment.

usaspending_api/search/delta_models/subaward_search.py
dpb-bah: It's leftover from AwardSearch, TransactionSearch, and SubawardSearch but I can include a comment to explain them in all the files.

collinwr (approved):

usaspending_api/search/delta_models/subaward_search.py
dporth-frb: ok

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -461,22 +465,40 @@
...
         fabs.place_of_performance_scope AS place_of_perform_scope,
         COALESCE(UPPER(bs.sub_place_of_perform_country_co), 'USA') AS sub_place_of_perform_country_co,
         pcc.country_name AS sub_place_of_perform_country_name,
+        -- For county codes, we need to make sure we extract the right digits out and ensure it's zero-padded to 3.
         LPAD(CAST(CAST(REGEXP_EXTRACT(pop.county_numeric, '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 3, '0') AS sub_place_of_perform_county_code,
         pop.county_name AS sub_place_of_perform_county_name,
         LEFT(COALESCE(sub_place_of_performance_zip, ''), 5) AS sub_place_of_perform_zip5,
         pop.census_code AS sub_place_of_perform_city_code,
-        LPAD(CAST(CAST(REGEXP_EXTRACT(UPPER(bs.sub_place_of_perform_congressio), '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 2, '0') AS sub_place_of_perform_congressio
+        -- For congressional districts, we need to make sure we extract the right digits out and ensure it's zero-padded to 2.
```
dporth-frb, line 494: I think this is knowledge duplicatoin here. Looking at the code we can tell that this is happening

> we need to make sure we extract the right digits out and ensure it's zero-padded to 2.

but why?

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -461,22 +465,40 @@
...
+                UPPER(bs.sub_legal_entity_country_code) <> 'USA'
+            ) THEN NULL
+            ELSE COALESCE(rl_cd_state_grouped.congressional_district_no, rl_zips.congressional_district_no, rl_cd_zips_grouped.congressional_district_no, rl_cd_city_grouped.congressional_district_no, rl_cd_county_grouped.congressional_district_no)
+        END) AS sub_legal_entity_congressional_current,
 
+        LATEST_CURRENT_CD.pop_congressional_code_current AS place_of_performance_congressional_current,
         fabs.place_of_performance_scope AS place_of_perform_scope,
         COALESCE(UPPER(bs.sub_place_of_perform_country_co), 'USA') AS sub_place_of_perform_country_co,
         pcc.country_name AS sub_place_of_perform_country_name,
+        -- For county codes, we need to make sure we extract the right digits out and ensure it's zero-padded to 3.
```
dporth-frb, line 489: I'm hesitant about repeating the `3` here. If the code changes we are also going to have to remember to change this comment too. That's two places an update needs to happen.

>  we need to make sure we extract the right digits out and ensure it's zero-padded to 3.

It's clear from looking at the code that this is happening so this may be knowledge duplication. Perhaps it'd be better if this comment answered why? If someone wants to know what, all they have to do is look at the code.

dpb-bah: > "Why are we using this regex and left padding for county codes?" If someone wants to know what, all they have to do is look at the code.

This ticket has nothing to do with that question. It's how we've always extracted congressional district code from the field (get the two digits, zero pad it). I don't know what more to explain. Do we just want to remove these regex comments?

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -539,6 +564,58 @@
...
+            AND pop_cd_city_grouped.state_abbreviation=UPPER(bs.sub_place_of_perform_state_code)
+        )
+    LEFT OUTER JOIN
+        global_temp.cd_city_grouped rl_cd_city_grouped ON (
+            rl_cd_city_grouped.city_name=UPPER(bs.sub_legal_entity_city_name)
+            AND rl_cd_city_grouped.state_abbreviation=UPPER(bs.sub_legal_entity_state_code)
+        )
+    LEFT OUTER JOIN
+        global_temp.cd_county_grouped pop_cd_county_grouped ON (
+            pop_cd_county_grouped.county_number=LPAD(CAST(CAST(REGEXP_EXTRACT(pop.county_numeric, '^[A-Z]*(\\d+)(?:\\.\\d+)?$', 1) AS SHORT) AS STRING), 3, '0')
```
dporth-frb, line 610: My rule of thumb is to keep comments to explaining why. I think there was less knowledge duplication before you added the comment.

> This ticket has nothing to do with that question
Can you explain why this story has nothing to do with my question when you've added new occurrences here?

> It's how we've always extracted congressional district code from the field (get the two digits, zero pad it).

So you're saying that because it's how we've always done it that any new place we do it more doesn't mean new code should follow best practices.

My original comment was suggesting you to add an explanation of why you are doing these transforms, that's it.

dpb-bah: So this was how we extracted congressional district from the field long ago and I'm simply keeping that logic when matching on fields. I could add a comment along the lines of "congressional district fields may include characters digits, so we're extracting the digits and making sure it's still zero padded" but that would be a guess. If we wanted to actually get to the root of that question, we'd need to analyze each raw field in production data to see its values looking for any edge cases and then determine if the transformation logic is necessary/needs to be updated everywhere, which is not in the scope.

collinwr (approved):

PR #3850, opened 2023-06-20
[DEV-9929] Use set comprehension instead of list comprehension

aguest-kc:
**Description:**
Remove duplicate Subaward IDs in the case where the same Subaward is associated with multiple disaster codes that are included in the API request. A Subaward ID should only be returned once, even if it's associated with multiple DEF codes in the request.

**Technical details:**
Use a set instead of a list to remove duplicate Subaward IDs from the subquery.

    - [ ] Backend
2. [x] Data validation completed
3. [x] Jira Ticket [DEV-9929](https://federal-spending-transparency.atlassian.net/browse/DEV-9929):

### DIFF
--- a/usaspending_api/search/filters/postgres/defc.py
+++ b/usaspending_api/search/filters/postgres/defc.py
@@ -66,7 +66,7 @@ class DefCodes:
 
             results = cursor.fetchall()
 
-        results = [result[0] for result in results]
+        results = {result[0] for result in results}
         q = Q(broker_subaward_id__in=results)
 
         return q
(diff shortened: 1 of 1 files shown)


usaspending_api/search/filters/postgres/defc.py
```diff
@@ -66,7 +66,7 @@ def build_def_codes_filter(cls, filter_values):
 
             results = cursor.fetchall()
 
-        results = [result[0] for result in results]
+        results = {result[0] for result in results}
```
collinwr, line 69: Would really like to see a comment on this explaining why we're using a set (to prevent duplicates)

aguest-kc: Added a comment explaining what the raw SQL is doing and why there can be duplicates which need to be removed.

collinwr (approved):

PR #3863, opened 2023-07-07
[DEV-10017] Add support for CD code 90 [hotfix][staging]

aguest-kc:
**Description:**
Add support for congressional district values of `90`. Add a function for combining multiple dictionaries in a list that have the same `code` and `name` value and sum their `amount` field values together into one dictionary.

**Technical details:**
Since the congressional district value of `90` is not in the database for any U.S. state, the transactions associated with this congressional code had the congressional district value (`90`) replaced with `None`. There is now a check to see if the amount of congressional districts for the given state is greater than 1. If this is true, then `90` is added to the list of valid congressional districts for that state and will no longer be replaced with `None`.

Added a function that checks all of the dictionaries in the provided list to see if certain keys have the same values. If they do, then these dictionaries are combined into one dictionary and their `amount` keys are summed together.

    - [ ] Backend
5. [x] Data validation completed
6. [x] Jira Ticket [DEV-10017](https://federal-spending-transparency.atlassian.net/browse/DEV-10017):

### DIFF
--- a/usaspending_api/search/v2/views/spending_by_category_views/spending_by_locations.py
+++ b/usaspending_api/search/v2/views/spending_by_category_views/spending_by_locations.py
@@ -1,21 +1,23 @@
 from abc import ABCMeta
+from collections import Counter
 from decimal import Decimal
-from django.db.models import QuerySet, F, TextField
-from django.db.models.functions import Concat
 from enum import Enum
-from typing import List
+from typing import Dict, List
 
+from django.db.models import F, QuerySet, TextField
+from django.db.models.functions import Concat
+
+from usaspending_api.recipient.models import StateData
 from usaspending_api.references.abbreviations import code_to_state, fips_to_code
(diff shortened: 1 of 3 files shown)


usaspending_api/search/v2/views/spending_by_category_views/spending_by_locations.py
```diff
@@ -103,6 +158,12 @@ def _key_to_geo_code(key):
                     "amount": int(bucket.get("sum_field", {"value": 0})["value"]) / Decimal("100"),
                 }
             )
+
+        # Combine all dicts in the `results` list that have the same values for
+        #   `code` and `name`. Sum their `amount` values together.
```
collinwr, line 163: Can you elaborate a little more on the "why" in this comment? Something about grouping results that don't match an entry in the reference table (null `code` and `name`) into a single result.

collinwr (approved):

PR #3887, opened 2023-08-09
[PIPE-489] Refactor COVID-19 Download to Spark SQL

kwhickey:
**Description:**
Converting the postgres SQL covid-19 download files to compatible Spark SQL, and creating spark functions and django job to orchestrate packaging results up as zipped csv(s).

**Technical details:**
TODO
The technical details for the knowledge of other developers. Any detailed caveats or specific deployment steps should be outlined here.

**Requirements for PR merge:**

1. [ ] Unit & integration tests updated
2. [ ] API documentation updated
3. [ ] Necessary PR reviewers:
    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
    - [ ] Domain Expert <OPTIONAL>
4. [ ] Matview impact assessment completed
5. [ ] Frontend impact assessment completed
6. [ ] Data validation completed
7. [ ] Appropriate Operations ticket(s) created
8. [ ] Jira Ticket [DEV-123](https://federal-spending-transparency.atlassian.net/browse/DEV-123):
    - [ ] Link to this Pull-Request
    - [ ] Performance evaluation of affected (API | Script | Download)
    - [ ] Before / After data comparison

**Area for explaining above N/A when needed:**
```
```

PR #3895, opened 2023-08-19
[PIPE-494 & PIPE-495] Covid Download on Databricks Implementation

dporth-frb:
**Description:**
This PR takes the dev work from https://github.com/fedspendingtransparency/usaspending-api/pull/3887 and completes it. Additionally, changes were made to enable integration of that PR into our existing Covid-19 Download generation job, changes were made to support multiple sql queries (https://federal-spending-transparency.atlassian.net/browse/PIPE-495), and some more work was done to productionize that other PR (included making the work re-use existing code and increasing it's orthogonality and modularity). 

This PR contains work for two stories, PIPE-494 and PIPE-495, it was easier working on both of these stories in a single branch due to how similar and coupled they are.

    - [x] Backend
8. [x] Jira Ticket [PIPE-495](https://federal-spending-transparency.atlassian.net/browse/PIPE-495):
9. [x] Jira Ticket [PIPE-494](https://federal-spending-transparency.atlassian.net/browse/PIPE-494):

5. The nature of this work does not require frontend's review
7. No operation tickets needed


usaspending_api/disaster/management/commands/generate_covid19_download.py
```diff
@@ -41,7 +41,42 @@ class Command(BaseCommand):
     readme_path = Path(settings.COVID19_DOWNLOAD_README_FILE_PATH)
     full_timestamp = datetime.strftime(datetime.now(timezone.utc), "%Y-%m-%d_H%HM%MS%S%f")
 
+    # KEY is the type of compute supported by this command
+    # key's VALUE are the strategies required by the compute type
+    # These strategies are used to change the behavior of this command
+    #   at runtime.
+    compute_flavors = {
+        ComputeFlavorEnum.AURORA.value: {
```
collinwr, line 49: Usually when we refer to a database version of something in code, we call it Postgres not Aurora. If we were to run this locally for instance, we wouldn't be running it in Aurora, we'd be running it on a local Postgres instance.

dporth-frb: Ok so you're saying you'd like me to change this value to postgres?

dporth-frb: Done

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,213 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
             spark.sql(sql_statement)
 
     logger.info(f"Created the reference views in the global_temp database")
+
+
+def load_csv_file(
```
collinwr, line 569: This function. seems a little redundant. All it seems to do is instantiate a logger, and otherwise pass the input parameters into another function. Why not instantiate the logger outside this function and use `write_csv_file()` directly?

collinwr: Actually, looks like `write_csv_file()` instantiates `logger` if it doesn't exist anyway

dporth-frb: This came from Keith's changes. I tried to keep his work as he put it.

dporth-frb: I'll remove it though. I agree with you.

dporth-frb: done

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,176 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+    spark: SparkSession,
+    parts_dir: str,
+    header: str,
+    max_rows_per_merged_file=EXCEL_ROW_LIMIT,
+    logger=None,
+    file_format=".csv",
+) -> None:
+    """PySpark impl of Hadoop 2.x copyMerge() (deprecated in Hadoop 3.x)
+    It uses the underlying Hadoop FileSystem API to combine all part files under a directory
+    into a single specified destination file.
```
collinwr, line 628: Is this combining the part files into a single file or multiple files based on the `max_rows_per_merged_file` parameter? 

Also, it's not clear based on this initial comment and arg descriptions where this "specified destination file" is. If it really is multiple files, what will they be named, and where can I expect to find them?

dporth-frb: This is from Keith's implementation.

dporth-frb: > It uses the underlying Hadoop FileSystem API to combine all part files under a directory
    into a single specified destination file
    
    If I update "destination file" here to the parts parameter does it make more sense?

dporth-frb: I've gone ahead and updated the documentation. Let me know if it's clear enough now.

collinwr: Mostly just pointing out the inaccuracy of that first sentence. It's sometimes true, but in many cases it will not be. Is there a better description that would fit all cases?

Maybe something like:

> Merges files from a provided input directory and then redivides them into multiple files based on a provided maximum row count parameter.

Also, when a reader is trying to figure out where these files are going to end up (where and what they will be named), they might not think to check the `max_rows_per_merged_file` argument description, which right now only covers what they'll be named, not where they will end up. Because a list of file paths are not being returned by this method, this information should be obvious to the reader. Can you make this more clear somewhere in the Pydoc comment? Maybe a notes section at the end?

dporth-frb: > they might not think to check the max_rows_per_merged_file argument description,

They should, its the contract of the function. When someone is trying to understand how to work with a function they need to read all of it's documentation.

On the note of not returning file paths, I propose we update the method to return a list of merged files. I believe that will be necessary anyway when this job starts running queries with large data. I believe once this function returns a path to the files everything will be clear. Do you agree?

dporth-frb: Changes made based on teams meeting.

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,177 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+                fs.delete(partial_merged_file_path, True)
+            out_stream = fs.create(partial_merged_file_path)
+            out_stream.writeBytes(header + "\n")
+            _merge_file_parts(fs, out_stream, conf, hadoop, partial_merged_file_path, parts_file_group.file_list)
+        finally:
+            if out_stream is not None:
+                out_stream.close()
+        try:
+            fs.rename(partial_merged_file_path, merged_file_path)
+        except Exception:
```
collinwr, line 710: We should probably log some sort of error if this Exception is caught

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,177 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+            if fs.exists(partial_merged_file_path):
+                fs.delete(partial_merged_file_path, True)
+            raise
+    return merged_file_names_created
```
collinwr, line 714: Looks like this variable actually contains a list of file paths. Your usage of the `hadoop_copy_merge()` also indicates this:

```
merged_file_paths = hadoop_copy_merge(
    spark=self.spark,
    parts_dir=s3_destination_path,
    header=header,
    max_rows_per_merged_file=EXCEL_ROW_LIMIT,
    logger=self._logger,
)
```

Can you update the variable and Pydoc return description to reflect this?

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,177 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+            Individual part files must NOT therefore be created with their own header.
+        max_rows_per_merged_file: Final CSV data will be subdivided into numbered files so that there is not more than
+            this many rows in any file written. Only if the total data exceeds this value will multiple files be
+                created with a pattern of ``{merged_file}_N.{extension}`` with N starting at 1.
+        logger: The logger to use. If one note provided (e.g. to log to console or stdout) the underlying JVM-based
+            Logger will be extracted from the ``spark`` ``SparkSession`` and used as the logger.
+        file_format: The format of the part files and the format of the final merged file
+
+    Returns:
+        a list of file names where each element is a merged file that was generated during the copy merge.
```
collinwr, line 643: For consistency, I'd like to see the leading `a` capitalized

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,177 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+    return df_record_count
+
+
+def hadoop_copy_merge(
+    spark: SparkSession,
+    parts_dir: str,
+    header: str,
+    max_rows_per_merged_file=EXCEL_ROW_LIMIT,
+    logger=None,
+    file_format=".csv",
```
collinwr, line 624: Might be a little tricky requiring the `.` as part of the `file_format`. See the Spark write's format as an example: https://sparkbyexamples.com/spark/spark-write-options/

Can you update the default value along with the instantiation of `merged_file_name`?

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
```diff
@@ -0,0 +1,197 @@
...
+                # See comment below about old date and time values cannot be parsed without these
+                "spark.sql.legacy.parquet.datetimeRebaseModeInWrite": "LEGACY",  # for dates at/before 1900
+                "spark.sql.legacy.parquet.int96RebaseModeInWrite": "LEGACY",  # for timestamps at/before 1900
+                "spark.sql.jsonGenerator.ignoreNullFields": "false",  # keep nulls in our json
+            }
+            self.spark = get_active_spark_session()
+            self.spark_created_by_command = False
+            if not self.spark:
+                self.spark_created_by_command = True
+                self.spark = configure_spark_session(**extra_conf, spark_context=self.spark)  # type: SparkSession
```
collinwr, line 143: I wonder if there is a way that we can reuse the spark session across files. Not a dealbreaker for this PR. Just something to consider in the future. Might want to see how this would affect performance.

usaspending_api/common/etl/spark.py
dporth-frb: I've addressed, but we need to include this in our style guide if this is something we expect people to follow.

dporth-frb: Here is an example of code that doesn't follow this style, it's existed for 3 years.
```
def _lookup_deleted_award_keys(
    client: Elasticsearch,
    lookup_key: str,
    value_list: list,
    config: dict,
    index: Optional[str] = None,
    lookup_chunk_size: int = 50000,
) -> list:
    """Derive a list of award keys given a target index, Lookup field, and lookup values

    This returns a list of all unique award keys, which are compiled from the ``ES_AWARDS_UNIQUE_KEY_FIELD`` field of
    any document in the given ``index`` that matches the query. The matching query is a terms query that will return
    the doc if its ``lookup_key`` field has any value provided in ``value_list``.

    Args:
        client (Elasticsearch): elasticsearch-dsl client for making calls to an ES cluster
        lookup_key (str): name of field in targeted elasticsearch index by which we are looking up docs. The field or
            sub-field provided MUST be of ``keyword`` type (or ``_id`` meta field)
        value_list (list): if lookup_key field has any of these values, the document will be returned from the lookup
        config (dict): collection of key-value pairs that encapsulates runtime arguments for this ES management task
        index (str): Optional name, alias, or pattern of index this query will target. Looks up via config if not
            provided
        lookup_chunk_size (int): the batch-size of terms value-array to be looked-up. Needs to be less
            than 65536 (max values for any terms query), and less than config["max_query_size"]

    Returns: list of values for the ES_AWARDS_UNIQUE_KEY_FIELD fields in the looked-up documents.
    """
```

dporth-frb: Here's another example
```
def extract_location(recipient_hash):
    """Extract the location data via the recipient hash

    Args:
        recipient_hash: uuid of the hash+duns to look up

    Returns:
        dict of location info
    """
```

dporth-frb:
> Left feedback requesting a few small changes.
> 
> Overall, I'm surprised by how some of this original code is structured. For example, seems a little confusing that `download_to_csv` is both outputing sql to a CSV than also adding that CSV to a zip file even though it's returning the CSV file names that are never used. I would have expected a later function to use the list of file names to add them to zip file, or else have the name changed to something like `download_to_csv_and_zip()`
> 
> I know that's something that predated your code, and we don't need to make major changes beyond the scope of your work. Just making an observation about the code. I think once the specific changes I've requested are addressed, we can get this merged in unless @kwhickey has any additional feedback

All requested changes have been made. Perhaps we should create a story in the backlog to refactor the poor structure of the original code.

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
```diff
@@ -0,0 +1,197 @@
...
+)
+from usaspending_api.download.filestreaming.zip_file import append_files_to_zip_file
+from usaspending_api.download.lookups import FILE_FORMATS
+from pyspark.sql import SparkSession
+from usaspending_api.common.etl.spark import hadoop_copy_merge, write_csv_file
+from usaspending_api.common.helpers.spark_helpers import configure_spark_session, get_active_spark_session
+from typing import List
+
+
+class AbstractToCSVStrategy(ABC):
```
kwhickey, line 27: Since this just contains 1 function ... it's like a Single Abstract Method (SAM) interface, aka "functional interface" ... and may be able to be simplified in python by just defining the alternate functions and passing them in at the point of execution rather than having to instantiate the concrete strategy classes and calling their function. Food for thought.

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
```diff
@@ -0,0 +1,197 @@
...
+            zip_process.start()
+            wait_for_process(zip_process, start_time, None)
+        except Exception as e:
+            raise e
+        finally:
+            Path(temp_file_path).unlink()
+        return [destination_path], count
+
+
+class DatabricksToCSVStrategy(AbstractToCSVStrategy):
```
kwhickey, line 116: Unless we're coupled specifically to a Databricks feature, prefer to name this `SparkToCSVStrategy`, or if this is really referencing the source data being in a Delta table, look to name it `DeltaToCSVStrategy`. (e.g. we should be able to execute this in our local envs with Spark and Delta, and no Databricks).

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
```diff
@@ -0,0 +1,197 @@
...
+        except Exception:
+            self._logger.exception("Exception encountered. See logs")
+            raise
+        finally:
+            delete_s3_object(s3_bucket_name, s3_destination_path)
+            if self.spark_created_by_command:
+                self.spark.stop()
+        append_files_to_zip_file(final_csv_data_file_locations, covid_profile_download_zip_path)
+        self._logger.info(f"Generated the following data csv files {final_csv_data_file_locations}")
+        return final_csv_data_file_locations, record_count
```
kwhickey, line 169: Maybe there was no other way around this, and it probably works, but it adds a great deal of inefficiency to what will be very large files getting:

1. built in s3
2. downloaded to EC2 filesystem
3. zipped in EC2
4. uploaded to s3

If it's a tradeoff of: 

1. elegantly fitting this new approach of sourcing data from S3 Delta tables (vs. from Postgres) into the existing django command so we can still keep with 1 command and the way that command handled zipping up the fin
2. versus: creating a 2nd (Spark+Delta specific) django command whose execution flow is slightly different than the 1st django command, but allows for more efficient flow to zipping up the final download data

Then my opinion is we should be opting for #2: efficiency over elegance.

To figure out if and how much efficiency is lost, I think we need to see how much time is spent for steps 2 through 4 for the full sized covid files and see if it can't be optimized by performing the Zip _on_ the S3 "FileSystem" using the hadoop copy-merge style code, like what was in the original code:

```
    try:
        # Get the new zipped bits to the target file.
        # If this file is supposed to already exist, there may be a short period of time where the file does NOT
        # exist, between when the delete completes and the rename completes.
        # S3 does not allow renaming to an existing file, and renames are done as copies of the full object
        if overwrite and fs.exists(zip_file_path):
            fs.delete(zip_file_path, True)
        fs.rename(partial_zipped_file_path, zip_file_path)
    except Exception:
        if fs.exists(partial_zipped_file_path):
            fs.delete(partial_zipped_file_path, True)  # drop partial zip file if rename completes or not
        raise

    logger.info(
        f"Completed zip of {len(merged_file_paths)} {extension} files into file "
        f"{str(zip_file_path)} in {(time.time() - zip_start):3f}s"
    )
```

e.g. given the size of the files, I think it'd be ideal if we could build and stage them all in S3, and perform the Zip _right in S3_ as the above code does, avoiding the download and upload.

usaspending_api/disaster/management/commands/generate_covid19_download.py
```diff
@@ -41,7 +43,42 @@ class Command(BaseCommand):
...
+                "disaster_covid19_file_a": "usaspending_api/disaster/management/sql/disaster_covid19_file_a.sql",
+                "disaster_covid19_file_b": "usaspending_api/disaster/management/sql/disaster_covid19_file_b.sql",
+                "disaster_covid19_file_d1_awards": "usaspending_api/disaster/management/sql/disaster_covid19_file_d1_awards.sql",
+                "disaster_covid19_file_d2_awards": "usaspending_api/disaster/management/sql/disaster_covid19_file_d2_awards.sql",
+                "disaster_covid19_file_f_contracts": "usaspending_api/disaster/management/sql/disaster_covid19_file_f_contracts.sql",
+                "disaster_covid19_file_f_grants": "usaspending_api/disaster/management/sql/disaster_covid19_file_f_grants.sql",
+            },
+            "download_to_csv_strategy": PostgresToCSVStrategy(logger=logger),
+        },
+        ComputeTypeEnum.DATABRICKS.value: {
```
kwhickey, line 62: See previous comment about avoiding use of "Databricks" in names.

usaspending_api/etl/tests/integration/test_spark_write_csv_file.py
```diff
@@ -0,0 +1,116 @@
...
+    file_prefix = "csv"
+    results_path = f"{obj_prefix}.{file_prefix}"
+    download_path = tmp_path / f"{file_timestamp}.{file_prefix}"
+    test_logger.info(f"Downloading test results from {results_path} to {download_path}")
+    _download_s3_object(
+        s3_bucket_name=bucket_name,
+        s3_obj_key=results_path,
+        download_path=download_path,
+        configured_logger=logger,
+    )
```
kwhickey, line 113: Please remove this function call and replace with `s3_helpers.download_s3_object`. Then you can get rid of the three private functions in this test.

kwhickey (changes requested):
Couple things lost in translation from the old `hadoop_copy_merge` code in how the files are bundled up as a "merge group" and combined, and dealing with empty files. Also I may need a walkthrough of how the file naming schema changed from the original code. 

Then also the approach to zipping parts is going to yield a lot of inefficiency when dealing with a COVID File D2 dataset, which usually ends up with 20+ individual CSV files that are 1M rows each. Or maybe more ... lost track of just how many assistance records get bundled in this download.

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,179 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+            hadoop.io.IOUtils.copyBytes(in_stream, out_stream, conf, False)
+        finally:
+            if in_stream:
+                in_stream.close()
+            if fs.exists(partial_merged_file_path):
+                fs.delete(partial_merged_file_path, True)
+
+
+# Helper to chunk up files into mergeable groups
+def _merge_grouper(items, group_size):
```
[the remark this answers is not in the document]
dporth-frb: This method is how you wrote it. I can add it though

usaspending_api/etl/tests/integration/test_spark_write_csv_file.py
dporth-frb: Those methods don't work for this test as you wrote it. This is copied from you, I'm assuming you created these extra functions for a reason in your PR. Instead of adding these to S3 because it's not within the scope of this story to create these new s3 helper functions I made them private.

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
dporth-frb: No it's a strategy. I'm using object oriented programming. Is it better to not do so?

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,179 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+            if fs.exists(partial_merged_file_path):
+                fs.delete(partial_merged_file_path, True)
+            out_stream = fs.create(partial_merged_file_path)
+            out_stream.writeBytes(header + "\n")
+            _merge_file_parts(fs, out_stream, conf, hadoop, partial_merged_file_path, parts_file_group.file_list)
+        finally:
+            if out_stream is not None:
+                out_stream.close()
+        try:
+            fs.rename(partial_merged_file_path, merged_file_path)
```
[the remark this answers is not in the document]
dporth-frb: This is from your code

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
```diff
@@ -0,0 +1,197 @@
+from abc import ABC, abstractmethod
```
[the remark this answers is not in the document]
dporth-frb: > In general we've been trying to get away from the use of helpers as a descriptor for a namespace/package or module

How would I have known that? Food for thought on improving team documentation practices.

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
collinwr: Good idea here 👍

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,179 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+            if file_path.getName().startswith("_"):
+                logger.debug(f"Skipping non-part file: {file_path.getName()}")
+                continue
+            logger.debug(f"Including part file: {file_path.getName()}")
+            part_files.append(f.getPath())
+    if not part_files:
+        logger.warn("Source directory is empty with no part files. Attempting creation of file with CSV header only")
+        out_stream = None
+        try:
+            out_stream = fs.create(hadoop.fs.Path(parts_dir), overwrite)
```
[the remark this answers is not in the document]
dporth-frb: > Not sure if this makes sense to write the header-only .csv file to the location of the `parts_dir`. The `parts_dir` would be a prefix in S3 (like a folder in the FileSystem), and in this case an empty one (or at least empty of any part files, it could have a `_SUCCESS` file).
> 
> Below, when there are part files to merge (combine) the write via `fs.create` is to `partial_merged_file_path`, which is then renamed to `merged_file_path`. So probably `merged_file_path` should be used here instead of `parts_dir`.
> 
> Which means you need to declare that `merged_file_name` and `merged_file_path` vars earlier so they can be used in this case of empty parts dir.

This behavior is how you implemented the method. I assumed you had a reason and did it as it should be.

usaspending_api/common/etl/spark.py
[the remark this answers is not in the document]
dporth-frb: parts_dir is the path to the parts directory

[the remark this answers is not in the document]
dporth-frb: > Definitely would be good to do some operational testing for LARGE (many 100s of part files) data such that it would yield many *_1.csv ... *_2.csv ... combined files

This is not within the scope of this story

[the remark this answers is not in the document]
kwhickey: This should be fine if you protect against `parts_dir` coming in with a trailing `/` e.g. by using `.rstrip("/")`. You can resolve after fixed.

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
kwhickey: Discussed in PR review and likely to push investigation of efficiency to next Story in this series, as a possible improvement.

usaspending_api/etl/tests/integration/test_spark_write_csv_file.py
dporth-frb: Ok done.

usaspending_api/common/etl/spark.py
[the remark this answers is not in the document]
dporth-frb: Fixed

usaspending_api/common/etl/spark.py
```diff
@@ -560,3 +564,179 @@ def create_ref_temp_views(spark: SparkSession, create_broker_views: bool = False
...
+            out_stream.writeBytes(header + "\n")
+        finally:
+            if out_stream is not None:
+                out_stream.close()
+        return
+
+    part_files.sort(key=lambda f: str(f))  # put parts in order by part number for merging
+    paths_to_merged_files = []
+    for parts_file_group in _merge_grouper(part_files, max_rows_per_merged_file):
```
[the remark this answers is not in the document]
dporth-frb: Addressed.

usaspending_api/common/etl/spark.py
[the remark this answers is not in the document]
dporth-frb: addressed.

usaspending_api/disaster/helpers/covid_download_csv_strategies.py
dporth-frb: Addressed.

usaspending_api/disaster/management/commands/generate_covid19_download.py
dporth-frb: Addressed.

collinwr (dismissed):

collinwr (approved):

PR #3909, opened 2023-08-31
[PIPE-496][DEV-10173] Copy Data From Delta to Postgres Fix; Additional Data Dictionary Logging

collinwr:
# Description
This PR addresses two tickets:

## PIPE-496
This creates a new function `_download_and_copy()` to replace the `_stream_and_copy()` function in the `load_table_from_delta.py` Databricks job. The previous function streamed a S3 file into memory via the `s3_client`, whereas the new version downloads the file to the filesystem. 

This should hopefully address the data length mismatch error we've frequently been seeing when trying to copy the `transaction_search` table from Delta Lake to Postgres

## DEV-10173
I went ahead and included this ticket into the PR, so we can get a better idea of why my logging logic doesn't seem to be helping with the timeout issue. 

**Technical details:**

    - [ ] Backend
7. N/A -Appropriate Operations ticket(s) created
8. [x] Jira Ticket [PIPE-496](https://federal-spending-transparency.atlassian.net/browse/PIPE-496]), [DEV-10173](https://federal-spending-transparency.atlassian.net/browse/DEV-10173]):

```
```


usaspending_api/common/csv_stream_s3_to_pg.py
```diff
@@ -94,6 +96,50 @@ def _stream_and_copy(
         raise exc
 
 
+def _download_and_copy(
```
dporth-frb, line 99: There is a lot of code duplication between this function and _stream_and_copy. Specifically, all parameters, lines 110-111, lines 119 to 122, lines 125 to 128, and lines 130 to 142 

I get the impression you copy an pasted the stream method and only changed the resources that are being managed.

For example, it appears the key differences between these methods are the resources that are being managed while the work on the resource is the same. In other words, the copy_expert arguments are duplicated. Can you implement something that maybe reduces the code duplication and works on any resources?

dporth-frb (changes requested):
Requesting responses not feedback, yet

usaspending_api/common/csv_stream_s3_to_pg.py
dporth-frb, line 99: Where are unit and integration tests for this new function?

dporth-frb (changes requested):

usaspending_api/download/filestreaming/download_generation.py
```diff
@@ -747,12 +747,24 @@ def add_data_dictionary_to_zip(working_dir, zip_file_path):
...
+            logger.info(
+                f"Attempt {attempt + 1} of {retry_count + 2} failed to download Data Dictionary. Error: {e}"
+            )
             if attempt < retry_count:
                 time.sleep(settings.DATA_DICTIONARY_DOWNLOAD_RETRY_COOLDOWN)
                 continue
             else:
+                logger.error(f"Failed to download Data Dictionary after {retry_count + 1} attempts.")
                 raise
+        except Exception as e:
```
dporth-frb, line 760: What's the reason you do not want to retry in case of exceptions other than the types `HTTPError, RemoteDisconnected`?

usaspending_api/common/helpers/s3_helpers.py
```diff
@@ -87,15 +89,47 @@ def multipart_upload(bucketname, regionname, source_path, keyname):
     transfer.upload_file(source_path, bucketname, Path(keyname).name, extra_args={"ACL": "bucket-owner-full-control"})
 
 
-def download_s3_object(bucket_name: str, key: str, file_path: str, region_name: str = settings.USASPENDING_AWS_REGION):
```
dporth-frb, line 90: Why add the knowledge of retry logic here. I think it'd be better to add retry logic to the code that calls this method. For example, my covid-19 download work doesn't need this retry logic

dporth-frb (changes requested):

usaspending_api/common/helpers/s3_helpers.py
dporth-frb: Essentially, you've taken this very simple function and made it more complex solely for the sake of table from Delta Lake to Postgres copy.

dporth-frb (changes requested):

usaspending_api/common/csv_stream_s3_to_pg.py
collinwr: You're right. Quite a bit of code duplicate here. I think I'd rather just remove `_stream_and_copy()` all together because it's no longer being used.

collinwr: Still need to get this in

usaspending_api/download/filestreaming/download_generation.py
collinwr: It's a good practice to plan for specific types of Exceptions and how to handle them. In this case, we're expecting the `RetrieveFileFromUri()` function to throw some sort of network error. That's really the only reason why we'd want to try another request.

I added the extra catch (`except`) because some sort of error seems to be slipping through. In this case of a unexpected error, we should log and raise it, so it can be addressed. 

Here are a few examples of exceptions we wouldn't want to retry:
1. `ValueError`: This might be raised if the provided URI is malformed or if some other parameter value is out of an expected range. Retrying won't fix a malformed URI.
3. `PermissionError`: This could be raised if the function tries to save a file to a local directory without appropriate permissions.
4. `MemoryError`: Raised when an operation runs out of memory. If the file is too large to fit into memory, for instance, retrying is unlikely to help unless there are external factors affecting available memory.

usaspending_api/common/helpers/s3_helpers.py
collinwr: I thought this function could benefit from retry logic. I think it's a pretty common case that if we're trying to download a file from S3, we'd want the ability to retry, even in your Covid-19 download code.

I knew that code used this function, and considered setting the `retry_count` to `0`, but ultimately decided to let it use the default of `3` because we might as well let it retry.

usaspending_api/common/csv_stream_s3_to_pg.py
dporth-frb: I don't think deleting the function is the best idea. We already put in the work to create it. Can we anticipate we'll never, ever need this functionality in the feature? I think it'd be best to put in the good work now vs. later.

usaspending_api/download/filestreaming/download_generation.py
dporth-frb: Ok makes sense.

usaspending_api/common/helpers/s3_helpers.py
dporth-frb: Ok, I'm for it.

usaspending_api/common/csv_stream_s3_to_pg.py
dporth-frb: That said, please feel free to override my request. Just want to make sure my suggestion is known.

collinwr: Generally, we don't leave code in the codebase that's not being used. Because it is more code to maintain. If someone needs to reference that code or add it back in the future, the history is available in Github.

dporth-frb: Ok

dporth-frb (approved):

PR #3912, opened 2023-09-07
[DEV-10150] Fixes File C to D Linkage Deletion Bug

dporth-frb:
**Description:**
When a submission is deleted, after the update_file_c_linkages_in_delta.py command runs, Awards associated with File C records from the deleted submission, have their update_date fields set to the time the command ran.

    - [x] Backend
8. [x] Jira Ticket [DEV-10150](https://federal-spending-transparency.atlassian.net/browse/DEV-10150):

1. Existing unit and integration tests already exist and are still valid for the code that was changed.
5. The nature of this work does not require frontend's review
7. No operation tickets needed


usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
```diff
@@ -85,6 +85,36 @@ def handle(self, *args, **options):
...
+            self.spark.sql(identify_faba_deletions_query)
+            # Update Awards whose file C submissions are now deleted
+            delete_sql = """
+                MERGE INTO int.awards AS aw
+                USING identify_faba_deletions_query AS deletes
+                ON aw.id = deletes.award_id
+                WHEN MATCHED
+                    THEN UPDATE SET aw.update_date = NOW();
+            """
+            self.spark.sql(delete_sql)
```
collinwr, line 116: Where you able to come up with a way to avoid updating Awards twice?

dporth-frb: I'm not confident that having two merge statements will negatively impact performance. Overall, the same number of records are being merged whether the two merge statements are combined or not.

dporth-frb: I've one ahead and combined the merge statements like you requested.

collinwr: > I'm not confident that having two merge statements will negatively impact performance. Overall, the same number of records are being merged whether the two merge statements are combined or not.

The same number of records may be merged, but the number of parquet files being rewritten could change with with multiple merges. 

For example, say we have 100 awards evenly distributed in order across 10 parquet files (Awards 1-10 in File 1, etc). Say the first merge updates Awards 6 and 35. This will cause files 1 and 3 to be rewritten as files 11 and 12 (with 1 and 3 being invalidated). In the second merge, Awards 4 and 38 are updated. This would cause files 11 and 12 to be rewritten, resulting in a total of 4 file rewrites, where performing a single merge with all 4 updates would have resulted in 2 rewrites.

collinwr: As far as your implementation goes, one thing I'd want to double check is `identify_faba_deletions_query` actually contains records by the time `all_updated_awards` is created. One concern that I have is because `identify_faba_deletions_query` is a view on top of the `int` and `raw` schemas, when the `int` table is rewritten to reflect the current state of the `raw` table, the underlying query will no longer return any results. 

May have to materialize the results somehow by pulling out the IDs into memory before recreating the `int` table.

I'd like to see a test case covering this scenario. I have a pretty robust testing framework setup for this command. Should hopefully be easy to update it to include this case. Let me know if you'd like me to walk you through it

usaspending_api/etl/tests/integration/test_update_file_c_linkages_in_delta.py
```diff
@@ -17,7 +17,7 @@
 common_faba_info = {"submission_id": "0", "distinct_award_key": "fake_key"}
 
 award_records = [
-    {"id": 1, "expected_update": False},
+    {"id": 1, "expected_update": True},
```
collinwr, line 20: This Award was already being used by a few other test cases, where it is important to ensure that it's not being updated. Can you define a 12th award to use for your test case.

collinwr: Same for FABA ID 5. Can you create a new FABA record for your test case?

dporth-frb: This is one of the consequences of having a test, test multiple things.

dporth-frb: I'll add an id 12

usaspending_api/etl/management/commands/update_file_c_linkages_in_delta.py
dporth-frb: Done

usaspending_api/etl/tests/integration/test_update_file_c_linkages_in_delta.py
```diff
@@ -149,6 +150,13 @@
     {"financial_accounts_by_awards_id": 22, "fain": "fain_200", "uri": "uri_1", "award_id": 1, "expected_award_id": 1},
     # Matches two Awards based on URI when both FAIN/URI are populated - No Update
     {"financial_accounts_by_awards_id": 23, "fain": "fain_200", "uri": "uri_2", "expected_award_id": None},
+    {
```
collinwr, line 153: Can you leave a comment above this item to describe the test case it's used for? Also I don't think the `piid`, `parent_award_id` or `expected_award_id` are necessary here.

dporth-frb: Added a comment to describe the test case

collinwr (approved):

collinwr (approved):

PR #3921, opened 2023-09-21
DEV-10143: Update SubawardSearch w/ County Cols

dpb-bah:
**Description:**
Adding several prime award county columns and updating the subaward county columns in `rpt.subaward_search` (having them be based on the zip data on the broker side instead of deriving it based on city and state data on USAS).

**Technical details:**
* Includes a migration to rpt.subaward_search

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-10143](https://federal-spending-transparency.atlassian.net/browse/DEV-10143):

```
```


usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -81,6 +87,8 @@
     "sub_place_of_perform_state_code": {"delta": "STRING", "postgres": "TEXT"},
     "sub_place_of_perform_state_name": {"delta": "STRING", "postgres": "TEXT"},
     "sub_place_of_performance_zip": {"delta": "STRING", "postgres": "TEXT"},
+    "sub_place_of_performance_county_code": {"delta": "STRING", "postgres": "TEXT"},
```
dporth-frb, line 90: Our code base hard codes the pattern of these these legal entity and performance location columns. It requires that they follow this pattern, "sub_place_of_perform_" and "sub_legal_entity_". This column name goes against the pattern and I anticipate it will cause issues. For example, see the file `/workspaces/usaspending-api/usaspending_api/search/v2/views/spending_by_geography.py` method `post` on line 70.

dporth-frb (changes requested):

usaspending_api/search/delta_models/subaward_search.py
dpb-bah: If that's the case, the downloads would need to update too right? Since they are already existing and pointing to the old names? Wondering why this would be a problem now and not already since they're already like this.

dpb-bah: Ah I see. The columns we were trying to update had that subtle name change. Fixing.

dpb-bah: Ah I see. The columns we were trying to update had that subtle name change. Good catch.

dpb-bah: With the new columns, I kept them as `place_of_performance_county_*` as we currently don't use them in the code but can change them to the `place_of_perform_county_*` versions. Do you have a preference?

dporth-frb: >  we currently don't use them in the code

That's true, but if we do end up using them in the places that hard code this pattern it'll matter.

I think `place_of_perform_` is good

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -44,6 +46,8 @@
     "place_of_perform_state_code": {"delta": "STRING", "postgres": "TEXT"},
     "place_of_perform_state_name": {"delta": "STRING", "postgres": "TEXT"},
     "place_of_performance_zip": {"delta": "STRING", "postgres": "TEXT"},
+    "place_of_perform_county_code": {"delta": "STRING", "postgres": "TEXT"},
```
dporth-frb, line 49: Why did you move these key value pairs?

dpb-bah: Those columns were originally derived on the USASpending code and the order of the columns had those last. Now that they are pulled from the Broker, I moved them to the section where we're pulling other broker values directly (sans major transformations/derivations).

dpb-bah: I'm assuming you are referring to the sub_place_* and the sub_legal_*. The prime award county fields are new.

dporth-frb: Correct. Why does order matter?

dpb-bah: It groups the columns into their respective sections for clarity/simplicity (see the comments)

-- Broker Subaward Table Meta
-- Prime Award Fields (from Broker)
-- Subaward Fields (from Broker)
-- Additional Prime Award Fields (from Broker)
-- Additional Subaward Fields (from Broker)
-- USAS Links (and associated derivations)
-- USAS Derived Fields

So we're moving those updated sub county cols from USAS Derived Fields to Subaward Fields. The order of the models at the top must match the order of the columns selected in the populate query.

dporth-frb (approved):

PR #3922, opened 2023-09-22
[PIPE-504]: Change data dict download to use S3 directly

dporth-frb:
**Description:**
A common issue in the Production Nightly Pipeline is the failure to download the Data Dictionary during the Covid19 Download step. The failure is usually due to a timeout or other connection issue when fetching the Data Dictionary from the file system usaspending.gov exposes. As a workaround, this PR changes how we retrieve the data dictionary, which is directly from its source in S3, in hopes that these connection issues will go away.

**Technical details:**
The change was simple to change to using the s3 download, but we had to make a few config changes in order to get the unit tests to work in Travis CI and to run locally. The unit tests and local operations will interact with MinIO whenever S3 is called. A copy of the data dictionary file was added to repo and the MinIO container will copy that file to into the correct bucket upon startup using docker-compose. 

    - [x] Backend
    - [x] Frontend <OPTIONAL>
    - [x] Operations <OPTIONAL>
8. [x] Jira Ticket [PIPE-504](https://federal-spending-transparency.atlassian.net/browse/PIPE-504):


docker-compose.yml
```diff
@@ -204,7 +205,19 @@ services:
     environment:
       MINIO_ACCESS_KEY: ${MINIO_ACCESS_KEY:-usaspending}
       MINIO_SECRET_KEY: ${MINIO_SECRET_KEY:-usaspender}
-    command: server --address ":10001" --console-address ":10002" /data
+    entrypoint: >
+      /bin/sh -c "
+        if [ ! -d data/dti-da-public-files-nonprod ]; then
```
dporth-frb, line 210: These blocks can be simplified further. There is no need to check if the directory already exists because mkdir will create it only if it doesn't already exist.

> The mkdir command will create any folders that do not exist in the specified path

dporth-frb: You can remove the if thens, as long as an error doesn't occur and interrupt execution.

usaspending_api/common/helpers/s3_helpers.py
```diff
@@ -117,11 +117,11 @@ def download_s3_object(
             logger.info(
                 f"Attempt {attempt + 1} of {retry_count + 1} failed to download {key} from bucket {bucket_name}. Error: {e}"
             )
-            if attempt <= retry_count:
+            if attempt < retry_count:
```
collinwr, line 120: Good catch with this change

collinwr (approved):

PR #3920, opened 2023-09-21
[DEV-10253] Parallel test runs to speed up continuous integration

kwhickey:
**Description:**
Reduce time it takes for CI to run the full test suite. 
Allow developers also to gain parallel test performance locally simply by calling `make tests`

**Technical details:**
- Gained speedups in THREE approaches
    1. Use `pytest-xdist` to run parallel pytest sessions on multiple underlying processes
    2. Employ matrixed builds in Travis CI to take on different subsets of unit and integration tests to divide the workloads up
    3. Separate CI steps into Travis CI `stages` to factor out redundant steps like `pip install` and `docker build` and allow static code analysis to run checks in parallel
- Refactored tests for concurrent process issues and race conditions
- Added a lot of `_fill_optional=True` to baker test models to fix tests
    - Found that if a django `call_command(...)` call was made to a command that used a `.sql` file with a `COMMIT` statement, it was causing the test model data to be committed, and constraint to be evaluated, which would fail if required fields were missing values.
- Refactored fixtures to setup separate tests databases per pytest-xdist worker 
- In `usaspending_api/tests/integration/test_setup_of_spark_dependencies.py`, added a singular test to be run before concurrent spark tests, so that we are not writing dependencies to the same location concurrently and failing due to a file write race condition
- Mocking connections to the download database, so it connects back to the main test database (the one specifically setup for an instance of a parallel test session worker) during tests
- Moving all improperly categorized automated integration tests in to their proper `/tests/integration/` folder

See [example passing Travis CI build](https://app.travis-ci.com/github/fedspendingtransparency/usaspending-api/builds/266321759) with new stages and jobs:
image

**TODO**
_Items that may still be left to do_
1. [x] Review if code coverage files created separately by each Travis matrixed build are being appended/combined to yield an _overall_ code coverage value by Code Climate. If not, figure out how to accumulate coverage from each matrixed build.
    - After a test run, that passed, it seems to stick with the first build run's code coverage (in that case 28% only). Needs fixing
    - See [Code Climate with Travis CI parallel tests](https://docs.codeclimate.com/docs/travis-ci-test-coverage#example-2) example
        1. At the end of each build job, it copies that job's partial coverage up to a external data store (s3 in this case)
        2. Then copies all files from that store (for all jobs finished to-date) down locally
        4. Then sums all
        5. Then updates code climate with the sum as of that job's completion
        6. Last job that finishes should be publishing the overall sum
[...]

PR #3925, opened 2023-09-25
[DEV-9596] Add testing libraries to support usaspending-test container

aguest-kc:
**Description:**
Add the required testing libraries so that the `usaspending-test` container can be run without errors.

**Technical details:**
Added a requirements file named **requirements-testing.txt** that has the testing libraries needed for the `usaspending-test` container to run without errors. Updated the Dockerfile to install the libraries in **requirements-testing.txt** when building the usaspending-backend Docker image, since this is the image that the `usaspending-test` container is based on.

    - [ ] Backend
2. [x] Jira Ticket [DEV-9596](https://federal-spending-transparency.atlassian.net/browse/DEV-9596):


Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN echo "$(python3 --version)"
 ##### Copy python packaged
 WORKDIR /dockermount
 COPY requirements/ /dockermount/requirements/
-RUN python3 -m pip install -r requirements/requirements.txt
+RUN python3 -m pip install -r requirements/requirements.txt -r requirements/requirements-testing.txt
```
dporth-frb, line 38: I don't think we want to put installation of requirements-testing.txt here. If we do, won't these testing libraries be installed on every image associated with this Dockerfile?

dporth-frb (dismissed):
Left a question that probably will turn into a TODO

Dockerfile
aguest-kc: They will be installed in every image associated with our Dockerfile, but that's also the Dockerfile that the usaspending-test container is built on.

The other alternatives would be to create a separate Dockerfile specifically for the usaspending-test container that would install these libraries or I could add to the `command` option in the docker-compose file to run `pip install` before running `pytest`, but that's not the standard way to deal with requirements in Docker.

dporth-frb: I think that it's a bad idea for images to have libraries they don't need. I believe it's best practice for images to be the smallest size possible. Let's look into alternative solutions.

dporth-frb: Here is a snipped from Docker development best practices.

> To keep your production image lean but allow for debugging, consider using the production image as the base image for the debug image. Additional testing or debugging tooling can be added on top of the production image.

aguest-kc: I agree about keeping images as small as possible, but I wouldn't consider our images "small" now and these packages aren't using up tons of extra space. The most reliable alternative would be to have a completely separate Dockerfile just for the usaspending-test container, but that adds another Dockerfile that needs to be maintained in the future.

I have no issue with this new Dockerfile alternative, but it will mean another file to keep track of during things like Python version updates.

collinwr (dismissed):

Dockerfile.testing
```diff
@@ -0,0 +1,18 @@
...
+# in the parent usaspending-api directory.
+
+FROM usaspending-backend:latest
+
+WORKDIR /usaspending-api
+
+# Prevent a Pytest RuntimeError where it expects the parent directory to be the name of an installed Django app
+# This will cause a RuntimeError because `dockermount` isn't the name of an installed Django app, so we move the files
+#   to a `/usaspending-api` parent directory which is the name of an installed Django app.
+RUN mv /dockermount/* /usaspending-api/
```
kwhickey, line 16: I don't feel like this should be necessary. `usaspending-api` I don't believe is either our django "project" or one of it's "apps". Maybe `usaspending_api`?  What's the error you get? It may just be a matter of:
1) setting the environment variable `DJANGO_SETTINGS_MODULE`. You can do that in `Dockerfile` with
```ENV DJANGO_SETTINGS_MODULE=usaspending_api.settings```

2) Making sure the `PYTHONPATH` is resolving all dependencies where the pip install is happening. The place where `pytest` is invoked should be added to the `PYTHONPATH` automatically, but make sure that that place is set as your `WORKDIR`.

Also, if you wanted to name it something other than `dockermount` consider what we do in the AMI images for the api backend: `/data-act/backend/usaspending-api`

Honestly, we should just move where the files are mounted in the base `usaspending-backend` image to get rid of the oddly named "dockermount" directory.

File: Dockerfile.testing [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+
+RUN pip install -r requirements/requirements-testing.txt
```
kwhickey, line 18: See comment below about using `requirements-dev.txt` instead of creating a new file with redundant dependencies.

requirements/requirements-testing.txt
```diff
@@ -0,0 +1,7 @@
+docker==5.0.3
+mock==3.0.5
+model-bakery==1.5.*
+pyspark==3.2.1
+pytest-cov==2.10.1
+pytest-django==4.2.0
+pytest==6.2.*
```
kwhickey, line 7: Typically a `-dev` or `[dev]` set of requirements is what is used for environments that run tests, like continuous integration environments and local development environments. 

We have that with `requirements-dev.txt`, which should have most or all of these dependencies and then maybe a few others.

I think the intention of the `usaspending-ci` docker compose service (container) was to allow tests to run locally

docker-compose.yml
```diff
@@ -49,8 +49,10 @@ services:
   usaspending-test:
     profiles:
       - test  # must pass --profile test to docker-compose for this to come up, or use docker-compose run
-    image: usaspending-backend  # when an image by this name is not found in the local repo, and it is forced to build, it will use this as the tag
-    build: .
+    image: usaspending-testing  # when an image by this name is not found in the local repo, and it is forced to build, it will use this as the tag
+    build:
+      context: .
+      dockerfile: Dockerfile.testing
```
kwhickey, line 55: See down on line 65:
```
command: pytest
```

While doing these changes, I suggest moving the command to the one we've curated in the `Makefile`
```
make tests
```
(which is improved in PR #3920 )

kwhickey (changes requested):
Some feedback inline. 

Can you also please update the PR description to state the problem that prompted this change?

Also `README.txt` should be updated with the suggested or optional way in which to use this container to run tests if that changed (e.g. you must build its image first)

requirements/requirements-testing.txt
aguest-kc: My original thought was to use the existing `requirements-dev.txt` file, but it would install a few extra non-testing libraries into the testing image. I realize that this current approach would mean keeping to requirement files up-to-date if we were to change the pinned version of a library and that would add some unnecessary maintenance overhead.

I'm perfectly happy to use `requirements-dev.txt` instead of creating a new requirements file as long as there are no objects to a few extra non-testing packages being added to the testing image?

Dockerfile.testing
aguest-kc: This is the error message from `pytest`:

```
RuntimeError: Model class dockermount.usaspending_api.etl.tests.integration.test_model.TestModel doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

aguest-kc: Setting the `DJANGO_SETTINGS_MODULE=usaspending_api.settings` did not fix the error. I'll double check the `PYTHONPATH`. I'm pretty sure it's set correctly, but I'll verify.

docker-compose.yml
aguest-kc: Updated the docker-compose file to run `make tests` instead of `pytest`

Dockerfile.testing
aguest-kc: I left this line in the Dockerfile.testing file, because I haven't been able to find a way to prevent `pytest` from throwing this error whenever the files are in a parent directory with a name other than **usaspending-api**

collinwr (approved):

PR #3927, opened 2023-09-27
[DEV-10158] File C to D linkage performance improvements

aguest-kc:
**Description:**
Updated the `UPDATE` queries in the following SQL files to be more performant:
* update_file_c_linkages_fain_and_uri.sql
* update_file_c_linkages_fain.sql
* update_file_c_linkages_piid.sql
* update_file_c_linkages_uri.sql

**Technical details:**
Updated the SQL queries to use `JOIN`s instead of subqueries to increase performance and decrease the runtime of each SQL script. There are before and after timing comparisons in a table in the JIRA ticket.

    - [ ] Backend
8. [x] Jira Ticket [DEV-10158](https://federal-spending-transparency.atlassian.net/browse/DEV-10158):

```
1. Unit and integration tests don't need to be updated for this change.
2. API documentation doesn't need to be updated since no endpoints changed.
5. Frontend isn't impacted by these changes.
7. No operations changes need to be done as part of this PR.
```


usaspending_api/etl/management/sql/c_file_linkage/update_file_c_linkages_fain.sql
```diff
@@ -14,16 +14,20 @@ WHERE
...
+            SELECT fain, uri, award_id, financial_accounts_by_awards_id, submission_id
+            FROM financial_accounts_by_awards AS faba_sub
+        ) AS faba_sub
+        JOIN (
+            SELECT fain
+            FROM {file_d_table} as aw_sub
+            GROUP BY fain
+            HAVING count(*) = 1
+        ) AS aw_sub
+        ON aw_sub.fain = faba_sub.fain
```
collinwr, line 27: Looks like the `UPPER()` function was removed from this comparison (same for other queries). We talked about doing this, but if we do, we'll need to take a few more steps. 

1. We've already confirmed that string fields on the `financial_accounts_by_awards` table are being uppercased in the Submission Loader, but @aguest-kc did find that we have a few records with non uppercased fields from 2018. A data change ticket will be needed to correct this. 
2. We need to perform the same analysis on the Awards table? Are there any fields with lowercase values? If not, is there a way to guarantee that this will always be the case? ex: Is there an upstream step that is capitalizing these values?

dporth-frb: Why not just add the `UPPER()` function back and leave it at that? I think it was removed on accident.

aguest-kc: I confirmed with Daniel that the Broker does accept lowercased values for these fields, because they want to keep the data in the same format that it was submitted in. He also thinks that we already uppercase the data during the pipeline.

I think we should fix the few rows that are currently lowercased, but also add the `UPPER()` calls back in just to be sure we're comparing the data in the same format.

dporth-frb: It was definitely removed by accident, here is the original

`WHERE UPPER(aw_sub.fain) = UPPER(faba_sub.fain)`

We should just add it back.

aguest-kc: We uppercase these values as part of the pipeline, so these `UPPER()` calls aren't necessary. I'm creating a data change ticket to correct some old rows that do have lowercase values in the FABA table, but then everything will be in uppercase format.

dporth-frb: I am going to validate performance improvements by not having this UPPER. If there are no performance improvement, I don't see this being within the scope of this story.

collinwr (approved):

PR #3944, opened 2023-10-20
[PIPE-509] Reduce Calls to load-transactions-in-delta Django Command

dporth-frb:
**Description:**
The unit tests in the file `test_load_transactions_in_delta.py` were optimized to not make so many calls that use a Django command. These commands take the majority of the time, so by reducing even a few calls, it helped reduce the overall runtime by over 2 minutes.

**Technical details:**
Some tests were combined and other parts of the code were refactored so that certain functionality could be tested without needing to use the Django command. There is a new unit test file that can run independently from the integration tests.

    - [x] Backend
8. [x] Jira Ticket [PIPE-509](https://federal-spending-transparency.atlassian.net/jira/software/c/projects/DEV/boards/25?selectedIssue=PIPE-509):

```
2. API docs do not need update as these changes only affected unit tests
4. Matview was not impacted by any of these changes.
5. Frontend was not impacted by any of these changes
6. None of the current changes would affect the data
7. No operational changes need to be done as part of this PR.
```

### DIFF
--- a/usaspending_api/etl/management/commands/load_transactions_in_delta.py
+++ b/usaspending_api/etl/management/commands/load_transactions_in_delta.py
@@ -210,6 +210,7 @@ class Command(BaseCommand):
         """
         Return a list of the award ids corresponding to the transaction_unique_ids that are about to be deleted.
         """
+        table_exists = self.spark._jsparkSession.catalog().tableExists(f"int.award_id_lookup")
         sql = f"""
             WITH txns_to_delete AS (
                 {self.award_id_lookup_delete_subquery}
@@ -225,10 +226,15 @@ class Command(BaseCommand):
         #       returned by a SELECT subquery inside the 'IN').  Thus, this code should return a dataframe directly,
         #       create a temporary view from the dataframe in award_id_lookup_post_delete, and use that temporary
         #       view to either do a subquery in the 'IN' clause or to JOIN against.
-        possibly_modified_award_ids = [str(row["award_id"]) for row in self.spark.sql(sql).collect()]
+        if table_exists:
+            possibly_modified_award_ids = [str(row["award_id"]) for row in self.spark.sql(sql).collect()]
+        else:
+            raise Exception(f"Table: int.{self.etl_level} does not exist.")
+
         return possibly_modified_award_ids
 
     def delete_records_sql(self):
+        table_exists = self.spark._jsparkSession.catalog().tableExists(f"int.{self.etl_level}")
         if self.etl_level == "transaction_id_lookup":
             id_col = "transaction_id"
             subquery = """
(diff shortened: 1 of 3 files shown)


usaspending_api/etl/tests/integration/test_load_transactions_in_delta.py
```diff
@@ -1345,11 +1324,6 @@ def unexpected_paths_source_tables_only_test_core(self):
         # First, load the source tables to Delta
         _load_tables_to_delta(self.s3_data_bucket)
 
-        # 1. Test calling load_transactions_in_delta with etl-level of transaction_f[ab|pd]s before calling with
-        # etl-level of initial_run.
-        with raises(pyspark.sql.utils.AnalysisException, match=f"Table or view not found: int.{self.etl_level}"):
-            call_command("load_transactions_in_delta", "--etl-level", self.etl_level)
-
         # 2. Call load_transactions_in_delta with etl-level of initial_run first, but without first loading
```
dpb-bah, line 1327: If we're removing `# 1.` here, can we update this to `# 1.` and the `# 3.` on line 1346 to `# 2.`?

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -225,10 +226,14 @@ def award_id_lookup_pre_delete(self):
         #       returned by a SELECT subquery inside the 'IN').  Thus, this code should return a dataframe directly,
         #       create a temporary view from the dataframe in award_id_lookup_post_delete, and use that temporary
         #       view to either do a subquery in the 'IN' clause or to JOIN against.
+        if not table_exists:
```
dpb-bah, line 229: Instead of having these two checks in these functions that essentially run before everything else, would it make more sense to refactor/extract this check outside them? It could be in the `handle` function after the `if self.etl_level == "initial_run":` and before the `if self.etl_level == "award_id_lookup":`. It would still run before everything and it would be dynamic `etl_level` based like it is in `delete_records_sql` to cover all the scenarios. Also could include a comment explaining why we're doing this check so early on to save time.

dpb-bah (dismissed):
Clever way of reducing the calls while still preserving the test scenario logic.

usaspending_api/etl/tests/integration/test_load_transactions_in_delta.py
cthomas-frb: Updated

usaspending_api/etl/management/commands/load_transactions_in_delta.py
dporth-frb: Where are these "two" checks that you speak of?

dporth-frb: Are you referring to the table exists check in two different functions?

dpb-bah: Correct, those two can be refactored to a single check just before the two different functions.

cthomas-frb: Done.

PR #3960, opened 2023-11-13
[DEV-10148] New Endpoint for Award Counts by Agency

dporth-frb:
**Description:**
GSA generates a yearly report with award counts for the previous FY by agency and award type. Their current method requires them to run dozens of separate searches and manually copy the award counts by type into a table. This allows them to count prime award summaries where the date range from the earliest to the latest transaction overlaps with the FY. This process takes them several hours per year. They need to continue pulling these same numbers once a year indefinitely. If we change how advance search counts prime award summaries (only count awards with a transaction in the selected FY) they would no longer be able to use that feature for their purposes. There are several possible solutions to support this use case. The ideal solution for GSA would be to create a custom GET endpoint which would return all the numbers they need using their current technique to count awards. This PR addresses that ideal solution.

    - [x] Backend
8. [x] Jira Ticket [DEV-10148](https://federal-spending-transparency.atlassian.net/browse/DEV-10148):

### DIFF
--- a/usaspending_api/agency/tests/integration/test_award_count.py
+++ b/usaspending_api/agency/tests/integration/test_award_count.py
@@ -2,7 +2,6 @@ import pytest
 from model_bakery import baker
 
 from rest_framework import status
-from usaspending_api.common.helpers.fiscal_year_helpers import current_fiscal_year
 from usaspending_api.search.tests.data.utilities import setup_elasticsearch_test
 
 url = "/api/v2/agency/awards/count/{filters}"
@@ -52,7 +51,7 @@ def award_data(db):
         latest_transaction_id=10,
         awarding_agency_id=ag1.id,
         awarding_toptier_agency_code=ta1.toptier_code,
-        awarding_toptier_agency_name='Department',
+        awarding_toptier_agency_name="Department",
         funding_agency_id=ag2.id,
         funding_toptier_agency_code=ta2.toptier_code,
     )
@@ -66,7 +65,7 @@ def award_data(db):
         latest_transaction_id=20,
         awarding_agency_id=ag1.id,
         awarding_toptier_agency_code=ta1.toptier_code,
-        awarding_toptier_agency_name='Department',
+        awarding_toptier_agency_name="Department",
         funding_agency_id=ag2.id,
         funding_toptier_agency_code=ta2.toptier_code,
     )
@@ -80,7 +79,7 @@ def award_data(db):
         latest_transaction_id=30,
         awarding_agency_id=ag1.id,
         awarding_toptier_agency_code=ta1.toptier_code,
-        awarding_toptier_agency_name='Department',
+        awarding_toptier_agency_name="Department",
         funding_agency_id=ag2.id,
         funding_toptier_agency_code=ta2.toptier_code,
     )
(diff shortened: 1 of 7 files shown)


usaspending_api/agency/v2/views/award_count.py
```diff
@@ -0,0 +1,120 @@
...
+        time_period_obj = AwardSearchTimePeriod(
+            default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_SEARCH_MIN_DATE
+        )
+        new_awards_only_decorator = NewAwardsOnlyTimePeriod(
+            time_period_obj=time_period_obj, query_type=_QueryType.AWARDS
+        )
+        filter_options["time_period_obj"] = new_awards_only_decorator
+        for awarding_toptier_agency_name in self._toptier_awarding_agencies_to_include:
+            awarding_toptier_agency_code = (
+                ToptierAgency.objects.filter(name=awarding_toptier_agency_name).first().toptier_code
```
collinwr, line 89: Instead of querying toptier agency for each record in the `toptier_awarding_agencies_to_include_results` list, could this be refactored to use a single query? 

Alternatively, why not update the hardcoded list to include codes instead of names to avoid the query altogether?

collinwr: Never mind about the second suggestion. I see that the code is only used for the final response, not for the ES query.

usaspending_api/api_contracts/contracts/v2/agency/awards/count.md
```diff
@@ -0,0 +1,105 @@
+FORMAT: 1A
+HOST: https://api.usaspending.gov
+
+# Count of awards for Agencies [/api/v2/agency/awards/count{?fiscal_year,group}]
+
+Return the count of Awards under Agencies
```
collinwr, line 6: Should call out that the Awards are grouped by award types. Maybe something like

```
Returns the count of Awards grouped by Award Type under Agencies
```

usaspending_api/agency/tests/integration/test_award_count.py
```diff
@@ -0,0 +1,171 @@
...
+    assert len(results) == 1
+
+    resp = client.get(url.format(filters="?fiscal_year=2021"))
+    results = resp.data["results"]
+    assert resp.status_code == status.HTTP_200_OK
+    assert len(results) == 1
+
+
+@pytest.mark.django_db
+def test_award_count_cfo_agencies_only(client, monkeypatch, award_data, helpers, elasticsearch_award_index):
```
collinwr, line 165: I know the test is passing, so it must be working, but I don't follow the logic. 
I don't see a year being passed in as a filter, nor do I see the current fiscal year being patched, which I expect would result in the actual current fiscal year (24) being used used here. The Award associated with the Department of Commerce (a CFO agency) takes place in 2021, so I would expect it not to appear in the results.

What am I missing here? Does it have to do with the patched Fiscal Year lingering around from the first test?

dporth-frb: Good question. This was an oversight. I agree with you that it should require a patched fiscal year. Let me add in the patch

dporth-frb: Addressed

usaspending_api/api_contracts/contracts/v2/agency/awards/count.md
dporth-frb: Addressed

collinwr (dismissed):

collinwr (approved):

PR #3970, opened 2023-12-03
[DEV-10292] Add location autocomplete API Contract

aguest-kc:
**Description:**
Created a new API Contract for a new Location autocomplete endpoint for Award Search 2.0.

**Technical details:**
Created a new API Contract for a new Location autocomplete endpoint for Award Search 2.0.

1. [x] API documentation updated
    - [ ] Backend
    - [x] Frontend
3. [x] Frontend impact assessment completed
4. [x] Jira Ticket [DEV-10292](https://federal-spending-transparency.atlassian.net/browse/DEV-10292):

### DIFF
--- a/usaspending_api/api_contracts/contracts/v2/autocomplete/location.md
+++ b/usaspending_api/api_contracts/contracts/v2/autocomplete/location.md
@@ -1,0 +1,107 @@
+FORMAT: 1A
+HOST: https://api.usaspending.gov
+
+# Location Autocomplete [/api/v2/autocomplete/location/]
+
+This endpoint is used by the Location autocomplete filter on the Advanced Search page.
+
+## POST
+
+This route sends a request to the backend to retrieve locations matching the specified search text.
+
++ Request (application/json)
+    + Schema
+
+            {
+                "$schema": "http://json-schema.org/draft-04/schema#",
+                "type": "object"
+            }
+
+    + Attributes (object)
+        + `limit` (optional, number)
+            + Default: 10
+        + `search_text` (required, string)
+    + Body
+
+            {
+                "search_text": "Den"
+            }
+
++ Response 200 (application/json)
+    + Attributes (object)
+        + `results` (required, array[Locations], fixed-type)
+    + Body
+
+            {
+                "results": [
+                    "countries": [
+                        {
+                            "country_name": "Denmark"
+                        },
+                        {
+                            "country_name": "Sweden"
+                        }
+                    ],
+                    "cities": [
+                        {
+                            "city_name": "Denver",
+                            "state_name": "Colorado",
+                            "country_name": "United States"
+                        },
+                        {
+                            "city_name": "Gadsden",
+                            "state_name": "Alabama",
+                            "country_name": "United States"
+                        },
+                        {
+                            "city_name": "Camden",
+                            "state_name": "Arkansas",
+                            "country_name": "United States"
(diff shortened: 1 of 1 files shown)


aguest-kc:
Waiting to merge until the new endpoint is also ready to merge.

usaspending_api/api_contracts/contracts/v2/autocomplete/location.md
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+    + Attributes (object)
+        + `limit` (optional, number)
+            + Default: 10
+        + `search_text` (required, string)
+    + Body
+
+            {
+                "search_text": "Den"
```
anjenkin, line 27: not sure if this question is premature, just curious if there are any details around the search text (ie. min, max string length, case sensitivity, etc)

anjenkin (changes requested):
Looks good from a frontend perspective.  Just one question inline.

anjenkin (commented):
Looks good from the frontend perspective. Just one question inline.

usaspending_api/api_contracts/contracts/v2/autocomplete/location.md
aguest-kc: I thought we had a minimum length setting in our existing `/api/v2/autocomplete` endpoints, but that doesn't appear to be the case.

For these Award Search 2.0 endpoints, I was thinking of a **minimum character count** of **3** to help reduce the number of API requests and Elasticsearch queries we need to handle. It would also help to reduce the number of results that the user would need to sort through.

For **case sensitivity**, I was planning on uppercasing the `search_text` value on the backend since the Elasticsearch data that we're querying is stored in uppercase format.

As for **max string length** and **special characters** I didn't take those two cases into account. I assumed we'd handle special characters like we do with other endpoints and treat them as literal text and not special characters. And I don't believe we have a max string length on other endpoints, but we could add something if we need it.

aguest-kc: It sounds like the questions are:

- Is 3 characters a good minimum input for these autocomplete endpoints?
- Is there any reason **not** to uppercase all input on the backend, regardless of the user-submitted casing?
- Should we have a max limit on the `search_text` parameter?
- Should we treat special characters as normal text like we do with other endpoints?

anjenkin: Ok, yeah I think a minimum character count of 3 makes sense.  This isn't standardized across the current filters.  For example NAICs and PSC, the search is triggered after two characters are entered.  This makes sense for NAICs since there are two character codes.

For special characters, I'm only really concerned commas.  Just wondering since commas will be displayed on the frontend in the results list.  Though we can handle commas on the frontend before the request is sent to the autocomplete endpoint.

Thanks for clarifying.

aguest-kc: Commas will be an interesting case. During the implementation we'll have to think though how the code should break that down from a search perspective. The code won't inherently know that a search for `Denver, Colorado` means a city of `Denver` and a state of `Colorado`, so what's the best way to perform that search?

anjenkin: ok

anjenkin (approved):

PR #3984, opened 2023-12-21
[DEV-10427] Add an Elasticsearch Index for Recipients

aguest-kc:
**Description:**
Create a new Elasticsearch index for Recipients to support future Award Search 2.0 autocomplete functionality.

**Technical details:**
Create a new Elasticsearch index named `recipients`. Update the `elasticsearch_indexer` and `elasticsearch_indexer_for_spark` commands to support a new `--load-type` of `recipient`.

    - [ ] Backend
2. [x] Matview impact assessment completed
3. [x] Data validation completed
4. [x] Jira Ticket [DEV-10427](https://federal-spending-transparency.atlassian.net/browse/DEV-10427):


usaspending_api/etl/management/commands/elasticsearch_indexer.py
```diff
@@ -221,7 +223,63 @@ def parse_cli_args(options: dict, es_client) -> dict:
...
+
+            stored_date_key: The `name` in the `external_data_type` table to use when referencing this Elasticsearch
+                index. This table, in combination with the `external_data_load_date` table, are used to keep track
+                of when data was last loaded into a destination (database table or Elasticsearch index).
+
+            unique_key_field: Name of the Elasticsearch field that will be unique across all records.
+
+            write_alias: Name to use for the Elasticsearch alias that wil be used to write data.
+                Note: This should be defined in settings.py
+    """
```
collinwr, line 280: This documentation on the config was a nice addition. Thank you!

usaspending_api/etl/es_recipient_profile_template.json
```diff
@@ -0,0 +1,71 @@
+{
+  "settings": {
+    "index.mapping.ignore_malformed": true,
+    "index.max_result_window": null,
```
collinwr, line 4: What's the intention behind setting this value to `null`? I see this is the case on other indexes, but I'm not sure `null` is a valid value for this field. 

https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html

usaspending_api/etl/es_recipient_profile_template.json
```diff
@@ -0,0 +1,71 @@
...
+    "index.refresh_interval": -1,
+    "index": {
+      "number_of_shards": 5,
+      "number_of_replicas": 0,
+      "sort.field": [
+        "recipient_name.keyword",
+        "uei.keyword"
+      ],
+      "sort.order": [
```
collinwr, line 13: Is the sort to support sorting and paging results for the API queries

usaspending_api/etl/management/commands/elasticsearch_indexer.py
```diff
@@ -286,6 +344,27 @@ def set_config(passthrough_values: list, arg_parse_options: dict) -> dict:
...
+            "extra_null_partition": False,
+            "field_for_es_id": "id",
+            "initial_datetime": default_datetime,
+            "max_query_size": settings.ES_RECIPIENTS_MAX_RESULT_WINDOW,
+            "optional_predicate": "",
+            "primary_key": "id",
+            "query_alias_prefix": settings.ES_RECIPIENTS_QUERY_ALIAS_PREFIX,
+            "required_index_name": settings.ES_RECIPIENTS_NAME_SUFFIX,
+            "sql_view": settings.ES_RECIPIENTS_ETL_VIEW_NAME,
+            "stored_date_key": ...,
```
collinwr, line 364: What does this `...` represent?

usaspending_api/etl/es_recipient_profile_template.json
aguest-kc: I kept this as `null` since it's what we have in the other index template files. We define an actual value for our indices in **settings.py** as `<index>_MAX_RESULT_WINDOW` and set them to 50,000. My only assumption is that we don't want to define this value in multiple places?

usaspending_api/etl/es_recipient_profile_template.json
aguest-kc: `index.sort.field` defines which field(s) to sort by when indexing documents and `index.sort.order` defines what order (asc or desc) to sort the field(s) by. These settings just affect how Elasticsearch internally organizes the terms and has nothing to do with how results are returned when a query is executed. They're just related to how ES divides up the records across the shards in the cluster.

As part of a query, Elasticsearch returns documents based on a relevance score named `_score` and the results are returned in descending order based on their `_score` value.

usaspending_api/etl/management/commands/elasticsearch_indexer.py
aguest-kc: `...` is usually used when omitting a value, but it's not the same as `None`. In this case `None` is probably the correct value to use since these `stored_date_key` isn't used for those indices.

If we use `...` and did a simple `if stored_date_key: <do something>` check then the `if` statement would pass since `...` isn't `None` or `False`.

For example:
```python
>>> x = ...

>>> type(x)
<class 'ellipsis'>

>>> y = None

>>> type(y)
<class 'NoneType'>

>>> if x: print("Value here")
Value here

>>> if y: print("Value here")
>>>
```

I can switch the `...` out for `None` in a commit to make them clearer.

usaspending_api/etl/management/commands/elasticsearch_indexer.py
aguest-kc: Thanks! I hope the documentation makes it easier for others to update these configs, because I had to search through the code base to understand a few of these fields.

dporth-frb (dismissed):
Missing unit/integration tests.

usaspending_api/etl/elasticsearch_loader_helpers/transform_data.py
```diff
@@ -137,6 +136,10 @@ def transform_covid19_faba_data(worker: TaskSpec, records: List[dict]) -> List[d
     return list(results.values())  # don't need the dict key, return a list of the dict values
 
 
+def transform_recipient_profile_data(worker: TaskSpec, records: List[dict]) -> List[dict]:
```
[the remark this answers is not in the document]
aguest-kc: The elasticsearch indexer command config expects a `data_transform_func` value that points to a function to run to transform the records. The management command expects this to be a valid function, so I just added a basic function that just returns the records since we don't need to transform them.

```python
config = {
            "base_table": "recipient_profile",
            "base_table_id": "id",
            "create_award_type_aliases": False,
            "data_transform_func": transform_recipient_profile_data,
        }
```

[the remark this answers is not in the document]
dporth-frb: Is there a better way to go about this besides defining a obsolete method?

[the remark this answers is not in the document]
dporth-frb: Perhaps you could take advantage of `lambda` in the config json.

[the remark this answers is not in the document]
dporth-frb: Another idea is to update the django command to allow None (or something besides a shell of a function).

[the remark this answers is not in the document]
aguest-kc: I updated the management command to support a value of `None` whenever the `data_transform_func` value isn't going to be needed.

usaspending_api/conftest.py
```diff
@@ -588,3 +588,21 @@ def fake_sqs_queue():
     q.purge()
     q.reset_instance_state()
     remove_unittest_queue_data_files(q)
+
+
+@pytest.fixture
+def elasticsearch_recipient_index(db):
```
collinwr, line 594: Tests look good. I'd maybe just move this up with the other ES fixtures.

aguest-kc: Just pushed a commit with the new fixture moved up with the other ES fixtures.

usaspending_api/etl/es_recipient_profile_template.json
aguest-kc: Removed these sort options since they aren't going to help the performance during index time and we aren't sorting our results in the same way, so there's no benefit to having these values.

collinwr (approved):

PR #3979, opened 2023-12-08
[DEV-10294]; Award Search 2.0 Agency Autocomplete Api Contracts

dporth-frb:
**Description:**
To support award search 2.0 agency autocomplete we are creating two new endpoints. Each endpoint returns the same structure of data, the difference is one is for funding and one is for awarding data.

Any agencies (toptier agencies), sub-agencies (sub-tier agencies), or offices that match the provided search text will be in the response. Additionally, any agencies, sub-agencies, or offices will include their associated sub-agencies, offices, or agencies.

PR #3995, opened 2024-01-21
[DEV-10455] Implements Award Search 2.0 Autocomplete Endpoints

dporth-frb:
**Description:**
The purpose of this story is to create and implement two new endpoints that will support the Award Search 2.0's agency autocomplete feature.

**Technical details:**
Please read the following for performance evaluation.

> Performance seems reasonable. In cases where the search text is one letter, "L", the response time is at most ~1000 miliseconds. "L" is a very generic search text, which means there are many records that match the search text. Parsing and returning all those matching records, with their children in under ~3 seconds seems reasonable. In other cases where the search text is specific, "African Development Foundation", the response time is at most ~600 miliseconds. Overall, I believe this performance is adequate.

Additionally, I've taken the changes from the api contract PR and rolled those changes into this PR. The old PR, which is now closed can be found here https://github.com/fedspendingtransparency/usaspending-api/pull/3979. This decision made since to me since the api contracts shouldn't be merged without the corresponding BE API logic implemented.

Lastly, this PR should be merged after https://github.com/fedspendingtransparency/usaspending-api/pull/3993.

    - [ ] Backend
8. [x] Jira Ticket [DEV-10455)](https://federal-spending-transparency.atlassian.net/browse/DEV-10455):


dporth-frb:
I plan to ignore the following code climate complaint. 

> File autocomplete.py has 338 lines of code (exceeds 250 allowed). Consider refactoring.

autcomplete.py with 338 lines of code for a module is consistent with our code base and is not unreasonable for the `autocomplete.py` module.

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -83,6 +83,194 @@ def agency_autocomplete(self, request):
 
         return Response({"results": results})
 
+    def agency_office_autocomplete(self, request):
```
collinwr, line 86: To me it seems like there are a lot of queries going on here for what seems to be pretty denormalized data. For example, for the Toptier Agency section, if you don't do the `DISTINCT` operation in the original query below, you actually already have everything you need to populate the subtier agency and offices section. 
```
toptier_agency_search_text_matches = AgencyOfficeAutocompleteMatview.objects.filter(
        toptier_agency_filter
).order_by(*order_by)
```
Why requery for this same data?

In theory, only one query is necessary to get everything you need for the whole response if you match the search text against all possible matching fields, but I'm not sure about how the performance of this would compare to running one query for each of the three sections. Let's discuss tomorrow.

collinwr (dismissed):

PR #4002, opened 2024-01-30
DEV-10428 Create autocomplete endpoint for Recipients

ayubshahab:
**Description:**
To support the new autocomplete features in Award Search 2.0 we created a new endpoint for searching through Recipients. This new endpoint builds off of dev-10427, which created the Recipient Elasticsearch index for Recipient profiles.

TESTING
 A new testing file titled test_recipients_autocomplete_v2.py is created to test for any matches in the recipient_name, recipient_level or uei fields and return all of the matches.

**Technical details:**

    - [x] Backend
4. [x] Matview impact assessment completed - N/A
5. [x] Frontend impact assessment completed - N/A
6. [x] Data validation completed - N/A
7. [x] Appropriate Operations ticket(s) created - N/A
8. [x] Jira Ticket [DEV-10428](https://federal-spending-transparency.atlassian.net/browse/DEV-10428):

```
```

### DIFF
--- a/usaspending_api/references/v2/views/recipients.py
+++ b/usaspending_api/references/v2/views/recipients.py
@@ -1,0 +1,87 @@
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from collections import OrderedDict
+from elasticsearch_dsl import Q as ES_Q, A
+
+from usaspending_api.common.cache_decorator import cache_response
+
+from usaspending_api.common.elasticsearch.search_wrappers import RecipientSearch
+from usaspending_api.search.v2.es_sanitization import es_sanitize
+from usaspending_api.common.validator.tinyshield import validate_post_request
+
+models = [
+    {"key": "search_text", "name": "search_text", "type": "text", "text_type": "search", "optional": False},
+    {"key": "limit", "name": "limit", "type": "integer", "max": 500, "optional": True, "default": 10},
+    {"key": "recipient_levels", "name": "recipient_levels", "type": "array", "items": {"type": "string"}, "optional": True},
+]
+
+@validate_post_request(models)
+class RecipientAutocompleteViewSet(APIView):
+    """
+    This endpoint is used for the Recipient autocomplete filter which returns a list of recipients matching the specified search text.
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md"
+
+    @cache_response()
+    def post(self, request, format=None):
+        search_text, recipient_levels = prepare_search_terms(request.data)
+        limit = request.data["limit"]
+
+        query = create_elasticsearch_query(search_text, recipient_levels, limit)
+        sorted_results = query_elasticsearch(query)
+        response = OrderedDict([("count", len(sorted_results)), ("results", sorted_results[:limit])])
+
+        return Response(response)
+
+
+def prepare_search_terms(request_data):
+    fields = [request_data["search_text"], request_data.get("recipient_levels", [])]
+    return [es_sanitize(field).upper() if isinstance(field, str) else field for field in fields]
+
+def create_elasticsearch_query(search_text, recipient_levels, limit):
+    query = create_es_search(search_text, recipient_levels)
+    query.aggs.bucket("recipients", A("terms", field="recipient_name.keyword", size=limit))
+    return query
+
+ # search_text = "kimberly"  # Test recipient name. There is a recipient with this name and a level of "R"
+def create_es_search(search_text, recipient_levels):
+    ES_RECIPIENT_SEARCH_FIELDS = ["recipient_name", "uei"]
+
+    if recipient_levels:
+        # Only return the matches where the `recipient_level` field contains the filter value from
+        #   the HTTP request.
+        query = ES_Q(
+            "bool",
+            must=[
+                ES_Q("match", recipient_level="C"),
+                ES_Q("query_string", query=f"*{search_text}*", fields=ES_RECIPIENT_SEARCH_FIELDS)
+            ]
+        )
+    else:
+        # If we're NOT filtering on the `recipient_level` field then return all results matching the given search text
+        query = ES_Q("query_string", query=f"*{search_text}*", fields=ES_RECIPIENT_SEARCH_FIELDS)
+
+    search = RecipientSearch().filter(query)
+    return search
+
+def query_elasticsearch(query):
+    hits = query.handle_execute()
+    results = []
+    if hits and hits["hits"]["total"]["value"] > 0:
+        results = parse_elasticsearch_response(hits)
+        results = sorted(results, key=lambda x: x.pop("hits"), reverse=True)
+    return results
+
+def parse_elasticsearch_response(hits):
+    results = []
+    for recipient in hits["aggregations"]["recipients"]["buckets"]:
+        results.append(
+            OrderedDict(
+                [
+                    ("recipient_name", recipient["key"]),
+                    ("hits", recipient["doc_count"]),
+                ]
+            )
+        )
+    return results
--- a/usaspending_api/references/v2/views/recipients.py
+++ b/usaspending_api/references/v2/views/recipients.py
@@ -10,9 +10,28 @@ from usaspending_api.search.v2.es_sanitization import es_sanitize
 from usaspending_api.common.validator.tinyshield import validate_post_request
 
 models = [
-    {"key": "search_text", "name": "search_text", "type": "text", "text_type": "search", "optional": False},
-    {"key": "limit", "name": "limit", "type": "integer", "max": 500, "optional": True, "default": 10},
-    {"key": "recipient_levels", "name": "recipient_levels", "type": "array", "items": {"type": "string"}, "optional": True},
+    {
+        "name": "search_text", 
+        "key": "search_text", 
+        "type": "text", 
+        "text_type": "search", 
+        "optional": False},
+    {
+        "name": "limit", 
+        "key": "limit", 
+        "type": "integer", 
+        "max": 500, 
+        "optional": True, 
+        "default": 10},
+    {
+        "name": "recipient_levels",
+        "key": "recipient_levels",  
+        "type": "array", 
+        "array_type":"text", 
+        "text_type": "search", 
+        "items": {"type": "string"}, 
+        "optional": True
+     },
 ]
 
 @validate_post_request(models)
@@ -26,10 +45,15 @@ class RecipientAutocompleteViewSet(APIView):
     @cache_response()
     def post(self, request, format=None):
         search_text, recipient_levels = prepare_search_terms(request.data)
+        print("SEARCH TEXT", search_text)
+        print("RECIPIENT LEVELS", recipient_levels)
         limit = request.data["limit"]
+        print("LIMIT", limit)
 
         query = create_elasticsearch_query(search_text, recipient_levels, limit)
+        print("QUERY:", query)
         sorted_results = query_elasticsearch(query)
+        print("Sorted Results:", sorted_results)
         response = OrderedDict([("count", len(sorted_results)), ("results", sorted_results[:limit])])
 
         return Response(response)
@@ -41,6 +65,7 @@ def prepare_search_terms(request_data):
 
 def create_elasticsearch_query(search_text, recipient_levels, limit):
     query = create_es_search(search_text, recipient_levels)
+    print("CREATE ES SEARCH: ", query)
     query.aggs.bucket("recipients", A("terms", field="recipient_name.keyword", size=limit))
     return query
 
@@ -51,11 +76,24 @@ def create_es_search(search_text, recipient_levels):
     if recipient_levels:
         # Only return the matches where the `recipient_level` field contains the filter value from
         #   the HTTP request.
+        # query = ES_Q(
+        #     "bool",
+        #     must=[
+        #         ES_Q("match", recipient_level="C"),
+        #         ES_Q("query_string", query=f"*{search_text}*", fields=ES_RECIPIENT_SEARCH_FIELDS)
+        #     ]
+        # )
+
+         # Use a bool query with should clauses for each recipient level
+        should_clauses = [
+            ES_Q("match", recipient_level=level)
+            for level in recipient_levels
+        ]
         query = ES_Q(
             "bool",
             must=[
-                ES_Q("match", recipient_level="C"),
-                ES_Q("query_string", query=f"*{search_text}*", fields=ES_RECIPIENT_SEARCH_FIELDS)
+                *should_clauses,
+                ES_Q("query_string", query=f"*{search_text}*", fields=ES_RECIPIENT_SEARCH_FIELDS),
             ]
         )
     else:
--- a/usaspending_api/references/tests/integration/test_recipients_autocomplete_v2.py
+++ b/usaspending_api/references/tests/integration/test_recipients_autocomplete_v2.py
@@ -1,0 +1,75 @@
+import json
+import pytest
+
+from django.conf import settings
+from model_bakery import baker
+
+@pytest.fixture
+def transaction_data_fixture(db):
+    baker.make(
+        "search.RecipientSearch", 
+        id=1,
+        recipient_hash="sample_hash_1",
+        recipient_level="R",
+        recipient_name="Sample Recipient 1",
+        uei="UEI1"
+    )
+
+    baker.make(
+        "search.RecipientSearch",  
+        id=2,
+        recipient_hash="sample_hash_2",
+        recipient_level="C",
+        recipient_name="Sample Recipient 2",
+        uei="UEI2"
+    )
+
+def test_recipient_search_matches_found(client, monkeypatch, transaction_data_fixture, elasticsearch_transaction_index):
+    monkeypatch.setattr(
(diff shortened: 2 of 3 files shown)


usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
```diff
@@ -0,0 +1,61 @@
...
+            + Default: 10
+        + `search_text` (required, string)
+        + `recipient_levels` (optional, array[string])
+    + Body
+
+            {
+                "search_text": "Holdings"
+            }
+
++ Response 200 (application/json)
```
dporth-frb, line 31: I would add a messages field here. Even though it won't be used...right now, it may be used in the future. I ran into this issue recently where an existing endpoint we deprecated didn't have a messages field so I had to change the api contract to include it. Please see, https://github.com/fedspendingtransparency/usaspending-api/pull/3995/files#diff-e35f25abf9df4847c3431eb10b6a969760b808de98bde1b7c96410f00d2d7284, for an example.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+def prepare_search_terms(request_data):
+    fields = [request_data["search_text"], request_data.get("recipient_levels", [])]
+    return [es_sanitize(field).upper() if isinstance(field, str) else field for field in fields]
+
+
+def create_es_search(search_text, recipient_levels, limit):
+    ES_RECIPIENT_SEARCH_FIELDS = ["recipient_name", "uei"]
+
+    query = ES_Q()
+    # Build the should_clauses list
```
dporth-frb, line 55: This comment is unnecessary. With a split second glance I can see that's what is happening. I'd limit your in line code comments to explain why you're doing something.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+        response = OrderedDict([("count", len(results)), ("results", results)])
+        return Response(response)
+
+
+def prepare_search_terms(request_data):
+    fields = [request_data["search_text"], request_data.get("recipient_levels", [])]
+    return [es_sanitize(field).upper() if isinstance(field, str) else field for field in fields]
+
+
+def create_es_search(search_text, recipient_levels, limit):
```
dporth-frb, line 51: This method probably should be a behavior of the class `RecipientAutocompleteViewSet` and not just a function belonging to this module. Additionally, you can probably make this method private via a leading underscore.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+    def post(self, request, format=None):
+        search_text, recipient_levels = prepare_search_terms(request.data)
+        limit = request.data["limit"]
+        query = create_es_search(search_text, recipient_levels, limit)
+        results = query_elasticsearch(query)
+        response = OrderedDict([("count", len(results)), ("results", results)])
+        return Response(response)
+
+
+def prepare_search_terms(request_data):
```
dporth-frb, line 46: > This method probably should be a behavior of the class RecipientAutocompleteViewSet and not just a function belonging to this module. Additionally, you can probably make this method private via a leading underscore.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+
+    else:
+        query = search_text_should_clause
+
+    query = RecipientSearch().query(query)[:limit]
+    return query
+
+
+def query_elasticsearch(query):
```
dporth-frb, line 88: > This method probably should be a behavior of the class RecipientAutocompleteViewSet and not just a function belonging to this module. Additionally, you can probably make this method private via a leading underscore.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+    hits = query.handle_execute()
+    results = []
+    if hits and hits["hits"]["total"]["value"] > 0:
+        results = parse_elasticsearch_response(hits)
+    return results
+
+
+def parse_elasticsearch_response(hits):
```
dporth-frb, line 96: > This method probably should be a behavior of the class RecipientAutocompleteViewSet and not just a function belonging to this module. Additionally, you can probably make this method private via a leading underscore.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+        ]
+
+        query = ES_Q(
+            "bool", must=[ES_Q("bool", should=recipient_should_clause), ES_Q("bool", should=search_text_should_clause)]
+        )
+
+    else:
+        query = search_text_should_clause
+
+    query = RecipientSearch().query(query)[:limit]
```
dporth-frb, line 84: I don't see an order by clause anywhere. That's problematic with this `[:limit]` because results could be in a different order each request, thus causing different results between requests with the same limit.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+        return Response(response)
+
+
+def prepare_search_terms(request_data):
+    fields = [request_data["search_text"], request_data.get("recipient_levels", [])]
+    return [es_sanitize(field).upper() if isinstance(field, str) else field for field in fields]
+
+
+def create_es_search(search_text, recipient_levels, limit):
+    ES_RECIPIENT_SEARCH_FIELDS = ["recipient_name", "uei"]
```
dporth-frb, line 52: This variable is locally scoped to this function. Typically, you only make global, static variables all caps. I don't think all caps are necessary here.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+                should=[ES_Q("match", recipient_level=level) for level in recipient_levels],
+                minimum_should_match=1,
+            )
+        ]
+
+        query = ES_Q(
+            "bool", must=[ES_Q("bool", should=recipient_should_clause), ES_Q("bool", should=search_text_should_clause)]
+        )
+
+    else:
```
dporth-frb, line 81: Instead of an `if...else` here. Why not just initialize the variable `query` to `search_text_should_clause` and include a single `if` statement. If the `if` statement evaluates to true then it will override the initial value. Doing so will make code that's easier to read (especially since the else is one line).

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+        query = search_text_should_clause
+
+    query = RecipientSearch().query(query)[:limit]
+    return query
+
+
+def query_elasticsearch(query):
+    hits = query.handle_execute()
+    results = []
+    if hits and hits["hits"]["total"]["value"] > 0:
```
dporth-frb, line 91: Before accessing the keys  "hits", "total", and "value", you are not checking for them to exist in the dictionary. Not checking for these keys to exist is potentially dangerous, e.g. although unlikely imagine a case where they do not exist your code would cause a `KeyError` to be thrown.

usaspending_api/references/v2/views/recipients.py
dporth-frb, line 96: You should include pydocs and type hints on this function.

usaspending_api/references/v2/views/recipients.py
dporth-frb, line 88: You should include pydocs and type hints on this function.

usaspending_api/references/v2/views/recipients.py
dporth-frb, line 51: You should include pydocs and type hints on this function.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,110 @@
...
+class RecipientAutocompleteViewSet(APIView):
+    """
+    This endpoint is used for the Recipient autocomplete filter which returns a list of recipients matching the
+    specified search text.
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md"
+
+    @cache_response()
+    def post(self, request, format=None):
```
dporth-frb, line 37: You should include pydocs and type hints on this function.

dporth-frb (changes requested):

usaspending_api/references/v2/views/recipients.py
ayubshahab: If the if statement evaluates to true, we don't want to override the initial value. if there are recipient_levels, the ES_Q query has a must with two shoulds inside (search_text_should_clause) & (recipient_should_clause). the else portion does not have to worry about the recipient_should_clause. Need the if-else

dporth-frb:
Lastly, you've created a few functions (that will become methods once you move them into a class) that can be unit tested. I suggest creating unit tests for these functions/methods.

usaspending_api/references/v2/views/recipients.py
ayubshahab: I thought we don't have to explicitly write out and order by clause because It would go by the score that elastic search generates for each hit and it automatically sorts them in desc order, ordering by relevance.

usaspending_api/references/v2/views/recipients.py
ayubshahab: I guess I could set it to override when the if statement evaluates to true, I would just be reusing the query and setting the should=query (on line 78 where we need to set the should =  search_text_should_clause). But this would be confusing for future developers.

dporth-frb: Hmm I don't think you understand my point. Leave the `if` remove the `else` and move the contents from the `else` right above the `if`.

ayubshahab: OH thats what you meant. Okay will do!

usaspending_api/references/v2/views/recipients.py
dporth-frb: Can you point me to elastic search documentation that tells us this?

usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
ayubshahab: Okay so if I understand this correctly, just adding the message field here and nothing else? Don't have to change the API itself right?

usaspending_api/references/v2/views/recipients.py
ayubshahab: What do we mean by "a behavior of the class RecipientAutocompleteViewSet and not just a function belonging to this module"?

ayubshahab (commented):
Please see the comments below.

usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
dporth-frb: If you add a new field to the api contract you will have to populate it right? It's ok if it's an empty list. See my link for an example

usaspending_api/references/v2/views/recipients.py
dporth-frb: When we talk about object oriented programming, "behaviors" mean methods of a class. To be exact, a behavior is something an object does. So a method is a behavior of a class.

dporth-frb: Just make them methods of the class not functions.

usaspending_api/references/v2/views/recipients.py
dporth-frb: I mean document the function, see existing method docs as an example, and yes "type hints is just specifying parameter input types". The multi-line comment circled in red is what I call a pydoc
image

usaspending_api/references/v2/views/recipients.py
aguest-kc: By default, Elasticsearch returns results based on each result's `_score` value in descending order. The `_score` field represents how relevant Elasticsearch thinks that result is to the provided search query. So there is an order by/sort happening here, it's just happening by default and behind-the-scenes.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,165 @@
...
+
+        search_text, recipient_levels = self._prepare_search_terms(request.data)
+        limit = request.data["limit"]
+        query = self._create_es_search(search_text, recipient_levels, limit)
+        results = self._query_elasticsearch(query)
+        response = OrderedDict([("count", len(results)), ("results", results), ("messages", [""])])
+        return Response(response)
+
+    @staticmethod
+    def _prepare_search_terms(request_data: Dict[str, Union[str, List[str]]]) -> List[Union[str, List[str]]]:
```
dporth-frb, line 65: Why is this a staticmethod?

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,165 @@
...
+            request_data: The request data containing search text and optional recipient levels.
+
+        Returns:
+            A list containing the sanitized search text and recipient levels.
+        """
+        fields = [request_data["search_text"], request_data.get("recipient_levels", [])]
+        return [es_sanitize(field).upper() if isinstance(field, str) else field for field in fields]
+
+    @staticmethod
+    def _create_es_search(search_text: str, recipient_levels: List[str], limit: int) -> RecipientSearch:
```
dporth-frb, line 79: > Why is this a staticmethod?

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,165 @@
...
+                )
+            ]
+            # if there are recipient levels, then any of the options from the recipient levels as well as the search text should match  # noqa: E501
+            query = ES_Q("bool", must=[ES_Q("bool", should=recipient_should_clause), ES_Q("bool", should=query)])
+
+        query = RecipientSearch().query(query)[:limit]
+        return query
+
+    @staticmethod
+    def _query_elasticsearch(query: RecipientSearch) -> List[Dict[str, Any]]:
```
dporth-frb, line 118: > Why is this a staticmethod?

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,165 @@
...
+            and "hits" in hits
+            and "total" in hits["hits"]
+            and "value" in hits["hits"]["total"]
+            and hits["hits"]["total"]["value"] > 0
+        ):  # noqa: E501
+            results = RecipientAutocompleteViewSet._parse_elasticsearch_response(hits)
+        return results
+
+    @staticmethod
+    def _parse_elasticsearch_response(hits: Dict[str, Any]) -> List[Dict[str, Any]]:
```
dporth-frb, line 142: > Why is this a staticmethod?

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,165 @@
...
+
+        Args:
+            query: The Elasticsearch search query.
+
+        Returns:
+            A dictionary containing results from the Elasticsearch search query.
+        """
+        hits = query.handle_execute()
+        results = []
+        # if hits and hits["hits"]["total"]["value"] > 0:
```
dporth-frb, line 130: I think you meant to remove this inline code comment right?

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,165 @@
...
+                    "search_text": str
+                    "recipient_levels": [""] (optional)
+                    "limit": int (optional)
+                }
+            format: The format of the response (default=None).
+
+        Returns:
+            Returns a list of Recipients matching the input search_text and recipient_levels, if passed in.
+        """
+
```
dporth-frb, line 56: Remove this whitespace

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md"
+
+    @cache_response()
+    def post(self, request: Request, format=None) -> Response:
+
+        """
+        Handle POST requests to the endpoint.
+
```
dporth-frb, line 43: Remove this whitespace

dporth-frb (changes requested):

dporth-frb:
Overall, looks really good after your changes. I found a few more small changes for you to make.

Also, it looks like you are still missing unit tests.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,159 @@
...
```
dporth-frb, line 43: whitespace is still here. Is your auto formatter adding it?

usaspending_api/references/v2/views/recipients.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    This endpoint is used for the Recipient autocomplete filter which returns a list of recipients matching the
+    specified search text.
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md"
+
+    @cache_response()
+    def post(self, request: Request, format=None) -> Response:
+
```
dporth-frb, line 40: whitespace is still here. Is your auto formatter adding it?

dporth-frb (approved):

aguest-kc (approved):

PR #4003, opened 2024-02-01
DEV-10455; Accommodates how usaspending.gov is currently using this endpoint

dporth-frb:
usaspending.gov was using the `awarding_office` and `funding_office` endpoints without a trailing forward slash. However, the lack of a trailing forward slash in use of these endpoints did not match our api contracts. Unfortunately, the code that powers these endpoints did not match the api contracts so usaspending.gov had no issue using the endpoints. In sprint 180 these endpoints were updated to match our api contracts and to enable two new, similar, endpoints to be created. This PR reverts the trailing forward slash so that usaspending.gov does not have to update itself. However, this PR requires us to update our api contracts to match how usaspending.gov wants to use the `awarding_office` and `funding_office` endpoints.

### DIFF
--- a/usaspending_api/api_contracts/contracts/v2/autocomplete/awarding_agency.md
+++ b/usaspending_api/api_contracts/contracts/v2/autocomplete/awarding_agency.md
@@ -1,7 +1,7 @@
 FORMAT: 1A
 HOST: https://api.usaspending.gov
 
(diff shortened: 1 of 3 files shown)


dporth-frb:
FE will be addressing this as a hotfix on their end.

PR #4009, opened 2024-02-12
[DEV-10588] Implements Unlinked Awards Endpoint

dporth-frb:
**Description:**
The purpose of this PR is to create an endpoint that accepts an agency’s toptier agency id and provides that agency’s unlinked award download files.

**Technical details:**
During development, I found a similar endpoint to this new one, and the way it handles agency id is to accept the value in the request body. As a result, I made the decision to implement this new endpoint in the same manner. One decision enforced by this feature, is that we will only ever have one file returned by this endpoint.

    - [ ] Backend
8. [x] Jira Ticket [DEV-10588](https://federal-spending-transparency.atlassian.net/browse/DEV-10588):

1. We have no established practice for testing bulk download endpoints. That said, I did include unit tests where possible.

3. The nature of this work does not require PR reviews from these groups


collinwr (commented):
This does seem a little different than what the ticket originally called for:
> Given this endpoint, when a user sends a GET request to it and provides a specific toptier agency id, then it returns the latest version of the download zip package for that specific agency

I would have expected the endpoint to return the latest zip file directly instead of a list of urls to the zip files. 

That being said, the implemented solution is closer to our existing patterns in some regard. Our Covid download url, for example, doesn't return the zip file directly, instead returning some metadata (status url, file url, etc), which can be used to download the file.

I do wonder if we need to return a history (multiple files, as you've done below), or just the single latest file? Any value you can see in returning this list?

usaspending_api/download/v2/list_unlinked_awards_downloads.py
```diff
@@ -0,0 +1,106 @@
...
+
+        agency_check = ToptierAgency.objects.filter(toptier_code=toptier_code).values(
+            "toptier_code", "name", "abbreviation"
+        )
+        if not agency_check or len(agency_check.all()) == 0:
+            raise InvalidParameterException(f"Agency not found given the toptier code: {toptier_code}")
+        agency = agency_check[0]
+
+        # Populate regex
+        download_prefix = f"{agency['name'].replace(' ', '_')}_unlinked_awards"
```
collinwr, line 43: This prefix is a little different than what is suggested in DEV-10589's description `Department_of_Veterans_Affairs_UnlinkedAwards_2024-01-29_H17M24S53136993.csv`. 

I think the format in the ticket is a little more consistent with out other download naming conventions (Pascal Case on the type of Download).

usaspending_api/common/helpers/endpoint_documentation.py
```diff
@@ -45,7 +47,7 @@ def _traverse_urls(base, url_patterns):
                 results.append((cleaned, url))
 
     _traverse_urls("/", urls.urlpatterns)
-
+    print(results)
```
collinwr, line 50: Was this left here on purpose?

usaspending_api/common/helpers/endpoint_documentation.py
```diff
@@ -115,6 +118,7 @@ def validate_docs(url, url_object, master_endpoint_list):
         if re.fullmatch(pattern, endpoint):
             break
     else:
-        messages.append("No URL found in {} that matches {} ({})".format(ENDPOINTS_MD, url, qualified_name))
+        if url not in _EXCLUDED_URLS:
+            messages.append("No URL found in {} that matches {} ({})".format(ENDPOINTS_MD, url, qualified_name))
```
collinwr, line 122: Nice idea and work in this file.

usaspending_api/download/v2/list_unlinked_awards_downloads.py
dporth-frb: This format is what the wanted though. This format came out of the call we all had together.

usaspending_api/common/helpers/endpoint_documentation.py
dporth-frb: Thank you, I think it establishes a way forward for future needs too. Like we could have used this when we wanted to remove the agency office autocomplete.

usaspending_api/download/v2/list_unlinked_awards_downloads.py
dporth-frb: Done

usaspending_api/common/helpers/endpoint_documentation.py
dporth-frb: Done

collinwr (approved):

PR #4034, opened 2024-03-05
[DEV-10453] Check for modified awards and transactions before FY2008

aguest-kc:
**Description:**
Adds a new check to the `elasticsearch_indexer` and `elasticsearch_indexer_for_spark` commands when using the `--process-deletes` flag. This checks for transactions/awards that have been modified in the past 3 days and that have an `action_date` before FY2008 If any transactions/awards are found then we attempt to delete those records from Elasticsearch.

This is for a scenario where a transaction/award is create with an `action_date` on or after 2007-10-01, is then added to Elasticsearch and then the transaction/award's `action_date` is later changed to be before 2007-10-01. This causes the transaction/award record in Elasticsearch to contain the old data and not be updated. The only solution to this is a full reindex of the affected index.

**Technical details:**
Adds a new check to the `elasticsearch_indexer` and `elasticsearch_indexer_for_spark` commands when using the `--process-deletes` flag. This checks for transactions/awards that have been modified in the past 3 days and that have an `action_date` before FY2008 If any transactions/awards are found then we attempt to delete those records from Elasticsearch.

    - [ ] Backend
8. [x] Jira Ticket [DEV-10453](https://federal-spending-transparency.atlassian.net/browse/DEV-10453):

```
2. API documentation updated

Matviews are not affected by this change.

```


usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -285,36 +313,45 @@ def delete_awards(
...
-            format_log(
-                f"No related awards found will be deleted. All derived awards are still in the DB.",
-                action="Delete",
-                name=task_id,
-            )
+            format_log(f"Derived {award_keys_len} award keys from transactions in ES", action="Delete", name=task_id)
         )
+        deleted_award_kvs = _check_awards_for_deletes(award_keys, spark)
+        deleted_award_kvs_len = len(deleted_award_kvs)
+        if deleted_award_kvs_len == 0:
```
collinwr, line 327: Might could initialize this to zero before the `if/else` block it's contained, then move it out of the conditional to prevent the nested conditions.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -352,13 +392,26 @@ def delete_transactions(
...
+    pre_fy2008_transactions_len = len(pre_fy2008_transactions)
+
+    if deleted_tx_keys is not None and pre_fy2008_transactions_len > 0:
+        tx_keys_to_delete = list(set([*deleted_tx_keys] + pre_fy2008_transactions))
+    elif deleted_tx_keys is None and pre_fy2008_transactions_len > 0:
+        tx_keys_to_delete = pre_fy2008_transactions
+    elif deleted_tx_keys is not None and pre_fy2008_transactions_len == 0:
+        tx_keys_to_delete = [*deleted_tx_keys]
+    else:
+        tx_keys_to_delete = []
```
collinwr, line 409: This block of conditionals seems clunky to me. Could this be simplified? I know the datastructures are set up a little different (so the same exact approach can't be used), but the `delete_awards()` approach of combining the two potential sets of deleted awards seems much cleaner
```
    values_list = list(
        set(
            [v for d in deleted_award_kvs for v in d.values()]
            + [v for d in updated_awards_pre_fy2008 for v in d.values()]
        )
    )
```

collinwr (approved):

PR #4055, opened 2024-03-25
[PIPE-514]: Prototype Persisting Changes from Staging Tables into Live Table

dporth-frb:
**Description:**
PIPE-513 is adding the ability to write change from the Delta Award Search to two staging tables in Postgres (one for inserts/updates and the other for deletes). This PR of takes these tables and persist them to the live Award Search table in Postgres.

1. [ ] Unit & integration tests updated (IN PROGRESS)
    - [ ] Backend
8. [x] Jira Ticket [PIPE-514](https://federal-spending-transparency.atlassian.net/browse/PIPE-514):


usaspending_api/etl/swap_in_new_table_strategies.py
```diff
@@ -0,0 +1,955 @@
...
+
+            qualified_upsert_postgres_table = f"{temp_schema_name}.{upsert_table_name}"
+            qualified_delete_postgres_table = f"{temp_schema_name}.{delete_table_name}"
+            self._validate_live_tables(cursor, table)
+            self._validate_temp_tables(cursor, temp_table_list)
+
+            self._execute_upserts(cursor, qualified_dest_table, qualified_upsert_postgres_table)
+            self._execute_deletes(cursor, qualified_dest_table, qualified_delete_postgres_table)
+
+            # Begin refreshing views
```
collinwr, line 118: For regular views, no recreation is necessary because a view queries the source data at runtime.

Materialized views can simply be refreshed rather than recreated.

usaspending_api/etl/swap_in_new_table_strategies.py
```diff
@@ -0,0 +1,955 @@
...
+                WHERE  c.relname = {temp_table_name}
+                AND    c.relkind = 'r'
+                );
+            """
+            cursor.execute(validation_query)
+            validation_results = ordered_dictionary_fetcher(cursor)
+            if len(validation_results) == 0 or validation_results[0]["exists"] is False:
+                raise ValueError(f"Swap in new table command cannot proceed because {temp_table_name} does not exist.")
+
+    def _execute_upserts(self, cursor, qualified_dest_table, qualified_upsert_postgres_table):
```
collinwr, line 231: There are a few issues with this method:
- cursor is executing itself rather than the prepared sql
- columns are still hardcoded to `AWARD_SEARCH_COLUMNS` (although this may have been a temporary workaround with the `pyspark` blocker)
- Not sure why you're overwriting the last three columns on the `INSERT` with timestamps and a `NULL`. The last three columns of `AWARD_SEARCH_COLUMNS` are `total_iija_outlay`, `total_iija_obligation`, and `total_outlays`. 
- I believe the `INSERT` statement will need table qualifiers on both the `INSERT` and `VALUES` lists.

usaspending_api/broker/models.py
```diff
@@ -37,3 +37,37 @@ class Meta:
         managed = True
         # Schema is set to `public` in migration `0003_deltatableloadversion`
         db_table = "delta_table_load_version"
+
+
+class LoadTracker(models.Model):
```
collinwr, line 42: I'd encourage you to combine these three tables into one if possible. Do the types and steps need to be tracked in separate tables?

Speaking from experience with `external_data_load_date` and `external_data_type`, I've always wished these two tables were originally created as a single table, so I don't have to join them every time I want to pull meaningful information from them.

Also, since you've taken the step to create a log of some of this history, what do you think about adding some extra info for incremental loads here like the Delta Load version or number of rows inserted/updated?

usaspending_api/etl/swap_in_new_table_strategies.py
```diff
@@ -0,0 +1,955 @@
...
+
+            # We assume both temp tables have the same schema, but let's double check
+            if temp_schema_name != schemas_lookup.get(upsert_table_name):
+                raise ValueError(
+                    f"The temp tables by the names {delete_table_name} and {upsert_table_name} must reside in the same schema. Currently, they do not."
+                )
+
+            qualified_upsert_postgres_table = f"{temp_schema_name}.{upsert_table_name}"
+            qualified_delete_postgres_table = f"{temp_schema_name}.{delete_table_name}"
+            self._validate_live_tables(cursor, table)
```
collinwr, line 112: I noticed something I left out of the AC here. No need to implement it now though, I'll add a new improvement ticket to track it:

We may want to validate that the schemas match between upsert table and live table like we do on the full load approach.

usaspending_api/broker/models.py
dporth-frb: > I'd encourage you to combine these three tables into one if possible. Do the types and steps need to be tracked in separate tables?

The tables are normalized right now. That's why I did what I did. Why should we favor denormalization over normalization for these table?

> Also, since you've taken the step to create a log of some of this history, what do you think about adding some extra info for incremental loads here like the Delta Load version or number of rows inserted/updated?

I made the very intentional decision to not because at this point in our pipeline Delta tables have no relation to anything being done. I believe it's going to better for a code base to not make a process that is unrelated to delta tables have to know about delta tables.

usaspending_api/etl/swap_in_new_table_strategies.py
dporth-frb: addressed

usaspending_api/etl/swap_in_new_table_strategies.py
dporth-frb: > cursor is executing itself rather than the prepared sql

Can you show me an example, my `cursor.execute` code matches other code.

dporth-frb: > columns are still hardcoded to AWARD_SEARCH_COLUMNS (although this may have been a temporary workaround with the pyspark blocker)

Why is this an issue? I used AWARD_SEARCH_COLUMNS variable based off how other parts of our code base do it.

dporth-frb: > Not sure why you're overwriting the last three columns on the INSERT with timestamps and a NULL. The last three columns of AWARD_SEARCH_COLUMNS are total_iija_outlay, total_iija_obligation, and total_outlays.

Again this was based off other existing code in our code base. Let me take a deeper look.

dporth-frb: I've addressed
* I believe the INSERT statement will need table qualifiers on both the INSERT and VALUES lists.
* cursor is executing itself rather than the prepared sql

collinwr: In other parts of the code, we pull things like `AWARD_SEARCH_COLUMNS` out of the `TABLE_SPEC`. In the current state, the `_execute_upeserts()` method is hardcoded to only work for one table

usaspending_api/broker/models.py
dporth-frb: If it's such a big deal, I could create a view on top of these tables that denormalizes it to save us the trouble. That would solve the problem you stated while still allowing us to follow best practices.

usaspending_api/etl/swap_in_new_table_strategies.py
dporth-frb: Oh gotcha, ok

usaspending_api/broker/models.py
collinwr: > The tables are normalized right now. That's why I did what I did. Why should we favor denormalization over normalization for these tables?

I do believe normalization is an important aspect of database design, but in this case I can't see much value added. The tables `load_tracker_type` and `load_tracker_step` only consist of a single field (other than their id fields): `name`. Why are these lists of names important to normalize? To me normalization becomes more important if there are multiple data points that are being tracked in the side table that need to stay consistent with each other. (`address_id`, `address_street`, `address_zip`, etc) or even (`type_code` and `type_name` in the case of `award_category`). But for these single fields, it adds overhead for developers and the database (although database impact is negligible). 

As I mentioned above, the cost of normalization here is going to be that developers are going to have to juggle a few tables every time they want to pull information out of that table (unless they memorize all the ids).

dporth-frb: Here's a solution to your problem of cost to developers. I could create a view on top of these tables that denormalizes it to save us the trouble. That would solve the problem you stated while still allowing us to follow best practices.

collinwr (changes requested):
I did notice that the last version loaded to `last_delta_table_load_version` is no longer being updated by the `swap_in_new_table.py` command. This is crucial for the operation of `load_table_from_delta.py`

usaspending_api/broker/models.py
collinwr: I'm good with that compromise 👍

usaspending_api/etl/swap_in_new_table_strategies.py
dporth-frb: addressed everything in this thread

usaspending_api/broker/models.py
dporth-frb: Done

usaspending_api/etl/management/sql/swap_in_new_table/update_live_tables.py
```diff
@@ -0,0 +1,6 @@
+update_live_tables_sql = """
+UPDATE {dest_table} AS d
```
collinwr, line 2: Missing the `SET` phrase for an `UPDATE` query

usaspending_api/etl/swap_in_new_table_strategies.py
```diff
@@ -0,0 +1,944 @@
...
+                        f"will not be dropped. Swapped-out partitions are not deleted after the swap due to the risk "
+                        f"of their live parent partitioned table still receiving traffic while having one of its "
+                        f"partitions deleted. If '{self.curr_table_parent_partitioned_table}' is also being swapped, "
+                        f"then this partition will be deleted when it is deleted. Otherwise delete "
+                        f"'{self.curr_table_name}{self.old_suffix}' manually."
+                    )
+                else:
+                    self._drop_old_table_sql(cursor)
+
+            # If `delta_table_load_version_key` exists in the table_spec, this table could potentially be loaded incrementally,
```
collinwr, line 371: As this is included in both strategies, I think it would make more sense to be moved into the main command to remove code duplication.

collinwr: (or into the base class)

usaspending_api/etl/swap_in_new_table_strategies.py
```diff
@@ -0,0 +1,944 @@
...
+            mv_s = "MATERIALIZED " if dep_view["is_matview"] else ""
+            cursor.execute(
+                f"DROP {mv_s}VIEW IF EXISTS {self.temp_schema_name}.{dep_view['dep_view_name']}{self.source_suffix};"
+            )
+            cursor.execute(
+                f"CREATE {mv_s}VIEW {self.temp_schema_name}.{dep_view['dep_view_name']}{self.source_suffix} "
+                f"AS ({dep_view_sql});"
+            )
+
+    def _dependent_views(self, cursor, curr_schema_name, curr_table_name):
```
collinwr, line 415: I believe methods like this that are implemented in both strategies should be moved into the base class

usaspending_api/etl/management/sql/swap_in_new_table/update_live_tables.py
```diff
@@ -0,0 +1,6 @@
...
+{set_cols}
+FROM {upsert_temp_table} AS s
+WHERE ON {join_condition};
```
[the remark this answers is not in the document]
dporth-frb: addressed

usaspending_api/etl/management/sql/swap_in_new_table/update_live_tables.py
dporth-frb: addressed

usaspending_api/etl/swap_in_new_table_strategies.py
```diff
@@ -0,0 +1,944 @@
...
+from django.core.management import BaseCommand
+from typing import List
+from usaspending_api.etl.management.helpers.table_specifications import DATABRICKS_GENERATED_TABLE_SPEC as TABLE_SPEC
+from usaspending_api.search.delta_models.award_search import AWARD_SEARCH_COLUMNS
+from usaspending_api.etl.management.sql.swap_in_new_table.update_live_tables import update_live_tables_sql
+from usaspending_api.etl.management.sql.swap_in_new_table.insert_to_live_tables import insert_to_live_tables_sql
+from usaspending_api.etl.management.sql.swap_in_new_table.delete_from_live_tables import delete_from_live_tables_sql
+from usaspending_api.etl.management.sql.swap_in_new_table.dependent_views import detect_dep_view_sql
+
+COLUMN_SPECS = {"award_search": AWARD_SEARCH_COLUMNS}
```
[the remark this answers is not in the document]
dporth-frb: addressed

usaspending_api/etl/swap_in_new_table_strategies.py
dporth-frb: Be warned moving this to a shared class couples these classes together. I made the change you requested

usaspending_api/etl/swap_in_new_table_strategies.py
dporth-frb: Be warned moving this to a shared class couples these classes together. I made the change you requested

usaspending_api/etl/swap_in_new_table_strategies.py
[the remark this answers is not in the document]
collinwr: This is still referencing a new set of data about the table `TABLE_COLUMN_SPECS` and not the already existing `TABLE_SPEC` definition.  I believe everything needed for this class should already be there. `AWARD_SEARCH_COLUMNS` and `unique_identifiers` should cover the needs with some tweaking.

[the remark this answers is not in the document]
dporth-frb: It's an entirely new data structure that stores new data not available. I'm confused why this is a problem?

aguest-kc (approved):

usaspending_api/etl/swap_in_new_table_strategies.py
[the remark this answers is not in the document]
collinwr: I would have preferred to keep this configuration all in one place. I believe `unique_identifiers` would have provided enough info to replace `delete_column`, `null_column`, and `join_condition`. This would make it easier to perform the tech debt of refactoring the TABLE_SPEC info into the Django model files. 

I'm not going to let this hold up the PR review though.

collinwr (approved):

collinwr (changes requested):
Unit tests are failing due to some changes with the Load Tracker

dporth-frb:
Closing PR for a clean one later

PR #4069, opened 2024-04-08
[DEV-9234] Use is_final_balances_for_fy instead of final_of_fy for total_budgetary_resources calculations

aguest-kc:
**Description:**
Update the calculation for the total budgetary resource amount on the `/v2/federal_accounts/` endpoint to use the `is_final_balances_for_fy` column from the submission_attributes table instead of the `final_of_fy` column from the appropriation_account_balances table.

**Technical details:**
Update the calculation for the total budgetary resource amount on the `/v2/federal_accounts/` endpoint to use the `is_final_balances_for_fy` column from the submission_attributes table instead of the `final_of_fy` column from the appropriation_account_balances table.

    - [ ] Backend
8. [x] Jira Ticket [DEV-9234](https://federal-spending-transparency.atlassian.net/browse/DEV-9234):

```
2. API documentation updated

```

### DIFF
--- a/usaspending_api/accounts/tests/integration/test_federal_account_v2.py
+++ b/usaspending_api/accounts/tests/integration/test_federal_account_v2.py
@@ -85,33 +85,64 @@ def fixture_data(db):
     ta3 = baker.make("accounts.TreasuryAppropriationAccount", federal_account=fa3, tas_rendering_label="tas-label-3")
     ta4 = baker.make("accounts.TreasuryAppropriationAccount", federal_account=fa4, tas_rendering_label="tas-label-4")
 
+    dabs100 = baker.make(
+        "submissions.DABSSubmissionWindowSchedule", submission_reveal_date="2017-11-01", submission_fiscal_year=2017
+    )
+    sub101 = baker.make(
+        "submissions.SubmissionAttributes",
+        submission_id="101",
+        reporting_fiscal_year=2017,
+        is_final_balances_for_fy=True,
+        submission_window_id=dabs100.id,
+        reporting_period_start="2017-06-01",
+    )
+    sub102 = baker.make(
+        "submissions.SubmissionAttributes",
+        submission_id="102",
+        reporting_fiscal_year=2017,
+        is_final_balances_for_fy=False,
+        submission_window_id=dabs100.id,
+        reporting_period_start="2017-06-01",
+    )
+    sub103 = baker.make(
+        "submissions.SubmissionAttributes",
+        submission_id="103",
+        reporting_fiscal_year=2017,
+        is_final_balances_for_fy=True,
+        submission_window_id=dabs100.id,
+        reporting_period_start="2017-06-01",
+    )
+
     baker.make(
         "accounts.AppropriationAccountBalances",
         final_of_fy=True,
         treasury_account_identifier=ta0,
         total_budgetary_resources_amount_cpe=1000,
-        submission__reporting_period_start="2017-06-01",
+        # submission__reporting_period_start="2017-06-01",
+        submission=sub101,
     )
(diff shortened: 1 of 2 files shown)


usaspending_api/accounts/views/federal_accounts_v2.py
```diff
@@ -674,7 +688,9 @@ def post(self, request, format=None):
                 account_number=F("federal_account_code"),
                 budgetary_resources=Subquery(
                     AppropriationAccountBalances.objects.filter(
-                        final_of_fy=True,
+                        submission__reporting_fiscal_year=fy,
+                        submission__submission_window__submission_reveal_date__lte=now(),
```
collinwr, line 692: I think this check might be redundant because the `is_final_balances_for_fy` field is only set to true after the submission is revealed. Taking this out could save a join.

usaspending_api/accounts/views/federal_accounts_v2.py
```diff
@@ -674,7 +688,9 @@ def post(self, request, format=None):
                 account_number=F("federal_account_code"),
                 budgetary_resources=Subquery(
                     AppropriationAccountBalances.objects.filter(
-                        final_of_fy=True,
+                        submission__reporting_fiscal_year=fy,
```
collinwr, line 691: I wonder if we need both this and line 694:
```
submission__reporting_period_start__fy=fy,
```
Could this be limited to one or the other?

aguest-kc: I meant to remove one of these. We use `reporting_fiscal_year` on line 508 to get the submissions for the treasury accounts and `reporting_period_start__fy` was already a filter on the AppropriationAccountBalance object before my change.

So is there any strong case for using one over the other? The only real difference I can see is that `reporting_fiscal_year` is always going to be an integer that represents the fiscal year whereas `reporting_period_start__fy` is a date column that we call the `FY()` database function on to get the appropriate fiscal year value.

It seems like using `reporting_fiscal_year` would prevent an unnecessary DB function call, but either seems like it would work.

usaspending_api/accounts/views/federal_accounts_v2.py
aguest-kc: The submission window check was removed.

usaspending_api/accounts/views/federal_accounts_v2.py
aguest-kc: I removed the check on line 694 and kept `submission_reporting_fiscal_year`. If there's a reason to use `reporting_period_start__fy` instead then just let me know and I can change that.

collinwr: This solution sounds good. Thank you!

usaspending_api/accounts/tests/integration/test_federal_account_v2.py
```diff
@@ -85,33 +85,64 @@ def fixture_data(db):
...
         submission__reporting_period_start="2017-03-01",
+        submission=sub102,
     )
     baker.make(
         "accounts.AppropriationAccountBalances",
         final_of_fy=True,
         treasury_account_identifier=ta0,
         total_budgetary_resources_amount_cpe=2000,
-        submission__reporting_period_start="2017-06-01",
+        # submission__reporting_period_start="2017-06-01",
```
collinwr, line 137: Did you mean to leave these commented out (two places)? I think they can be removed.

aguest-kc: I don't even remember commenting these out. I removed the two commented out lines and removed the `submission__reporting_period_start` values from the other AppropriationAccountBalances models since they're not used in the query any more.

collinwr (approved):
Approved! Sorry I forgot to follow up yesterday!

PR #4071, opened 2024-04-09
[DEV-9633] Use latest_transaction_search instead of latest_transaction

aguest-kc:
**Description:**
Use the `latest_transaction_search` field instead of the `latest_transaction` field in spending_filter to reduce the number of joins to the transaction_search table.

**Technical details:**
Use the `latest_transaction_search` field instead of the `latest_transaction` field in spending_filter to reduce the number of joins to the transaction_search table. Previously, this filter was joining to the `transaction_search` table multiples times through different views and this was causing poor performance. This change uses the `transaction_search` table directly to reduce those JOINs to just one.

    - [ ] Backend
8. [x] Jira Ticket [DEV-9633](https://federal-spending-transparency.atlassian.net/browse/DEV-9633):

```
1. Unit & integration tests updated
No new tests are needed for this change since we're still using the same table for the same data, just reducing the number of JOIN statements.

2. API documentation updated

```

### DIFF
--- a/usaspending_api/spending_explorer/v2/filters/spending_filter.py
+++ b/usaspending_api/spending_explorer/v2/filters/spending_filter.py
@@ -1,9 +1,9 @@
 import logging
 
 from django.db.models import Q
-from usaspending_api.references.models import ToptierAgency
-from usaspending_api.common.exceptions import InvalidParameterException
 
+from usaspending_api.common.exceptions import InvalidParameterException
+from usaspending_api.references.models import ToptierAgency
 
 logger = logging.getLogger(__name__)
 
@@ -64,10 +64,7 @@ def spending_filter(alt_set, queryset, filters, _type):
 
             # recipient
             elif key == "recipient":
-                alt_set = alt_set.filter(
-                    Q(award__latest_transaction__assistance_data__awardee_or_recipient_legal=value)
-                    | Q(award__latest_transaction__contract_data__awardee_or_recipient_legal=value)
-                )
+                alt_set = alt_set.filter(Q(award__latest_transaction_search__recipient_name_raw=value))
 
             # award, award_category
             elif key == "award" or key == "award_category":
@@ -106,8 +103,7 @@ def spending_filter(alt_set, queryset, filters, _type):
             elif key == "recipient":
                 queryset = queryset.filter(
                     treasury_account__in=alt_set.filter(
-                        Q(award__latest_transaction__contract_data__awardee_or_recipient_legal=value)
-                        | Q(award__latest_transaction__assistance_data__awardee_or_recipient_legal=value)
+                        Q(award__latest_transaction_search__recipient_name_raw=value)
                     ).values_list("treasury_account_id", flat=True)
                 )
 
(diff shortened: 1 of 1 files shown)


usaspending_api/spending_explorer/v2/filters/spending_filter.py
```diff
@@ -64,10 +64,7 @@ def spending_filter(alt_set, queryset, filters, _type):
 
             # recipient
             elif key == "recipient":
-                alt_set = alt_set.filter(
-                    Q(award__latest_transaction__assistance_data__awardee_or_recipient_legal=value)
-                    | Q(award__latest_transaction__contract_data__awardee_or_recipient_legal=value)
-                )
+                alt_set = alt_set.filter(Q(award__latest_transaction_search__recipient_name_raw=value))
```
dporth-frb, line 67: As your story told you to do, I see you updated to `latest_transaction_search`, but why did you change this query from what it used to do (beyond the `latest_transaction_search` update)?

usaspending_api/spending_explorer/v2/filters/spending_filter.py
```diff
@@ -106,8 +103,7 @@ def spending_filter(alt_set, queryset, filters, _type):
             elif key == "recipient":
                 queryset = queryset.filter(
                     treasury_account__in=alt_set.filter(
-                        Q(award__latest_transaction__contract_data__awardee_or_recipient_legal=value)
-                        | Q(award__latest_transaction__assistance_data__awardee_or_recipient_legal=value)
+                        Q(award__latest_transaction_search__recipient_name_raw=value)
```
dporth-frb, line 106: Same question here.

usaspending_api/spending_explorer/v2/filters/spending_filter.py
aguest-kc: `latest_transaction__assistance_data` and `latest_transaction__contract_data` are both views on top of `transaction_search`. One is a view where `is_fpds = False` and the other view where `is_fpds = True`. Since the `is_fpds` column isn't null-able, there's no reason to join to the `transaction_search` twice like we were before.

collinwr, line 67: I believe the existing columns (`awardee_or_recipient_legal`), on the two tables, are currently mapped to `recipient_name` not `recipient_name_raw`

aguest-kc: It looks like you're correct. I just used `recipient_name_raw` since that's the column that the ticket mentioned. I'll update this query to use recipient name.

This is from the FPDS to Transaction Search mappings:
```jsonc
# transaction_fpds col name : transaction_search col name
"awardee_or_recipient_legal": "recipient_name
```

aguest-kc: Updated the recipient column

collinwr: Thanks Andy! I didn't see that the ticket mentioned the `_raw` field.

collinwr (approved):

PR #4074, opened 2024-04-15
Pipe 518 python 3.12 upgrade

ayubshahab:
**Description:**

**Technical details:**

    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>

```
```


usaspending_api/search/migrations/0041_alter_awardsearch_award_alter_summarystateview_duh.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 5.0.4 on 2024-04-15 19:24
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
collinwr, line 7: Hey Ayub, we purposely avoided creating a FK here in the database. We simulated the relationship in Django here: 
https://github.com/fedspendingtransparency/usaspending-api/blob/qat/usaspending_api/search/migrations/0008_awardsearch_table.py#L116

collinwr: After looking more, I can see that this probably has to do with adding `serlialize=False` to primary keys. Not sure if we need to update this migration to include the change to Award Search in a state migration like in the linked migration

collinwr: After researching a bit, I think we should be in good shape here. The change should only apply to the Django state, not the database schema. 

Can you please double check that this is the case in your environment after applying the migration? (No FK exists from `award_search` to `awards`)

usaspending_api/settings.py
```diff
@@ -313,6 +313,7 @@ def _configure_database_connection(environment_variable, test_options=None):
     config = dj_database_url.parse(os.environ.get(environment_variable), conn_max_age=CONNECTION_MAX_SECONDS)
     config["OPTIONS"] = {**config.setdefault("OPTIONS", {}), **default_options}
     config["TEST"] = {**test_options, "SERIALIZE": False}
+    config["DISABLE_SERVER_SIDE_CURSORS"] = True
```
collinwr, line 316: Why did you make this change? It can have serious performance implications.

usaspending_api/references/migrations/0067_alter_office_office_code.py
```diff
@@ -0,0 +1,18 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("references", "0066_alter_office_unique_index_code_only"),
+    ]
+
+    operations = [
```
collinwr, line 12: We do already have a migration that accomplishes this, but it is done with a SQL script, so I'm not sure if it is registering in the Django state. Can you check if this causes any problems when applying migrations?

usaspending_api/settings.py
ayubshahab: The Disable Server Side Cursor needed to be set to True to fix the DJango Cursor errors we were seeing after updating Django to the newest version. Local testing shows no differences in performance, however I have only tested a handful of apis, but I am not sure if testing in local environment is enough to test performance.

ayubshahab:
Since we decided on updating to python 3.10.12 from python 3.8.16. The feature branch `pipe-518-python-3.12-upgrade` itself can be stashed away until we are ready to update to python 3.12.

PR #4087, opened 2024-05-15
Changes Python version from 3.10.13 to 3.10.12

collinwr:
**Description:**
Aligning Python version with Databricks runtime version 14.3

https://docs.databricks.com/en/release-notes/runtime/14.3lts.html


ayubshahab:
Everything looks good. No additional tests are failing in between python 3.10.13 and 3.10.12. Looks like we can safely downgrade the minor version.

ayubshahab (approved):

PR #4091, opened 2024-05-16
[PIPE-524] Non-Spark Integration test failures

ayubshahab:
**Description:**
This ticket involved updating many failing tests that were erroring out as a result from updating to Python 3.10.12. This included test_load_to_from_delta as well as test_load_rosetta. The major issue was `django.test.testcases.DatabaseOperationForbidden: Database queries to 'data_broker' are not allowed in this test` 

**Technical details:**
Updated python from 3.8.16 to 3.10.12

    - [x] Backend
5. [x] Jira Ticket [PIPE-524](https://federal-spending-transparency.atlassian.net/browse/PIPE-524):

```
```


usaspending_api/transactions/agnostic_transaction_loader.py
```diff
@@ -156,8 +156,10 @@ def generate_ids_from_broker(self):
         with psycopg2.connect(dsn=get_broker_dsn_string()) as connection:
             with connection.cursor("usaspending_data_transfer") as cursor:
                 cursor.execute(sql.strip("\n"))
+                # fetchmany returns a list of tuples that can be iterated via next()
                 while True:
-                    id_list = [id[0] for id in cursor.fetchmany(size=self.chunk_size)]
+                    tuples = cursor.fetchmany(size=self.chunk_size)
+                    id_list = [id[0] for id in tuples]
```
collinwr, line 162: Any reason for this change? Looks to be the same other than the results from the `fetchmany()` call being stored in a temporary variable, `tuples`

.travis.yml
```diff
@@ -155,7 +155,15 @@ jobs:
         name: ws3
         paths:
           - coverage.*.xml
-      # Inherits "global" job phases defined below (e.g. before_install, install, before_script, script, etc.)
+    script:
```
collinwr, line 158: I think it would be better to figure out why there are tests that are now going longer than 10 minutes without output than to change this timeout setting.

If we do determine there are scenarios that will results in these logging gaps, we may want to update the default `script` instead of these overrides.

usaspending_api/transactions/agnostic_transaction_loader.py
ayubshahab: Oh i did that when I was debugging but I can revert this change no problem if you'd like.

.travis.yml
ayubshahab: The main one that was going longer than 10 minutes without output was the load_to_from_delta one. I added the travis_wait in the other one as a precautionary step but it doesn't seem to have any affects now.

ayubshahab: Looking at other builds using 3.8.16 I see that the load_to_from_delta and other select jobs almost always take closer to 20 minutes so I think keeping this setting doesn't do us any harm

collinwr (approved):

PR #4097, opened 2024-05-28
[DEV-11015][Staging] CFDA url update

collinwr:
Staging version of https://github.com/fedspendingtransparency/usaspending-api/pull/4093

### DIFF
--- a/usaspending_api/references/v2/views/cfda.py
+++ b/usaspending_api/references/v2/views/cfda.py
@@ -51,7 +51,7 @@ class CFDAViewSet(APIView):
         while not cdfas and remaining_tries:
             remaining_tries -= 1
             response = post(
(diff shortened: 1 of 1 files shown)


usaspending_api/references/v2/views/cfda.py
```diff
@@ -51,7 +51,7 @@ def get_cfdas(self):
         while not cdfas and remaining_tries:
             remaining_tries -= 1
             response = post(
-                "https://www.grants.gov/grantsws/rest/opportunities/search/cfda/totals",
```
dporth-frb, line 54: Would it be better to make this URL a setting/configuration? Usually data that can change is best stored in settings/configs.

collinwr: Yeah, that's probably a good call. I'm going to close the PR and make the change in the QAT version of this PR

PR #4115, opened 2024-06-18
[DEV-10977] Support Subaward Query Type

dporth-frb:
**Description:**
DEV-10976 added a new Elasticsearch index for Subawards that will be used for searching Subawards in advanced search instead of the current approach that uses Postgres. In order to support this goal, this PR updates the query_with_filter.py classes to support the _QueryType of SUBAWARDS

    - [x] Backend
8. [x] Jira Ticket [DEV-10977](https://federal-spending-transparency.atlassian.net/browse/DEV-10977):


collinwr (changes requested):
Hey Desmond, I started reviewing your updates to `query_with_filters.py` quite a few of the filters need special handling for the new `SUBAWARDS` query type. Here are some examples. I didn't go through the full list of filters though.
1. The `award_type_codes` filter should be using the `prime_award_type` field from the subaward index
2. The `recipient_search_text` and `recipient_id` filters looks like they need some special handling for Subawards
3. The location based filters for subawards should have their own subaward based location fields. (ex `sub_legal_entity_country_code`, `sub_place_of_perform_country_co`, `sub_legal_entity_congressional_current`)
4. The `award_amount` filter should be using `subaward_amount`

I'd recommend checking the `subaward_filter()` function in `sub_award.py` to see how each filter is currently being handled for Subawards then map those to the new Elasticsearch index you created. So far most of the necessary fields seem to be included in the index.

usaspending_api/database_scripts/etl/subaward_es_view.sql
```diff
@@ -2,106 +2,112 @@ DROP VIEW IF EXISTS subaward_es_view;
...
+    s.treasury_account_identifiers,
+    s.sub_ultimate_parent_unique_ide,
+    s.sub_ultimate_parent_uei,
+    a.disaster_emergency_fund_codes
 FROM
-	rpt.subaward_search
-WHERE action_date >= '2007-10-01';
\ No newline at end of file
+	rpt.subaward_search s
+LEFT JOIN rpt.award_search a
```
collinwr, line 111: Shoot! This creates a new dependency, meaning that the Subaward ES indexer can't run until the Award Search Delta Table finishes refreshing. Definitely not an easy way around it though. I think we have two options:
1. Leave this JOIN here and move the Subaward Reindex step back in the pipeline to after the Award Search Delta Table refresh
2. Calculate the DEFCs in the Subaward Search Delta Table refresh. `int.awards` is already being joined to, but that table doesn't have DEFC information. An extra join would have to be added to `financial_accounts_by_awards`

I think option 2 would be the better long term solution, but it would lead to significant scope creep. I'm okay with sticking to option 1 if you can create a ticket for option 2 that we can implement if we need to based on pipeline runtimes.

dporth-frb: I like option 1

dporth-frb:
> Hey Desmond, I started reviewing your updates to `query_with_filters.py` quite a few of the filters need special handling for the new `SUBAWARDS` query type. Here are some examples. I didn't go through the full list of filters though.
> 
> 1. The `award_type_codes` filter should be using the `prime_award_type` field from the subaward index
> 2. The `recipient_search_text` and `recipient_id` filters looks like they need some special handling for Subawards
> 3. The location based filters for subawards should have their own subaward based location fields. (ex `sub_legal_entity_country_code`, `sub_place_of_perform_country_co`, `sub_legal_entity_congressional_current`)
> 4. The `award_amount` filter should be using `subaward_amount`
> 
> I'd recommend checking the `subaward_filter()` function in `sub_award.py` to see how each filter is currently being handled for Subawards then map those to the new Elasticsearch index you created. So far most of the necessary fields seem to be included in the index.

I have pushed changes to address every bullet point in this comment.

usaspending_api/common/query_with_filters.py
```diff
@@ -506,6 +559,23 @@ def _generate_covid_iija_es_queries_transactions(cls, def_code_field, covid_filt
...
+            & ES_Q("range", **{"sub_action_date": {"gte": datetime.strftime(enactment_date, "%Y-%m-%d")}})
+            for def_code, enactment_date in covid_filters.items()
+        ]
+        iija_es_queries = [
+            ES_Q("match", **{def_code_field: def_code})
+            & ES_Q("range", **{"sub_action_date": {"gte": datetime.strftime(enactment_date, "%Y-%m-%d")}})
+            for def_code, enactment_date in iija_filters.items()
+        ]
+        print(covid_es_queries)
+        print(iija_es_queries)
```
collinwr, line 575: Leftover print statements here

usaspending_api/common/query_with_filters.py
```diff
@@ -162,8 +178,12 @@ def generate_elasticsearch_query(cls, filter_values: List[str], query_type: _Que
...
 
         for filter_value in filter_values:
-            award_type_codes_query.append(ES_Q("match", type=filter_value))
-
+            if query_type == _QueryType.SUBAWARDS:
+                type_ = "prime_award_type"
+            else:
+                type_ = "type"
+            award_type_codes_query.append(ES_Q("match", **{type_: filter_value}))
+        print(award_type_codes_query)
```
collinwr, line 186: Print statement left here

collinwr:
> I have pushed changes to address every bullet point in this comment.

Hey @dporth-frb, thank you for fixing these filters. I'll review these changes. 

Did you take a look at the other filters? It seems like there are other incompatibles with Subawards that need to be addressed. I went ahead and looked at the remaining filters, and I think `award_ids` might be the only remaining filter that needs to be updated. Subaward ES index doesn't have `display_award_id` used for the `award_ids` filter.

usaspending_api/common/query_with_filters.py
```diff
@@ -227,33 +247,50 @@ class _RecipientSearchText(_Filter):
...
-            # If the recipient name ends with a period, then add a regex query to find results ending with a
-            #   period and results with a period in the same location but with characters following it.
-            # Example: A query for COMPANY INC. will return both COMPANY INC. and COMPANY INC.XYZ
-            if upper_recipient_string.endswith("."):
-                recipient_search_query.append(recipient_name_query)
-                recipient_search_query.append(
-                    ES_Q({"regexp": {"recipient_name.keyword": f"{upper_recipient_string.rstrip('.')}\\..*"}})
-                )
+        if query_type == _QueryType.SUBAWARDS:
+            fields = ["sub_awardee_or_recipient_uniqu", "sub_awardee_or_recipient_uei"]
```
collinwr, line 252: These DUNS and UEI fields should not be used in the recipient name query.

usaspending_api/common/query_with_filters.py
```diff
@@ -227,33 +247,50 @@ class _RecipientSearchText(_Filter):
...
+                # If the recipient name ends with a period, then add a regex query to find results ending with a
+                #   period and results with a period in the same location but with characters following it.
+                # Example: A query for COMPANY INC. will return both COMPANY INC. and COMPANY INC.XYZ
+                if upper_recipient_string.endswith("."):
+                    recipient_search_query.append(recipient_name_query)
+                    recipient_search_query.append(
+                        ES_Q({"regexp": {"recipient_name.keyword": f"{upper_recipient_string.rstrip('.')}\\..*"}})
+                    )
+                else:
+                    recipient_search_query.append(recipient_name_query)
```
collinwr, line 293: There is quite a bit of duplicated logic other than field names, some of which seems to be missing for the new Subawards implementation (recipient name period logic and some of the extra sanitization logic).

I'd recommend setting these field names in variables in the condition statement, then using the same code block to handle all query types

usaspending_api/common/query_with_filters.py
```diff
@@ -316,7 +359,12 @@ def generate_elasticsearch_query(cls, filter_values: List[dict], query_type: _Qu
                     raise InvalidParameterException(INCOMPATIBLE_DISTRICT_LOCATION_PARAMETERS)
                 if location_value is not None:
                     location_value = location_value.upper()
-                    location_query.append(ES_Q("match", **{f"recipient_location_{location_key}": location_value}))
+                    if query_type == _QueryType.SUBAWARDS:
```
collinwr, line 362: Recipient Location updates look good. Place of Performance Location still needs to be updated though. I think the Sub Place of Performance County Code might be missing from the ES index (I might have missed it though looking for the wrong name), so it may need to be added for this to work.

usaspending_api/common/query_with_filters.py
dporth-frb: What should be used then? I used these because I thought it was following the existing pattern.

usaspending_api/common/query_with_filters.py
dporth-frb: Done

dporth-frb:
> > I have pushed changes to address every bullet point in this comment.
> 
> Hey @dporth-frb, thank you for fixing these filters. I'll review these changes.
> 
> Did you take a look at the other filters? It seems like there are other incompatibles with Subawards that need to be addressed. I went ahead and looked at the remaining filters, and I think `award_ids` might be the only remaining filter that needs to be updated. Subaward ES index doesn't have `display_award_id` used for the `award_ids` filter.

Addressed.

usaspending_api/common/query_with_filters.py
dporth-frb: Addressed.

usaspending_api/common/query_with_filters.py
collinwr: I think `sub_awardee_or_recipient_legal` would be good

collinwr (approved):

PR #4123, opened 2024-07-03
DEV-9224 Fixed spending_by_category endpoint to allow conditional tra…

ayubshahab:
**Description:**
Recieved 404 errors when trying to query the api/v2/search/spending_by_category/recipient/ endpoint. The endpoint functions correctly when querying it without the trailing slash - i.e. api/v2/search/spending_by_category/recipient. A conditional argument regarding the trailing slash is added to the regex. 

**Technical details:**

    - [x] Backend
8. [x] Jira Ticket [DEV-9224](https://federal-spending-transparency.atlassian.net/browse/DEV-9224):

```
```


usaspending_api/search/v2/urls_spending_by_category.py
```diff
@@ -35,7 +35,7 @@
     re_path(r"^funding_subagency", FundingSubagencyViewSet.as_view()),
     re_path(r"^naics", NAICSViewSet.as_view()),
     re_path(r"^psc", PSCViewSet.as_view()),
-    re_path(r"^recipient$", RecipientViewSet.as_view()),
+    re_path(r"^recipient/?$", RecipientViewSet.as_view()),
```
ayubshahab, line 38: Please refer to the comment made on ticket: 
https://federal-spending-transparency.atlassian.net/browse/DEV-9224

Without the `$`, `the re_path(r"^recipient", RecipientViewSet.as_view())` pattern matches any URL that starts with `recipient`, regardless of what follows it. This is because the pattern does not enforce the end of the string with a `$`. `"^recipient"` will match any string that starts with recipient, including `recipientdlasdgjk` for instance. To restrict the URL to match exactly recipient with an optional trailing slash, we need to use a regex that enforces this constraint, which is `/?$`. This should be done to all the urls to ensure the respective APIs only respond to calls that match exactly (with the conditional trailing slash included ofcourse. )

collinwr (approved):

PR #4126, opened 2024-07-09
[DEV-11139] Add UEI to response for spending_by_category/recipient

ayubshahab:
**Description:**
The `usaspending_api/api_contracts/contracts/v2/search/spending_by_category/recipient.md` contract and its respective endpoint includes a uei field for each recipient returned in the results section. If no uei is associated with that recipient, then null is returned. 

**Technical details:**

    - [x] Backend
8. [x] Jira Ticket [DEV-11139](https://federal-spending-transparency.atlassian.net/browse/DEV-11139):

```
```


usaspending_api/api_contracts/contracts/v2/search/spending_by_category/recipient.md
```diff
@@ -92,8 +94,9 @@ This endpoint returns a list of the top results of Recipients sorted by the tota
     The id is the database key.
 + `name` (required, string, nullable)
 + `code` (required, string, nullable)
-    `code` is a user-displayable code (such as a program activity or NAICS code, but **not** a database ID). When no such code is relevant, return a `null`.
+    `code` is a user-displayable code (such as a program activity or DUNS code, but **not** a database ID). When no such code is relevant, return a `null`.
```
collinwr, line 97: Ayub, I have a minor change for this one. I think it would be better to call out that this is actually a `DUNS` code rather than as an example of a potential type of code it could be. Could be something like this:
> The Recipient's DUNS code. If it doesn't have a DUNS code, this field will be `null`

collinwr (approved):

PR #4134, opened 2024-07-21
[DEV-10979]: Migrates Subaward Downloads to Elasticsearch

dporth-frb:
**Description:**
Currently, advanced search is using Elasticsearch to search on Transactions and Prime Awards, but Subawards are still being searched using Postgres, resulting in poor performance (and inconsistency in our codebase). In order to improve performance and consistency, Subawards should be searched using Elasticsearch. This PR is the downloads portion of this effort.

**Technical details:**
Config PR required: https://github.com/fedspendingtransparency/usaspending-config/pull/1620

    - [x] Backend
8. [x] Jira Ticket [DEV-10979](https://federal-spending-transparency.atlassian.net/browse/DEV-10979):


usaspending_api/api_contracts/contracts/v2/download/awards.md
```diff
@@ -139,7 +139,8 @@ This route sends a request to the backend to begin generating a zipfile of award
 + `award_type_codes` (optional, array[string])
 + `agencies` (optional, array[Agency], fixed-type)
 + `contract_pricing_type_codes` (optional, array[string])
-+ `elasticsearch_keyword` (optional, string)
```
dporth-frb, line 142: Renamed this because the previous name was terribly not descriptive. I concluded this change was OK because this filter is only used by our code base (internally). Even if a request comes in with this filter, it won't be used.

usaspending_api/common/query_with_filters.py
```diff
@@ -379,10 +462,15 @@ class _PlaceOfPerformanceScope(_Filter):
 
     @classmethod
     def generate_elasticsearch_query(cls, filter_value: str, query_type: _QueryType, **options) -> ES_Q:
-        # If an ES record has a pop_country_code of "USA" OR if it has any value for the pop_state_code field
-        #   then it's considered domestic. Since we only support domestic states/territories then any
-        #   pop_state_code value means the pop_country_code would be "USA".
-        pop_scope_query = ES_Q("match", pop_country_code="USA") | ES_Q("exists", field="pop_state_code")
+        if query_type == _QueryType.SUBAWARDS:
+            pop_scope_query = ES_Q("match", sub_place_of_perform_country_co="USA") | ES_Q(
+                "match", sub_place_of_perform_country_name="UNITED STATES"
```
aguest-kc, line 467: `sub_place_of_perform_country_co` and `sub_place_of_perform_country_name` aren't the names of the fields in Elasticsearch, they're `sub_pop_country_code` and `sub_pop_country_name`.

It looks like you used their database column names here.

usaspending_api/common/query_with_filters.py
```diff
@@ -310,7 +388,12 @@ class _RecipientScope(_Filter):
 
     @classmethod
     def generate_elasticsearch_query(cls, filter_value: str, query_type: _QueryType, **options) -> ES_Q:
-        recipient_scope_query = ES_Q("match", recipient_location_country_code="USA")
+        if query_type == _QueryType.SUBAWARDS:
+            recipient_scope_query = ES_Q("match", sub_legal_entity_country_code="USA") | ES_Q(
+                "match", sub_legal_entity_country_name="UNITED STATES"
```
aguest-kc, line 393: `sub_legal_entity_country_code` and `sub_legal_entity_country_name` aren't the names of the fields in Elasticsearch, they're `sub_recipient_location_country_code` and `sub_recipient_location_country_name`.

It looks like you used their database column names here.

aguest-kc (changes requested):

usaspending_api/common/query_with_filters.py
```diff
@@ -830,7 +919,11 @@ def generate_awards_elasticsearch_query(cls, filters: dict, **options) -> ES_Q:
 
     @classmethod
     def generate_subawards_elasticsearch_query(cls, filters: dict, **options) -> ES_Q:
-        return cls._generate_elasticsearch_query(filters, _QueryType.SUBAWARDS, **options)
+        cls.filter_lookup[_Keywords.underscore_name] = _SubawardsKeywords
+        cls.filter_lookup[_SubawardsPrimeSubAwardTypes.underscore_name] = _SubawardsPrimeSubAwardTypes
+        resp = cls._generate_elasticsearch_query(filters, _QueryType.SUBAWARDS, **options)
+        cls.filter_lookup[_Keywords.underscore_name] = _Keywords
```
collinwr, line 925: Is it safe to temporarily override this map attribute at the class level? I see you are setting it back to `_Keywords` at the end, but could these situations occur:
- `generate_awards_elasticsearch_query`, `generate_transactions_elasticsearch_query` or `generate_accounts_elasticsearch_query` is called while `generate_subawards_elasticsearch_query` is running
- A failure occurs in `cls._generate_elasticsearch_query(filters, _QueryType.SUBAWARDS, **options)`, which would mean that it never gets reset back? 

On another note, since `cls.filter_lookup[_Keywords.underscore_name] = _SubawardsKeywords`is not being removed from the map at the of the function, would it make sense to go ahead and include it in the main `filter_lookup` definition. 

With these in mind, it might be a good idea to add some checks to these two new types of Subaward-specific filters to make sure QueryType is Subawards

usaspending_api/common/query_with_filters.py
```diff
@@ -120,6 +165,26 @@ def generate_elasticsearch_query(cls, filter_values: List[str], query_type: _Que
         return ES_Q("dis_max", queries=keyword_queries)
 
 
+class _TransactionKeywordSearch(_Filter):
```
collinwr, line 168: Doesn't seem like this is compatible with a `_QueryType` of `transactions`. I think we should either:
- Make this compatible with Transactions
- Make it clear that it's not compatible with Transactions and potentially throw an error if this is attempted like you did with `recipient_id` for `subaward`

Also, I don't think this filter is functional yet because it is being filtered out during request validations. I think adding it to `AWARD_FILTER` in `usaspending-api/usaspending_api/common/validator/award_filter.py` may help with this. Be careful though because Transaction based endpoints use the same object to validate their requests. As far as I know, before this PR all filters in this file were compatible with all `_QueryType`s, with the exception of `subaward` and `award` not supporting `recipient_id`.

usaspending_api/common/query_with_filters.py
dporth-frb: > Is it safe to temporarily override this map attribute at the class level?

Yes I checked this. In what case would generate_awards_elasticsearch_query, generate_transactions_elasticsearch_query or generate_accounts_elasticsearch_query be called while generate_subawards_elasticsearch_query is running? Our application is single threaded. And in jenkins any async calls each have their own python interprettor, so class attributes are not shared.

dporth-frb: > On another note, since cls.filter_lookup[_Keywords.underscore_name] = _SubawardsKeywordsis not being removed from the map at the of the function, would it make sense to go ahead and include it in the main filter_lookup definition.

No doesn't make sense, no reason to.

usaspending_api/common/query_with_filters.py
dporth-frb: The name of the class indicates what it does. It's not the right interpretation to think that because Transaction is in the class name that it's referring to QueryTypes.

This class name is meant to show that a keyword search is being done on transactions.

dporth-frb: I don't like either of those options.

usaspending_api/download/helpers/elasticsearch_download_functions.py
```diff
@@ -190,3 +198,29 @@ def query(cls, filters: dict, download_job: DownloadJob) -> QuerySet:
...
+            default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_MIN_DATE
+        )
+        filter_options["time_period_obj"] = time_period_obj
+        base_queryset = DBSubawardSearch.objects.all()
+        cls._populate_download_lookups(filters, download_job, **filter_options)
+        queryset = base_queryset.extra(
+            tables=["download_job_lookup"],
+            where=[
+                '"download_job_lookup"."download_job_id" = %s '
+                'AND "download_job_lookup"."lookup_id" = "transaction_search"."transaction_id"'
```
collinwr, line 221: Should this be using a Subaward table and ID instead of Transactions?

collinwr (changes requested):
Overall, this looks great! Definitely a complex subject area, so thanks for digging in and figuring this out.

I did leave a few comments about requested changes or suggestions

usaspending_api/common/query_with_filters.py
collinwr: > Yes I checked this. In what case would generate_awards_elasticsearch_query, generate_transactions_elasticsearch_query or generate_accounts_elasticsearch_query be called while generate_subawards_elasticsearch_query is running? Our application is single threaded. And in jenkins any async calls result in two seperate Python processes that have their own python interpreter, so class attributes are not shared.

Sounds good. I wasn't sure about this, so I just wanted to make sure that it was considered

collinwr: > No doesn't make sense, what's the concern here?

I think we should deviate from base cases (in this case defining our filters in `filter_lookup`) as little as possible and only if there is a specific reason. In this case, it's going to permanently end up in `filter_lookup` after `generate_subawards_elasticsearch_query()` is run for the first time, which indicates that it is safe to stay there. In that case, why not follow the base case and define it with the other filters?

usaspending_api/common/query_with_filters.py
collinwr: >The name of the class hints at what it does, that was my goal at least and why I renamed it from it's old, existing name. It's not the right interpretation to think of the name as meaning Transaction from QueryTypes.
This class name is meant to show that a keyword search is being done on transactions.

I think the rename was a good move and works. My point here is that it's breaking the pattern of the other filters to support all `_QueryTypes` without any indication of doing so. In the `recipient_id` filter, you left a comment and threw an exception when trying to use an incompatible `_QueryType`, which I thought was really useful for readers of the code. 
```
if query_type == _QueryType.SUBAWARDS:
    # Subawards did not support "recipient_id" before migrating to elastic search
    # so this behavior is honored here.
    raise InvalidParameterException(
        f"Invalid filter: {_RecipientId.underscore_name} is not supported for subaward queries."
    )
```

collinwr: > What you see is what existed before this story. If it is not supported now it means it never was. Adding support for it is not within the scope of this ticket. FYI this filter is some sort of internal concept, our code made me believe it's not actually a filter we should have in our api contracts.

I was just calling this out because you already did most of the work implementing this filter: renaming it in contracts and some usages across the codebase, implementing the filter function. I thought you wanted to see it go live and might have just missed this one piece of it.

I'm okay if you don't want to make this filter functional. Can you please just remove it from the API contracts then and add a ticket to the tech-debt backlog to decide whether we can remove the filter altogether or not?

dporth-frb: It was already in API contracts. I shouldn't remove it. Doing so would require FE approval and collaboration, right? Is this more scope creep?

No, it's not right for me to make this filter functional because it scope creeps my story. I'm happy to work on an additional story that has AC for it to be functional

usaspending_api/download/helpers/elasticsearch_download_functions.py
dporth-frb: No

usaspending_api/download/lookups.py
```diff
@@ -106,6 +107,16 @@
         "filter_function": subaward_download,
         "annotations_function": subaward_annotations,
     },
+    "elasticsearch_sub_awards": {
```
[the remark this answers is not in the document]
dporth-frb: It is being used.

usaspending_api/common/query_with_filters.py
dporth-frb: Other filters do not need it. That's why. I can make the change, however I need a better reason beyond, it's always been done this way. 

Overall, I'm not too happy with the design pattern of this class anyway.  The way the class is, is not modular and is hard to extend. This subawards work showed me this because we are having to create new filter queries in a hacky way instead of extending and overwriting existing behaviors. That's why I implemented how I did.

usaspending_api/download/lookups.py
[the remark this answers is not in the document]
collinwr: Looking at the codebase, the only other place I see it being used it if a single DEFC is passed to the Disaster Download Validator. Why wasn't this usage also updated to use the new `elasticsearch_sub_awards` mapping type?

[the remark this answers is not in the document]
dporth-frb: Where do you see that? I must have missed it. I was referring to it being used in some of our tests.

usaspending_api/common/query_with_filters.py
dporth-frb: Good catch. Fixed

usaspending_api/common/query_with_filters.py
dporth-frb: Good catch. Fixed

usaspending_api/download/lookups.py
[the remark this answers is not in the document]
collinwr: Check out `DisasterDownloadValidator`. There is a section that adds it as a field in the `download_types` field.

I didn't realize it was being used in downloads, so should be good to keep it in this `lookups.py` file. Would be good to update the `DisasterDownloadValidator` though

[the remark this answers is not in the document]
dporth-frb: Fixed.

usaspending_api/download/helpers/elasticsearch_download_functions.py
collinwr: I took a deeper dive look at this file and set of classes, and I want to push back on this a bit. 

Take a look at the logic of `_populate_download_lookups()`. 

One of the first things it does is create an Elasticsearch search object based on the `_search_type()` then applies filters and requests ids based on the `_source_field`, which is `broker_subaward_id` in this case.

The `_get_download_ids_generator()` executes the search and returns the list of ids. This methods' signature needs to be updated to allow for `SubawardSearch` as a search type. Those ids are stored in the `download_job_lookup` table. 

So going back to my original feedback, if `broker_subaward_id` is the field being stored in `lookup_id`, why would we be joining it to `transaction_id`  on the `transaction_search` table when querying data from `DBSubawardSearch`?

dporth-frb: This was copied from subawards filter, as it originally was. It shouldn't be using subaward search because that's not how it was originally. I can see that lookup_id might need to be changed. But that's it, let me look into it.

dporth-frb: Ok fixed.

usaspending_api/common/query_with_filters.py
collinwr: If you feel it's important queries shouldn't have access to it, then it should be removed from `cls.filter_lookup` after `cls._generate_elasticsearch_query` is run, similar to how the keywords filter is reset. 

I'm okay with whatever you decide on the items in this thread. Just trying to provide some feedback. The solid request I'd like to make is to please add comments to these new filters, including `_TransactionKeywordSearch` explaining what `_QueryTypes` they each are or are not compatible with.

usaspending_api/common/query_with_filters.py
collinwr: I see your point of view about the API contracts. 

The only thing that was not addressed was my feedback around making it clear that `_TransactionKeywordSearch` is not compatible with the transaction `_QueryType`. I'm okay if you don't want to enforce this in the code, but please leave a comment on that filter calling out what this filter does (which is not immediately obvious) and that it doesn't support the transaction `_QueryType`.

usaspending_api/common/query_with_filters.py
collinwr: Thank you!

usaspending_api/download/lookups.py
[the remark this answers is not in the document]
collinwr: Thanks, looks good

usaspending_api/download/helpers/elasticsearch_download_functions.py
collinwr: Main change looks good, but `_get_download_ids_generator()` method signature still needs to be updated to include `SubawardSearch`

collinwr: Thank you!

usaspending_api/common/query_with_filters.py
collinwr: This is compatible with the award query type as well.

aguest-kc (dismissed):

aguest-kc (approved):

collinwr (approved):
Thanks for following up on that comment! Approved!

PR #4135, opened 2024-07-23
[DEV-11100] Add DUNS to recipient profile index in Elastic-search

ayubshahab:
**Description:**
To support Advanced Search 2.0 we need to support searching for recipients by their DUNS value. Currently only Recipient Name, UEI, and Recipient Level are supported by Elasticsearch, so DUNS was also added. 

**Technical details:**

    - [x] Backend
4. [x] Jira Ticket [DEV-11100](https://federal-spending-transparency.atlassian.net/browse/DEV-11100):

```
```

### DIFF
--- a/usaspending_api/database_scripts/etl/recipient_profile_delta_view.sql
+++ b/usaspending_api/database_scripts/etl/recipient_profile_delta_view.sql
@@ -2,11 +2,15 @@ DROP VIEW IF EXISTS recipient_profile_delta_view;
 
 CREATE VIEW recipient_profile_delta_view AS
 SELECT
-  "id",
-  "recipient_hash",
-  "recipient_name",
-  "uei",
-  "recipient_level",
-  "duns"
+  rp.id,
+  rp.recipient_hash,
+  rp.recipient_name,
+  rp.uei,
+  rp.recipient_level,
+  rl.duns
 FROM
-  "recipient_profile"
+  recipient_profile rp
+JOIN
+  recipient_lookup rl
+ON
+  rp.recipient_hash = rl.recipient_hash;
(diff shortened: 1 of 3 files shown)


usaspending_api/database_scripts/etl/recipient_profile_delta_view.sql
```diff
@@ -2,10 +2,15 @@ DROP VIEW IF EXISTS recipient_profile_delta_view;
...
-  "recipient_hash",
-  "recipient_name",
-  "uei",
-  "recipient_level"
+  rp.id,
+  rp.recipient_hash,
+  rp.recipient_name,
+  rp.uei,
+  rp.recipient_level,
+  rl.duns
```
collinwr, line 10: I think you should be able to get this from the `recipient_unique_id` field on `recipient_profile`, so you don't need to join to `recipient_lookup`. Might want to trace that back through in the code to double check though

aguest-kc: It looks like `recipient_unique_id` is the same as DUNS, for the recipients with a DUNS. In the views that populate the recipient_profile table we `SELECT duns AS recipient_unique_id`.

Out of curiosity, any reason this field is named `recipient_unique_id` instead of `duns` if it's actually DUNS? `recipient_unique_id` is really vague.

collinwr: I'm not completely sure, but I think this might be from before the migration from DUNS To UEI. Because we only had one unique identifier (DUNS), we just called it `recipient_unique_id`. Not that we have two, this is pretty confusing. 

@ayubshahab could you create a ticket to rename usages of `recipient_unique_id` to `duns` and add it to the tech debt backlog?

ayubshahab: I have implemented the necessary changes :)

collinwr (approved):

PR #4137, opened 2024-07-24
[Dev-11101] add duns recipient autocomplete

ayubshahab:
**Description:**
To support recipient autocomplete in Advanced Search 2.0 we need to add the ability to search for recipients by their DUNS value. Currently only Recipient Name, UEI, and Recipient Level are supported by API endpoint using Elasticsearch so the `/api/v2/autocomplete/recipient` is updated to allow duns to be searched as well as be returned in the response object.

**Technical details:**

    - [x] Backend
8. [x] Jira Ticket [DEV-11101](https://federal-spending-transparency.atlassian.net/browse/DEV-11101):

```
```


usaspending_api/references/v2/views/recipients.py
```diff
@@ -68,22 +76,25 @@ def _prepare_search_terms(self, request_data: Dict[str, Union[str, List[str]]])
...
+    def _create_es_search(
+        self, search_text: str, recipient_levels: List[str], duns: str, limit: int
+    ) -> RecipientSearch:
         """
         Create an Elasticsearch search query for recipient autocomplete.
 
         Args:
             search_text: The search text entered by the user.
             recipient_levels: The list of recipient levels to filter by.
+            duns: Any specific duns key specified by the user.
```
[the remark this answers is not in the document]
ayubshahab: I wasn't sure if that was needed to be repeated here since recipient_levels is also optional but the comment above doesn't exactly outline that. I suppose I can update the description for both.

usaspending_api/references/v2/views/recipients.py
```diff
@@ -106,6 +117,10 @@ def _create_es_search(self, search_text: str, recipient_levels: List[str], limit
             # if there are recipient levels, then any of the options from the recipient levels as well as the search text should match  # noqa: E501
             query = ES_Q("bool", must=[ES_Q("bool", should=recipient_should_clause), ES_Q("bool", should=query)])
 
+        if duns:
+            # Add a must clause for the duns field if it is provided
```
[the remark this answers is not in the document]
ayubshahab: Gotcha, I will take it out completely as it honestly seems unnecessary.

usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
```diff
@@ -7,7 +7,7 @@ This endpoint is used by the Recipient autocomplete filter on the Advanced Searc
 
 ## POST
 
-This route sends a request to the backend to retrieve recipients matching the specified search text. The endpoint will search for `search_text` in the `recipient_name` and `uei` fields. The `recipient_levels` value can also be included to further filter results, but is not required.
+This route sends a request to the backend to retrieve recipients matching the specified search text. The endpoint will search for `search_text` in the `recipient_name` and `uei` fields. The `recipient_levels` value can also be included to further filter results, but is not required. The endpoint also supports further querying the data based on the `duns` field but it also not required.
```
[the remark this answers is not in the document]
ayubshahab: Will do.

usaspending_api/references/v2/views/recipients.py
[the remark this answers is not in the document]
ayubshahab: The str | None optional parameters syntax is only supported in python 3.10 and above. So unfortunately do not have the capacity to do that right now. I have left the type as None and left a comment in the code explaining that it is by default null unless a str value is passed in. I believe that is sufficient.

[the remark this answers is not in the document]
ayubshahab: Hmm no i think I'll just leave it as type string and remove the comment. Reading the api contract a dev should be able to clearly see what that the duns parameter is optional

usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
```diff
@@ -32,6 +32,10 @@ This route sends a request to the backend to retrieve recipients matching the sp
                             "type": "string"
                         },
                         "description": "An array of recipient levels to filter results. E.g., ['P', 'R']"
+                    },
+                    "duns":{
+                        "type": "string",
+                        "description": "The specific duns value to filter by. E.g. '001013197' "
```
collinwr, line 38: `duns` should not be it's own new field in the request body. The existing `search_text` field should be used to search for `duns` in addition to it's existing search fields `uei` and `recipient_name`

collinwr (changes requested):
Hey Ayub, I wanted to provide some context around this. The frontend has a single search box where a user can enter text for recipients. That's why all these types of filters are routed through the same `search_text` field. Sorry I didn't help clarify this earlier! I should have caught it.

usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
```diff
@@ -60,16 +60,19 @@ This route sends a request to the backend to retrieve recipients matching the sp
                         "recipient_name": "ABC Holdings Inc.",
                         "recipient_level": "P",
                         "uei": "ABCDEF12345"
+                        "duns": null
```
dpb-bah, line 63: For one of these, could you include an example DUNS? Like `123456789'

usaspending_api/references/v2/views/recipients.py
```diff
@@ -151,6 +154,7 @@ def _parse_elasticsearch_response(self, hits: Dict[str, Any]) -> List[Dict[str,
                         ("recipient_name", recipient["recipient_name"]),
                         ("uei", recipient["uei"]),
                         ("recipient_level", recipient["recipient_level"]),
+                        ("duns", recipient["duns"] if "duns" in recipient else None),
```
dpb-bah, line 157: Is `recipient` a dict? If so you can probably use the `.get()` syntax to spare the `if else` check. If `recipient` is something special with ES though you can disregard.

dpb-bah (changes requested):

usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
ayubshahab: Yes I have updated the response with proper examples from our data.

usaspending_api/references/v2/views/recipients.py
ayubshahab: there is no .get() function but I can access the variable directly such as `recipient.duns` which i guess works the same

ayubshahab: Nope, i was mistaken, it needs to be `recipient["duns"] if "duns" in recipient else None`

dpb-bah (approved):

PR #4160, opened 2024-08-13
[DEV-11207] updated summary state view model

ayubshahab:
**Description:**
In order to support the new changes to the State Profile Pages, the summary_state_view matview in Postgres is updated to include an additional field total_outlays. The `total_outlays` is added to `summary_state_view` delta model as well as the postgres django model. 

**Technical details:**

    - [x] Backend
8. [x] Jira Ticket [DEV-11207](https://federal-spending-transparency.atlassian.net/browse/DEV-11207):

```
```


aguest-kc:
@ayubshahab if PR #4163 gets merged in before this PR you'll need to rename your migration to be 0042, because Ron's PR also has a migration named 0041 within the **search** app.

usaspending_api/common/matview_manager.py
```diff
@@ -32,14 +32,6 @@
...
             },
         ),
-        (
-            "summary_state_view",
-            {
-                "model": mv.SummaryStateView,
-                "json_filepath": str(JSON_DIR / "summary_state_view.json"),
-                "sql_filename": "summary_state_view.sql",
-            },
-        ),
```
aguest-kc, line 42: @collinwr will this require a change to the nightly pipeline config? I know we have the SummaryStateView matview files being generated as part of the pipeline, but it doesn't look like we're using those .sql files. I don't think this is going to break anything in the pipeline but just wanted a second set of eyes to confirm that.

dpb-bah (approved):

usaspending_api/common/matview_manager.py
aguest-kc: As a note, a couple of files that should be cleaned up in the config repo that reference the summary_state_view matview:
- matview_runner.groovy
  - `MATVIEW_LIST`
- parameterized_nightly_pipeline_docker_shared.groovy
  - `OPTIONAL_MATVIEWS`

collinwr: I believe `summary_state_view.json` can also be removed if we are removing traces of this being a matview.

usaspending_api/transactions/delta_models/summary_state_view.py
```diff
@@ -87,17 +88,37 @@
...
+        COUNT(*) AS counts,
+        CAST(
+            COALESCE(
+                SUM(matching_awards.total_outlays),
+                0
+            ) AS NUMERIC(23, 2)
+        ) AS total_outlays
     FROM
         int.transaction_normalized
+    JOIN
```
collinwr, line 100: Is this join being used? Looks like the reference to `as2` below is a completely separate instance of Award Search

usaspending_api/transactions/delta_models/summary_state_view.py
```diff
@@ -87,17 +88,37 @@
...
         int.transaction_fpds ON (transaction_normalized.id = transaction_fpds.transaction_id)
     LEFT OUTER JOIN
         int.transaction_fabs ON (transaction_normalized.id = transaction_fabs.transaction_id)
+    LEFT JOIN
+        (
+            SELECT
+                as2.award_id,
+                as2.total_outlays
+            FROM
+                rpt.award_search as2
```
collinwr, line 112: If this is referencing Award Search, then the `load_query_to_delta-summary_state_view` job in the pipeline needs to be moved to after Award Search is refreshed with `load_query_to_delta-award_search`. This will require a separate PR to the `usaspending-config` repo

usaspending_api/transactions/delta_models/summary_state_view.py
```diff
@@ -87,17 +88,37 @@
                 0
             ) AS NUMERIC(23, 2)
         ) AS face_value_loan_guarantee,
-        COUNT(*) AS counts
+        COUNT(*) AS counts,
+        CAST(
+            COALESCE(
+                SUM(matching_awards.total_outlays),
```
collinwr, line 94: This approach of summing the total outlays for matching awards will result in lot's of duplicate funds, and will not work. Many transactions are associated with the same awards, which means the funds associated with an award will be included in this sum for each associated transaction. 

We'll need to either dedupe the awards here or restructure the query to be based on Awards instead of Transactions.

usaspending_api/etl/tests/integration/test_load_to_from_delta.py
```diff
@@ -831,14 +847,37 @@ def test_load_table_to_delta_for_sam_recipient(spark, s3_unittest_data_bucket, p
...
     ]
     create_and_load_all_delta_tables(spark, s3_unittest_data_bucket, tables_to_load)
+    verify_delta_table_loaded_to_delta(
+        spark, "award_search", s3_unittest_data_bucket, load_command="load_query_to_delta"
+    )
+
+    # We now want to load the award_search table that we created above along with other tables needed to create award_search
+    # Then create the summay_state_view table and populate it using the load_query_to_delta command
+    tables_to_load = ["transaction_fabs", "transaction_fpds", "transaction_normalized", "award_search"]
+    create_and_load_all_delta_tables(spark, s3_unittest_data_bucket, tables_to_load)
```
collinwr, line 878: It seems like there is some duplication going on here as far as what tables are created and loaded in spark. Can this be trimmed down at all?

collinwr: Also, we may be able to use the `load_dict_to_delta_table()` function in `spark_helpers.py` to mock out `award_search` in Spark instead of creating all those dependent tables. Each of those tables will require 2 extra spark jobs to be run (one for creation, one for population), which are pretty time consuming. 

I don't want this to hold up the PR because the PR is urgent, but this is something to think about it if it can be done quickly. You'd only need to mock out the fields required for the Summary State View query.

usaspending_api/transactions/delta_models/summary_state_view.py
ayubshahab: It wasn't being used. I have taken that line out.

usaspending_api/transactions/delta_models/summary_state_view.py
ayubshahab: After a meeting discussion, we decided to move forward with the approach of running multiple queries where we will let the `summary_state_view` table get created and populated as usual, with the exception of creating the `total_outlays` column, but not calculating the values. Then in a subsequent query, we will use the `distinct_awards` from the `summary_state_view` table and map those to the `award_search` table to generate the total_outlays.

usaspending_api/transactions/delta_models/summary_state_view.py
```diff
@@ -87,7 +90,8 @@
                 0
             ) AS NUMERIC(23, 2)
         ) AS face_value_loan_guarantee,
-        COUNT(*) AS counts
+        COUNT(*) AS counts,
+        0 AS total_outlays  -- Default value for new column
```
collinwr, line 94: Might make sense to make this a `NULL` instead of `0`. We want to make sure that we can distinguish between the case that a state actually has $0 in funds vs. if this field isn't populated by the subsequent queries for some reason that we can d

ayubshahab: I have made the correction.

usaspending_api/etl/tests/integration/test_load_to_from_delta.py
ayubshahab: I will create a tech debt ticket for this and link it here.

collinwr (approved):

PR #4190, opened 2024-09-10
[DEV-11452] Update DEFC logic to filter with existing defc filter

ronaldongbah:
**Description:**
Continuing on DEV-11452, the Spending by DEFC endpoint needed additional changes that included logic to ensure that the defc are filtering properly whether or not if the user was searching by transactions or subawards.

**Technical details:**
A hardcoded filter was added under the `post` method in `spending_by_category.py` that passes the `def_codes` filter with all of the DEFCs.

    - [ ] Backend
2. [x] Jira Ticket [DEV-11452](https://federal-spending-transparency.atlassian.net/browse/DEV-11452):

```
2. API documentation updated
No API documentations needed to be updated.

4. Matview impact assessment completed.
No matviews were affected by this change.

6. Data validation completed
No data validation is needed by this change.

8. Appropriate Operations ticket(s) created
No operation tickets needed for these changes.

```

### DIFF
--- a/usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
+++ b/usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
@@ -25,6 +25,7 @@ from usaspending_api.common.query_with_filters import QueryWithFilters
 from usaspending_api.common.validator.award_filter import AWARD_FILTER
 from usaspending_api.common.validator.pagination import PAGINATION
 from usaspending_api.common.validator.tinyshield import TinyShield
+from usaspending_api.references.models import DisasterEmergencyFundCode
 from usaspending_api.search.v2.elasticsearch_helper import (
     get_number_of_unique_terms_for_transactions,
     get_scaled_sum_aggregations,
@@ -65,6 +66,18 @@ class AbstractSpendingByCategoryViewSet(APIView, metaclass=ABCMeta):
 
         original_filters = request.data.get("filters")
         validated_payload = TinyShield(models).block(request.data)
+
+        # Custom logic to support Subawards search for DEFC
+        # Check if it's the Spending by DEFC endpoint
+        if self.category.name == "defc":
+            # Get list of defc
+            def_codes = list(DisasterEmergencyFundCode.objects.values_list("code", flat=True))
+            # Add the filter
+            if original_filters is None:
+                validated_payload["filters"] = {}
+                validated_payload["filters"]["def_codes"] = def_codes
+            else:
+                validated_payload["def_codes"] = def_codes
 
         raw_response = self.perform_search(validated_payload, original_filters)
 
--- a/usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
+++ b/usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
@@ -67,17 +67,13 @@ class AbstractSpendingByCategoryViewSet(APIView, metaclass=ABCMeta):
         original_filters = request.data.get("filters")
         validated_payload = TinyShield(models).block(request.data)
 
-        # Custom logic to support Subawards search for DEFC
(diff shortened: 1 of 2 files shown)


usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
```diff
@@ -66,6 +67,18 @@ def post(self, request: Request) -> Response:
         original_filters = request.data.get("filters")
         validated_payload = TinyShield(models).block(request.data)
 
+        # Custom logic to support Subawards search for DEFC
+        # Check if it's the Spending by DEFC endpoint
+        if self.category.name == "defc":
```
collinwr, line 72: Here instead of checking if the whole set of original filters is None, we should be checking to see if "def_codes" exists as a filter on validated_payload.

usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
```diff
@@ -66,6 +67,18 @@ def post(self, request: Request) -> Response:
         original_filters = request.data.get("filters")
         validated_payload = TinyShield(models).block(request.data)
 
+        # Custom logic to support Subawards search for DEFC
+        # Check if it's the Spending by DEFC endpoint
```
collinwr, line 71: This logic applies to both Subawards and Transactions. See the follow following new language. 

Because Subawards and Transactions are associated with DEFCs via their Prime Award, which contains a list of all DEFCs from its associated File C records, we need to limit results to records with a `sub_action_date` and `action_date`, respectively, earlier than the enactment date of the public law for each DEFC. This provides a more accurate breakdown of funds by DEFC. This can be accomplished by simply adding an extra filter for all DEFCs because DEFC filtering logic already takes this into account in `query_with_filters.py`. Because we are grouping by DEFC, it is safe to filter down to only records that include one.

collinwr (changes requested):

usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
```diff
@@ -66,6 +67,14 @@ def post(self, request: Request) -> Response:
         original_filters = request.data.get("filters")
         validated_payload = TinyShield(models).block(request.data)
 
+        # Because Subawards and Transactions are associated with DEFCs via their Prime Award, which contains a list of all DEFCs from its associated File C records, we need to limit results to records with a sub_action_date and action_date, respectively, earlier than the enactment date of the public law for each DEFC. This provides a more accurate breakdown of funds by DEFC. This can be accomplished by simply adding an extra filter for all DEFCs because DEFC filtering logic already takes this into account in query_with_filters.py. Because we are grouping by DEFC, it is safe to filter down to only records that include one.
+        # Check if it's the Spending by DEFC endpoint
+        if self.category.name == "defc":
+            # Get the list of DEFCs and add the filter if it doesn't already exist
+            if "def_codes" not in validated_payload["filters"]:
```
collinwr, line 74: I think that we need to double check that `filters` exists as an element in `validated_payload` as well. Sorry I didn't call it out earlier, but I just tested this, and an error will be thrown if someone doesn't provide any filters.

usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
```diff
@@ -66,6 +67,14 @@ def post(self, request: Request) -> Response:
         original_filters = request.data.get("filters")
         validated_payload = TinyShield(models).block(request.data)
 
+        # Because Subawards and Transactions are associated with DEFCs via their Prime Award, which contains a list of all DEFCs from its associated File C records, we need to limit results to records with a sub_action_date and action_date, respectively, earlier than the enactment date of the public law for each DEFC. This provides a more accurate breakdown of funds by DEFC. This can be accomplished by simply adding an extra filter for all DEFCs because DEFC filtering logic already takes this into account in query_with_filters.py. Because we are grouping by DEFC, it is safe to filter down to only records that include one.
```
collinwr, line 70: Could you please split this comment up across a few different lines? Around column 100-110 could be a good cutoff point to make sure that the entire comment can be read in an editor without the need to scroll to the right. I'll take an action item to make sure this is in our style guide.

usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
ronaldongbah: No worries. Added back a similar check

PR #4202, opened 2024-09-30
[DEV-11402] Update download count endpoint

ayubshahab:
**Description:**
The changes made in this ticket makes it so that calling the `/api/v2/download/count` endpoint will return the number of records that would be included in a download based on the current spending_level filter (Awards, Transactions, or Subawards), which has a default of transactions. This would allow users to download 3 Awards that have more than 500,000 Transactions, because the endpoint would return 3 in this example regardless of how many Transactions make up those 3 Awards.

**Technical details:**

    - [x] Backend
8. [x] Jira Ticket [DEV-11402](https://federal-spending-transparency.atlassian.net/browse/DEV-11402):

```
```

### DIFF
--- a/usaspending_api/api_contracts/contracts/v2/download/count.md
+++ b/usaspending_api/api_contracts/contracts/v2/download/count.md
@@ -24,7 +24,7 @@ Returns the number of transactions that would be included in a download request 
                 + `subawards`
                 + `transactions`
                 + `awards`
-                + Default: `transactions`
+            + Default: `transactions`
     + Body
 
 
--- a/usaspending_api/api_contracts/contracts/v2/download/count.md
+++ b/usaspending_api/api_contracts/contracts/v2/download/count.md
@@ -94,7 +94,7 @@ Returns the number of transactions that would be included in a download request 
 + `extent_competed_type_codes`: `A` (optional, array[string])
 + `tas_codes` (optional, array[TASCodeObject], fixed-type)
 + `treasury_account_components` (optional, array[TreasuryAccountComponentsObject], fixed-type)
-+ `spending_level` (required, array[string])
++ `spending_level` (optional, array[string])
     Possible values in this list are either awards, transaction, or subawards
 
 ## TimePeriodObject (object)
--- a/usaspending_api/api_contracts/contracts/v2/download/count.md
+++ b/usaspending_api/api_contracts/contracts/v2/download/count.md
@@ -19,7 +19,11 @@ Returns the number of transactions that would be included in a download request 
 
     + Attributes (object)
         + `filters` (required, AdvancedFilterObject)
-        + `spending_level` (optional, SpendingLevelObject)
+        + `spending_level` (optional, enum[string])
+            + Members
+                + `subawards`
+                + `transactions`
+                + `awards`
     + Body
 
 
@@ -39,21 +43,22 @@ Returns the number of transactions that would be included in a download request 
                     "recipient_type_names": ["higher_education"],
                     "place_of_performance_locations": [{"country": "USA", "state": "WI"}],
                     "psc_codes": {"require": [["Product","10","1035"]]}
-                }
+                },
+                "spending_level": "awards"
             }
 
 + Response 200 (application/json)
     + Attributes (object)
-        + `transaction_rows_gt_limit` (required, boolean)
-            A boolean returning whether the transaction count is over the maximum row limit.
         + `calculated_transaction_count` (required, number)
             The calculated count of all transactions which would be included in the download files.
         + `maximum_transaction_limit` (required, number)
             The current allowed maximum number of transactions in a row-limited download. Visit https://www.usaspending.gov/download_center/custom_award_data to download larger volumes of data.
-        + `spending_level` (required, string)
-            The spending_level provided by the user or the default value of transactions.
+        + `transaction_rows_gt_limit` (required, boolean)
+            A boolean returning whether the transaction count is over the maximum row limit.
         + `calculated_count` (required, number)
             The calculated count of Awards, Transactions, or Subawards included in the download files.
+        + `spending_level` (required, string)
+            The spending_level provided by the user or the default value of transactions.
         + `maximum_limit` (required, number)
             The current allowed maximum number of Awards, Transactions, or Subawards in a row-limited download. Visit https://www.usaspending.gov/download_center/custom_award_data to download larger volumes of data.
         + `rows_gt_limit` (required, boolean)
@@ -103,13 +108,6 @@ Returns the number of transactions that would be included in a download request 
 + `extent_competed_type_codes`: `A` (optional, array[string])
 + `tas_codes` (optional, array[TASCodeObject], fixed-type)
 + `treasury_account_components` (optional, array[TreasuryAccountComponentsObject], fixed-type)
-
-## SpendingLevelObject (object)
-+ `spending_level` (optional, enum[string])
-    + Members
-        + `subawards`
-        + `transactions`
-        + `awards`
 
 ## TimePeriodObject (object)
 See the Transaction Search category defined in [TransactionSearchTimePeriodObject](../../../search_filters.md#transaction-search-time-period-object)
--- a/usaspending_api/api_contracts/contracts/v2/download/count.md
+++ b/usaspending_api/api_contracts/contracts/v2/download/count.md
@@ -94,6 +94,8 @@ Returns the number of transactions that would be included in a download request 
 + `extent_competed_type_codes`: `A` (optional, array[string])
 + `tas_codes` (optional, array[TASCodeObject], fixed-type)
 + `treasury_account_components` (optional, array[TreasuryAccountComponentsObject], fixed-type)
++ `spending_level` (required, array[string])
+    Possible values in this list are either awards, transaction, or subawards
 
 ## TimePeriodObject (object)
 See the Transaction Search category defined in [TransactionSearchTimePeriodObject](../../../search_filters.md#transaction-search-time-period-object)
--- a/usaspending_api/api_contracts/contracts/v2/download/count.md
+++ b/usaspending_api/api_contracts/contracts/v2/download/count.md
@@ -24,6 +24,7 @@ Returns the number of transactions that would be included in a download request 
                 + `subawards`
                 + `transactions`
                 + `awards`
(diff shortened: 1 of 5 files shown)


usaspending_api/download/v2/download_count.py
```diff
@@ -52,15 +64,33 @@ def post(self, request):
...
+        else:
+            options = {}
+            time_period_obj = AwardSearchTimePeriod(
+                default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_SEARCH_MIN_DATE
+            )
+            new_awards_only_decorator = NewAwardsOnlyTimePeriod(
+                time_period_obj=time_period_obj, query_type=_QueryType.TRANSACTIONS
+            )
+            options["time_period_obj"] = new_awards_only_decorator
+            filter_query = QueryWithFilters.generate_transactions_elasticsearch_query(filters, **options)
```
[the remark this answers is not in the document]
ayubshahab: Yes! Good catch, I forgot to update that to awards haha. Fixed it now :)

usaspending_api/download/v2/download_count.py
```diff
@@ -52,15 +64,33 @@ def post(self, request):
...
             search = TransactionSearch().filter(filter_query)
-            total_count = search.handle_count()
+            total_transaction_count = search.handle_count()
+        else:
+            options = {}
+            time_period_obj = AwardSearchTimePeriod(
+                default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_SEARCH_MIN_DATE
+            )
+            new_awards_only_decorator = NewAwardsOnlyTimePeriod(
+                time_period_obj=time_period_obj, query_type=_QueryType.TRANSACTIONS
```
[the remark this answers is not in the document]
ayubshahab: Yes! Good catch, I forgot to update that to awards haha. Fixed it now :)

usaspending_api/api_contracts/contracts/v2/download/count.md
```diff
@@ -94,6 +94,8 @@ Returns the number of transactions that would be included in a download request
 + `extent_competed_type_codes`: `A` (optional, array[string])
 + `tas_codes` (optional, array[TASCodeObject], fixed-type)
 + `treasury_account_components` (optional, array[TreasuryAccountComponentsObject], fixed-type)
++ `spending_level` (required, array[string])
```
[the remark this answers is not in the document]
ayubshahab: I updated the type to be enum[string] and have removed it from the filters object. It has its own object now.

usaspending_api/download/v2/download_count.py
```diff
@@ -52,15 +64,33 @@ def post(self, request):
...
 
         result = {
-            "calculated_transaction_count": total_count,
+            "calculated_transaction_count": total_transaction_count,
+            "calculated_award_count": total_award_count,
+            "calculated_subaward_count": total_subaward_count,
             "maximum_transaction_limit": settings.MAX_DOWNLOAD_LIMIT,
-            "transaction_rows_gt_limit": total_count > settings.MAX_DOWNLOAD_LIMIT,
+            "transaction_rows_gt_limit": total_transaction_count > settings.MAX_DOWNLOAD_LIMIT,
             "messages": get_generic_filters_message(
```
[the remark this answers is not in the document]
aguest-kc: I missed adding the A/C of adding the deprecation message to this endpoint, because `subawards` isn't a valid filter according to the API contract for this endpoint.

@ayubshahab this is the deprecation message for `subawards` that I added to the other tickets:
> 'subawards' will be deprecated in the future. Set ‘spending_level’ to ‘subawards’ instead. See documentation for more information.

usaspending_api/download/v2/download_count.py
```diff
@@ -52,15 +64,33 @@ def post(self, request):
...
+            total_subaward_count = 0
+        elif total_award_count is None:
+            total_award_count = 0
 
         result = {
-            "calculated_transaction_count": total_count,
+            "calculated_transaction_count": total_transaction_count,
+            "calculated_award_count": total_award_count,
+            "calculated_subaward_count": total_subaward_count,
             "maximum_transaction_limit": settings.MAX_DOWNLOAD_LIMIT,
```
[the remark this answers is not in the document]
ayubshahab: I'm wondering if that is necessary or not as I am not sure if we even have a maximum limit for when the spending_level is set to awards or subawards.

usaspending_api/download/v2/download_count.py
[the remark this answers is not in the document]
ayubshahab: Message has been added.

sethstoudenmier:
Adding comment as a follow up from ad-hoc conversation with @ayubshahab :

To avoid confusing users that might see this endpoint called without a `spending_type` provided (given we are using a default value of `transactions`) the `spending_type` will be carried forward into the response object and the response fields with `transaction` in the name will have a new value created that makes them type agnostic. TODO would be to remove the old transaction fields after the frontend make necessary updates to use the new fields.

usaspending_api/api_contracts/contracts/v2/download/count.md
```diff
@@ -95,6 +100,13 @@ Returns the number of transactions that would be included in a download request
 + `tas_codes` (optional, array[TASCodeObject], fixed-type)
 + `treasury_account_components` (optional, array[TreasuryAccountComponentsObject], fixed-type)
 
+## SpendingLevelObject (object)
```
sethstoudenmier, line 107: This feels like a nit-pick but I would move this up into the `Attributes` entirely. By referring to this as an `object` you are saying that the request should be:

```json
{
   "filters": { },
   "spending_level": {
      "spending_level": "transactions"
   } 
}
```

usaspending_api/api_contracts/contracts/v2/download/count.md
```diff
@@ -47,6 +48,10 @@ Returns the number of transactions that would be included in a download request
             A boolean returning whether the transaction count is over the maximum row limit.
         + `calculated_transaction_count` (required, number)
             The calculated count of all transactions which would be included in the download files.
+        + `calculated_awards_count` (required, number)
```
sethstoudenmier: Don't forget to update the response to match the new format.

usaspending_api/download/tests/integration/test_download_count.py
sethstoudenmier, line 1: Can you also add some test cases for the new `spending_level` attribute?

usaspending_api/api_contracts/contracts/v2/download/count.md
```diff
@@ -19,6 +19,7 @@ Returns the number of transactions that would be included in a download request
 
     + Attributes (object)
         + `filters` (required, AdvancedFilterObject)
+        + `spending_level` (optional, SpendingLevelObject)
     + Body
```
sethstoudenmier, line 25: Could you add `spending_level` to the body example?

usaspending_api/download/v2/download_count.py
```diff
@@ -53,14 +67,32 @@ def post(self, request):
...
+            options["time_period_obj"] = new_awards_only_decorator
+            filter_query = QueryWithFilters.generate_awards_elasticsearch_query(filters, **options)
+            search = AwardSearch().filter(filter_query)
+            total_count = search.handle_count()
 
         if total_count is None:
             total_count = 0
 
         result = {
+            # TO-DO: the following fields that follow the transaction_ naming convention will be removed
```
sethstoudenmier, line 87:
```suggestion
            # TODO: the following fields that follow the transaction_ naming convention will be removed
```

Minor recommended change because most IDEs will auto highlight `TODO`

sethstoudenmier (changes requested):

usaspending_api/api_contracts/contracts/v2/download/count.md
```diff
@@ -19,6 +19,11 @@ Returns the number of transactions that would be included in a download request
 
     + Attributes (object)
         + `filters` (required, AdvancedFilterObject)
+        + `spending_level` (optional, enum[string])
```
sethstoudenmier, line 22: Can you also add the `Default` here.

Example https://github.com/fedspendingtransparency/usaspending-api/blob/bb493c82f36798f07e1ead75ec9d0c2a5c293ffe/usaspending_api/api_contracts/contracts/v2/subawards.md?plain=1#L15

usaspending_api/api_contracts/contracts/v2/download/count.md
```diff
@@ -38,17 +43,26 @@ Returns the number of transactions that would be included in a download request
...
 
 + Response 200 (application/json)
     + Attributes (object)
-        + `transaction_rows_gt_limit` (required, boolean)
-            A boolean returning whether the transaction count is over the maximum row limit.
         + `calculated_transaction_count` (required, number)
             The calculated count of all transactions which would be included in the download files.
         + `maximum_transaction_limit` (required, number)
             The current allowed maximum number of transactions in a row-limited download. Visit https://www.usaspending.gov/download_center/custom_award_data to download larger volumes of data.
+        + `transaction_rows_gt_limit` (required, boolean)
```
sethstoudenmier, line 56: Don't forget to put these in the sample body response down below with sample data as well

usaspending_api/download/v2/download_count.py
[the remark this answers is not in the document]
sethstoudenmier: @ayubshahab I don't see any new messages in the response from the API

sethstoudenmier (changes requested):

collinwr (dismissed):

usaspending_api/api_contracts/contracts/v2/download/count.md
```diff
@@ -19,6 +19,12 @@ Returns the number of transactions that would be included in a download request
 
     + Attributes (object)
         + `filters` (required, AdvancedFilterObject)
+        + `spending_level` (optional, enum[string])
+            + Members
+                + `subawards`
+                + `transactions`
+                + `awards`
+                + Default: `transactions`
```
sethstoudenmier, line 27: The indentation is off; `Default` should be at the same level as the `Members` key.
Screenshot 2024-10-02 091929

sethstoudenmier (changes requested):

sethstoudenmier (approved):

PR #4204, opened 2024-10-01
[DEV-11525] Update recipient states endpoint [hotfix][staging]

aguest-kc:
**Description:**
Return all states and set their award count and spending amounts to 0 if no results were found in the previous database query. This is a common case on the first day of a new fiscal year when we don't have any spending for that fiscal year yet.

**Technical details:**

    - [x] Backend
8. [x] Jira Ticket [DEV-11525](https://federal-spending-transparency.atlassian.net/browse/DEV-11525):

```
2. API documentation updated

```

### DIFF
--- a/usaspending_api/recipient/tests/integration/test_state.py
+++ b/usaspending_api/recipient/tests/integration/test_state.py
@@ -3,11 +3,11 @@ import datetime
 import decimal
 
 import pytest
-from model_bakery import baker
 
 # Core Django imports
 # Third-party app imports
 from rest_framework import status
+from model_bakery import baker
 
 # Imports from your apps
 from usaspending_api.common.helpers.fiscal_year_helpers import generate_fiscal_year
--- a/usaspending_api/recipient/tests/integration/test_state.py
+++ b/usaspending_api/recipient/tests/integration/test_state.py
@@ -2,12 +2,12 @@
 import datetime
 import decimal
 
-# Core Django imports
+import pytest
+from model_bakery import baker
 
+# Core Django imports
 # Third-party app imports
 from rest_framework import status
-from model_bakery import baker
-import pytest
 
 # Imports from your apps
 from usaspending_api.common.helpers.fiscal_year_helpers import generate_fiscal_year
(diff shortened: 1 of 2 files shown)


usaspending_api/recipient/v2/views/states.py
```diff
@@ -112,15 +111,29 @@ def get_all_states(year=None, award_type_codes=None, subawards=False):
...
-        results = [
-            {
-                "pop_state_code": row["pop_state_code"],
-                "total": row["total"],
-                "count": len(set(row["distinct_awards"].split(","))),
-                "total_outlays": row["outlay_total"],
-            }
-            for row in list(queryset)
-        ]
+        if len(queryset) > 0:
```
sethstoudenmier, line 114: I think this should be expanded to not just if there are no results, but even if there is only a single result. This list of states is the main entry point into the State Profile page which allows users to view more than just the latest FY.

usaspending_api/recipient/v2/views/states.py
```diff
@@ -112,15 +111,29 @@ def get_all_states(year=None, award_type_codes=None, subawards=False):
...
+                    "total": row["total"],
+                    "count": len(set(row["distinct_awards"].split(","))),
+                    "total_outlays": row["outlay_total"],
+                }
+                for row in list(queryset)
+            ]
+        # In the case where there were no results, return a list of state codes with 0 for the amounts
+        # This is a common case on the first day of a new fiscal year
+        else:
+            queryset = SummaryStateView.objects.all().distinct("pop_state_code").values("pop_state_code")
```
sethstoudenmier, line 127: Probably not much of a concern since the two queries this will run against the DB are relatively short. However, if there are any performance concerns you could also query one with a LEFT OUTER JOIN between SummaryStateView that returns all unique states and then SummaryStateView that captures the values per state. Then in the response set values to their default.

sethstoudenmier (changes requested):
I think the comment about returning all states regardless is the only comment work blocking on. The single query comment is potential improvement if anything runs slow; thankfully since this is state data we know that the queryset size is capped.

Last thing, could you add a test case that validates this returns all states and sets values to 0 for those that aren't present?

usaspending_api/recipient/v2/views/states.py
aguest-kc: Updated this to check for more than 1 result.

usaspending_api/recipient/tests/integration/test_state.py
```diff
@@ -2,12 +2,12 @@
...
 import decimal
 
-# Core Django imports
+import pytest
+from model_bakery import baker
 
+# Core Django imports
 # Third-party app imports
 from rest_framework import status
-from model_bakery import baker
```
sethstoudenmier, line 9: I only complain about this because you moving those around makes that comment not hold as much value now.

Any reason to move those from under the `# Third-party app imports`

usaspending_api/recipient/v2/views/states.py
```diff
@@ -112,15 +111,29 @@ def get_all_states(year=None, award_type_codes=None, subawards=False):
...
-        results = [
-            {
-                "pop_state_code": row["pop_state_code"],
-                "total": row["total"],
-                "count": len(set(row["distinct_awards"].split(","))),
-                "total_outlays": row["outlay_total"],
-            }
-            for row in list(queryset)
-        ]
+        if len(queryset) > 1:
```
sethstoudenmier, line 114: My original concern still exists here. If you had a queryset that contains only data for Virginia then you still run into the problem where the remainder of the states will not appear and thus have no formal way for a user to reach that page.

I would think that this should be updated that regardless of the queryset size, we want to always return all states and merge whatever data is applicable for the selected time period.

usaspending_api/recipient/tests/integration/test_state.py
aguest-kc: This is my linter moving around imports. I'm not sure why it moved the `model_baker` import. I'll undo this import move.

usaspending_api/recipient/v2/views/states.py
aguest-kc: Got it. I was misunderstanding what your previous comment was saying.

usaspending_api/recipient/v2/views/states.py
```diff
@@ -111,16 +110,29 @@ def get_all_states(year=None, award_type_codes=None, subawards=False):
...
                 "total_outlays": row["outlay_total"],
             }
-            for row in list(queryset)
+            for row in list(fiscal_year_queryset)
         ]
+
+        # Get all other states and return `0` for their award count and totals
+        all_states_queryset = SummaryStateView.objects.all().distinct("pop_state_code").values("pop_state_code")
+        for row in all_states_queryset:
+            if row["pop_state_code"] not in [state["pop_state_code"] for state in results]:
```
sethstoudenmier, line 126: > [state["pop_state_code"] for state in results]

Can we break this out and define it outside of this `for` loop? This will recreate the list for each iteration of the loop

sethstoudenmier (changes requested):

sethstoudenmier (approved):

PR #4211, opened 2024-10-18
Updating README and local config

sethstoudenmier:
**Description:**
Update the README and other aspects of the `usaspending-api` codebase to support local development.

**Technical details:**
Below is a brief list of changes going file by file in order to give a high-level reasoning of "why".

* `.env.template`
   * The wrong database name was used for the Broker DB
   * Updated the rest of the DB string to create the Broker DB for tests on the USAspending DB container
* `.gitignore`
   * Change to make sure we are ignoring nested directories as well
* `CONTRIBUTING.md`
   * Update reference to Travis CI to be GitHub Actions
* `Dockerfile`
   * The `es_configure` job relies on `curl` to query the ES indexes
   * Without `curl` the `elasticsearch_indexer` job fails to create the templates for the ES indexes and thus creates many mis-typed fields on the indexes
* `Dockerfile.spark`
   * Similar changes that were made to the `Dockerfile` to change the base image used
   * Also had to update the version of OpenJDK; use AWS Corretto JDK 8 as it is the same version used by Databricks and this is a LTS version maintained by AWS
   * Some new installs were needed to support the different image
   * Old installs were left in place to avoid breaking things
* `Dockerfile.testing`
   * Update to the OpenJDK version similar to `Dockerfile.spark`
   * Remove the logic that changed the working directory in favor of having the volume named `/usaspending-api` instead of `/dockermount`
* Makefile
   * Update references of `docker-compose` to instead use `docker compose` (read more [here](https://docs.docker.com/compose/releases/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2))
   * Added back reference to spark-hive in Makefile; I incorrectly removed this in the Python upgrade
* `README.md`
   * Some cleanup
   * Update all shell commands to support copy and paste as-is
   * Made changes to reflect the steps that I took when setting everything up
* `docker-compose.yaml`
   * Removed the `version` as that is deprecated
   * Updated the `usaspending-test` volume to be `/usaspending-api` in the container to support hot reload when testing
   * Change the folder create for bulk downloads on Minio; Minio folders cannot have an underscore in the name
* `pyproject.toml`
   * Add the current directory as pythonpath to support tests having issues to run
* `spark_helpers.py`
   * Added a check for the "SparkSession" attribute on the JVM; this projects against some cases where during cleanup the SparkSession was already closed / removed by another parallel test build
* `local.py`
   * Minio doesn't support bucket names with underscore
* `conftest.py`
   * The test suite handles creation of a Broker DB for tests via a docker run command. The network was proving to be a problem when attempting to run this via the docker containers exclusively. As a workaround, when using the test container, it will define a "PROJECT_SRC_PATH" env that is used to assign the created Broker containers to the same network, allowing reference by name.
* `conftest_spark.py`
   * Increase the memory allocated to Spark to support the test cases

    - [x] Backend
8. N/A Jira Ticket

```
```


Makefile
```diff
@@ -161,51 +161,51 @@ endif
...
 
 .PHONY: docker-compose
-docker-compose: ## Run an arbitrary docker-compose command by passing in the Docker Compose profiles in the "profiles" variable, and args in the "args" variable
+docker-compose: ## Run an arbitrary docker compose command by passing in the Docker Compose profiles in the "profiles" variable, and args in the "args" variable
 	# NOTE: The .env file is used to provide environment variable values that replace variables in the compose file
 	#       Because the .env file does not live in the same place as the compose file, we have to tell compose explicitly
 	#       where it is with "--project_directory". Since this is called from the root Makefile, using ./ points to the dir
 	#       of that Makefile
-	docker-compose ${profiles} --project-directory . --file ${docker_compose_file} ${args}
+	docker compose ${profiles} --project-directory . --file ${docker_compose_file} ${args}
```
ayubshahab, line 169: Why did we remove the - between docker compose? I have historically always used docker-compose. Is there a difference? For example: `docker-compose up usaspending-db usaspending-es`

ayubshahab (approved):

Makefile
aguest-kc: @ayubshahab Recent updates to `docker` have added `compose` as a sub-command of `docker` itself instead of being a standalone executable.

sethstoudenmier:
> @sethstoudenmier What do you think about adding an `init.sql` script to run some common SQL commands when setting up the DB for the first time, since we're updating the docs? There are some common issues and their solutions don't seem to be documented publicly.
> 
> * `ALTER USER usaspending SET SEARCH_PATH TO "public", "raw", "int", "temp", "rpt";`
>   
>   * This prevents a common error about tables outside of the `public` schema not being found.
> * Create the `etl_user`
>   
>   * This user is needed for migrations to run.
> * Create the `readonly` user
>   
>   * This user is needed for migrations to run

Funny you should say that, I actually had that in place originally. The problem is that it conflicts with the Travis CI/CD build process when we use the `--reuse-db` flag because it attempts to create users that already exist. Instead I opted to document that in the README.

Makefile
sethstoudenmier: @ayubshahab here is the doc to help explain: https://docs.docker.com/compose/releases/migrate/

aguest-kc:
Ahh.. that makes sense with the CI/CD process. I missed the changes to the README, because it was collapsed. I'm happy with those steps being documented!

zachflanders-frb (dismissed):
👏

Dockerfile.spark
```diff
@@ -1,37 +1,24 @@
...
 ARG SPARK_VERSION=3.2.1
 ARG PROJECT_LOG_DIR=/logs
 
-RUN yum -y update && yum clean all
-# sqlite-devel added as prerequisite for coverage python lib, used by pytest-cov plugin
-RUN yum -y install wget gcc openssl-devel bzip2-devel libffi libffi-devel zlib-devel sqlite-devel
-RUN yum -y groupinstall "Development Tools"
+# Install dependencies
+RUN apt update && \
+    apt install -y coreutils gcc wget openssl libssl-dev libbz2-dev build-essential
```
zachflanders-frb, line 10:
```suggestion
    apt install -y coreutils gcc wget openssl libssl-dev libbz2-dev build-essential curl
```

docker-compose.yml
```diff
@@ -259,10 +263,10 @@ services:
 
   spark-worker:
```
zachflanders-frb, line 264: I found I had to add the following environmental variables to the spark worker:
```yaml
      PYTHONPATH: /project
      DJANGO_SETTINGS_MODULE: "usaspending_api.settings"
      DATABASE_URL: postgres://${USASPENDING_DB_USER}:${USASPENDING_DB_PASSWORD}@${USASPENDING_DB_HOST}:${USASPENDING_DB_PORT}/data_store_api
```

zachflanders-frb (approved):

PR #4212, opened 2024-10-18
[DEV-11086] Add new PYA and PARK fields to Files B and C

collinwr:
**Description:**
Broker is introducing 4 new fields to be to both their file b and c tables. We should update our corresponding tables and loaders to ensure these fields are read. 

**Technical details:**
Field and table details can be found in ticket. 

This should not be merged until broker adds these fields 

1. [x] Unit & integration tests updated - Made small change to the `test_load_multiple_submissions` test to ensure that the build will fail until broker adds these fields. Other than that, we don't have a comprehensive test that ensures that each individual field was successfully loaded from broker during the submission loader. 
    - [ ] Backend
7. [ ] Appropriate Operations ticket(s) created - Need to create deploy coordination for running migrations
8. [x] Jira Ticket [DEV-11086](https://federal-spending-transparency.atlassian.net/browse/DEV-11086):

```
```


usaspending_api/awards/models/financial_accounts_by_awards.py
```diff
@@ -17,6 +17,8 @@ class AbstractFinancialAccountsByAwards(DataSourceTrackedModel):
     parent_award_id = models.TextField(blank=True, null=True)
     fain = models.TextField(blank=True, null=True)
     uri = models.TextField(blank=True, null=True)
+    prior_year_adjustment = models.TextField(blank=True, null=True)
+    pa_reporting_key = models.TextField(blank=True, null=True)
```
zachflanders, line 21: Do we expect a certain set of values for these fields (e.g. can we use [`choices`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#choices))?

zachflanders (approved):

usaspending_api/awards/models/financial_accounts_by_awards.py
sethstoudenmier, line 21: Since there was a question about what `pa_` stands for it might be good to make use of `help_text` here similar to this spot on Award model: https://github.com/fedspendingtransparency/usaspending-api/blob/master/usaspending_api/awards/models/award.py#L54

usaspending_api/financial_activities/models.py
```diff
@@ -13,6 +13,8 @@ class AbstractFinancialAccountsByProgramActivityObjectClass(DataSourceTrackedMod
     treasury_account = models.ForeignKey(
         "accounts.TreasuryAppropriationAccount", models.CASCADE, related_name="program_balances", null=True
     )
+    prior_year_adjustment = models.TextField(blank=True, null=True)
+    pa_reporting_key = models.TextField(blank=True, null=True)
```
sethstoudenmier, line 17: Similar use of `help_text` here would be nice.

sethstoudenmier (approved):

usaspending_api/awards/models/financial_accounts_by_awards.py
collinwr: Choices is a good idea when we're accepting something from a source that hasn't been validated (maybe user submission). In this case, we're reading data from the broker, which has already performed it's own set of validations. 

We could still potentially add options here, but if anything changes on the schema side (a new PARK is added, we'd need to make sure to update this before hand, or else the submission loader would fail). 

I'd also suspect that in this case, there might be too many options to include here. We have quite a few program activities. 

I can add `help_text` as Seth suggested though.

usaspending_api/financial_activities/migrations/0010_auto_20241018_1526.py
```diff
@@ -0,0 +1,38 @@
...
+            model_name='financialaccountsbyprogramactivityobjectclass',
+            name='ussgl480110_reinstated_del_cpe',
+            field=models.DecimalField(blank=True, decimal_places=2, max_digits=23, null=True),
+        ),
+        migrations.AddField(
+            model_name='financialaccountsbyprogramactivityobjectclass',
+            name='ussgl490110_reinstated_del_cpe',
+            field=models.DecimalField(blank=True, decimal_places=2, max_digits=23, null=True),
+        ),
+        # Update the download view to include the new columns
```
sethstoudenmier, line 34: Since the view uses `SELECT *` on File B is this needed?

sethstoudenmier (approved):
LGTM! One small comment but I won't block over it

usaspending_api/financial_activities/migrations/0010_auto_20241018_1526.py
sethstoudenmier: Nevermind, I see that this is what was required to fix the failing test cases. Can disregard

PR #4219, opened 2024-11-06
[DEV-11608] - Add county FIPS code to location autocomplete

zachflanders-frb:
**Description:**
From JIRA: 
> The frontend has a list of counties that contains both the FIPS and county name. Currently we provide the county name in the /api/v2/autocomplete/location endpoint, but this can be problematic in the case of either bad data or if the frontend list uses a different name. To help with some of the bad data we should provide the FIPS (five digit code) in the response to be consistent between the front and backend.

These changes enable users to search `/autocomplete/location`  using both the county name and the county fips code. 

**Technical details:**
The response returns a list of counties in the following format:
```json
[
    {"county_fips": "12039", "county_name": "GADSDEN", "state_name": "FLORIDA", "country_name": "UNITED STATES"},
    {"county_fips": "13039", "county_name": "CAMDEN", "state_name": "GEORGIA", "country_name": "UNITED STATES"}
]
```

    - [x] Backend
    - [x] Frontend <OPTIONAL>
8. [x] Jira Ticket [DEV-11608](https://federal-spending-transparency.atlassian.net/browse/DEV-11608):

```
I don't believe these are needed:

```


aguest-kc:
You will need to complete step 7, **Appropriate Operations ticket(s) created**. Since this ticket changes the structure of the Elasticsearch index, it will need to be recreated and repopulated.

We can go over that process together.

aguest-kc (dismissed):

anjenkin (approved):

usaspending_api/references/v2/views/location_autocomplete.py
```diff
@@ -192,26 +182,32 @@ def _format_county_results(self, es_results: List[Hit]) -> Union[List, None]:
...
         """
-
         if len(es_results) > 0:
-            counties = []
-            for doc in es_results:
-                for county in doc.meta.highlight.counties:
-                    counties.append(
-                        {"county_name": county, "state_name": doc.state_name, "country_name": doc.country_name}
-                    )
+            counties = [
```
sethstoudenmier, line 198: This seems to work exactly as expected. With all of the nested looped though we should make sure to validate against large QAT dataset before marking the ticket done. Not going to block on this though.

sethstoudenmier (changes requested):
One small request change about the regex usage. Open to discussion if there is a reason to not limit the specific character count.

sethstoudenmier (approved):
Pulled down code and validated that the endpoint works as expected. While I did not personally test further that this works with Databricks, it is my understanding this was tested in Databricks.

PR #4225, opened 2024-11-19
[DEV-11086] Fix handling of lowercase PYA and duplicate Object Class case

sethstoudenmier:
**Description:**
While testing PYA it was noticed that USAspending needs to handle the possibility of a lowercase value for PYA. Additionally, similar to DEFC, PYA should also be taken into account when considering the case of a duplicate Object Class.

**Technical details:**
File B loader in the Submission Loader was updated to covert `prior_year_adjustment` and `pa_reporting_key` both to uppercase values ingested from Broker. This logic should also be applied to all other text fields to prevent the case of bad data, but that is out of the scope of this ticket; plan is to create tech debt.

Additionally,  `prior_year_adjustment` and `pa_reporting_key` were added to our handling of the duplicate Object Class case and is demonstrated in the updated test cases.

    - [ ] Backend
8. [x] Jira Ticket [DEV-11086](https://federal-spending-transparency.atlassian.net/browse/DEV-11086):

```
```


aguest-kc (approved):

sethstoudenmier:
> Makes sense to me. The only thing I would ask is if the same uppercase treatment should be applied to the new fields in File C even though they're not currently being used for any calculations.
> 
> Not a blocker for this ticket, but it could be something to think about in the future.

I called that out briefly in the PR description, but to summarize yes we need to take that into account. After talking with the Broker team we should not assume that values are in a specific case when we pull it into USAspending. I have a TODO item to create a tech debt ticket.

PR #4242, opened 2024-12-17
[Dev-10092] -  PSC references update

zachflanders-frb:
**Description:**
The USAs PSC filter does not include all PSC values.  This PR updates the load_psc command to update the PSC table with the most recent PSC Excel file found at https://www.acquisition.gov/psc-manual 

**Technical details:**
This code uses pandas instead of openpyxl to make the code more readable and somewhat less brittle and more resilient to changes.  `pandas.read_excel` supports using a url to the excel file, so I used the URL to the hosted file on www.acquisition.gov instead of keeping the file in the code repository.  Another option is that we may want to host this file ourselves in a storage bucket.

    - [x] Backend
8. [x] Jira Ticket [DEV-10092](https://federal-spending-transparency.atlassian.net/browse/DEV-10092):

```
```


usaspending_api/references/management/commands/load_psc.py
```diff
@@ -1,81 +1,134 @@
...
-from openpyxl import load_workbook
 
 
 class Command(BaseCommand):
     help = "Loads program information obtained from Excel file on https://www.acquisition.gov/PSC_Manual"
 
     logger = logging.getLogger("script")
-    default_directory = os.path.normpath("usaspending_api/references/management/commands/")
-    default_filepath = os.path.join(default_directory, "PSC_Data_June_2019_Edition_FINAL_6-20-19+DRW.xlsx")
+    default_filepath = "https://www.acquisition.gov/sites/default/files/manual/PSC%20April%202024.xlsx"
```
zachflanders-frb, line 16: Do we want to host this file in a storage bucket instead to ensure that it is available?

usaspending_api/references/management/commands/load_psc.py
```diff
@@ -1,81 +1,134 @@
...
+                num_updated += 1
+            psc.length = new_psc["length"]
+            psc.description = new_psc["description"]
+            psc.start_date = check_dates(psc.start_date, new_psc["start_date"])
+            psc.end_date = check_dates(psc.end_date, new_psc["end_date"])
+            psc.full_name = new_psc["full_name"]
+            psc.excludes = new_psc["excludes"]
+            psc.notes = new_psc["notes"]
+            psc.includes = new_psc["includes"]
             psc.save()
```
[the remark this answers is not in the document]
zachflanders-frb: Also, the this does not remove codes.  Do we want to modify the code the remove PSC codes if they do not appear in the new file?

usaspending_api/references/management/commands/load_psc.py
aguest-kc: I think that's a good idea. Plus, if we self-host it we know nothing is changing unexpectedly. We would just need to remember to go download and then upload the updated file whenever it's updated.

If we do decide to self-host this file, any thoughts on using a more standardized filename instead of one with the month and year in it? I see that we can use the `--path` flag to change which spreadsheet is being loaded, but it could be convenient to have a default value of `latest_psc.xlsx`. That way if/when we uploaded a new file we could just name it `latest_psc.xlsx` and not need to worry about updating `self.default_filepath` or updating the Jenkins job to specify a new `--path` value. I'm not tied to this idea, but it might make long-term maintenance of this command easier.

sethstoudenmier: > any thoughts on using a more standardized filename instead of one with the month and year in it?
Not entirely against the idea, but we would want to maintain an easy way to determine exactly what file we uploaded previously. That could be tags on the S3 object or a comment in the code.

usaspending_api/references/management/commands/load_psc.py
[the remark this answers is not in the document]
sethstoudenmier: >I assume this means perform the analysis before we update elastic search indexes that use the PSC model.

Would want to step through just the PSC table this loads into, you can ignore the usage of PSC. On that note, having an old version to compare is up to you and how you want to analyze the difference. It doesn't have to be anything long living.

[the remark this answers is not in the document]
sethstoudenmier: > Also, this does not remove codes. Do we want to modify the code the remove PSC codes if they do not appear in the new file?

I think a nuke-and-pave approach is appropriate here, but I would make sure to perform it all inside of a transaction so that if it fails we don't create a headache for active users.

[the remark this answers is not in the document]
zachflanders-frb: Something like this might be cleaner if we are taking a nuke-and-pave approach:
```python
with transaction.atomic():
    old = PSC.objects.all().delete()
    new = PSC.objects.bulk_create(new_pscs)
make_a_report(old, new)
```

requirements/requirements-app.txt
```diff
@@ -23,7 +23,7 @@ fiscalyear==0.2.0
 Markdown<3.0
 marshmallow==3.14.1
 numpy==1.20.*
-openpyxl==2.4.7
+openpyxl>=3
```
zachflanders-frb, line 26: pandas 1.3 has a [dependency](https://pandas.pydata.org/pandas-docs/version/1.3/getting_started/install.html#excel-files) on openpyxl >= 3.0.0.

usaspending_api/references/management/commands/load_psc.py
```diff
@@ -1,81 +1,176 @@
...
+        def _serialize_pscs(models: Iterable[PSC]) -> str:
+            return json.dumps([{k: str(v) for k, v in model.__dict__.items()} for model in models], indent=2)
+
+        if result.get("error"):
+            self.logger.error(result["error"])
+        else:
+            self.logger.info(f"Added: {_serialize_pscs(result['added'].values())}")
+            self.logger.info(f"Updated: {_serialize_pscs([v.get('new') for v in result['updated'].values()])}")
+            self.logger.info(f"Deleted: {_serialize_pscs(result['deleted'].values())}")
+            self.logger.log(
```
sethstoudenmier, line 51: Any reason to not use `self.logger.info` here?

usaspending_api/references/management/commands/load_psc.py
```diff
@@ -1,81 +1,176 @@
...
-    else:
-        psc.start_date = start_date
-    if psc.end_date:
-        if end_date:
-            if psc.end_date < end_date.date():
-                psc.end_date = end_date
-
-    else:
-        psc.end_date = end_date
+    except IOError as e:
```
sethstoudenmier, line 167: Since the result is the same regardless of the error is there any benefit to having the different exceptions called out?

sethstoudenmier (commented):
Overall looking good. Not going to approve or block right now as it is a draft.

usaspending_api/references/management/commands/load_psc.py
zachflanders-frb: Since we're not doing anything to handle the error and if we want it to raise the error, should we just remove the try except entirely and let it fail and raise the error if there is a problem?

sethstoudenmier: I don't see why not. Not sure why this would have silenced the error before.

usaspending_api/references/tests/integration/test_load_psc.py
```diff
@@ -1,10 +1,15 @@
...
 
+import pytest
 from django.core.management import call_command
-from datetime import datetime
-from usaspending_api.references.models import PSC
 from model_bakery import baker
 
+from usaspending_api.references.models import PSC
+
+logger = logging.getLogger(__name__)
```
sethstoudenmier, line 11: Any reason to keep this logger defined in the test?

sethstoudenmier (dismissed):
Approving with the assumption that this loader has been tested locally. One last comment about removing logger from test case, but not going to block over it.

usaspending_api/references/tests/integration/test_load_psc.py
zachflanders-frb: Good catch.

sethstoudenmier (approved):

PR #4243, opened 2024-12-17
[DEV-11489] Update query_with_filters to use ES for program_activities

aguest-kc:
**Description:**
Update the `_ProgramActivities` filter in `query_with_filters.py` to use the fields in Elasticsearch instead of building database queries.

**Technical details:**

    - [ ] Backend
8. [x] Jira Ticket [DEV-11489](https://federal-spending-transparency.atlassian.net/browse/DEV-11489):

```
API documentation updated
No API endpoints were updated as part of this change.

Matview impact assessment completed
```


usaspending_api/common/query_with_filters.py
```diff
@@ -624,26 +622,44 @@ class _ProgramActivities(_Filter):
...
-                query_filter_predicates.append(Q(program_activity__program_activity_name=filter_value["name"].upper()))
-            if "code" in filter_value:
-                query_filter_predicates.append(Q(program_activity__program_activity_code=filter_value["code"]))
-            award_ids_filtered_by_program_activities = FinancialAccountsByAwards.objects.filter(
-                *query_filter_predicates
-            )
-            award_ids = set(award_ids_filtered_by_program_activities.values_list("award_id", flat=True))
+            if "name" in filter_value and "code" in filter_value:
+                program_activity_match_queries.append(
+                    ES_Q(
```
sethstoudenmier, line 631: Not worth blocking over, but this could be cleaned up since you are repeating yourself a bit. For example, build a list of the `must` values and then at the end append the `ES_Q()` object. You could then simplify the conditionals down to check the `name`, check the `code`, and if the `must` list has values then append the `ES_Q`.

usaspending_api/search/migrations/0048_replace_pac_pan_columns_with_program_activities.py
```diff
@@ -0,0 +1,52 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('search', '0047_add_program_activity_to_search'),
+    ]
+
+    operations = [
```
sethstoudenmier, line 12: I believe this is the case, but can you confirm that these fields are not currently being used anywhere?

sethstoudenmier (commented):
@aguest-kc can you confirm if there are already tests for this functionality? If not then one should be added.

usaspending_api/search/migrations/0048_replace_pac_pan_columns_with_program_activities.py
aguest-kc: They are not being used anywhere else. Thankfully, these two columns haven't been around long enough to be used anywhere else.

usaspending_api/common/query_with_filters.py
aguest-kc: I've updated this logic.

aguest-kc:
> @aguest-kc can you confirm if there are already tests for this functionality? If not then one should be added.

Yes, there's already tests for this functionality and those are now passing. The test data just needed to be updated to be present in Elasticsearch as well as the database.

usaspending_api/search/migrations/0050_replace_pac_pan_columns_with_program_activities.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 3.2.25 on 2024-12-20 01:14
```
sethstoudenmier, line 1: Can you confirm that this doesn't attempt to generate any additional fields when using Django 4.2?

sethstoudenmier (approved):
Approving, but left one question would like an answer on before merging.

PR #4251, opened 2025-01-08
[DEV-11876] Update PARK field names

sethstoudenmier:
**Description:**
Rename some USAspending fields to match new Broker names.

**Technical details:**
The changes are pretty straightforward, but the migrations are a little bit involved. The order of events is as follows:
- Run `stage` migrations for both File B and C tables to create the new columns
- Run `backfill` script to populate the new field while minimizing the time that a lock is held
- Run `delete` migrations for both File B and C tables to remove the old columns

Afterwards, an additional step to VACUUM the File B and C tables will be added to the deploy coordination to help reduce bloat.

    - [x] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-11876](https://federal-spending-transparency.atlassian.net/browse/DEV-11876):

```
```

### DIFF
--- a/usaspending_api/database_scripts/job_archive/backfill_renamed_park_fields.py
+++ b/usaspending_api/database_scripts/job_archive/backfill_renamed_park_fields.py
@@ -1,0 +1,114 @@
+"""
+Jira Ticket Number: DEV-11876
+
+    Change Broker PARK columns in USAspending
+
+Expected CLI:
+
+    $ python3 usaspending_api/database_scripts/job_archive/backfill_renamed_park_fields.py
+
+Purpose:
+
+    Multiple fields were renamed in Broker and thus renamed in USAspending accordingly.
+    Performing a single RENAME operation will hold an AccessExclusiveLock on the DB and
+    potentially interfere with normal operations of the API and Downloads. To prevent any
+    outage this script iterates over all File B and C records to copy values from the old
+    columns to their renamed counterparts, taking a lock on the File B and C tables intermittently.
+
+"""
+
+import logging
+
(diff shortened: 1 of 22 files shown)


usaspending_api/database_scripts/job_archive/backfill_renamed_park_fields.py
```diff
@@ -0,0 +1,114 @@
...
+
+from os import environ
+from pathlib import Path
+from typing import Iterator, Tuple
+
+import psycopg2
+
+# Import our USAspending Timer component.  This will not work if we ever add
+# any Django specific stuff to the timing_helpers.py file.
+exec(Path("usaspending_api/common/helpers/timing_helpers.py").read_text())
```
zachflanders-frb, line 30: Why use `exec` here vs. just importing the classes?

zachflanders-frb (dismissed):

usaspending_api/database_scripts/job_archive/backfill_renamed_park_fields.py
sethstoudenmier: Honest answer is that I was lazy and reused one of the patterns we have for this type of work. After looking some more, I saw that we had a pattern of treating these as standalone scripts which brings into question the idea of even the `exec` line above. I don't see a reason why we should avoid importing so I am going to make the change and push it up.

On a side note, Travis PR check is being dumb, so I need to force it to re-run anyways.

zachflanders-frb (approved):

PR #4221, opened 2024-11-12
[DEV-11409] Grafana Migration

ayubshahab:
**Description:**
This ticket/PR was created to continue the Grafana Migration from DataDog. Specifically, we addressed the issue of subprocess tracing inside the bulk_downloader and further streamlined alot of the information we were capturing. Logs inside loki and traces inside tempo have also been aligned. The testing suite was also updated. This PR contains the initial changes implemented inside of DEV-11215. 

**Technical details:**
Since, Otel does not support filtering of traces and spans, there is some left over code that was being utilized with Datadog. That code has been left inside the tracing file in case of future uses but can be safely taken out if need be. 

**Additional Details:**
Previously, we were able to perform some foundational work regarding the Grafana Migration from Datadog. Baseline functionality was implemented so UsaSpending API endpoints could be traced and relevant information such as traces could be ingested in Tempo and logging could be ingested in Loki. This goal was accomplished with [DEV-11215.](https://federal-spending-transparency.atlassian.net/browse/DEV-11215) 

**Requirements for PR merge:**

1. [x] Unit & integration tests updated
2. [x] API documentation updated
3. [x] Necessary PR reviewers:
    - [x] Backend
7. [x] Appropriate Operations ticket(s) created
8. [x] Jira Ticket [DEV-11409](https://federal-spending-transparency.atlassian.net/browse/DEV-11409):
    - [x] Link to this Pull-Request

**Area for explaining above N/A when needed:**
```
```

PR #4261, opened 2025-01-28
[DEV-11409] Grafana Migration

collinwr:
**Description:**
This ticket/PR was created to continue the Grafana Migration from DataDog. Specifically, we addressed the issue of subprocess tracing inside the bulk_downloader and further streamlined alot of the information we were capturing. Logs inside loki and traces inside tempo have also been aligned. The testing suite was also updated. This PR contains the initial changes implemented inside of DEV-11215. 

**Technical details:**
Since, Otel does not support filtering of traces and spans, there is some left over code that was being utilized with Datadog. That code has been left inside the tracing file in case of future uses but can be safely taken out if need be. 

**Additional Details:**
Previously, we were able to perform some foundational work regarding the Grafana Migration from Datadog. Baseline functionality was implemented so UsaSpending API endpoints could be traced and relevant information such as traces could be ingested in Tempo and logging could be ingested in Loki. This goal was accomplished with [DEV-11215.](https://federal-spending-transparency.atlassian.net/browse/DEV-11215) 

This pull request is a recreation of this closed PR: https://github.com/fedspendingtransparency/usaspending-api/pull/4221

    - [x] Backend
8. [x] Jira Ticket [DEV-11409](https://federal-spending-transparency.atlassian.net/browse/DEV-11409):

```
```


requirements/requirements-app.txt
```diff
@@ -33,4 +32,39 @@ python-json-logger==2.0.7
...
+opentelemetry-instrumentation-botocore==0.49b2
+opentelemetry-instrumentation-dbapi==0.49b2
+opentelemetry-instrumentation-django==0.49b2
+opentelemetry-instrumentation-elasticsearch==0.49b2
+opentelemetry-instrumentation-grpc==0.49b2
+opentelemetry-instrumentation-logging==0.49b2
+opentelemetry-instrumentation-psycopg2==0.49b2
+opentelemetry-instrumentation-redis==0.49b2
+opentelemetry-instrumentation-requests==0.49b2
+opentelemetry-instrumentation-sqlite3==0.49b2
```
sethstoudenmier, line 57: Do we need sqlite here?

requirements/requirements-dev.txt
```diff
@@ -2,8 +2,8 @@ black==24.10.0
 click==8.1.7
 docker==7.0.0
 dredd-hooks==0.2.0
+importlib-metadata==8.5.0
```
sethstoudenmier, line 5: Can we keep this in alphabetical order?

usaspending_api/bulk_download/config/supervisord.conf
```diff
@@ -12,8 +12,23 @@ user=ec2-user
...
+    OTEL_PYTHON_LOG_CORRELATION="true",
+    OTEL_PYTHON_LOG_FORMAT="%(msg)s [span_id=%(otelSpanID)s trace_id=%(otelTraceID)s]",
+    OTEL_PYTHON_LOG_LEVEL="debug",
+    OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED="true",
+    OTEL_EXPORTER_OTLP_PROTOCOL="http/protobuf",
+    OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SANITIZE_FIELDS=".*session.*,set-cookie",
+    OTEL_EXPORTER_OTLP_TIMEOUT="30000",
+    OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST="content-length,content-type,host,origin,referer,ua-is-bot,user-agent,x-forwarded-for,x-requested-with,allow,cache-trace,is-dynamically-rendered,key,strict-transport-security",
+    OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE="content-length,content-type,host,origin,referer,ua-is-bot,user-agent,x-forwarded-for,x-requested-with,allow,cache-trace,is-dynamically-rendered,key,strict-transport-security",
+    GRPC_VERBOSITY="debug",
```
sethstoudenmier, line 29: Do we want this to be set to `debug` outside of local development instances?

usaspending_api/bulk_download/config/supervisord.conf
```diff
@@ -12,8 +12,23 @@ user=ec2-user
...
-    DOWNLOAD_DATABASE_URL="%(ENV_DOWNLOAD_DATABASE_URL)s"
-command=python3 manage.py download_sqs_worker
+    DOWNLOAD_DATABASE_URL="%(ENV_DOWNLOAD_DATABASE_URL)s",
+    # ==== [Open Telemetry Configuration] ====
+    OTEL_SERVICE_NAME="usaspending-downloader",
+    OTEL_TRACES_EXPORTER="otlp",
+    OTEL_METRICS_EXPORTER="otlp",
+    OTEL_PYTHON_LOG_CORRELATION="true",
+    OTEL_PYTHON_LOG_FORMAT="%(msg)s [span_id=%(otelSpanID)s trace_id=%(otelTraceID)s]",
+    OTEL_PYTHON_LOG_LEVEL="debug",
```
sethstoudenmier, line 22: Do we want this to be set to `debug` outside of local development instances?

usaspending_api/common/logging.py
```diff
@@ -204,3 +222,90 @@ def ensure_logging(
...
+        ("DEFUALT_DB_HOST", DATABASES["default"]["HOST"]),
+        ("TRACE_ENV", TRACE_ENV),
+    ]
+
+    add_custom_attribute_span_processors(trace.get_tracer_provider(), attribute_pairs)
+
+    # Set up the OTLP exporter
+    # Check out https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/
+    exporter = None
+    if os.getenv("USASPENDING_DB_HOST") == "127.0.0.1" and os.getenv("TOGGLE_OTEL_CONSOLE_LOGGING") == "True":
```
sethstoudenmier, line 289: Could we use the `settings.IS_LOCAL` instead? Because the host could be `localhost`, `127.0.0.1`, a container name, or even `0.0.0.0`.

usaspending_api/common/tests/unit/test_tracing.py
```diff
@@ -40,171 +82,185 @@ def caplog(caplog):
...
+                        log_record = log_queue.get(block=True, timeout=5)
+                        log_records.append(log_record)
+
+                    if log_queue.empty():
+                        draining = False
+
+                for log_record in log_records:
+                    caplog.handler.flush(log_record)
+            except Exception:
+                print("Error draining captured log queue when handling subproc TimeoutError")
```
sethstoudenmier, line 205: Can this print be removed?

usaspending_api/common/tests/unit/test_tracing.py
```diff
@@ -40,171 +82,185 @@ def caplog(caplog):
...
+                pass
+            raise mp.TimeoutError(f"subprocess {worker.name} did not complete in timeout")
```
sethstoudenmier, line 207: Why does it raise an error regardless after the `try...except`?

usaspending_api/download/filestreaming/download_generation.py
```diff
@@ -63,17 +67,55 @@ def generate_download(download_job: DownloadJob, origination: Optional[str] = No
...
         if limit is not None and limit > MAX_DOWNLOAD_LIMIT:
+
+            with SubprocessTrace(
+                name=f"job.{JOB_TYPE}.generate_download_{request_type}",
+                kind=SpanKind.INTERNAL,
+                service="bulk-download",
+            ) as limit_exceeded:
+                limit_exceeded.set_attributes(
+                    {
+                        "message": "Unable to process this download because it includes more than the current limit of {MAX_DOWNLOAD_LIMIT} records",
```
sethstoudenmier, line 110:
```suggestion
                        "message": f"Unable to process this download because it includes more than the current limit of {MAX_DOWNLOAD_LIMIT} records",
```

usaspending_api/download/management/commands/download_sqs_worker.py
```diff
@@ -1,110 +1,219 @@
...
-        raise DownloadJobNoneError(download_job_id)
+    download_job_details = download_job_to_log_dict(download_job)
+
+    with SubprocessTrace(
+        name=f"job.{JOB_TYPE}.download.retrieve",
+        kind=SpanKind.INTERNAL,
+        service="bulk-download",
+    ) as span:
+
+        span.set_attributes(
```
sethstoudenmier, line 182: Seems like this could be broken out into a method since it is such a long list of attributes that get set and it appears to be used in multiple places. Won't block over this, but a suggestion to help with maintenance.

usaspending_api/settings.py
```diff
@@ -61,6 +60,10 @@
 # Define local flag to affect location of downloads
 IS_LOCAL = True
 
+# Indicates which environment is sending traces to Grafana.
+# This will be overwritten by Ansible
+TRACE_ENV = "unspecified"
```
sethstoudenmier, line 65: Do we not want to define this in our different config files?

usaspending_api/wsgi.py
```diff
@@ -7,10 +7,82 @@
...
+def request_hook(span, environ):
+
+    if span and span.is_recording():
+        for header in HEADERS_TO_CAPTURE:
+            header_env_name = f"HTTP_{header.replace('-', '_').upper()}"
+            header_value = environ.get(header_env_name)
+            if header_value:
+                span.set_attribute(f"http.request.header.{header}", header_value)
+
+    if os.getenv("USASPENDING_DB_HOST") == "127.0.0.1" and os.getenv("TOGGLE_OTEL_CONSOLE_LOGGING") == "True":
```
sethstoudenmier, line 52: Could be a good spot to use `settings.IS_LOCAL` if Django is started by this point. If it is not, then we need to find a way to better capture that this is running on a local build.

usaspending_api/wsgi.py
```diff
@@ -7,10 +7,82 @@
...
+
+def response_hook(span, environ, status, response_headers):
+
+    if span and span.is_recording():
+        for header in HEADERS_TO_CAPTURE:
+            for response_header in response_headers:
+                if response_header[0].lower() == header:
+                    span.set_attribute(f"http.response.header.{header}", response_header[1])
+
+    if os.getenv("USASPENDING_DB_HOST") == "127.0.0.1" and os.getenv("TOGGLE_OTEL_CONSOLE_LOGGING") == "True":
```
sethstoudenmier, line 64: Same comment as above

sethstoudenmier (changes requested):

requirements/requirements-dev.txt
collinwr: Yep, got this fixed

requirements/requirements-app.txt
collinwr: What do you think about removing these as well?
- opentelemetry-instrumentation-asgi
- opentelemetry-instrumentation-dbapi
- opentelemetry-instrumentation-tortoiseorm

usaspending_api/common/logging.py
collinwr: Yes, I think this makes sense 👍

usaspending_api/common/tests/unit/test_tracing.py
collinwr: I think there are a few things going on here. The `if worker.is_alive()` block handles the case where the subprocess did not complete within the specified timeout period `worker.join(timeout=100)`. In this case, we ultimately want to fail the test, but we first want to drain the logs from our `log_queue` (a `multiprocessing.Queue`) to preserve that debugging information and potentially figure out why it missed the timeout window. After draining the logs, we're safe to go ahead and throw a meaningful exception to fail the test.

usaspending_api/common/tests/unit/test_tracing.py
collinwr: Based on the comment below about this block, what do you think about changing this from a print to a `logger.error`?

usaspending_api/settings.py
collinwr: We tried this but found that those config files were not actually being used outside of management commands. Otherwise, we could have just used the existing `ENV_CODE` variable. Double checking with @boozallendanny that we're also overwriting this outside of deploys (places like management commands)

usaspending_api/bulk_download/config/supervisord.conf
collinwr: Set this to `info`

usaspending_api/bulk_download/config/supervisord.conf
collinwr: Set this to `info`

usaspending_api/download/management/commands/download_sqs_worker.py
collinwr: I broke a few of these out. Let me know what you think. Might end up taking the logging back out of that new function

usaspending_api/settings.py
collinwr: Checked on this with Danny. Looks like another change is needed on the infra side to add the override to the bulk downloads. No tracing is taking place in other management commands.

sethstoudenmier (approved):
Could probably remove some of the additional requirements but not going to block over that. LGTM!

sethstoudenmier (dismissed):

sethstoudenmier (approved):

PR #4264, opened 2025-01-31
Update subaward ES test setup

aguest-kc:
**Description:**
Update the Elasticsearch test setup when using Subawards.

**Technical details:**

    - [ ] Backend

```
2. API documentation updated

8. Jira Ticket [DEV-123](https://federal-spending-transparency.atlassian.net/browse/DEV-123):
No JIRA ticket for this fix.
```

### DIFF
--- a/usaspending_api/conftest_helpers.py
+++ b/usaspending_api/conftest_helpers.py
@@ -108,13 +108,12 @@ class TestElasticSearchIndex:
         match self.index_type:
             case "award":
                 es_id_field = "award_id"
-                primary_key_field = "award_id"
             case "subaward":
                 es_id_field = "broker_subaward_id"
-                primary_key_field = "broker_subaward_id"
-            case _:
+            case "transaction":
                 es_id_field = "transaction_id"
-                primary_key_field = "transaction_id"
+            case _:
+                raise Exception("No value for the `_id` field in Elasticsearch has been set")
 
         self.worker = TaskSpec(
             base_table=None,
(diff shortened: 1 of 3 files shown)


usaspending_api/conftest_helpers.py
```diff
@@ -103,16 +104,28 @@ def __init__(self, index_type):
...
         }
+
+        match self.index_type:
+            case "award":
+                es_id_field = "award_id"
+                primary_key_field = "award_id"
+            case "subaward":
+                es_id_field = "broker_subaward_id"
+                primary_key_field = "broker_subaward_id"
+            case _:
```
sethstoudenmier, line 115: Might be worth matching on "transaction" and let the default case raise an error to help avoid any future indexes overlooking this.

usaspending_api/conftest_helpers.py
```diff
@@ -103,16 +104,28 @@ def __init__(self, index_type):
             "write_alias": self.index_name + "-load-alias",
             "process_deletes": True,
         }
+
+        match self.index_type:
+            case "award":
+                es_id_field = "award_id"
```
sethstoudenmier, line 110: Do these need to be different variables since they are always the same?

sethstoudenmier (approved):
Overall LGTM. Two comments but not worth blocking over

usaspending_api/conftest_helpers.py
aguest-kc: Added a case for `transaction` and raised an error in all other cases.

usaspending_api/conftest_helpers.py
aguest-kc: I consolidated these variables. I originally left them separate in the future case that we wanted to use a non-primary key field as the ES `_id` field's value it would be more obvious.

sethstoudenmier (approved):

PR #4265, opened 2025-01-31
[DEV-11770] [DEV-11771] - Add account_download create and load commands, Add generate spark download command

zachflanders-frb:
**Description:**
This PR updates the load_query_to_delta command to allow the creation and loading of a table that represents an unfiltered account download.

**Technical details:**


usaspending_api/download/delta_models/account_download.py
```diff
@@ -0,0 +1,383 @@
...
+        transaction_search.place_of_performance_zip4a AS primary_place_of_performance_zip_code,
+        transaction_search.cfda_number,
+        transaction_search.cfda_title,
+        transaction_search.product_or_service_code,
+        transaction_search.product_or_service_description AS product_or_service_code_description,
+        transaction_search.naics_code,
+        transaction_search.naics_description,
+        transaction_search.national_interest_action AS national_interest_action_code,
+        transaction_search.national_interest_desc AS national_interest_action,
+        submission_attributes.reporting_agency_name AS reporting_agency_name,
```
zachflanders-frb, line 158: Will need to add the following columns to allow filtering against these columns to create the final download.
 - `submission_id`
 - `reporting_fiscal_period`
 - `reporting_fiscal_quarter`
 - `reporting_fiscal_year`
 - `quarter_format_flag`

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -291,6 +296,26 @@
...
+    "account_download": {
+        "model": None,
+        "is_from_broker": False,
+        "source_query": [account_download_load_sql_string],
+        "source_database": None,
+        "source_table": None,
+        "destination_database": "rpt",
+        "swap_table": "account_download",
+        "swap_schema": "rpt",
+        "partition_column": "id",
```
sethstoudenmier, line 308: Verify or add `# TODO` to verify the partition_column

usaspending_api/common/helpers/download_csv_strategies.py
```diff
@@ -108,7 +116,7 @@ def download_to_csv(
             raise e
         finally:
             Path(temp_file_path).unlink()
-        return [destination_path], count
+        return CSVDownloadMetadata([destination_path], row_count)
```
sethstoudenmier, line 119: Not worth fixing as part of this work, but worth calling out so that we remember. This change to return a dataclass will mean that the current download functionality will need to take this new return into account as well.

usaspending_api/common/helpers/download_csv_strategies.py
```diff
@@ -179,12 +188,12 @@ def download_to_csv(
...
             raise
         finally:
-            delete_s3_object(s3_bucket_name, s3_destination_path)
+            delete_s3_objects(s3_bucket_name, key_prefix=f"{s3_bucket_sub_path}/{destination_file_name}")
             if self.spark_created_by_command:
                 self.spark.stop()
         append_files_to_zip_file(final_csv_data_file_locations, download_zip_path)
         self._logger.info(f"Generated the following data csv files {final_csv_data_file_locations}")
-        return final_csv_data_file_locations, record_count
+        return CSVDownloadMetadata(final_csv_data_file_locations, record_count, column_count)
```
sethstoudenmier, line 196: Same comment as above

usaspending_api/download/delta_models/account_download.py
```diff
@@ -0,0 +1,391 @@
...
+        END AS submission_period,
+        treasury_appropriation_account.budget_function_title AS budget_function,
+        treasury_appropriation_account.budget_subfunction_title AS budget_subfunction,
+        financial_accounts_by_awards.transaction_obligated_amount AS transaction_obligated_amount,
+        CASE
+            WHEN
+                (
+                    (
+                        submission_attributes.quarter_format_flag = TRUE
+                        AND submission_attributes.reporting_fiscal_quarter = 4
```
sethstoudenmier, line 198: You are really close to this SQL being done. This download query should not filter by anything that would be input by a user. In this case the Fiscal Quarter, Period, and Year fall into that category. Similar to how the filtering by Fiscal Year had to be moved into the next query, these fields do as well.

sethstoudenmier: Similar comment applies to anywhere that is filtering by a specific Fiscal Period or Quarter.

sethstoudenmier (changes requested):
Few more changes on the download query that I don't think will impact your captured metrics, but should be corrected so we don't carry forward that error.

PR #4271, opened 2025-02-11
[DEV-11464] Move Subaward Search in Spending By Award to Elastic Search

evan-voogd-frb:
**Description:**
Switches the subaward option in `spending_by_award` from using the database to query results to using Elastic Search.

**Technical details:**
Mirrors the award implementation in `spending_by_award` which already uses Elastic Search, with slight modifications due to the different fields that can be returned by the subaward endpoint. The methods used are exact analogues to the award methods.

    - [x] Backend
8. [x] Jira Ticket [DEV-11464](https://federal-spending-transparency.atlassian.net/browse/DEV-11464):

```


```


usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -374,6 +392,59 @@ def query_elasticsearch(self) -> list:
...
+            )
+        # no values, within result window, use regular elasticsearch
+        else:
+            search = (
+                SubawardSearch().filter(filter_query).sort(*sorts)[record_num : record_num + self.pagination["limit"]]
+            )
+
+        response = search.handle_execute()
+
+        return response
```
zachflanders-frb, line 446: Could we try refactoring some of this duplciated code into its own method.  Method signature could look something like
```python
def query_elasticsearch(search: AwardSearch | SubawardSearch, filter_query: ES_Q, sorts: list[dict[str, str]]) -> Response:
```

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -516,6 +587,55 @@ def construct_es_response_for_prime_awards(self, response) -> dict:
...
+            "page_metadata": {
+                "page": self.pagination["page"],
+                "hasNext": has_next,
+                "last_record_unique_id": last_record_unique_id,
+                "last_record_sort_value": str(last_record_sort_value),
+            },
+            "messages": get_generic_filters_message(
+                self.original_filters.keys(), [elem["name"] for elem in AWARD_FILTER_NO_RECIPIENT_ID]
+            ),
+        }
```
zachflanders-frb, line 637: Similar to above, some of this shared logic could be placed into its own method and then called by both of these `construct_es_response` methods.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -374,6 +392,59 @@ def query_elasticsearch(self) -> list:
 
         return response
 
+    def query_elasticsearch_subawards(self) -> list:
```
zachflanders-frb, line 395: This method returns a `Response`
```suggestion
    def query_elasticsearch_subawards(self) -> Response:
```

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -516,6 +587,55 @@ def construct_es_response_for_prime_awards(self, response) -> dict:
             ),
         }
 
+    def construct_es_response_for_subawards(self, response) -> dict:
```
zachflanders-frb, line 590: Can we add/expand the type hints?
```suggestion
    def construct_es_response_for_subawards(self, response: Response) -> dict[str, Any]:
```

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -286,7 +304,7 @@ def populate_response(self, results: list, has_next: bool, models: List[dict]) -
             "messages": get_generic_filters_message(self.original_filters.keys(), [elem["name"] for elem in models]),
         }
 
-    def query_elasticsearch(self) -> list:
+    def query_elasticsearch_awards(self) -> list:
```
zachflanders-frb, line 307:
```suggestion
    def query_elasticsearch_awards(self) -> Response:
```

zachflanders-frb (commented):
👏 This is looking good!  I added some comments mainly focused on making sure we are as specific as possible with type hints and also some ideas for reducing duplicated code by moving shared logic to its own method when possible.

usaspending_api/search/v2/views/spending_by_award.py
evan-voogd-frb: Is my latest commit what you had in mind?

zachflanders-frb: Yes, look good.  👍

usaspending_api/search/v2/views/spending_by_award.py
evan-voogd-frb: Just finished this one, is the new commit what you had in mind?

aguest-kc:
Just a reminder that this PR will require a couple of data change tickets to be created.

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -563,7 +566,16 @@
         CONCAT(rl_state_fips.fips, rl_county_fips.county_numeric) AS legal_entity_county_fips,
         CONCAT(pop_state_fips.fips, pop_county_fips.county_numeric) AS place_of_perform_county_fips,
         UPPER(COALESCE(fpds.place_of_perform_county_na, fabs.place_of_perform_county_na)) AS pop_county_name,
-        tas.program_activities
+        tas.program_activities,
+        CASE
```
sethstoudenmier, line 570: Since you are doing a LEFT JOIN down below doesn't this CASE statement have the possibility to return `-R` or `-C` without a Recipient Hash?

sethstoudenmier: For example, what if a Subaward either doesn't contain a UEI / DUNS or it does but it doesn't match a value in the Recipient Profile table.

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -696,6 +708,75 @@
...
+                    NAMED_STRUCT(
+                        'id', fa.id,
+                        'account_title', fa.account_title,
+                        'federal_account_code', fa.federal_account_code
+                    )
+                )
+            )
+        ) AS federal_accounts,
+        COLLECT_SET(
+        DISTINCT CONCAT(
```
sethstoudenmier, line 730: Minor cleanup, but can you indent contents of the COLLECT_SET

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -696,6 +708,75 @@
...
+            'main=', COALESCE(taa.main_account_code, ''),
+            'ata=', COALESCE(taa.allocation_transfer_agency_id, ''),
+            'sub=', COALESCE(taa.sub_account_code, ''),
+            'bpoa=', COALESCE(taa.beginning_period_of_availability, ''),
+            'epoa=', COALESCE(taa.ending_period_of_availability, ''),
+            'a=', COALESCE(taa.availability_type_code, '')
+        )
+        ) AS tas_paths,
+        COLLECT_SET(
+        CONCAT(
```
sethstoudenmier, line 744: Minor cleanup, but can you indent contents of the COLLECT_SET

usaspending_api/search/delta_models/subaward_search.py
```diff
@@ -696,6 +708,75 @@
...
+        global_temp.treasury_appropriation_account taa
+    INNER JOIN int.financial_accounts_by_awards faba ON (taa.treasury_account_identifier = faba.treasury_account_id)
+    INNER JOIN global_temp.federal_account fa ON (taa.federal_account_id = fa.id)
+    INNER JOIN global_temp.toptier_agency agency ON (fa.parent_toptier_agency_id = agency.toptier_agency_id)
+    LEFT JOIN global_temp.ref_program_activity rpa ON (faba.program_activity_id = rpa.id)
+    WHERE
+        faba.award_id IS NOT NULL
+    GROUP BY
+        faba.award_id
+    ) TREASURY_ACCT ON (TREASURY_ACCT.award_id = a.id)
```
sethstoudenmier, line 779: Does this have to live on `subaward_search`? Currently the View that powers the Elasticsearch Subaward Index joins `subaward_search` and `award_search`, which would allow you to reference the fields from this join in the subaward view. This is what I am referring to: https://github.com/fedspendingtransparency/usaspending-api/blob/master/usaspending_api/database_scripts/etl/subaward_es_view.sql#L118

usaspending_api/awards/v2/lookups/lookups.py
```diff
@@ -57,9 +57,9 @@
     "Sub-Award Amount": "subaward_amount",
     "Awarding Agency": "awarding_toptier_agency_name",
     "Awarding Sub Agency": "awarding_subtier_agency_name",
-    "Prime Award ID": "piid",
+    "Prime Award ID": "award_piid_fain",
     "Prime Recipient Name": "awardee_or_recipient_legal",
-    "prime_award_recipient_id": "_prime_award_recipient_id",
+    "prime_award_recipient_id": "prime_award_recipient_id",
```
sethstoudenmier, line 62: Similar question about this change applying to the grants down below.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -51,9 +51,13 @@
 from usaspending_api.common.exceptions import UnprocessableEntityException
 from usaspending_api.search.filters.elasticsearch.filter import _QueryType
 from usaspending_api.search.filters.time_period.decorators import NewAwardsOnlyTimePeriod
-from usaspending_api.search.filters.time_period.query_types import AwardSearchTimePeriod
+from usaspending_api.search.filters.time_period.query_types import AwardSearchTimePeriod, SubawardSearchTimePeriod
 from usaspending_api.submissions.models import SubmissionAttributes
 
+from typing import Any
```
sethstoudenmier, line 57: Can you move this towards the top? Usually we will try to format the imports in the following order:

* Python built-in
* 3rd party
* USAspending imports

Where within each group we format such that you import an entire module above the imports that are more specific. For example:

```
import os
from typing import List

import pytest

from usaspending_api... import <FUNC_NAME>
```

sethstoudenmier: I will admit it is not the most consistent so I won't block over this, but it would be good to try and be consistent.

sethstoudenmier (changes requested):
Initial pass on the bulk of the changes. I will need to wrap up the review tomorrow.

usaspending_api/awards/v2/lookups/lookups.py
```diff
@@ -57,9 +57,9 @@
     "Sub-Award Amount": "subaward_amount",
     "Awarding Agency": "awarding_toptier_agency_name",
     "Awarding Sub Agency": "awarding_subtier_agency_name",
-    "Prime Award ID": "piid",
+    "Prime Award ID": "award_piid_fain",
```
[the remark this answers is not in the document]
evan-voogd-frb: I made this change as `piid` wasn't in the Elastic Search index for local testing. If it's in the Delta Model it's possible that's an issue with my environment or something.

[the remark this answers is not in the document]
evan-voogd-frb: The reason this change is only in one mapping is that my code doesn't ever touch the grant_subaward_mapping and uses the contract_subaward_mapping as they have the same fields in the API contract. I can update the grant mapping as well, but it shouldn't affect functionality on the endpoint.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -116,14 +121,31 @@ def post(self, request):
...
         else:
-            self.last_record_unique_id = json_request.get("last_record_unique_id")
-            self.last_record_sort_value = json_request.get("last_record_sort_value")
-            raw_response = self.construct_es_response_for_prime_awards(self.query_elasticsearch())
+            return Response(self.construct_es_response_for_prime_awards(self.query_elasticsearch_awards()))
+
+    def post_subawards(self) -> Response:
+        try:
+            return Response(self.construct_es_response_for_subawards(self.query_elasticsearch_subawards()))
+        except KeyError:
```
sethstoudenmier, line 135: I believe this check on the request structure should be handled by line 101 where it validates the request data. Ideally we would update the TinyShield model (see [here](https://github.com/fedspendingtransparency/usaspending-api/blob/master/usaspending_api/common/validator/award_filter.py#L108)) to account for any changes to the incoming request structure.

sethstoudenmier (changes requested):
Some cleanup I won't block on, but then there are some requested changes with regards to the `try...except` and the mapping.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -77,6 +81,7 @@
         "type_code_to_field_map": {"procurement": contract_subaward_mapping, "grant": grant_subaward_mapping},
         "annotations": {"_prime_award_recipient_id": annotate_prime_award_recipient_id},
         "filter_queryset_func": subaward_filter,
+        "elasticsearch_type_code_to_field_map": contract_subaward_mapping,
```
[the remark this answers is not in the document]
evan-voogd-frb: I think for now I'm going to remove/rename.

usaspending_api/search/v2/views/spending_by_award.py
evan-voogd-frb: Interesting that line `validate_request_data` is supposed to do this - I added this try-catch as it was failing a test that checked for this. I'll look into why `validate_request_data` isn't passing this test.

usaspending_api/awards/v2/lookups/lookups.py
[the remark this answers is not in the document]
sethstoudenmier: > I made this change as `piid` wasn't in my Elastic Search index for local testing. If it's in the Delta Model it's possible that's an issue with my environment or something.

Ah I confused the delta model with the ES Index view. Good to move forward with the `award_piid_fain` field, but we should cleanup this use of contracts and grants for Subawards so that we don't create future confusion. If spending_by_award is the only place that uses this then I would suggest consolidating into a single mapping and remove the grants mapping for subawards.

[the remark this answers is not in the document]
evan-voogd-frb: Yep, that's how I'm going to move forward with it, thanks for pointing that out!

sethstoudenmier (changes requested):

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -179,6 +183,13 @@ def validate_request_data(request_data):
         for m in models:
             if m["name"] in ("award_type_codes", "fields"):
                 m["optional"] = False
+            elif "subawards" in request_data and request_data["subawards"] and m["name"] == "time_period":
```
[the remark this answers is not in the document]
evan-voogd-frb: Just saw your later comment, ignore this one.

sethstoudenmier (approved):

sethstoudenmier (approved):

PR #4284, opened 2025-02-18
[DEV-11590] Sort location autocomplete results by occurrences

aguest-kc:
**Description:**
Added an `occurrences` field to the location index in Elasticsearch. This new field is then used to sort the results that match the given `search_text`, in descending order. This puts the locations that appear the most in our Transaction data at the top of the search results.

**Technical details:**

    - [x] Backend
8. [x] Jira Ticket [DEV-11590](https://federal-spending-transparency.atlassian.net/browse/DEV-11590):

```
2. API documentation updated

```

### DIFF
--- a/usaspending_api/database_scripts/etl/location_delta_view.sql
+++ b/usaspending_api/database_scripts/etl/location_delta_view.sql
@@ -11,11 +11,15 @@ WITH country_cte AS (
 			ELSE
 				pop_country_name
 		END AS location_string,
-		'country' AS location_type
+		'country' AS location_type,
+        COUNT(*) AS occurrences
 	FROM
 		rpt.transaction_search
 	WHERE
 		pop_country_name IS NOT NULL
+    GROUP BY
+        location_string,
+        location_type
 	UNION
 	SELECT
 		CASE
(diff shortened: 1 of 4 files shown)


usaspending_api/database_scripts/etl/location_delta_view.sql
```diff
@@ -11,11 +11,15 @@ WITH country_cte AS (
 			ELSE
 				pop_country_name
 		END AS location_string,
-		'country' AS location_type
+		'country' AS location_type,
+        COUNT(*) AS occurrences
```
sethstoudenmier, line 15: Do you mind indenting this to be at the same level as the rest of the SELECT?

sethstoudenmier: Can you actually fix some of the indentation elsewhere in this SQL?

sethstoudenmier (changes requested):
Just some cleanup requests for the SQL

usaspending_api/database_scripts/etl/location_delta_view.sql
aguest-kc: I'm doing some testing and might end up just closing this PR. While this change gives us more "realistic" results in _some_ cases, in other cases it completely messes up the "realistic" results.

For example, if we sort by `occurrences` and you search for `Denver` then it does put `Denver, Colorado` at the top as you would (most likely) expect. However, if you then search for `New York City` it puts `Long Island City, New York` at the top because this location occurs more often than `New York City, New York` (which is actually the 5th result when sorted by occurrences).

There's not a perfect solution and in this testing it seems like this may end up causing confusion more often than not.

PR #4289, opened 2025-02-20
[DEV-11950] Refactor QueryWithFilters and search endpoints views to be more DRY

loreleitrimberger:
**Description:**
Refactors methods in query_with_filters.py to be more generic instead of having separate methods for each query type. Ex: ` generate_<transactions/subawards/accounts/awards>_eleasticsearch_query` is now `generate_elasticsearch_query` and `query_elasticsearch_for_<awards/transactions/subawards>` is `query_elasticsearch`. These are used throughout the code base, so most is implementing the new methods

**Technical details:**
I added an `__init__` method to the QueryWithFilters class to have the query type as an attribute of self to more easily consolidate the methods

    - [ ] Backend
8. [x] Jira Ticket [DEV-11950](https://federal-spending-transparency.atlassian.net/browse/DEV-11950)
    - N/A Performance evaluation of affect (API | Script | Download)
    - N/A Before / After data comparison

```
 data should stay the same
```


usaspending_api/agency/v2/views/award_count.py
```diff
@@ -129,7 +129,8 @@ def _get_award_counts_response(self, filters: List[dict], filter_options: dict)
         Returns:
             List[dict]: A dictionary containing award types and the number of their awards.
         """
-        filter_query = QueryWithFilters.generate_awards_elasticsearch_query(filters, **filter_options)
+        query_with_filters = QueryWithFilters(_QueryType.AWARDS)
```
sethstoudenmier, line 132: This comment is unrelated to this line specifically, but applies with its use of `_QueryType`.

If it is not too much of a pain, would you mind updating `_QueryType` to instead be `QueryType`? In our codebase we have a few instances where the `_` is prepended to signify a "private" method or property, but then it gets used in a lot of places outside of the file where it is defined. It would be good to start cleaning this up so that we are more intentional and confirming to Python standards.

usaspending_api/common/query_with_filters.py
```diff
@@ -893,34 +900,40 @@ def generate_elasticsearch_query(cls, filter_values: List[str], query_type: _Que
...
-        _SetAsideTypeCodes.underscore_name: _SetAsideTypeCodes,
-        _ExtentCompetedTypeCodes.underscore_name: _ExtentCompetedTypeCodes,
-        _DisasterEmergencyFundCodes.underscore_name: _DisasterEmergencyFundCodes,
-        _QueryText.underscore_name: _QueryText,
-        _NonzeroFields.underscore_name: _NonzeroFields,
-        _ProgramActivities.underscore_name: _ProgramActivities,
-    }
+
+    @property
+    def filter_lookup(self) -> Dict[str, _Filter]:
```
sethstoudenmier, line 905: Won't block over this but as of Python 3.9 the use of `dict` is preferred over `Dict` (see [here](https://docs.python.org/3/library/typing.html#typing.Dict)).

usaspending_api/common/query_with_filters.py
```diff
@@ -930,10 +943,38 @@ class QueryWithFilters:
...
+            self.default_options = {"nested_path": "financial_accounts_by_award"}
+        elif self.query_type == _QueryType.TRANSACTIONS:
+            time_period_obj = TransactionSearchTimePeriod(
+                default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_MIN_DATE
+            )
+        elif self.query_type == _QueryType.AWARDS:
+            time_period_obj = AwardSearchTimePeriod(
+                default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_MIN_DATE
+            )
+        else:
```
sethstoudenmier, line 959: Since this such a widely used class can you make this conditional check for subawards an `elif` instead of `else`? You can then throw an error in an else statement to avoid PyCharm or other IDEs from yelling at you. That way we can hopefully make this easier to follow in the future.

sethstoudenmier: Actually, building on that point should this do anything if the QueryType is for the location index? Probably don't want to use the `else` at all and instead only check for Subawards with another `elif`.

usaspending_api/common/query_with_filters.py
```diff
@@ -930,10 +943,38 @@ class QueryWithFilters:
...
+            time_period_obj = SubawardSearchTimePeriod(
+                default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_MIN_DATE
+            )
 
+        if time_period_obj is not None:
```
sethstoudenmier, line 964: Subawards do not support the New Awards Only filter, this should be limited to just Awards and Transactions query type

usaspending_api/common/query_with_filters.py
```diff
@@ -942,27 +983,27 @@ def _generate_elasticsearch_query(cls, filters: dict, query_type: _QueryType, **
         filters_copy = copy.deepcopy(filters)
 
         # tas_codes are unique in that the same query is spread across two keys
-        must_queries = cls._handle_tas_query(must_queries, filters_copy, query_type)
+        must_queries = self._handle_tas_query(must_queries, filters_copy, self.query_type)
```
sethstoudenmier, line 986: You should be able to update the `_handle_tas_query` method to be an instance method instead of a class method. And then that would also allow you to remove the `self.query_type` from the parameters as you could access it in the instance method.

usaspending_api/search/v2/views/spending_by_award_count.py
```diff
@@ -18,12 +18,10 @@
     get_generic_filters_message,
 )
 from usaspending_api.common.query_with_filters import QueryWithFilters
+from usaspending_api.search.filters.elasticsearch.filter import _QueryType
```
sethstoudenmier, line 21: Really small cleanup, can you keep this in alphabetical order for the usaspending modules?

sethstoudenmier (changes requested):
Overall changes look great! Not all comments are blocking as they are more about cleanup. Let me know if you have any questions.

usaspending_api/common/query_with_filters.py
```diff
@@ -930,10 +943,38 @@ class QueryWithFilters:
...
+            )
+        elif self.query_type == QueryType.AWARDS:
+            time_period_obj = AwardSearchTimePeriod(
+                default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_MIN_DATE
+            )
+        elif self.query_type == QueryType.SUBAWARDS:
+            time_period_obj = SubawardSearchTimePeriod(
+                default_end_date=settings.API_MAX_DATE, default_start_date=settings.API_MIN_DATE
+            )
+            self.default_options = dict()
```
sethstoudenmier, line 963: Did you mean to add the `time_period_obj` into the default options?

usaspending_api/common/query_with_filters.py
```diff
@@ -930,10 +943,38 @@ class QueryWithFilters:
...
 
+        if time_period_obj is not None and (self.query_type == QueryType.AWARDS or self.query_type == QueryType.TRANSACTIONS):
+            new_awards_only_decorator = NewAwardsOnlyTimePeriod(
+                time_period_obj=time_period_obj, query_type=self.query_type
+            )
+            self.default_options = {"time_period_obj": new_awards_only_decorator}
+        if self.query_type == QueryType.ACCOUNTS:
+            self.default_options = (
```
sethstoudenmier: You're already doing this on Line 950

usaspending_api/common/query_with_filters.py
```diff
@@ -930,10 +943,38 @@ class QueryWithFilters:
...
+                {"nested_path": "financial_accounts_by_award"}
+            )
+
+    def generate_elasticsearch_query(self, filters: dict, **options) -> ES_Q:
+        options = options if len(options) > 0 else self.default_options
```
sethstoudenmier, line 976: I am curious, what problems did you run into with your previous change that resulted in you going to this? Referencing this: https://github.com/fedspendingtransparency/usaspending-api/pull/4289/commits/30107e164cd2af3ab97351320bf4f3bd9215aa6d#diff-45ff8631ebb70d95ce8525bfa64ae4d79edb9d19b13088cafbdd0e996078b11fL976

sethstoudenmier: I believe what you meant to use in your previous commit was `options = {**self.default_options, **options}`, that way any user provided options will override the default options that are spread across the dictionary.

sethstoudenmier (changes requested):
Just some minor changes left

PR #4312, opened 2025-03-12
Ftr/dev 10981 spending by subawards grouped endpoint

loreleitrimberger:
**Description:**
Creates a new API endpoint `/api/v2/search/spending_by_subaward_grouped` that returns a list containing `award_id`, `award_generated_internal_id`, `subaward_count`, and `subaward_obligation`. This will be used to in Award Search 2.0 for the table results when filtered on subawards

**Technical details:**
This endpoint accepts all of the same filters as the `/api/v2/search/spending_by_award` endpoint when filtering by Subawards. A new api contract was made and from line 149 to the end is identical to spending_by_award since the filter objects are the same

    - [ ] Backend
    - N/A Frontend <OPTIONAL>
    - N/A Operations <OPTIONAL>
8. [x] Jira Ticket [DEV-10981](https://federal-spending-transparency.atlassian.net/browse/DEV-10981):
    - N/A Performance evaluation of affected (API | Script | Download)
    - N/A Before / After data comparison

```
```


usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,154 @@
...
+        search_unique = SubawardSearch().filter(filter_query)
+        search_unique.aggs.bucket("results", unique_award_key)
+        response_unique = search_unique.handle_execute()
+
+        if response_sum is None or response_unique is None:
+            raise Exception("Breaking generator, unable to reach cluster")
+
+        # response_sum and response_unique will have the same number indices since they're coming both being aggregated within the original es query and neither is compressing the results to a smaller list
+        results = []
+        count = 0
```
loreleitrimberger, line 134: I don't think this is a good way to do this. I wasn't sure how to grab the unique_award_key field along with the award_piid_fain in the aggregation. Getting them together should cause the same results. I'll keep looking into this

loreleitrimberger: From what I'm seeing multi-term might work but it's not being found when I try to use it

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,145 @@
...
+from usaspending_api.common.validator.award_filter import AWARD_FILTER_NO_RECIPIENT_ID
+from usaspending_api.common.validator.pagination import PAGINATION
+from usaspending_api.common.validator.tinyshield import TinyShield
+from usaspending_api.search.filters.elasticsearch.filter import QueryType
+from usaspending_api.search.filters.time_period.query_types import SubawardSearchTimePeriod
+
+logger = logging.getLogger(__name__)
+
+
+class subaward_grouped_model:
```
zachflanders-frb, line 26: I really like the use of a class here to represent this data structure. Could use `@dataclass ` decorator to simplify the syntax. Also python naming best practice is to use TitleCase in class names e.g. `SubawardGroupedModel`

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,145 @@
...
+    total amount of subaward obligation within each award and each award's generated internal id
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/search/spending_by_subaward_grouped.md"
+
+    @cache_response()
+    def post(self, request):
+        """Return all subawards matching given awards"""
+
+        self.original_filters = request.data.get("filters")
```
zachflanders-frb, line 47: In general, it is discouraged to create new properties on `self` outside of the `__init__` method.  You could add an `__init__` method to this class that calls the `APIView`'s `__init__` first with `super().__init__()` and then adds these properties e.g.
```python
        self.original_filters: dict[str, Any] = {}
        <set initial values of other properties of self here>
```

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,145 @@
...
+        search_sum.aggs.bucket("award_id", terms_aggregation)
+        response = search_sum.handle_execute()
+
+        if response is None:
+            raise Exception("Breaking generator, unable to reach cluster")
+
+        results = []
+        count = 0
+        for result in response["aggregations"]["award_id"]["buckets"]:
+            award_generated_internal_id = response["hits"]["hits"][count]["_source"]["unique_award_key"]
```
zachflanders-frb, line 130: Instead of defining a `count` variable and iterating it, you can use python's `enumerate` to get the count this way:
```suggestion
        for i, result in enumerate(response["aggregations"]["award_id"]["buckets"]):
            award_generated_internal_id = response["hits"]["hits"][i]["_source"]["unique_award_key"]
```
Can we also check to make sure that the `award_piid_fain` in `result.key` matches the value on the hit to ensure that we're getting the right `unique_award_key`?  Relying on this list of hits always being in the right order seems like it could be a little fragile.  Alternatively, this could be a nested `agg` by adding another `.metric("unique_award_key", "terms", field="unique_award_key")` and referenced like ` result.unique_award_key.buckets[0].key`

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
zachflanders-frb, line 1: Can we add type hints for the parameters and return values where possible?  Some of the methods in this file are missing type hints and some have them.

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,154 @@
...
+    total amount of subaward obligation within each award and each award's generated internal id
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/search/spending_by_subaward_grouped.md"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.original_filters: dict[str, Any] = {}
+        self.filters: dict[str, Any] = {}
+        self.subaward_filters: dict[str, Any] = {}
```
zachflanders-frb, line 49: The `subaward_filters` property does not seem to be used.

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,154 @@
...
+                "text_type": "search",
+                "default": "award_id",
+            },
+        ]
+
+        # Accepts the same filters as spending_by_award
+        spending_by_subaward_grouped_models.extend(copy.deepcopy(AWARD_FILTER_NO_RECIPIENT_ID))
+        spending_by_subaward_grouped_models.extend(
+            copy.deepcopy([model for model in PAGINATION if model["name"] != "sort"])
+        )
```
zachflanders-frb, line 104: This could be defined in the `__init__` as `self.models`.  Then it could be referenced here, which would also allow this method to just return the validated response instead of the response and the models.

usaspending_api/api_contracts/contracts/v2/search/spending_by_subaward_grouped.md
```diff
@@ -0,0 +1,291 @@
+FORMAT: 1A
+HOST: https://api.usaspending.gov
+
+# Spending By Subaward Grouped
+
+## POST
+
+This endpoint takes fields and award filters and returns a list containing filtered award ids, the number of subawards in each award, the total amount of obligations of all the subawards, and the award unique id
```
sethstoudenmier, line 8:
```suggestion
This endpoint takes award filters and returns a list containing filtered award ids, the number of subawards in each award, the total amount of obligations of all the subawards, and the award unique id.
```

usaspending_api/api_contracts/contracts/v2/search/spending_by_subaward_grouped.md
```diff
@@ -0,0 +1,291 @@
...
+        + `filters` (required, AdvancedFilterObject)
+        + `limit` (optional, number)
+            How many results are returned
+        + `page` (optional, number)
+            The page number that is currently returned.
+        + `sort`: Any field within the result can be sort (optional, enum[string])
+            The field on which to order `results` in the response.
+            + Default: `award_id`
+            + Members
+                + `award_id`
```
sethstoudenmier, line 28: Won't block over this, but if you have time would like to try and keep list like this and the response alphabetized at least in the documentation, so it is easier to find what you need.

zachflanders-frb (dismissed):

usaspending_api/search/tests/integration/test_spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,159 @@
...
+def test_spending_by_subaward_grouped_legacy_filter(client, monkeypatch, elasticsearch_subaward_index):
+    setup_elasticsearch_test(monkeypatch, elasticsearch_subaward_index)
+
+    resp = client.post(
+        "/api/v2/search/spending_by_subaward_grouped",
+        content_type="application/json",
+        data=json.dumps({"page": 1, "limit": 2, "sort": "award_id", "filters": legacy_filters()}),
+    )
+
+    assert resp.status_code == status.HTTP_200_OK
```
sethstoudenmier, line 104: Please update the test cases that check for a 200 status to validate some form of data on the response. We do use this pattern of test cases elsewhere in the codebase, but they don't test much in regard to functionality.

usaspending_api/search/tests/integration/test_spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,159 @@
...
+def test_spending_by_subaward_grouped_non_legacy_filter(client, monkeypatch, elasticsearch_subaward_index):
+    setup_elasticsearch_test(monkeypatch, elasticsearch_subaward_index)
+
+    resp = client.post(
+        "/api/v2/search/spending_by_subaward_grouped",
+        content_type="application/json",
+        data=json.dumps({"page": 1, "limit": 2, "sort": "award_id", "filters": non_legacy_filters()}),
+    )
+
+    assert resp.status_code == status.HTTP_200_OK
```
sethstoudenmier, line 117: Same as the above test case

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def build_elasticsearch_search_with_aggregation(self, filter_query: ES_Q) -> dict[str, Any]:
+        # Aggregate the ES query to group the subaward values by their prime award
+        terms_aggregation = A("terms", field="award_piid_fain")
+
+        # Get the unique_award_key for each prime award
+        terms_aggregation.metric("unique_award_key", "terms", field="unique_award_key")
+
+        # Sum the subaward amount within each prime award
+        terms_aggregation.metric("subaward_obligation", "sum", field="subaward_amount")
```
sethstoudenmier, line 126: We have to do some special handling of float sums with Elasticsearch otherwise you get value such as below. You can find examples in the different spending_by_ endpoints and spending_over_time.
```
{
            "award_id": "NNM17AA12C",
            "subaward_count": 228,
            "award_generated_internal_id": "CONT_AWD_NNM17AA12C_8000_-NONE-_-NONE-",
            "subaward_obligation": 28947569.990000002
        },
```

sethstoudenmier (changes requested):
Left some comments around cleanup that aren't blockers. However, the float sums do need to be addressed.

usaspending_api/search/v2/urls.py
```diff
@@ -23,4 +24,5 @@
     re_path(r"^spending_by_transaction", es.SpendingByTransactionVisualizationViewSet.as_view()),
     re_path(r"^spending_over_time", SpendingOverTimeVisualizationViewSet.as_view()),
     re_path(r"^transaction_spending_summary", es.TransactionSummaryVisualizationViewSet.as_view()),
+    re_path(r"^spending_by_subaward_grouped", SpendingBySubawardGroupedVisualizationViewSet.as_view()),
```
sethstoudenmier, line 27: Please put this in alphabetical order:

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -0,0 +1,151 @@
+import copy
+import logging
+from sys import maxsize
+from typing import Any
+from dataclasses import dataclass
+from decimal import Decimal
+from django.conf import settings
+from elasticsearch_dsl import A
```
sethstoudenmier, line 8: Won't block over this, but would be good to combine the two `elasticsearch_dsl` imports into one line. Realizing now how many places across the codebase we split this up for some reason.

sethstoudenmier (changes requested):
One really small change for the order of the URLs, otherwise LGTM!

sethstoudenmier (dismissed):

sethstoudenmier (approved):

PR #4313, opened 2025-03-12
[DEV-12237] - Update Spending by Transaction Grouped

zachflanders-frb:
**Description:**
This PR reworks this endpoint to support the new view on Advanced Search 2.0.  Previously this endpoint queried for individual transactions that make up a group, which created pagination problems.  The reworked endpoint just aggregates information about transactions in the group.

**Technical details:**
The keyword filter and sort parameter are no longer required.  The endpoint is able to search on `award_generated_internal_id` in addition to `award_id`, `transaction_count`, and `transaction_obligation`.

    - [ ] Backend    
8. [x] Jira Ticket [DEV-12237](https://federal-spending-transparency.atlassian.net/browse/DEV-12237):

```
```


usaspending_api/search/v2/views/spending_by_transaction_grouped.py
```diff
@@ -0,0 +1,165 @@
...
+            case "award_id":
+                key = item.key
+            case "transaction_count":
+                key = item.doc_count
+            case "award_generated_internal_id":
+                key = item.award_generated_internal_id.buckets[0].key
+            case "transaction_obligation":
+                key = item.transaction_obligation.value
+            case _:
+                raise ValueError("Invalid sort key")
```
zachflanders-frb, line 164: Codeclimate does not like the match/case syntax
image
This is [valid syntax in python 3.10](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching).

I could update to use an if/else statement, but I think the match/case syntax is better suited for this use.
> While an “imperative” series of instructions using nested “if” statements could be used to accomplish something similar to structural pattern matching, it is less clear than the “declarative” approach. Instead the “declarative” approach states the conditions to meet for a match and is more readable through its explicit patterns.

aguest-kc: This might be something we can update to in Github Actions, but if I remember correctly CodeClimate doesn't support this version of radon.

We force merged the last PR that failed because of this, since it's valid syntax.

evan-voogd-frb (dismissed):

.codeclimate.yml
```diff
@@ -55,6 +55,7 @@ engines:
         enabled: false
   radon:
     enabled: true
+    channel: beta
```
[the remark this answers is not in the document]
sethstoudenmier: We may need to revisit our use of codeclimate here or at least our use of radon. I ran into this during the python upgrade as well and even tried the SonarPython alternative, but it found a lot of issues we would need to address.

usaspending_api/api_docs/markdown/endpoints.md
```diff
@@ -26,171 +26,171 @@ The currently available endpoints are listed in the following table.
...
-|[/api/v2/awards/last_updated/](/api/v2/awards/last_updated/)|GET| Returns date of last update |
-|[/api/v2/budget_functions/list_budget_functions/](/api/v2/budget_functions/list_budget_functions/)|GET| Returns all Budget Functions associated with a TAS, ordered by Budget Function code |
-|[/api/v2/budget_functions/list_budget_subfunctions/](/api/v2/budget_functions/list_budget_subfunctions/)|POST| Returns all Budget Functions associated with a TAS, ordered by Budget Function code |
-|[/api/v2/bulk_download/awards/](/api/v2/bulk_download/awards/)|POST| Generates zip file for download of award data in CSV format |
-|[/api/v2/bulk_download/list_agencies/](/api/v2/bulk_download/list_agencies/)|POST| Lists all the agencies and the subagencies or federal accounts associated under specific agencies |
-|[/api/v2/bulk_download/list_monthly_files/](/api/v2/bulk_download/list_monthly_files/)|POST| Lists the monthly files associated with the requested params |
+| Endpoint | Methods | Description                                                                                                                                                              |
```
sethstoudenmier, line 29: Did you update the spacing to try and keep the last `|` on each line at the same spot? If so, I commend the effort, but this is one of those things that may be hard to keep perfect as future updates are made. Definitely won't block on this, but I would just set expectations for this file.

usaspending_api/search/v2/views/spending_by_transaction_grouped.py
```diff
@@ -0,0 +1,165 @@
...
+    the transactions to be filtered.
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/search/spending_by_transaction_grouped.md"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        models = [
+            {
+                "name": "fields",
```
sethstoudenmier, line 44: I don't believe this should be in use anymore. If that is the case, can you remove it from the TinyShield model and also the API Contract?

.codeclimate.yml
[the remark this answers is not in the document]
zachflanders-frb: @sethstoudenmier Should I revert this change since it doesn't result in the version that we would need for this to pass?  The documentation says that adding beta will result in radon version `4.3.2` , but that documentation was last updated 2 years ago, so I don't know how much to trust it.  I have so far not been able to verify which version is being used.

usaspending_api/search/v2/views/spending_by_transaction_grouped.py
```diff
@@ -0,0 +1,165 @@
...
+                "default": "transaction_obligation",
+            },
+        ]
+        models.extend(copy.deepcopy(AWARD_FILTER))
+        models.extend(copy.deepcopy([model for model in PAGINATION if model["name"] != "sort"]))
+        for m in models:
+            if m["name"] == "award_type_codes":
+                m["optional"] = False
+        self.tinysheld = TinyShield(models)
+        self.sort_by: str = "transaction_obligation"
```
sethstoudenmier, line 71: Not something to block over but setting this here could cause confusion since the default value is set in the TinyShield model and thus this is overwritten when the `sort` value is retrieved from the `validated_payload`.

usaspending_api/api_contracts/contracts/v2/search/spending_by_transaction_grouped.md
```diff
@@ -17,20 +17,25 @@ Searches for transaction records based on a provided set of filters and groups t
 
     + Attributes (object)
         + `filters` (required, AdvancedFilterObject)
-            These filters are applied at the Transaction level (`children`). They will not affect the Prime Awards (`results`) themselves. Must at least provide the following in the filters object, `keywords` and `award_type_codes`.
+            These filters are applied at the Transaction level. They will not affect the Prime Awards themselves. Must at least provide `award_type_codes` in the filters object.
         + `fields` (required, FieldNameObject)
```
sethstoudenmier, line 21: If we can remove `fields` then I would say remove this as well; and other usages following this in the contract

sethstoudenmier (changes requested):
Only requested change is on the `fields` array in the request if that is something we are safe to remove.

usaspending_api/api_docs/markdown/endpoints.md
zachflanders-frb: Something must have auto formatted this file.  I did not intend to make these changes. I only intended to update the description of `/api/v2/search/spending_by_transaction_grouped/`

sethstoudenmier (approved):

.codeclimate.yml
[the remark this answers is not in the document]
sethstoudenmier: @zachflanders-frb I am fine with this as you have it. Maybe it will pick up a new version when they support Python 3.10

usaspending_api/search/v2/views/spending_by_transaction_grouped.py
```diff
@@ -0,0 +1,157 @@
...
+            validated_payload["filters"]["keyword_search"] = [
+                es_minimal_sanitize(keyword) for keyword in validated_payload["filters"]["keywords"]
+            ]
+            validated_payload["filters"].pop("keywords")
+        query_with_filters = QueryWithFilters(QueryType.TRANSACTIONS)
+        filter_query = query_with_filters.generate_elasticsearch_query(validated_payload["filters"])
+        search = TransactionSearch().filter(filter_query)
+        (
+            search.aggs.bucket("group_by_prime_award", A("terms", field="display_award_id"))
+            .metric("transaction_obligation", A("sum", field="federal_action_obligation"))
```
sethstoudenmier, line 105: @zachflanders-frb I didn't remember this until a test of the spending_by_subaward_grouped endpoint pointed it out. We need to get a follow-up to this to handle Elasticsearch's sum of float fields. There are few endpoints that have similar logic that you should be able to pulled from. A simple `sum` aggregation on a float field can result with values such as the following:
```{
            "award_id": "0003",
            "transaction_count": 351,
            "transaction_obligation": 133045351.53999999,
            "award_generated_internal_id": "CONT_AWD_0003_9700_FA865010D7015_9700"
        },
        {
            "award_id": "0001",
            "transaction_count": 526,
            "transaction_obligation": 104008197.22999999,
            "award_generated_internal_id": "CONT_AWD_0001_9700_FA865016D5402_9700"
        },
```
(these are taken from my local when running your branch)

PR #4317, opened 2025-03-13
[DEV-12075] Narrow keyword and award description results

evan-voogd-frb:
**Description:**
Improves the keywords and description filters such that when searching `booz allen` it only returns results that include the exact search criteria, in its exact order. Additionally, it continues to support partial matches.

**Technical details:**
Accomplishes this by using `multi_match` queries with the `phrase_phrefix` type in a similar manner to #4262. Additionally, I added integration tests to ensure that this functionality stays consistent.

    - [ ] Backend
8. [X] Jira Ticket [DEV-12075](https://federal-spending-transparency.atlassian.net/browse/DEV-12075):

```
2. API documentation updated
6.  Frontend <OPTIONAL>
Doesn't impact endpoint interface

Doesn't impact matviews

No changes necessary to data
```

### DIFF
--- a/usaspending_api/common/query_with_filters.py
+++ b/usaspending_api/common/query_with_filters.py
@@ -40,11 +40,14 @@ class _SubawardsKeywords(_Filter):
 
         def keyword_parse(keyword):
             queries = []
-            queries.append(ES_Q("match", sub_awardee_or_recipient_legal=keyword))
-            queries.append(ES_Q("match", product_or_service_description=keyword))
-            queries.append(ES_Q("match", subaward_description=keyword))
+            fields = [
+                "sub_awardee_or_recipient_legal",
+                "product_or_service_description",
+                "subaward_description",
+                "subaward_number",
+            ]
+            queries.append(ES_Q("multi_match", query=keyword, fields=fields, type="phrase_prefix"))
             queries.append(ES_Q("match", award_piid_fain=keyword))
-            queries.append(ES_Q("match", subaward_number=keyword))
             if len(keyword) == 4 and PSC.objects.filter(code=keyword).exists():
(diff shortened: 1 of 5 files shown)


usaspending_api/search/tests/integration/test_spending_by_award.py
zachflanders-frb, line 1: Can we add a test to ensure that an award is not returned if it has the words in the wrong order? E.g. "test the" should not return any results,

zachflanders-frb (commented):
TIL about `multi_match` and `phrase_prefix`.  Nice job!

usaspending_api/common/query_with_filters.py
```diff
@@ -130,7 +137,7 @@ def generate_elasticsearch_query(cls, filter_values: str, query_type: QueryType,
             query += r"\*"
         else:
```
aguest-kc, line 138: The `if/else` block can be dropped, because `multi_match` doesn't support wildcards in the query text and will treat them as normal characters. So a search for `test` is being turned into a search for `test*`.

usaspending_api/common/query_with_filters.py
```diff
@@ -130,7 +137,7 @@ def generate_elasticsearch_query(cls, filter_values: str, query_type: QueryType,
...
             query += "*"
```
aguest-kc, line 139: The `if/else` block can be dropped, because `multi_match` doesn't support wildcards in the query text and will treat them as normal characters. So a search for `test` is being turned into a search for `test*`.

sethstoudenmier (approved):
After some brief local testing the changes LGTM.

aguest-kc (approved):

zachflanders-frb (approved):

PR #4339, opened 2025-04-03
[DEV-12376] Updated ES delete logging and corrected use of pipeline timestamp

sethstoudenmier:
**Description:**
Fix a bug in our pipeline around the use of the load dates for deleting records from Elasticsearch.

**Technical details:**
The changes fall into the following buckets:

* Test cases
* Support running these commands locally for test
* Add more verbose logging
   * `--verbosity` or (`-v`) is built into the Django command argparse; we simply use it to set our own `verbose` config regardless of verbosity level provided (see [here](https://docs.djangoproject.com/en/5.2/ref/django-admin/#cmdoption-verbosity))
* Cleanup and move around where some items are logged
* Update the ES Deletes to use the new `transaction_fabs` and `transaction_fpds` external load dates since the pipeline is running in Delta now
   * Future improvement would be to include the load date names as part of the argparse, but this was already involving a decent amount of changes

    - [ ] Backend
8. [x] Jira Ticket [DEV-12376](https://federal-spending-transparency.atlassian.net/browse/DEV-12376):

```
```

### DIFF
--- a/usaspending_api/broker/helpers/last_load_date.py
+++ b/usaspending_api/broker/helpers/last_load_date.py
@@ -57,7 +57,7 @@ def get_earliest_load_date(keys, default=None, format_func: callable = (lambda _
 
     if earliest_date is None:
         logger.warning(
-            format_func("No earliest load date could be calculated because no dates for keys `{keys}` were found!")
+            format_func(f"No earliest load date could be calculated because no dates for keys `{keys}` were found!")
         )
         return default
 
--- a/usaspending_api/broker/helpers/last_load_date.py
+++ b/usaspending_api/broker/helpers/last_load_date.py
@@ -9,7 +9,7 @@ from usaspending_api.common.helpers.date_helper import cast_datetime_to_utc
 logger = logging.getLogger("script")
 
 
-def get_last_load_date(key, lookback_minutes=None, default=None):
+def get_last_load_date(key, lookback_minutes=None, default=None, format_func: callable = (lambda _: _)):
     """
     Retrieve the last_load_date from the USAspending database.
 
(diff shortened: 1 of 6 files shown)


sethstoudenmier:
Attached are sample log outputs for Award and Transaction deletes when using verbose and non-verbose configuration.

[es_awards_delete.txt](https://github.com/user-attachments/files/19592932/es_awards_delete.txt)
[es_awards_delete_verbose.txt](https://github.com/user-attachments/files/19592933/es_awards_delete_verbose.txt)
[es_transactions_delete.txt](https://github.com/user-attachments/files/19592934/es_transactions_delete.txt)
[es_transactions_delete_verbose.txt](https://github.com/user-attachments/files/19592936/es_transactions_delete_verbose.txt)

usaspending_api/broker/helpers/last_load_date.py
```diff
@@ -56,13 +56,15 @@ def get_earliest_load_date(keys, default=None):
                 earliest_date = key_date
 
     if earliest_date is None:
-        logger.warning(f"No earliest load date could be calculated because no dates for keys `{keys}` were found!")
+        logger.warning(
+            format_func("No earliest load date could be calculated because no dates for keys `{keys}` were found!")
```
zachflanders-frb, line 60: Looks like this accidently droped the `f` for format string.
```suggestion
            format_func(f"No earliest load date could be calculated because no dates for keys `{keys}` were found!")
```

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -587,45 +545,47 @@ def _gather_modified_transactions_pre_fy2008(
...
+                    THEN CONCAT('ASST_TX_', afa_generated_unique)
             END AS generated_unique_transaction_id
         FROM
             {transactions_table}
         WHERE
-            etl_update_date > CURRENT_DATE - {days_delta}
+            etl_update_date > '{delete_window_start}'
             AND
             action_date < '2007-10-01'
         """
```
zachflanders-frb, line 562: Could consider parameterizing this with `%(delete_window_start)s`, `$(transactions_table)s` syntax and passing a dictionary to `execute_sql_statement`

zachflanders-frb (dismissed):
Looks good.  One general comment is that there are a lot a `config` parameters that are type hinted as a general `dict`.  It might be nice to use `dataclass`es or something similar to make it more clear what can go into these configs.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
sethstoudenmier: I don't feel too bad about using f-string here given we control the values that are going into the string. I like the idea of using format functionality built-in to psycopg2, but I would also need to add it for the Spark logic branch. This would mean having two copies of the SQL for the two different approaches to SQL formatting (`{VALUE}` vs `%(VALUES)s` in Spark vs psycopg2) or having two rounds of formatting, one to add in the specific pattern used when formatting and then another to add in the values.

It does look like psycopg2 supports `{}` for formatting SQL, but it is used for table and column names and not recommended for values.

zachflanders-frb (approved):

PR #4359, opened 2025-04-15
DEV-11858: Reload Transactions Fix

dpb-bah:
**Description:**
When reloading transactions with a nonexistent `raw.transaction_normalized`, it uses code that hasn't been updated since the Phase 3 deploy, which includes the assumption of `AnalysisException.desc`. This simply updates it to use `str(e)` which contains the message which the code checks for.

**Technical details:**
N/A

2. API documentation updated
    - [ ] Backend
    - [ ] Frontend <OPTIONAL>
    - [ ] Operations <OPTIONAL>
6. Data validation completed
8. [ ] Jira Ticket [DEV-11858](https://federal-spending-transparency.atlassian.net/browse/DEV-11858):
9. [x] Used for the test run on staging with success

```
```

### DIFF
--- a/usaspending_api/etl/management/commands/load_transactions_in_delta.py
+++ b/usaspending_api/etl/management/commands/load_transactions_in_delta.py
@@ -1145,7 +1145,7 @@ class Command(BaseCommand):
             except AnalysisException as e:
                 if re.match(
                     r"^\[TABLE_OR_VIEW_NOT_FOUND\] The table or view `raw`\.`transaction_normalized` cannot be found\..*$",
-                    e.desc,
+                    e.message,
                     re.MULTILINE,
                 ):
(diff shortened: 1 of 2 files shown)


sethstoudenmier (approved):
I was curious why this seems to work without the change but was also found to need the change in some cases. After some looking it would appear that pyspark has moved in favor of the `__str__` method. In versions such as `3.5.0` which we use right now the `desc` attribute is public on the class (see [here](https://github.com/apache/spark/blob/v3.5.0/python/pyspark/errors/exceptions/captured.py#L58)). However, in most recent versions they have marked it as internal and access it via the `__str__` method (see [here](https://github.com/apache/spark/blob/v4.0.0/python/pyspark/errors/exceptions/captured.py#L94)).

All of that to say, making this change will help to future proof any updates to our version of pyspark.

PR #4360, opened 2025-04-16
[DEV-12478] Update location autocomplete to limit by location_type [hotfix][staging]

aguest-kc:
**Description:**
The `/api/v2/autocomplete/location` endpoint now applies the `limit` value to each `location_type` instead of to the result as a whole.

    - [x] Backend
8. [x] Jira Ticket [DEV-12478](https://federal-spending-transparency.atlassian.net/browse/DEV-12478):

### DIFF
--- a/usaspending_api/api_contracts/contracts/v2/autocomplete/location.md
+++ b/usaspending_api/api_contracts/contracts/v2/autocomplete/location.md
@@ -19,7 +19,7 @@ This route sends a request to the backend to retrieve locations matching the spe
 
     + Attributes (object)
         + `limit` (optional, number)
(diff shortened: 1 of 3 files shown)


aguest-kc:
The `limit` filter was updated to default to 5 instead of 10 and that's noted in the API contract. This shouldn't affect the frontend since they send a `limit` of 5 anyway. It was lowered because it applies to every location type, so we _could_ send a max of 35 locations in a response.

5 results * 7 location types = 35 possible locations in the API response

sethstoudenmier (approved):
Tested locally and changes look good!

PR #4375, opened 2025-04-21
[DEV-12472] Add legacy unique award key that is generated using subtier code

sethstoudenmier:
**Description:**
This handles the generation of the unique Award key with subtier code for FABS, allowing users to continue using old links to the Award Profile page.

**Technical details:**
This PR is broken into two parts:

1. Create the `generated_unique_award_id_legacy` field on AwardSearch as part of the delta tables that are copied back to Postgres.

   Previously this was also updating the `vw_awards` VIEW to include the legacy ID. However, after thinking it through there is not much gained by adding this field and we are able to easily avoid adding by pointing the endpoint to the AwardSearch table instead.

2. Allow for the API's `/api/v2/awards/<AWARD_ID>/` to support old unique Award keys after we transition away from using the subtier code.

   This updates the endpoint logic to first check the current unique award key and then defaults to the legacy key.

4. [x] API documentation updated
5. [ ] Necessary PR reviewers:
    - [ ] Backend
6. [x] Matview impact assessment completed
7. [x] Frontend impact assessment completed
8. [x] Data validation completed
9. [x] Appropriate Operations ticket(s) created
10. [x] Jira Ticket [DEV-12472](https://federal-spending-transparency.atlassian.net/browse/DEV-12472):

```
```


usaspending_api/awards/v2/views/awards.py
```diff
@@ -55,18 +56,30 @@ def _parse_and_validate_request(self, provided_award_id: str) -> dict:
...
-            logger.info("No Award found with: '{}'".format(request_dict))
-            raise NotFound("No Award found with: '{}'".format(request_dict))
+        award_queryset = AwardSearch.objects.filter(**request_dict)
+
+        if not award_queryset and "generated_unique_award_id" in request_dict:
+            # This is handled as a different DB call to avoid the possibility that one Award's current unique key
+            # matches the legacy unique key of a different Award.
+            logger.info(f"No Award found with: {request_dict}. Checking against legacy unique award key.")
+            request_dict = {"generated_unique_award_id_legacy": request_dict["generated_unique_award_id"]}
+            award_queryset = AwardSearch.objects.filter(**request_dict)
```
zachflanders-frb, line 73: This seems like the best way to handle this.  I do wonder how many legacy unique keys match a current unique key of a different award.  If this is a fairly common occurrence, we could consider a pattern to prompt the user on which one they want.  Or some kind of message along the lines of "We also found this award using the legacy award unique key:"

PR #4384, opened 2025-05-01
Upgrading Opensearch to 2.9

jimmyihuoma1:
Upgrading Opensearch to 2.9

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -178,7 +178,7 @@ services:
       - transport.host=localhost
       - bootstrap.memory_lock=true
       - plugins.security.disabled=true
-      - "OPENSEARCH_JAVA_OPTS=-Xms6048m -Xmx6048m"  # Ensure Docker is allocated plenty of memory, otherwise this will fail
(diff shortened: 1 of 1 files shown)


aguest-kc:
The existing code seems to function correctly with OpenSearch 2.9. I ran into a few issues with settings in the docker-compose file, but those were fixed.

The only remaining issue is that the Kibana dashboard wants a username and password, even though the security plugin is disabled on the OpenSearch container. The documented default of admin/admin didn't work. I'll keep looking for a fix for this and testing OpenSearch as I have time.

aguest-kc:
> The only remaining issue is that the Kibana dashboard wants a username and password, even though the security plugin is disabled on the OpenSearch container. The documented default of admin/admin didn't work. I'll keep looking for a fix for this and testing OpenSearch as I have time.

I was able to disable the security on the OpenSearch Dashboard. This PR should be in a good spot for others to review and test locally. I've been running it locally for a while and I have not noticed any issues with our current code.

aguest-kc (approved):

sethstoudenmier (dismissed):

PR #4395, opened 2025-05-15
[DEV-12101] Add command for updating USAspending tables from Broker

sethstoudenmier:
**Description:**
Simple command for updating a single field in a table with a value from Broker.

**Technical details:**
This command was created to be generic but solves the problem of needing to update `unique_award_key` values from `published_fabs` on Broker to the `source_assistance_transaction` table on USAspending. We don't want to run as part of a single query because that will cause a lock on both tables that would interfere with updates on Broker.

    - [ ] Backend
8. [x] Jira Ticket [DEV-12101](https://federal-spending-transparency.atlassian.net/browse/DEV-12101):

```
```

[DEV-12101]: https://federal-spending-transparency.atlassian.net/browse/DEV-12101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/broker/management/commands/update_table_value_from_broker.py
```diff
@@ -0,0 +1,168 @@
...
+            )
+            logger.error(msg)
+            raise ValueError(msg)
+        logger.info(
+            f"Found range of {min_val} to {max_val} for Broker table "
+            f'"{self.broker_table_name}"."{self.broker_match_field}"'
+        )
+        return min_val, max_val
+
+    def run_update(self, min_id: int, max_id: int) -> None:
```
zachflanders-frb, line 109: Should this be wrapped in a `@transaction.atomic`?

zachflanders-frb (approved):

usaspending_api/broker/management/commands/update_table_value_from_broker.py
sethstoudenmier: Not in this case as I want to avoid maintaining a lock on the table. The update will take an "access exclusive" lock and block any other query against the table for the duration. Chunking it out allows operation on the table to function almost seamlessly while it works on updates.

PR #4410, opened 2025-06-03
[DEV-11901] award_financial spending_level will limit the DEFC returned based on filter

sethstoudenmier:
**Description:**
Limit the results to only those DEFC that were searched on when filtering to the `award_financial` spending_level.

**Technical details:**
When filtering by the `awards`, `transactions`, and `subawards` spending_level all of the DEFC on the filtered Award(s) and Transaction(s) are returned. This occurs because we return the cumulative sum of obligations and outlays on the Award / Transaction without breaking it down by DEFC. Now when a user filters by the `award_financial` spending_level it will instead only return the DEFC that are present in the filters because we can break it down by DEFC.

This change also allows us to support backwards compatibility with the current endpoint in production.

    - [ ] Backend
8. [x] Jira Ticket [DEV-11901](https://federal-spending-transparency.atlassian.net/browse/DEV-11901):

```
```

[DEV-11901]: https://federal-spending-transparency.atlassian.net/browse/DEV-11901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/search/v2/views/spending_by_category_views/spending_by_industry_codes.py
```diff
@@ -32,6 +32,8 @@ def build_elasticsearch_result(self, response: dict) -> List[dict]:
         # Get the codes
         if self.category.nested_path:
             response = response.get("nested_agg")
+        if self.category.filter_key_to_limit:
+            response = response.get("filter_agg")
```
zachflanders-frb, line 36: Since the `self.category` will have both properties in case of `fitler_agg` could do this just to be more clear about what takes precedence: 
```suggestion
        if self.category.filter_key_to_limit:
            response = response.get("filter_agg")
        elif self.category.nested_path:
            response = response.get("nested_agg")
```

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -746,15 +746,15 @@ def get_defc_row_values(self, row: dict[str, any]) -> dict[str, any]:
...
                     [
                         (
                             spending_by_defc[lookup_values["field_type"]]
                             if spending_by_defc["defc"] in self.def_codes_by_group[lookup_values["group_name"]]
                             else 0
                         )
-                        for spending_by_defc in row.get(field)
+                        for spending_by_defc in row[field]
                     ]
                 )
```
zachflanders-frb:
```python
result = {
    field: sum([
        spending_by_defc[lookup_values["field_type"]]
        for spending_by_defc in row[field]
        if spending_by_defc["defc"] in self.def_codes_by_group[lookup_values["group_name"]]
    ])
    for field, lookup_values in self.spending_by_defc_lookup.items()
    if field in row and row[field] is not None
}
```
Not consequential but could use a comprehension pattern for consistency rather than a combo of comprehension and accumulator pattern.  I don't think the `else 0` is needed because `sum([]) == 0`.  This was likely to cover the case where `row.get(field) is None` which you have eliminated.

zachflanders-frb (approved):
I made a couple of non-blocking clean up suggestions.

usaspending_api/search/v2/views/spending_by_category_views/spending_by_industry_codes.py
sethstoudenmier: It was intentional to have them as two `if` statements because they are chained. If a filter is nested, then we need to first step into the nested aggregation before we can access the filter.

zachflanders-frb: 👍 Oh yeah, I see that now

PR #4416, opened 2025-06-13
[DEV-12320] -  Update dependency management

zachflanders-frb:
**Description:**
~This is a draft PR for discussion around adding uv as a dependency manager.~
This PR updated the project to use `uv` as the python project/package manager.  

- The `Dockerfile` is updated to install `uv` and to sync the project dependencies. 
- The dependencies are now defined in `pyproject.toml` (including extra dependency groups for dev, server, ansible, etc.)
- The makefile is updated to use `uv run` for tests and to allow uv to manage the environment.
- The `uv.lock` file is added to the repo to ensure reproducibility of the environment.
- The readme is updated to reflect changes including instructions for running tests locally.

Jira: [DEV-12320](https://federal-spending-transparency.atlassian.net/browse/DEV-12320)

[DEV-12320]: https://federal-spending-transparency.atlassian.net/browse/DEV-12320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ dependencies = [
     "python-json-logger==2.0.7",
     "requests==2.31.*",
     "retrying==1.3.4",
+    "sqlparse==0.5.*",
     "urllib3==1.26.*",
     "xlrd3==1.1.0",
     # Opentelemetry Dependencies
--- a/uv.lock
+++ b/uv.lock
@@ -2287,6 +2287,7 @@ dependencies = [
     { name = "python-json-logger" },
     { name = "requests" },
     { name = "retrying" },
+    { name = "sqlparse" },
     { name = "urllib3" },
     { name = "xlrd3" },
 ]
@@ -2420,6 +2421,7 @@ requires-dist = [
     { name = "requests", specifier = "==2.31.*" },
     { name = "retrying", specifier = "==1.3.4" },
     { name = "setuptools", marker = "extra == 'server'", specifier = ">=68.1.2" },
+    { name = "sqlparse", specifier = "==0.5.*" },
     { name = "supervisor", marker = "extra == 'server'", specifier = "==4.1.0" },
     { name = "urllib3", specifier = "==1.26.*" },
     { name = "uwsgi", marker = "extra == 'server'", specifier = "==2.0.23" },


Makefile
```diff
@@ -52,65 +52,23 @@ printvars: ## Print the Environment variables present in calls to make, plus var
 	@printf "\n==== Environment Variables ====\n\n"
 	@printenv
 
-.python-version: ## Attempt to setup python using pyenv
```
zachflanders-frb, line 55: Let the docker container handle setting the python version.  Added a note about installing the python version with uv if running locally to the readme.

Makefile
```diff
@@ -52,65 +52,23 @@ printvars: ## Print the Environment variables present in calls to make, plus var
...
-  		pyenv install -s ${PYTHON_VERSION}; \
-  		pyenv local ${PYTHON_VERSION}; \
-  		python3 -V; \
-  		if [ "$$(python3 -V)" != "Python ${PYTHON_VERSION}" ]; then \
-  			echo "ERROR: pyenv was not able to set local python version to ${PYTHON_VERSION}"; \
-  			exit 1; \
-  		fi; \
-	fi;
-
-.venv: ## Ensure a virtual environment is established at .venv
```
zachflanders-frb, line 70: No need for a venv since we are already in a container or if running locally uv will handle the environment.

docker-compose.yml
```diff
@@ -52,7 +52,7 @@ services:
     image: usaspending-testing  # when an image by this name is not found in the local repo, and it is forced to build, it will use this as the tag
     build:
       context: .
-      dockerfile: Dockerfile.testing
```
zachflanders-frb, line 55: The Dockerfile filename should be <something>.Dockerfile by convention:
> Some projects may need distinct Dockerfiles for specific purposes. A common convention is to name these `<something>.Dockerfile`.

https://docs.docker.com/build/concepts/dockerfile/#filename

This will enable IDEs and github to highlight the syntax.

pyproject.toml
```diff
@@ -1,3 +1,133 @@
...
+    "Development Status :: 5 - Production/Stable",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3 :: Only",
+]
+dependencies = [
+    "asyncpg==0.29.*",
+    "attrs==23.2.*",
+    "boto3>=1.34,<1.36",
```
zachflanders-frb, line 26: boto3 was set to `1.34.*`  I loosened this a bit because it was conflicting with the ansible and/or awscli requirements.  In the Dockerfile these are pip installed as a subsequent step, so it was likely installing a different version of boto3 into the image during that step.

zachflanders-frb (commented):
We will need to update the README.md, especially regarding running tests not in a container.  I think we should include instructions for settings up a venv with the python version using uv to run tests locally, e.g. `uv venv --python 3.10.12 usaspending-api`.  I have not tested this yet.

Dockerfile
```diff
@@ -8,6 +8,8 @@
 
 FROM python:3.10.12-slim-bullseye
 
+COPY --from=ghcr.io/astral-sh/uv:0.7.19 /uv /uvx /bin/
```
sethstoudenmier, line 11: Need to follow up with OPS to make sure that pulling from GitHub container registry is alright with some of the planned changes.

Makefile
```diff
@@ -52,65 +52,26 @@ printvars: ## Print the Environment variables present in calls to make, plus var
...
-.PHONY: activate
-activate: ## Spit out the command to run to activate the virtual env, since you can't do it within a make shell process. Use this like: source $(make activate)
-	@echo ".venv/${venv_name}/bin/activate"
-
 .PHONY: local-dev-setup
-local-dev-setup: .python-version requirements-dev check-dependencies .ivy2 ## Setup python, virtual environment, and pip dependencies, then check version info
+local-dev-setup: uv-sync check-dependencies .ivy2 ## Setup virtual environment and pip dependencies, then check version info
 
 .PHONY: check-dependencies
 check-dependencies: ## Prints out the versions of dependencies in use
```
sethstoudenmier, line 65: If I am following correctly, uv will be managing the virtual environment and remove the need for us doing it ourself?

pyproject.toml
```diff
@@ -1,3 +1,134 @@
...
+    "opentelemetry-instrumentation-urllib==0.49b2",
+    "opentelemetry-instrumentation-urllib3==0.49b2",
+    "opentelemetry-instrumentation-wsgi==0.49b2",
+    "opentelemetry-propagator-aws-xray==1.0.1",
+    "opentelemetry-proto==1.28.2",
+    "opentelemetry-sdk==1.28.2",
+    "opentelemetry-semantic-conventions==0.49b2",
+    "opentelemetry-util-http==0.49b2",
+    # The above packages are listed and pinned to avoid using the 'opentelemetry-bootstrap -a install' command
+    "xlrd3==1.1.0",
```
sethstoudenmier: If you happen to need to do any further cleanup can we move this above the opentelemetry packages? It isn't something you changed but seems out of place down here.

pyproject.toml
```diff
@@ -1,3 +1,134 @@
...
+    "certifi==2024.7.4",
+    "databricks-sdk==0.44.1",
+    "dataclasses-json==0.6.*",
+    "dj-database-url==2.1.0",
+    "django-cors-headers==4.3.*",
+    "django-debug-toolbar==4.3.*",
+    "django-extensions==3.2.*",
+    "django-spaghetti-and-meatballs==0.4.*",
+    "Django==4.2.*",
+    "Django==4.2.*",
```
sethstoudenmier: You have Django listed twice

requirements/caching_requirements.txt
sethstoudenmier, line 1: Need to work with OPS to see where this file is used

sethstoudenmier: The fact that it calls out the `requirements-server.txt` makes me wonder if it was used directly

requirements/server_requirements.txt
sethstoudenmier, line 1: Similar to the caching requirements above, we need to double check if this is used directly anywhere

sethstoudenmier (changes requested):
Small change for the duplicate reference to Django, but mostly questions for OPS around if some of the files are referenced directly

sethstoudenmier (approved):

collinwr (approved):

PR #4420, opened 2025-06-18
[DEV-12727] Update Elasticsearch Award Index to Fix Duplicate Awards

loreleitrimberger:
**Description:**
When all transactions of an award are deleted but at least one transaction from the award is added, Elasticsearch deletes the award and creates a new award instead of recycling the Award's ID. A new Award is created with the same Generated ID so the old award isn't deleted and new award is inserted

**Technical details:**
From looking into the code base, I believe the changes need to be made in within `load_transactions_in_delta.py` in `award_id_lookup_delete_subquery`. The part of the query I edited where is checking if there exists a transaction with that doesn't have a null value in correction_delete_indicatr (value of D means the transaction is deleted, null is new transaction). If it finds that there's a value that exists then that award isn't included in award ids to delete

    - [ ] Backend
    - N/A Frontend <OPTIONAL>
    - N/A Operations <OPTIONAL>
8. [ ] Jira Ticket [DEV-12727](https://federal-spending-transparency.atlassian.net/browse/DEV-12727):

```
```

[DEV-12727]: https://federal-spending-transparency.atlassian.net/browse/DEV-12727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -86,7 +86,9 @@ class Command(BaseCommand):
         FROM aidlu_fabs AS aidlu LEFT JOIN raw.published_fabs AS pfabs ON (
             aidlu.transaction_unique_id = ucase(pfabs.afa_generated_unique)
         )
-        WHERE pfabs.afa_generated_unique IS NULL
+        WHERE pfabs.afa_generated_unique IS NULL AND NOT EXISTS (SELECT 1 FROM aidlu_fabs  AS aidlu LEFT JOIN raw.published_fabs AS pfabs ON (
```
sethstoudenmier: Whenever `pfabs.afa_generated_unique IS NULL` criteria is met it should be guaranteed that a record won't exist in the subquery since it relies on joining via the `pfabs.afa_generated_unique`.

usaspending_api/etl/elasticsearch_loader_helpers/delete_data.py
```diff
@@ -464,6 +465,12 @@ def _gather_deleted_transaction_keys(
             and not x.key.startswith("staging")
             and delete_window_start <= x.last_modified <= end_date
         )
+        # matches the pattern YYYY-mm-DD_FABSdeletions_{TIME_IN_SECONDS}.csv
+        or (
+            re.search(r"^\d{4}-(0[1-9]|1[0-2])-(0?[1-9]|[12][0-9]|3[01])_FABSdeletions_.*.csv", x.key)
```
sethstoudenmier:
```suggestion
            re.fullmatch(r"^\d{4}-(0[1-9]|1[0-2])-(0?[1-9]|[12][0-9]|3[01])_FABSdeletions_.*\.csv", x.key)
```
Small change to use `fullmatch` to be extra cautious since you have such a thorough regex defined. And then escaping the `.` at the end to make sure we match on exactly `.csv`.

PR #4429, opened 2025-07-01
[DEV-12772] -  Treasury Account Spark Downloads

zachflanders-frb:
**Description:**
This PR adds Treasury Account Downloads to the generate spark downloads command.

**Technical details:**
The treasury account downloads were able to use the account download delta table with the addition of a few columns specific to the treasury download.

    - [ ] Backend
8. [ ] Jira Ticket [DEV-12772](https://federal-spending-transparency.atlassian.net/browse/DEV-12772):

[DEV-12772]: https://federal-spending-transparency.atlassian.net/browse/DEV-12772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/download/tests/integration/test_account_download_dataframe_builder.py
+++ b/usaspending_api/download/tests/integration/test_account_download_dataframe_builder.py
@@ -89,9 +89,7 @@ def federal_account_models(db):
     baker.make("accounts.FederalAccount", pk=3, agency_identifier="345", main_account_code="0333")
 
 
-@patch(
-    "usaspending_api.download.management.commands.delta_downloads.builders.get_submission_ids_for_periods"
-)
+@patch("usaspending_api.download.management.commands.delta_downloads.builders.get_submission_ids_for_periods")
 def test_federal_account_download_dataframe_builder(mock_get_submission_ids_for_periods, spark, account_download_table):
     mock_get_submission_ids_for_periods.return_value = [1, 2, 4, 5]
     account_download_filter = AccountDownloadFilter(
@@ -108,9 +106,7 @@ def test_federal_account_download_dataframe_builder(mock_get_submission_ids_for_
     assert sorted(result_df.gross_outlay_amount_FYB_to_period_end.to_list()) == [100, 200]
 
 
-@patch(
-    "usaspending_api.download.management.commands.delta_downloads.builders.get_submission_ids_for_periods"
-)
+@patch("usaspending_api.download.management.commands.delta_downloads.builders.get_submission_ids_for_periods")
 def test_filter_federal_by_agency(mock_get_submission_ids_for_periods, spark, account_download_table, agency_models):
     mock_get_submission_ids_for_periods.return_value = [1, 2, 4, 5]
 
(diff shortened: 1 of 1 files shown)


usaspending_api/download/management/commands/generate_spark_download.py
zachflanders-frb, line 1: TODO: Should be able to get the download level and type from the json response saved on the download job.

usaspending_api/download/management/commands/delta_downloads/award_financial/builders.py
```diff
@@ -99,36 +80,105 @@ class Condition:
...
+            | (sf.col("transaction_obligated_amount") != 0)
+        )
+
     @staticmethod
     def collect_concat(col_name: str, concat_str: str = "; ") -> Column:
         return sf.concat_ws(concat_str, sf.sort_array(sf.collect_set(col_name))).alias(col_name)
 
+    @property
+    @abstractmethod
+    def source_df(self) -> DataFrame: ...
```
zachflanders-frb: Black and Flake8 disagree on this.  I vote we override flake8.

zachflanders-frb: More detail here: https://github.com/psf/black/issues/4173

usaspending_api/download/management/commands/delta_downloads/builders.py
sethstoudenmier, line 1: Only question I have is around file structure and maybe you have already thought about this. The File A and B downloads will be different columns than the File C (Award Financial). This means that ideally, they could re-use the same dataframe filter logic, but the columns will differ. This gets me to my actual question:

Should we have this structured such that the Abstract classes are broken apart from the specific download implementations? For example, right now you have the Federal and Treasury account implementations in this file for the File C download. But if File A and B are going to need different implementations, would we want to group them into this same file?

sethstoudenmier (approved):
Approving because the changes look good to me, but I do have a conceptual question about structure as build this out more.

usaspending_api/download/management/commands/delta_downloads/builders.py
zachflanders-frb: The organization of these classes feels like it could evolve.  It seems like the overall decision tree looks something like 
```mermaid
graph LR;
A((Account Level)) --> B[Federal Account];
A((Account Level)) --> C[Treasury Account];
B[Federal Account] --> D((Submission Type));
C[Treasury Account] --> E((Submission Type));
D((Submission Type)) --> F[Account Download];
D((Submission Type)) --> G[Object Class];
D((Submission Type)) --> H[Award Financial];
E((Submission Type)) --> I[Account Download];
E((Submission Type)) --> J[Object Class];
E((Submission Type)) --> K[Award Financial];
```

  In some of my WIP on File A/B downloads I have this on the abstract class:
```python
    @property
    @abstractmethod
    def account_balances(self) -> DataFrame: ...
    
    @property
    @abstractmethod
    def object_class_program_activity(self) -> DataFrame: ...

    @property
    @abstractmethod
    def award_financial(self) -> DataFrame: ...

    @property
    def source_dfs(self) -> list[DataFrame]:
        return [getattr(self, submission_type) for submission_type in self.submission_types]
```

Many of the filtering methods and other useful functions can be reused across the different submission types.

PR #4439, opened 2025-07-17
[DEV-11765] Fix subaward zip bug

sethstoudenmier:
## Description:

Use the 5-digit zip code for subawards instead of the larger form zip code.

## Technical Details:

There are a few small changes grouped into this PR:
* Use the `sub_pop_zip5` instead of `sub_pop_zip` when filtering the zip code on Subawards. This is a pattern we use for both Awards and Transactions, and I was unable to see a reason why we would use the longer form zip code at this time. Optionally, we could support querying the longer form zip code (and changes were initially queued up for this), but that adds extra functionality that I don't believe was intended at this time.
* Some minor cleanup of the location filters that could use some additional refactoring in order to be more DRY, but I wanted to limit the changes in this PR as it was already touching some items outside of scope.
* When initially writing test cases I noticed an issue with the sorting on `Sub-Award ID` field. Sorting on `Sub-Award ID` was previously sorting on the Prime Award's value for `display_award_id` which is confusing because the `Sub-Award ID` is actually the `subaward_number`. This was corrected initially for the purpose of test cases although that has since changed. But this change was left in place as it is a small bug that improves UX on the site and API.
* Slight change to the Primary Place of Performance zip code logic that will use the `sub_pop_zip5` value when the `sub_pop_zip` value is not 5 or 9 characters. This was causing a weird UI bug where a Subaward would appear in the filter, but the zip code wouldn't show on the page because the `sub_pop_zip` often has extra characters in it.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [x] Jira Ticket(s)
    1. [DEV-11765](https://federal-spending-transparency.atlassian.net/browse/DEV-11765)

[DEV-11765]: https://federal-spending-transparency.atlassian.net/browse/DEV-11765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/database_scripts/etl/subaward_es_view.sql
```diff
@@ -87,6 +87,7 @@ SELECT
 	s.sub_place_of_perform_state_code AS sub_pop_state_code,
 	s.sub_place_of_perform_state_name AS sub_pop_state_name,
     s.sub_place_of_perform_county_code AS sub_pop_county_code,
+    s.sub_place_of_perform_zip5 AS sub_pop_zip5,
```
sethstoudenmier, line 90: While I really want to correct the formatting on this file, it would greatly inflate the changeset for the PR.

zachflanders-frb (approved):

PR #4447, opened 2025-07-30
Ftr/dev 12639 update custom download

loreleitrimberger:
## Description:
This has file c (award_financial) use Spark to process a download job when award_financial is a submission_type

## Technical Details:
Uses the experimental header HTTP-X-DOWNLOAD-API to go through spark
If Files A and B are included in the request, they still go through the SQS queue

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [X] Jira Ticket(s)
    1. [DEV-12639](https://federal-spending-transparency.atlassian.net/browse/DEV-12639)

[DEV-12639]: https://federal-spending-transparency.atlassian.net/browse/DEV-12639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/common/experimental_api_flags.py
```diff
@@ -29,6 +31,12 @@ def is_experimental_elasticsearch_api(request: Request) -> bool:
     return request.META.get(EXPERIMENTAL_API_HEADER) == ELASTICSEARCH_HEADER_VALUE
 
 
+def is_experimental_download_api(request: Request) -> bool:
+    """
+    Returns True or False depending on if the expected_header_value matches what is sent with the request
+    """
+    return request.headers.get(DOWNLOAD_API_HEADER) == DOWNLOAD_HEADER_VALUE
```
loreleitrimberger: Using the same format as the experimental_elasticsearch_api wasn't coming through for me (using META.get and _ instead of -), but doing it this way did

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -68,7 +70,34 @@ def post(
...
+                download_job.json_request = json.dumps(str_to_json_original)
+                download_job.save()
+                self.process_request(download_job)
+                # add award_financial back for the final response
+                str_to_json_original['filters']['submission_types'].append('award_financial')
+                str_to_json_original['download_types'].append('award_financial')
+                download_job.json_request = json.dumps(str_to_json_original)
+                download_job.save()
+        else:
+            self.process_request(download_job)
```
loreleitrimberger: I'm not sure if this is the best way to go about it, but this way the spark job is reading from the original download_job without files a or b in it and then the download_job runs without file c

usaspending_api/common/experimental_api_flags.py
sethstoudenmier: Django does some handling of the request headers (see https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.HttpRequest.META).

In short, the ES experimental header would normally be supplied to the API as `X-Experimental-API` and Django would convert it under `request.META` to be `HTTP_X_EXPERIMENTAL_API`.

With that said you should also be good to use request.headers in this case, but I would recommend changing the expected value to no include "HTTP" as that is something that Django prepends for the "request.META".

usaspending_api/common/experimental_api_flags.py
```diff
@@ -29,6 +32,13 @@ def is_experimental_elasticsearch_api(request: Request) -> bool:
     return request.META.get(EXPERIMENTAL_API_HEADER) == ELASTICSEARCH_HEADER_VALUE
 
 
+def is_experimental_download_api(request: Request) -> bool:
```
sethstoudenmier, line 33: I would recommend sharing the same header value of `X-Experimental-API` (converted to `HTTP_X_EXPERIMENTAL_API` in the `request.META` by Django) and have different values `elasticsearch` vs `download`. You could then continue to have two different functions, but each would serve its own purpose based on the header value.

This is more of a functional programming approach, but for cases like this I think it is perfectly fine.

usaspending_api/search/v2/views/new_awards_over_time.py
```diff
@@ -148,7 +148,10 @@ def format_results(self, es_results):
             if self.group == "month":
                 time_period = {"fiscal_year": f"{date.year}", self.group: f"{date.month}"}
             elif self.group == "quarter":
-                time_period = {"fiscal_year": f"{date.year}", self.group: f"{int(date.month/3) + (date.month % 3>0)}"}
+                time_period = {
```
sethstoudenmier: Is it correct to assume that this change is due to some linter?

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -85,6 +115,17 @@ def process_request(self, download_job: DownloadJob):
             queue = get_sqs_queue(queue_name=settings.BULK_DOWNLOAD_SQS_QUEUE_NAME)
             queue.send_message(MessageBody=str(download_job.download_job_id))
 
+    def process_account_download_in_spark(self, download_job: DownloadJob):
+        """
+        Process File C downloads through spark instead of sqs for better performance
+        """
+        spark_jobs = SparkJobs(LocalStrategy()) if IS_LOCAL else SparkJobs(DatabricksStrategy())
+        spark_jobs.start(
+            job_name="download_delta_table-award_search",
```
sethstoudenmier: We never decided on a name for the download, but I would recommend changing this to `api_download-accounts`. Open to other ideas but my thought process is that we want to distinguish this as a download generated by the API, and it is an "accounts" download.

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -68,7 +71,34 @@ def post(
         )
 
         log_new_download_job(request, download_job)
-        self.process_request(download_job)
+        if (
```
sethstoudenmier, line 78: I would recommend moving this block into the `process_request` function to keep that logic grouped together.

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -68,7 +71,34 @@ def post(
...
+            is_experimental_download_api(request)
+            and json_request["request_type"] == "account"
+            and "award_financial" in json_request["download_types"]
+        ):
+            # run spark download with only award_financial in download type
```
sethstoudenmier: This is something I think we have normed on in the past few weeks. If the download contains the "award_financial" type then we want to run all files in that download as a Spark download.

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -68,7 +71,34 @@ def post(
...
+            str_to_json_original = json.loads(download_job.json_request).copy()
+            str_to_json_award_financial = json.loads(download_job.json_request)
+            str_to_json_award_financial["download_types"] = ["award_financial"]
+            str_to_json_award_financial["filters"]["submission_types"] = ["award_financial"]
+            download_job.json_request = json.dumps(str_to_json_award_financial)
+            download_job.save()
+            # goes to spark for File C account download
+            self.process_account_download_in_spark(download_job=download_job)
+            # remove award_financial from json request to run download with remaining types
+            if len(json_request["download_types"]) > 1:
```
sethstoudenmier: This is similar to the comment above

usaspending_api/download/tests/integration/test_download_accounts.py
```diff
@@ -497,3 +497,28 @@ def test_empty_array_filter_fail(client, download_test_data):
     assert (
         "Field 'filters|def_codes' value '[]' is below min '1' items" in resp.json()["detail"]
     ), "Incorrect error message"
+
+
+@pytest.mark.django_db(databases=[settings.DOWNLOAD_DB_ALIAS, settings.DEFAULT_DB_ALIAS])
+def test_file_c_spark_download(client, download_test_data):
```
sethstoudenmier: Is this actually hitting the Spark download? I don't see usage of the experimental flag so I believe this should be hitting Postgres.

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -85,6 +115,17 @@ def process_request(self, download_job: DownloadJob):
             queue = get_sqs_queue(queue_name=settings.BULK_DOWNLOAD_SQS_QUEUE_NAME)
             queue.send_message(MessageBody=str(download_job.download_job_id))
 
+    def process_account_download_in_spark(self, download_job: DownloadJob):
+        """
+        Process File C downloads through spark instead of sqs for better performance
+        """
+        spark_jobs = SparkJobs(LocalStrategy()) if IS_LOCAL else SparkJobs(DatabricksStrategy())
+        spark_jobs.start(
```
sethstoudenmier: I would log details about the spark job here so that it is captured outside of Spark for debugging purposes.

sethstoudenmier (changes requested):

usaspending_api/download/v2/base_download_viewset.py
zachflanders-frb: Do we want to simply check for the header value and send all files to spark if the header value is present?  `qat` already supports `account_download` (file A) downloads and will soon support  `object_class_program_activity` (file B) downloads.

```python
if is_experimental_download_api(request):
    self.process_account_download_in_spark(download_job)
else:
    self.process_request(download_job)
```

usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -24,6 +26,7 @@
 from usaspending_api.download.v2.request_validations import DownloadValidatorBase
 from usaspending_api.routers.replicas import ReadReplicaRouter
 from usaspending_api.submissions.models import DABSSubmissionWindowSchedule
+from usaspending_api.settings import IS_LOCAL
```
zachflanders-frb: You can use `settings.IS_LOCAL` since `settings` is already imported above.

usaspending_api/search/v2/views/new_awards_over_time.py
loreleitrimberger: Yes I wouldn't have touched that

usaspending_api/download/management/commands/delta_downloads/builders.py
sethstoudenmier, line 1: Why did you have to make changes to this file?

sethstoudenmier (approved):

usaspending_api/common/spark/jobs.py
```diff
@@ -4,11 +4,17 @@
 from typing import TYPE_CHECKING, Generator
 
 from databricks.sdk import WorkspaceClient
+from django.conf import settings
 from django.core.management import call_command
 
-from usaspending_api.common.helpers.spark_helpers import configure_spark_session, get_active_spark_session
 from usaspending_api.common.spark.configs import LOCAL_EXTENDED_EXTRA_CONF
 
+if settings.IS_LOCAL:
```
zachflanders-frb, line 12: `IS_LOCAL` is hardcoded to True:
https://github.com/fedspendingtransparency/usaspending-api/blob/7e9f7970940f0aa78e0c02941e6373ae81f8519f/usaspending_api/settings.py#L61
This is causing the import below to happen, which errors in not spark contexts because `pyspark` is not a defined dependency of the app.

@loreleitrimberger @sethstoudenmier @aguest-kc

PR #4457, opened 2025-08-08
[DEV-12966] [DEV-12967] Update the spending by transaction and subaward grouped endpoints

aguest-kc:
## Description:
The `/v2/search/spending_by_subaward_grouped/` and `/v2/search/spending_by_transaction_grouped/` endpoints are updated to support proper pagination and to improve performance.

## Technical Details:
These endpoints are updated query the Awards index in Elasticsearch instead of the Subawards or Transaction index. This improves performance by avoiding the need to create aggregations. This also allows us to do the result sorting in Elasticsearch without hitting the 65,535 bucket limit for a sort, which we were running into when using the Transaction index before.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-12966](https://federal-spending-transparency.atlassian.net/browse/DEV-12966)
    2. [DEV-12967](https://federal-spending-transparency.atlassian.net/browse/DEV-12967)


usaspending_api/awards/migrations/0095_auto_20220617_1620.py
sethstoudenmier, line 1: Assuming this changed from your linter?

usaspending_api/awards/migrations/0098_auto_20221215_2029.py
```diff
@@ -7,55 +7,63 @@
...
+                        on_delete=django.db.models.deletion.CASCADE,
+                        primary_key=True,
+                        serialize=False,
+                        to="search.awardsearch",
+                    ),
                 )
-            ]
+            ],
         ),
+        # Moved to migration #0111 so that we can pull in a new column added to award_search
```
sethstoudenmier, line 52:
```suggestion
        # Moved to migration #0112 so that we can pull in a new column added to award_search
```

usaspending_api/awards/migrations/0098_auto_20221215_2029.py
```diff
@@ -7,55 +7,63 @@
...
+        # migrations.RunSQL(
+        #     sql=vw_awards_sql,
+        #     reverse_sql="DROP VIEW IF EXISTS vw_awards",
+        #     # Without this, Django will try to actually change the old table names in another migration
+        #     # This says we've already done it.
+        #     state_operations=[
```
sethstoudenmier, line 58: I believe when we chatted I had mentioned we probably didn't need the `state_operations` here, but thinking on it more I think I was incorrect. It could cause some problems especially for someone standing up the repo for the first time since this wouldn't track the renaming of the model table.

All of that to say, I think the state_operations either needs to stay here in this migration or follow the VIEW SQL to the 0112 migration.

usaspending_api/etl/es_award_template.json
```diff
@@ -63,7 +63,12 @@
         "null_value": "NULL"
       },
       "type_description": {
-        "type": "keyword"
+        "type": "text",
```
sethstoudenmier, line 66: Why is this change needed?

sethstoudenmier: Primary concern is that changing keyword to text with nested keyword could break items that use it.

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -463,7 +461,8 @@ def update_awards(self):
                     SUM(tn.indirect_federal_sharing)    AS total_indirect_federal_sharing,
                     -- Transaction FPDS Fields
                     SUM(CAST(fpds.base_and_all_options_value AS NUMERIC(23, 2))) AS base_and_all_options_value,
-                    SUM(CAST(fpds.base_exercised_options_val AS NUMERIC(23, 2))) AS base_exercised_options_val
+                    SUM(CAST(fpds.base_exercised_options_val AS NUMERIC(23, 2))) AS base_exercised_options_val,
+                    COUNT(tn.award_id) AS transaction_count
```
sethstoudenmier: Realistically it doesn't matter but to avoid confusion I would recommend changing this to be a count on the Transaction ID

usaspending_api/etl/management/commands/load_transactions_in_delta.py
```diff
@@ -1492,6 +1491,7 @@ def prepare_orphaned_award_temp_table():
                     list(AWARDS_COLUMNS),
                 ),
                 ("transaction_id", "transaction_id", "id", "id"),
+                strict=False,
```
sethstoudenmier: This is the default behavior, right?

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -115,36 +106,56 @@ def construct_es_response(self, results: list[SubawardGroupedModel]) -> dict[str
...
-            results.append(item)
-        results = self.sort_by_attribute(results)
-        return [result.__dict__ for result in results]
-
-    # default sorting is to sort by the award_id, default order is desc
-    def sort_by_attribute(self, results: list[SubawardGroupedModel]) -> list[SubawardGroupedModel]:
-        reverse = True if self.pagination["sort_order"] == "asc" else False
-        return sorted(results, key=lambda result: getattr(result, self.pagination["sort_key"]), reverse=reverse)
+
+        sorted_results = sorted(
```
sethstoudenmier: All sorting should be occurring inside of Elasticsearch if I remember correctly.

usaspending_api/search/v2/views/spending_by_transaction_grouped.py
```diff
@@ -92,38 +90,59 @@ def post(self, request: Request) -> Response:
...
+                    "award_id": source["display_award_id"],
+                    "transaction_count": source["transaction_count"],
+                    "transaction_obligation": Decimal(source["generated_pragmatic_obligation"]).quantize(
+                        Decimal(".01") if source["generated_pragmatic_obligation"] else 0.0
+                    ),
+                    "award_generated_internal_id": source["generated_unique_award_id"],
+                }
+            )
+
+        sorted_results = sorted(
```
sethstoudenmier: Similar to subawards, if I remember correctly, all sorting should occor inside of Elasticsarch.

usaspending_api/awards/migrations/0095_auto_20220617_1620.py
aguest-kc: Yeah, that's from my linter.

usaspending_api/awards/migrations/0098_auto_20221215_2029.py
aguest-kc: Good catch.

usaspending_api/search/v2/views/spending_by_transaction_grouped.py
```diff
@@ -92,38 +90,59 @@ def post(self, request: Request) -> Response:
...
+            results,
+            key=lambda x: x[self.sort_by],
+            reverse=False if validated_payload["order"] == "asc" else True,
+        )
```
zachflanders-frb: This could be a list comp:
```python
results = [
    {
        "award_id": hit["_source"]["display_award_id"],
        "transaction_count": hit["_source"]["transaction_count"],
        "transaction_obligation": Decimal(hit["_source"]["generated_pragmatic_obligation"]).quantize(
            Decimal(".01") if hit["_source"]["generated_pragmatic_obligation"] else 0.0
        ),
        "award_generated_internal_id": source["generated_unique_award_id"],
    }           
    for hit in es_response.hits.hits
]
```

usaspending_api/etl/management/commands/load_transactions_in_delta.py
aguest-kc: It is. My formatter always adds this and I usually manually remove it. I missed it this time, but I've removed it in a new commit.

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
```diff
@@ -115,36 +106,56 @@ def construct_es_response(self, results: list[SubawardGroupedModel]) -> dict[str
...
+                    "award_id": source["display_award_id"],
+                    "subaward_count": source["subaward_count"],
+                    "subaward_obligation": (
+                        Decimal(source["total_subaward_amount"]).quantize(Decimal(".01"))
+                        if source["total_subaward_amount"]
+                        else 0.0
+                    ),
+                    "award_generated_internal_id": source["generated_unique_award_id"],
+                }
             )
```
zachflanders-frb: This could also be a list comp.
```python
results = [
    {
        "award_id":hit["_source"]["display_award_id"],
        "subaward_count": hit["_source"]["subaward_count"],
        "subaward_obligation": (
            Decimal(hit["_source"]["total_subaward_amount"]).quantize(Decimal(".01"))
            if hit["_source"]["total_subaward_amount"]
            else 0.0
        ),
        "award_generated_internal_id": hit["_source"]["generated_unique_award_id"],
    }           
    for hit in es_response.hits.hits
]
```

zachflanders-frb (dismissed):

zachflanders-frb:
Approved assuming @sethstoudenmier's requested changes are resolved first.

usaspending_api/search/v2/views/spending_by_subaward_grouped.py
aguest-kc: It does. I noticed some weirdness after the `for` loop ran and the result order was changing, but I believe that was due to another issue and I forgot to remove this `sorted()`. I've removed this in a new commit.

usaspending_api/search/v2/views/spending_by_transaction_grouped.py
aguest-kc: Removed in a new commit.

usaspending_api/etl/management/commands/load_transactions_in_delta.py
aguest-kc: I updated the SQL to count the transaction IDs instead of award IDs.

usaspending_api/etl/es_award_template.json
aguest-kc: I was running into an error in `query_with_filters` when either a `keywords` or `description` filter was passed into the API request, but I've tested the endpoint with this set back to a keyword type and everything appears to work as expected.

I don't remember what the error was, but it doesn't look like the `type_description` field even gets included in the generated ES queries so this should be fine to revert.

usaspending_api/awards/migrations/0098_auto_20221215_2029.py
aguest-kc: I added the `state_operations` to migration 0112

usaspending_api/etl/es_award_template.json
aguest-kc: I was able to reproduce the error I was getting from the `_KeywordSearch()` class in `query_with_filters.py`:

> "Can only use phrase prefix queries on text fields - not on [type_description] which is of type [keyword]"

Changing this to a text type with a nested keyword type would be consistent with the mapping for the `type_description` field in the ES transaction template.

It seems like the `type_description` filter hasn't been compatible with the Awards index for a while, but the other endpoints that use the Awards index don't support the `keywords` filter so we haven't seen this behavior before.

sethstoudenmier (approved):

zachflanders-frb (approved):

PR #4470, opened 2025-08-22
[DEV-13016] add new endpoint

loreleitrimberger:
## Description:
This is creating a new endpoint in v2 for returning federal account totals for program activities. It's to replace the v1 endpoint's (`/api/v1/tas/categories/total/`) functionality for this specific purpose (v1 endpoint has additional functionality, so it isn't being replace entirely) and add implementation for using PARK or PAC/PAN. 

## Technical Details:
The will return a list of results that contain the obligation (sum of obligations_incurred_by_program_object_class_cpe for the File B records that belong to the last submission; should use is_final_balance_for_fy), name (PARK or PAN name), code (PARK/PAC), and type (enum of either PARK/PAC) of a program activity. 

A federal account is given as part of the url, and supports `time_period`, `object_class`, and `program_activity` as filters

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [X] Jira Ticket(s)
    1. [DEV-13016](https://federal-spending-transparency.atlassian.net/browse/DEV-13016)

### DIFF
--- a/usaspending_api/accounts/v2/views/federal_account_program_activities_total.py
+++ b/usaspending_api/accounts/v2/views/federal_account_program_activities_total.py
@@ -8,14 +8,13 @@ from usaspending_api.common.helpers.generic_helper import get_pagination_metadat
 from usaspending_api.common.query_with_filters import QueryWithFilters
 from usaspending_api.financial_activities.models import FinancialAccountsByProgramActivityObjectClass
 
+
 class FederalAccountProgramActivitiesTotal(PaginationMixin, FederalAccountBase):
     """
     Retrieve a list of program activity totals for a federal account.
     """
 
-    endpoint_doc = (
-        "usaspending_api/api_contracts/contracts/v2/views/federal_account_program_activities_total.md"
-    )
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/views/federal_account_program_activities_total.md"
     filters: dict
 
     def __init__(self, **kwargs):
@@ -24,15 +23,16 @@ class FederalAccountProgramActivitiesTotal(PaginationMixin, FederalAccountBase):
         self.sortable_columns = ["obligations", "code", "name", "type"]
 
     def post(self, request: Request, *args, **kwargs):
-        self.json_request = request.data
-        self.filters = self.json_request.get("filter", [])
+        self.filters = request.data.get("filter", [])
         query = self.get_filter_query()
         results = (
             FinancialAccountsByProgramActivityObjectClass.objects.filter(
-                Q(treasury_account__federal_account__federal_account_code=self.federal_account_code,
-                submission__is_final_balances_for_fy=True,
-
-            )).filter(query)
+                Q(
+                    treasury_account__federal_account__federal_account_code=self.federal_account_code,
+                    submission__is_final_balances_for_fy=True,
+                )
+            )
+            .filter(query)
             .annotate(
                 obligations=Sum("obligations_incurred_by_program_object_class_cpe"),
                 code=Coalesce(
@@ -58,7 +58,7 @@ class FederalAccountProgramActivitiesTotal(PaginationMixin, FederalAccountBase):
         page_metadata = get_pagination_metadata(len(results), self.pagination.limit, self.pagination.page)
         return Response(
             {
-                "results": results[self.pagination.lower_limit: self.pagination.upper_limit],
+                "results": results[self.pagination.lower_limit : self.pagination.upper_limit],
                 "page_metadata": page_metadata,
             }
         )
@@ -70,7 +70,10 @@ class FederalAccountProgramActivitiesTotal(PaginationMixin, FederalAccountBase):
             end_date = self.filters["time_period"][0]["end_date"]
             query &= Q(reporting_period_start__gte=start_date, reporting_period_end__lte=end_date)
         if "object_class" in self.filters:
-            query &= Q(Q(object_class__object_class_name__in=self.filters["object_class"]) | Q(object_class__major_object_class_name__in=self.filters["object_class"]))
+            query &= Q(
+                Q(object_class__object_class_name__in=self.filters["object_class"])
+                | Q(object_class__major_object_class_name__in=self.filters["object_class"])
+            )
         if "program_activity" in self.filters:
             if "PARK" in self.filters["program_activity"]:
                 query &= Q(program_activity_reporting_key__isnull=False)
--- a/usaspending_api/accounts/v2/views/federal_account_program_activities_total.py
+++ b/usaspending_api/accounts/v2/views/federal_account_program_activities_total.py
@@ -1,0 +1,81 @@
+from django.db.models import Case, Q, TextField, When, Value, Sum
+from django.db.models.functions import Coalesce
+from rest_framework.response import Response
+from rest_framework.request import Request
+
+from usaspending_api.accounts.v2.views.federal_account_base import FederalAccountBase, PaginationMixin
+from usaspending_api.common.helpers.generic_helper import get_pagination_metadata
+from usaspending_api.common.query_with_filters import QueryWithFilters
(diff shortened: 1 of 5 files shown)


usaspending_api/accounts/v2/views/federal_account_program_activities_total.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    filters: dict
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.default_sort_column = "obligations"
+        self.sortable_columns = ["obligations", "code", "name", "type"]
+
+    def post(self, request: Request, *args, **kwargs):
+        self.filters = request.data.get("filter", [])
```
sethstoudenmier: Need to implement the TinyShield validation for the filters in the request body. We have a couple of patterns for this that you can lean on (such as this example in [agency_base.py](https://github.com/fedspendingtransparency/usaspending-api/blob/master/usaspending_api/agency/v2/views/agency_base.py#L32)).

Additionally, can you make sure to expect a plural "filters" instead of a singular "filter".

sethstoudenmier: If you could implement this validation similar to the agency_base.py example that will allow us to carry forward your logic for all federal account endpoints we need to create.

usaspending_api/accounts/v2/views/federal_account_program_activities_total.py
```diff
@@ -0,0 +1,86 @@
...
+        query = self.get_filter_query()
+        results = (
+            FinancialAccountsByProgramActivityObjectClass.objects.filter(
+                Q(
+                    treasury_account__federal_account__federal_account_code=self.federal_account_code,
+                    submission__is_final_balances_for_fy=True,
+                )
+            )
+            .filter(query)
+            .values("submission_id")
```
sethstoudenmier: The query in close, but it shouldn't be grouping by the submission_id. You can actually have multiple annotation and values statements in order to have the ORM create exactly what you want.

usaspending_api/accounts/v2/views/federal_account_program_activities_total.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+        page_metadata = get_pagination_metadata(len(results), self.pagination.limit, self.pagination.page)
+        return Response(
+            {
+                "results": results[self.pagination.lower_limit : self.pagination.upper_limit],
+                "page_metadata": page_metadata,
+            }
+        )
+
+    def get_filter_query(self) -> Q:
```
sethstoudenmier: Similar to the comment at the top of this file, it would be good to implement this on the base class similar to the agency_base.py example so that other endpoints can be more DRY.

usaspending_api/accounts/v2/views/federal_account_program_activities_total.py
```diff
@@ -0,0 +1,86 @@
...
+        if "time_period" in self.filters:
+            start_date = self.filters["time_period"][0]["start_date"]
+            end_date = self.filters["time_period"][0]["end_date"]
+            query &= Q(reporting_period_start__gte=start_date, reporting_period_end__lte=end_date)
+        if "object_class" in self.filters:
+            query &= Q(
+                Q(object_class__object_class_name__in=self.filters["object_class"])
+                | Q(object_class__major_object_class_name__in=self.filters["object_class"])
+            )
+        if "program_activity" in self.filters:
```
sethstoudenmier: In this case we don't want to filter by if it is PAC/PAN or PARK, but we want to filter by the actual values.

usaspending_api/api_contracts/contracts/v2/federal_accounts/federal_account_totals/program_activities.md
```diff
@@ -0,0 +1,106 @@
...
+# Program Activities Total [/api/v2/federal_accounts/{federal_account_code}/program_activities/total{?limit,page}]
+
+This endpoint returns an array of each program activity's obligation, code, name, and type for the specified federal account 
+
+## POST
+
++ Parameters
+    + `federal_account_code`: `431-0500` (required, string)
+        Federal account code consisting of the AID and main account code
+    + `limit`: 10 (optional, number)
```
sethstoudenmier: Since this is a POST with a request body we can keep the values in the json request instead of the query parameters.

sethstoudenmier (changes requested):

usaspending_api/accounts/v2/views/federal_account_base.py
```diff
@@ -12,6 +13,60 @@ class FederalAccountBase(APIView):
...
+        ]
+
+        return TinyShield(model).block(request_data)
+
+    def get_filter_query(self, validated_data) -> Q:
+        query = Q()
+        filters = validated_data.get("filters")
+        if filters is not None:
+            if "time_period" in filters:
+                start_date = filters["time_period"][0]["start_date"]
```
sethstoudenmier: This is only filtering by the first time period

usaspending_api/accounts/v2/views/federal_account_base.py
```diff
@@ -12,6 +13,60 @@ class FederalAccountBase(APIView):
...
+                query &= Q(
+                    Q(object_class__object_class_name__in=filters["object_class"])
+                    | Q(object_class__major_object_class_name__in=filters["object_class"])
+                )
+            if "program_activity" in filters:
+                query &= Q(program_activity_reporting_key__code__in=filters["program_activity"]) | Q(
+                    program_activity__program_activity_code__in=filters["program_activity"]
+                )
+            else:
+                query &= Q(Q(program_activity_reporting_key__isnull=False) | Q(program_activity__isnull=False))
```
sethstoudenmier: Can we break the not-null check into its own function that is combined with the other filters when implemented? My concern with it here is that if something such as a Federal Account object class or TAS endpoint were to use this then it is incorrectly filtering to not-null program activities.

usaspending_api/api_contracts/contracts/v2/federal_accounts/federal_account_totals/program_activities.md
```diff
@@ -0,0 +1,102 @@
...
+    + Body
+        
+            {
+                "filters": {
+                    "time_period": [
+                        {
+                            "start_date": "2019-10-01",
+                            "end_date": "2020-09-30"
+                        }],
+                    "program_activity": ['0001']
```
sethstoudenmier:
```suggestion
                    "program_activity": ["0001"]
```

usaspending_api/api_contracts/contracts/v2/federal_accounts/federal_account_totals/program_activities.md
```diff
@@ -0,0 +1,102 @@
...
+# Data Structure
+
+## ProgramActivityTotalsFilterObject (object)
++ `federal_account_code`: `431-0500` (required, string)
+Federal account code consisting of the AID and main account code
++ `time_period`: `[
+                        {
+                            "start_date": "2019-10-01",
+                            "end_date": "2020-09-30"
+                        }]`(optional, array[TimePeriod], fixed-type)
```
sethstoudenmier, line 78: If you are going to reference TimePeriod I would either define it below or link to it

usaspending_api/api_contracts/contracts/v2/federal_accounts/federal_account_totals/program_activities.md
```diff
@@ -0,0 +1,102 @@
...
+                    "hasNext": False,
+                    "hasPrevious": False
+                  },
+              }
+
+
+# Data Structure
+
+## ProgramActivityTotalsFilterObject (object)
++ `federal_account_code`: `431-0500` (required, string)
```
sethstoudenmier: This is included in the parameters up above as part of the URI and not the request body

PR #4471, opened 2025-08-24
[DEV-12593] Set one recipient level for Awards

aguest-kc:
## Description:
The `spending_by_category/recipient` endpoint returns one recipient level instead of an array of recipient levels or `None` if there is no recipient level.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-12593](https://federal-spending-transparency.atlassian.net/browse/DEV-12593)

API documentation updated (examples listed below)

Appropriate Operations ticket(s) created
No data change tickets are required for this change.

[DEV-12593]: https://federal-spending-transparency.atlassian.net/browse/DEV-12593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/search/tests/integration/spending_by_category/test_spending_by_recipient.py
```diff
@@ -386,3 +382,56 @@ def test_recipient_search_text_duns(client, monkeypatch, elasticsearch_transacti
...
+            {
+                "amount": 0.0,
+                "recipient_id": "8301261f-8fd1-49d1-9720-8afa3ce6e26f-P",
+                "name": "P RECIPIENT",
+                "code": "DUNS010",
+                "uei": "UEI010",
+                "total_outlays": 0.0,
+            },
+            {
+                "amount": 0.0,
```
sethstoudenmier, line 425: This is actually a good point brought up by the test case. Can you add logic that sets the return of `reicpient_id` to `None` if a level isn't found? This should only occur for our special cases such as `MULTIPLE RECIPIENTS`. Additionally, this allows the frontend to carry forward their logic that greys out the recipient's name (avoiding a link to recipient page) when the ID isn't found.

PR #4474, opened 2025-08-28
[DEV-13147] Restructure the Spark downloads and fix some bugs

sethstoudenmier:
## Description:

This restructures the implementation for the spark downloads and fixes bugs that are a blocker to the release of Spark based downloads.

## Technical Details:

Below is a list of changes in this PR. While this could have been broken down into multiple tickets and PRs, the bulk of the changeset comes from restructuring the spark downloads.

1. Restructure implementation

This moves away from a builder pattern in favor of an abstract factory in an attempt to handle the different combinations that are possible with our downloads. Additionally, it moves the associated code closer to the root of the download app to make it easier to find and reduce length of import statements.

2. Race condition with handling of reader / writer instances

This is a case where a DownloadJob record is created by a request (using the writer) and then read from the DB (using the reader) prior to the replication of the DownloadJob to the reader instances.

3. Downloads fail when more than one submission type is provided

When more than a single submission type is provided the zip file name is changed to a neutral name, while the CSVs remain the same. Without handling this the download would fail to generate if more than one submission type was specified.

4. Downloads fail to retry

The downloads would only work if a DownloadJob was marked as "ready". This change allows for downloads marked as both "ready" and "failed" to begin processing.

5. File A (account balances) downloads should ignore DEFC filter

File A doesn't contain a field for DEFC and thus the filter is ignored when downloading File A. Currently when providing the DEFC filter, the File A downloads will fail to process because the DEFC field cannot be found.

6. Rename File C downloads from "account_download" to "award_financial"

This name was carried over from the proof of concept, but we should rename this to be consistent with how the downloads are used and the data returned.

7. Dates are returned as timestamps

File A date fields are currently returned as timestamps because they are not formatted to be dates.

## Requirements for PR Merge:

9. [ ] API documentation updated (examples listed below)
    1. API Contracts
    2. API UI
    10. Comments
11. [ ] Appropriate Operations ticket(s) created
12. [x] Jira Ticket(s)
    1. [DEV-13147](https://federal-spending-transparency.atlassian.net/browse/DEV-13147)


usaspending_api/download/delta_downloads/account_balances.py
zachflanders-frb, line 1: There will be some conflicts with https://github.com/fedspendingtransparency/usaspending-api/pull/4475. You will be able to remove the global temp tables and use the delta table `account_balances_download` .  Several of the column names are aliased to their final name in the download table, so the groupby, agg, and select column names will need to be updated.

zachflanders-frb (dismissed):
👏 This looks great.

usaspending_api/download/delta_downloads/account_balances.py
```diff
@@ -0,0 +1,272 @@
...
+            sf.col("quarter_format_flag"),
+            sf.concat(sf.lit("FY"), sf.col("reporting_fiscal_year"), sf.lit("Q"), sf.col("reporting_fiscal_quarter")),
+        ).otherwise(
+            sf.concat(
+                sf.lit("FY"),
+                sf.col("reporting_fiscal_year"),
+                sf.lit("P"),
+                sf.lpad(sf.col("reporting_fiscal_period"), 2, "0"),
+            )
+        )
```
zachflanders-frb: I don't think this is needed here anymore since it is used in the `load_query_to_delta` command to create the download table.

PR #4475, opened 2025-08-28
[DEV-12587] - Add Account Balances (File A) Download Delta Table

zachflanders-frb:
## Description:
This PR adds the File A (account balances) Delta Table to support spark based account downloads.

## Technical Details:
The account balances table uses the pyspark dataframe api for the `load_query_to_delta` command and a pyspark schema for the `create_delta_table` command.  These commands have been updated to support this methodology. 

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-12587](https://federal-spending-transparency.atlassian.net/browse/DEV-12587)

[DEV-12587]: https://federal-spending-transparency.atlassian.net/browse/DEV-12587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -415,6 +438,7 @@ def handle(self, *args, **options):
             "spark.sql.legacy.parquet.datetimeRebaseModeInWrite": "LEGACY",  # for dates at/before 1900
             "spark.sql.legacy.parquet.int96RebaseModeInWrite": "LEGACY",  # for timestamps at/before 1900
             "spark.sql.jsonGenerator.ignoreNullFields": "false",  # keep nulls in our json
+            "hive.metastore.disallow.incompatible.col.type.changes": "false",
```
zachflanders-frb: I added this line in order to resolve an error where loading the data was resulting in a `InvalidOperationException(message:The following columns have types incompatible with the existing columns in their respective positions` when the data was being loaded into the existing empty delta table.  If we added the `delta` package as a dependency we could use the merge method on the DeltaTable class ([see here for example](https://docs.delta.io/delta-update/#modify-all-unmatched-rows-using-merge)).  I tested this locally but did not want to add a dependency to the app without having a discussion about it first.

usaspending_api/download/delta_models/account_balances_download.py
```diff
@@ -0,0 +1,170 @@
+from pyspark.sql import DataFrame, SparkSession
+from pyspark.sql import functions as sf, Column
```
sethstoudenmier: Why two different lines to import from the same place?

usaspending_api/etl/management/commands/create_delta_table.py
```diff
@@ -79,18 +81,33 @@ def handle(self, *args, **options):
...
-
-        # Define Schema Using CREATE TABLE AS command
-        spark.sql(
-            TABLE_SPEC[destination_table]["delta_table_create_sql"].format(
-                DESTINATION_TABLE=destination_table_name,
-                DESTINATION_DATABASE=destination_database,
-                SPARK_S3_BUCKET=spark_s3_bucket,
-                DELTA_LAKE_S3_PATH=CONFIG.DELTA_LAKE_S3_PATH,
+        if isinstance(table_spec["delta_table_create_sql"], str):
+            spark.sql(f"create database if not exists {destination_database};")
```
sethstoudenmier: Should this line be moved outside of the conditional in case the database doesn't exist for the DataFrame approach?

usaspending_api/etl/management/commands/load_query_to_delta.py
sethstoudenmier: Does changing this allow for the table to loaded with a schema that doesn't match the schema we provided? For example, if someone accidently loaded an ID field into a name field would this silence those errors?

sethstoudenmier (changes requested):
Marking as requesting changes, but this may not actually need any changes.

usaspending_api/etl/management/commands/create_delta_table.py
zachflanders-frb: ~I don't think this line or the next one are actually needed at all since the table is immediate recreated with a `CREATE OR REPLACE TABLE` or a `df.write.format("delta").mode("overwrite")`~

On second thought this might be necessary, and yes it should be moved outside of the conditional block.

usaspending_api/etl/management/commands/load_query_to_delta.py
zachflanders-frb: Somewhere along the line I changed something that fixed the error and this line is no longer needed.  Perhaps I had some small inconsistency in my schema that I fixed.

PR #4490, opened 2025-09-11
Ftr/dev 9331 naics year retired

loreleitrimberger:
## Description:
This adds a `year_retired` field for naics codes that no longer exist and adds to the load_naics command to populate the field. It also updates the endpoints 
`/api/v2/search/spending_by_category/naics`
`/api/v2/autocomplete/naics`
`/api/v2/references/naics`
to include year_retired in the their results.

## Technical Details:
For the `year_retired` the year will always end up being the next naics year because we update the `naics.year` field to be the youngest year it exists in because of this logic in load_naics
```
if not created:
        if int(naics_year) > int(obj.year):
            NAICS.objects.filter(pk=naics_code).update(description=naics_desc, year=naics_year)
```
so if the naics year is 2012 and doesn't exist in 2017, `year_retired` = 2017

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
4. [x] Appropriate Operations ticket(s) created: [DEV-13448](https://federal-spending-transparency.atlassian.net/browse/DEV-13448)
5. [X] Jira Ticket(s)
    1. [DEV-9331](https://federal-spending-transparency.atlassian.net/browse/DEV-9331)

[DEV-9331]: https://federal-spending-transparency.atlassian.net/browse/DEV-9331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DEV-13448]: https://federal-spending-transparency.atlassian.net/browse/DEV-13448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/references/management/commands/load_naics.py
```diff
@@ -70,6 +71,17 @@ def load_single_naics(naics_code, naics_year, naics_desc):
             NAICS.objects.filter(pk=naics_code).update(description=naics_desc, year=naics_year)
 
 
+def load_naics_year_retired():
```
sethstoudenmier, line 81: I believe we should only have one record for each NAICS. This would mean that aggregating on that max year should be just the year for each NAICS record. Did you happen to see anything different in the data?

sethstoudenmier: To capture the NAICS that were retired in 2017 we would need to load in the 2017 file and see which values aren't in the file but are in our NAICS table.

sethstoudenmier: Nevermind, I misread the queue below

usaspending_api/references/management/commands/load_naics.py
```diff
@@ -70,6 +71,17 @@ def load_single_naics(naics_code, naics_year, naics_desc):
...
+    oldest_naics_year = NAICS.objects.aggregate(models.Max("year"))["year__max"]
+    retired_naics = NAICS.objects.filter(year__lt=oldest_naics_year)
```
sethstoudenmier: Small update, but I would add a check on where the retired year is null so that you don't keep passing over the data that has already been updated. Could optionally log the number of records that were marked as retired

usaspending_api/references/management/commands/load_naics.py
```diff
@@ -70,6 +71,17 @@ def load_single_naics(naics_code, naics_year, naics_desc):
...
+    naics_years = list(NAICS.objects.all().values_list("year", flat=True).distinct().order_by("year"))
+
+    for naics in retired_naics:
+        previous_naics_year_index = naics_years.index(naics.year)
+        naics.year_retired = naics_years[previous_naics_year_index + 1]
+        NAICS.objects.filter(pk=naics.pk).update(year_retired=naics.year_retired)
```
sethstoudenmier: Can you rework this to move the database call for updating outside of the loop? While this is a small table and shouldn't be a problem, I would generally try to avoid a query in a loop that will iterate this much.

sethstoudenmier (changes requested):

sethstoudenmier (approved):

PR #4497, opened 2025-09-24
[DEV-13454] - Transaction search improvements

zachflanders-frb:
## Description:
This PR improves the transaction search query by:
 - Using pyspark dataframe api and delta python library instead of query strings.
 - Removing the subquery in the incremental load.
 - Adding a hash key to compare rows when merging new data during an incremental load.

## Technical Details:
See comments for technical details.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)
    1. [DEV-13454](https://federal-spending-transparency.atlassian.net/browse/DEV-13454)


usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,931 @@
...
+    pop_state_lookup = (
+        state_data.groupBy("code", "name", "fips")
+        .agg(sf.max(sf.col("id")))
+        .select(
+            sf.col("code").alias("pop_state_code"),
+            sf.col("fips").alias("pop_state_fips"),
+        )
+    )
+
+    fed_and_tres_acct = (
```
zachflanders-frb, line 124: I left the logic for these subqueries unchanged from the original query.  Further tuning could be accomplished by exploring whether refactoring these subqueries into groupby clauses would improve performance.

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,931 @@
...
+            *amounts_cols,
+            *recipient_cols,
+            *recipient_location_cols,
+            *place_of_performance_cols,
+            *accounts_cols,
+            *officer_amounts_cols,
+            *fabs_cols,
+            *fpds_cols,
+        )
+        .withColumn("merge_hash_key", sf.xxhash64("*"))
```
zachflanders-frb, line 896: I chose [this hashing algorithm](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.xxhash64.html) to balance out uniqueness vs performance.  [Md5](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.md5.html) or [sha2](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.sha2.html) could provide better uniqueness guarantees.   Second opinions are welcome here.

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,931 @@
...
+def load_transaction_search_incremental(
+    spark: SparkSession, destination_database: str, destination_table_name: str
+) -> None:
+    target = DeltaTable.forPath(spark, f"{delta_root_path}/{destination_database}/{destination_table_name}").alias("t")
+    source = transaction_search_dataframe(spark).alias("s")
+    (
+        target.merge(source, "s.merge_hash_key = t.merge_hash_key")
+        .whenNotMatchedInsertAll()
+        .whenNotMatchedBySourceDelete()
+        .execute()
```
zachflanders-frb: Here we are able to compare against just the `merge_hash_key` instead of the prior version that merged on `transaction_id` and then compared each field in the matching rows.

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
zachflanders-frb: On further reading it seems like `hash` and `xxhash64` can result in collisions. We could use both `transaction_id` and `merge_hash_key` to ensure uniqueness or use md5/sha to produce the hashes.
image
https://preshing.com/20110504/hash-collision-probabilities/

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,914 @@
...
+    return sf.regexp_replace(sf.md5(sf.upper(col)), "^(.{8})(.{4})(.{4})(.{4})(.{12})$", "$1-$2-$3-$4-$5")
+
+
+def extract_numbers_as_string(col: Column, length: int = 2, pad: str = "0") -> Column:
+    return sf.lpad(
+        sf.regexp_extract(col, r"^[A-Z]*(\d+)(?:.\d+)?$", 1).cast(ShortType()).cast(StringType()), length, pad
+    )
+
+
+def transaction_search_dataframe(spark: SparkSession) -> DataFrame:
```
sethstoudenmier, line 26: This is already a huge improvement over the previous SQL approach.  What are your thoughts on an approach that helps to break this apart more instead of having a single function that contains all of this? Not something to do as part of this work but would want to get your thoughts so we can capture it as a potential rework for future.

For example, would having a classmethod that returns the dataframe make sense for grouping all of this logic under a single item (class vs function) and then allow breaking parts of it out into additional classmethods?

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,914 @@
...
+    tfabs = spark.table("int.transaction_fabs")
+    awards = spark.table("int.awards")
+    rcfda = spark.table("global_temp.references_cfda")
+    rlu = spark.table("rpt.recipient_lookup")
+    prl = spark.table("rpt.recipient_lookup").select(
+        sf.col("recipient_hash").alias("parent_recipient_hash"),
+        sf.col("legal_business_name").alias("parent_recipient_name"),
+    )
+    rp = spark.table("rpt.recipient_profile")
+    aa = spark.table("global_temp.agency")
```
sethstoudenmier, line 40: Since this is the awarding agency does it make sense to handle it similar to `fa` where you have it restricted to just the toptier and subtier IDs, and then rename them to be `awarding_*`?

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,914 @@
...
+    )
+    saa = (
+        spark.table("global_temp.subtier_agency")
+        .withColumn("awarding_subtier_agency_abbreviation", sf.col("abbreviation"))
+        .withColumn("awarding_subtier_agency_name", sf.col("name"))
+    )
+    aa_id = spark.table("global_temp.agency").withColumn("awarding_toptier_agency_id", sf.col("id"))
+    fa = (
+        spark.table("global_temp.agency")
+        .withColumn("funding_toptier_agency_id", aa.toptier_agency_id)
```
sethstoudenmier, line 54: Why use `aa.*` in this case?

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,914 @@
...
+        spark.table("global_temp.toptier_agency")
+        .withColumn("awarding_toptier_agency_abbreviation", sf.col("abbreviation"))
+        .withColumn("awarding_toptier_agency_name", sf.col("name"))
+    )
+    saa = (
+        spark.table("global_temp.subtier_agency")
+        .withColumn("awarding_subtier_agency_abbreviation", sf.col("abbreviation"))
+        .withColumn("awarding_subtier_agency_name", sf.col("name"))
+    )
+    aa_id = spark.table("global_temp.agency").withColumn("awarding_toptier_agency_id", sf.col("id"))
```
sethstoudenmier, line 51: For cases such as this and `fa_id` what your thoughts on moving the filter on `toptier_flag` up to the definition of the table instead of when joining?

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,914 @@
...
+        .withColumn("funding_toptier_agency_id", sf.col("toptier_agency_id"))
+        .withColumn("funding_toptier_agency_abbreviation", sf.col("abbreviation"))
+        .withColumn("funding_toptier_agency_name", sf.col("name"))
+    )
+    sfa = (
+        spark.table("global_temp.subtier_agency")
+        .withColumn("funding_subtier_agency_abbreviation", sf.col("abbreviation"))
+        .withColumn("funding_subtier_agency_name", sf.col("name"))
+    )
+    w = Window.partitionBy(aa.toptier_agency_id).orderBy(aa.toptier_flag.desc(), aa.id.asc())
```
sethstoudenmier, line 69: Similar as above, why use `aa.*` in this case when the window is created for `fa`?

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,914 @@
...
+    )
+
+    fed_and_tres_acct = (
+        faba.join(
+            treasury_appropriation_account,
+            treasury_appropriation_account.treasury_account_identifier == faba.treasury_account_id,
+            "inner",
+        )
+        .join(federal_account, federal_account.id == treasury_appropriation_account.federal_account_id, "inner")
+        .join(taa, federal_account.parent_toptier_agency_id == taa.toptier_agency_id, "inner")
```
sethstoudenmier, line 131: I don't think this needs to be changed for this body of work, but more serves as example in support of breaking some of this out from a single function (TBD on what that looks like). In the context of FABA we normally refer to `treasury_appropriation_account` as `TAA` in raw SQL. However, for the dataframe approach we already have the `taa` variable defined for the toptier awarding agency.

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
```diff
@@ -0,0 +1,914 @@
...
+        .join(sfa, fa.funding_subtier_agency_id == sfa.subtier_agency_id, "leftouter")
+        .join(fa_id, ((fa_id.toptier_agency_id == tfa.funding_toptier_agency_id) & (fa_id.row_num == 1)), "leftouter")
+        .join(prl, prl.parent_recipient_hash == parent_recipient_hash, "leftouter")
+        .join(
+            recipient_hash_and_levels,
+            (sf.col("recipient_hash") == recipient_hash_and_levels.recipient_level_hash)
+            & ~(
+                sf.col("legal_business_name").isin(
+                    [
+                        "MULTIPLE RECIPIENTS",
```
sethstoudenmier, line 798: Good case to use this lookup instead now that we have the dataframe approach: https://github.com/fedspendingtransparency/usaspending-api/blob/cf57d470ea1e55916f5b064c78e1949aa36e7b37/usaspending_api/recipient/v2/lookups.py#L17

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
sethstoudenmier: If I am thinking about this right a duplicate hash would mean that a Transaction will appear to have no changes when in fact it did and thus it won't be updated. A duplicate across multiple Transaction records shouldn't matter since we rely on the Transaction ID first for matches. Given the low impact if it occurs and the microscopic chance of it occurring, I am personally fine with this approach.

zachflanders-frb: I updated the code to use both the `transaction_id` and `merge_hash_key` to merge on, so really it's only comparing two values (hashes from the rows with the matching transaction ids in the source and target dataframes), so the changes of a collision are less than 1 in 10<sup>18</sup>.

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
zachflanders-frb: This has been updated to merge on both `transaction_id` and `merge_hash_key` to prevent collisions.

usaspending_api/transactions/delta_models/transaction_search_dataframe.py
zachflanders-frb: Yes, I think that would definitely be a good next step.  I tried to keep the structure the same as the sql script while porting it over to use the dataframe api so that I did not confuse myself, but I think there are opportunities for streamlining and breaking it up.

PR #4499, opened 2025-09-25
Ftr/dev 13328 spark download specify columns

loreleitrimberger:
## Description:
Spark downloads doesn't support the columns filter. These changes make it so when you add columns to the json request only those ones appear in the downloads.

## Technical Details:
This implementation was probably the simplest way to go about it, but I'm not sure if necessarily the best. I tried a different way as well where I had added a columns field to AbstractDownload and AbstractAccountDownloadFactory similar to spark and filter properties and changing what columns to get within object_class_program activity that was getting messy and wasn't working quite right, but this solution so that's why I went with it.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
3. [X] Data validation completed (examples listed below)
5. [X] Jira Ticket(s)
    1. [DEV-13328](https://federal-spending-transparency.atlassian.net/browse/DEV-13328)

[DEV-13328]: https://federal-spending-transparency.atlassian.net/browse/DEV-13328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/download/tests/integration/test_download_accounts.py
+++ b/usaspending_api/download/tests/integration/test_download_accounts.py
@@ -506,8 +506,6 @@ def test_file_c_spark_download_columns(client, download_test_data, s3_unittest_d
                 assert list(df.columns) == ["owning_agency_name", "federal_account_name"]
 
 
-
-
 @pytest.mark.django_db(databases=[settings.DOWNLOAD_DB_ALIAS, settings.DEFAULT_DB_ALIAS])
 def test_file_c_spark_download_unknown_columns(
     client, download_test_data, s3_unittest_data_bucket, hive_unittest_metastore_db
--- a/usaspending_api/download/v2/base_download_viewset.py
+++ b/usaspending_api/download/v2/base_download_viewset.py
@@ -62,7 +62,9 @@ class BaseDownloadViewSet(APIView):
 
         final_output_zip_name = create_unique_filename(json_request, origination=origination)
         download_job = DownloadJob.objects.create(
-            job_status_id=JOB_STATUS_DICT["ready"], file_name=final_output_zip_name, json_request=json.dumps(json_request)
+            job_status_id=JOB_STATUS_DICT["ready"],
+            file_name=final_output_zip_name,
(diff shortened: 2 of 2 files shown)


usaspending_api/download/management/commands/generate_spark_download.py
```diff
@@ -65,6 +65,7 @@ class Command(BaseCommand):
     should_cleanup: bool
     spark: SparkSession
     working_dir_path: Path
+    columns: []
```
zachflanders-frb: In general use `list` instead of `[]` when declaring types, also this could be `None`.
```suggestion
    columns: list | None
```

usaspending_api/download/management/commands/generate_spark_download.py
```diff
@@ -147,13 +149,20 @@ def process_download(self):
...
         try:
             spark_to_csv_strategy = SparkToCSVStrategy(logger)
             zip_file_path = self.working_dir_path / f"{self.download_zip_file_name}.zip"
             download_request = self.get_download_request()
+            if self.columns is not None:
+                for download in download_request.download_list:
+                    source = download.dataframe.columns
+                    valid_columns = all(col in source for col in self.columns)
+                    if not valid_columns:
+                        raise InvalidParameterException("Unknown columns in: {}".format(self.columns))
```
zachflanders-frb: We could switch this to invalid columns to give a better indication of which columns are invalid in the error message:
```suggestion
                    invalid_columns = [col for col in self.columns if col not in source]
                    if invalid_columns:
                        raise InvalidParameterException("Unknown columns: {}".format(invalid_columns))
```

usaspending_api/download/tests/integration/test_download_accounts.py
zachflanders-frb, line 1: Can we write a test that inspects the download dataframe to ensure that it only has the requested columns and that they are in the right order?

PR #4508, opened 2025-10-03
[DEV-13394] Allow for the SQS queue to handle Spark Downloads

sethstoudenmier:
## Description:

This both allows the queue to support Spark downloads and also enables the ability to use a Priority queue for triggering of Spark downloads.

## Technical Details:

Our downloads currently use a single SQS queue. The download worker processes pull messages from this queue and then run them against the Postgres DB. For the case of Spark downloads, we will be handing the processing of these requests over to Databricks but still need to handle logic around the concurrency limit on jobs in Databricks.

The work around for this limit is the following:
* a worker pulls a message from the queue
* if the message is for Spark it will attempt to start the download in Databricks
* after some amount of time the job is checked to see if it was started or skipped
* if the job was started then the worker will loop and pull a new message
* if the job was NOT started (e.g., skipped) then the worker will add the message back to the queue
* lastly, if the concurrency limit is reached OR there are no messages in the priority queue then a message from the normal queue will be pulled for processing

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [x] Jira Ticket(s)
    1. [DEV-13394](https://federal-spending-transparency.atlassian.net/browse/DEV-13394)


Dockerfile.download
zachflanders-frb, line 1: The convention is to name this `<something>.Dockerfile`, (.Dockerfile as the extension), which would result in github and IDEs highlighting this code. (https://docs.docker.com/build/concepts/dockerfile/#filename)

usaspending_api/common/spark/jobs.py
```diff
@@ -48,44 +42,77 @@ def client(self) -> WorkspaceClient:
...
+            job_status = job_run.state.life_cycle_state
+            if job_status not in expected_job_statuses:
+                # Databricks jobs have a start-up time associated that will always run longer than the short wait time
+                # associated with this loop. As a result we only check for three statuses to ensure that the job has
+                # been picked up for processing.
+                job_status = None
+            if job_status is None:
+                time.sleep(5)
+                # Fetch the job_run for next iteration
+                job_run = self.client.jobs.get_run(job_run_id)
```
zachflanders-frb, line 92: What would happen if a job was `PENDING` when this ran but then transitioned to `SKIPPED` later after it exited this loop?  Would it look like it was picked up for processing, but actually it was skipped? It looks like it could transition from `PENDING` -> `SKIPPED` https://docs.databricks.com/aws/en/reference/jobs-2.0-api#runlifecyclestate

sethstoudenmier: That is a good catch on the documentation. I updated it to not allow a PENDING state. I mistook "PENDING" as a required state during the cluster start-up. However, after testing further it turns out that the job is actually in a "RUNNING" state even though the cluster is booting up and says "PENDING" on the job details screen.

Dockerfile.download
sethstoudenmier: I actually remembered you saying this previously in your open work for implementing UV. It seems contradictory, but if you don't mind, I would like to include this change in the UV PR and switch all Dockerfiles to be consistent at the same time.

PR #4511, opened 2025-10-06
[DEV-13558] - upload csv to delta

zachflanders-frb:
## Description:
This PR adds a new command to load a csv into a delta table.

## Technical Details:
The table must already be defined in the destination tables.  There are options for overriding the database name, table name, and path of the csv, but the schema is not overwritable.  The main reason for this is that it would be difficult to support overwriting the schema in the command line because it would need to be a path to a file and then we would need to parse the file.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-13558](https://federal-spending-transparency.atlassian.net/browse/DEV-13558)

~4. [ ] Appropriate Operations ticket(s) created~
        - ~**This will need a data change ticket to upload the world_cities.csv file to s3. We will need to determine the correct/best location on s3 for this csv.**~ This PR has been updated to require a source path instead of referencing a predefined path. 

[DEV-13558]: https://federal-spending-transparency.atlassian.net/browse/DEV-13558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/etl/management/commands/load_csv_to_delta.py
```diff
@@ -0,0 +1,117 @@
...
+    schema: StructType
+    path: str
+
+
+destination_tables = {
+    "world_cities": CsvTableMetadata(
+        db_name="raw",
+        table_name="world_cities",
+        schema=world_cities_schema,
+        path=f"{csv_root}/raw/world_cities/world_cities.csv",
```
zachflanders-frb: We will need a data change ticket to upload this file to s3 for the different environments.

usaspending_api/etl/management/commands/load_csv_to_delta.py
```diff
@@ -0,0 +1,117 @@
...
+        self.load_csv()
+
+        if spark_created_by_command:
+            self.spark.stop()
+
+    def load_csv(self):
+        delta_path = f"{delta_root}/{self.csv_metadata.db_name}/{self.csv_metadata.table_name}"
+        self.spark.sql(f"CREATE DATABASE IF NOT EXISTS {self.csv_metadata.db_name};")
+        df = self.spark.read.csv(self.csv_metadata.path, header=True, schema=self.csv_metadata.schema)
+        df.write.saveAsTable(
```
zachflanders-frb, line 112: The A/C includes saving this file to parquet in addition to a delta table.  I question whether this is needed since it would seem to duplicate data.  However, we could add this after the `saveAsTable` method call if we want to also save it as a parquet file:
```python
df.write.parquet(
    f"{csv_root}/{self.csv_metadata.db_name}/{self.csv_metadata.table_name}/{self.csv_metadata.table_name}.parquet"
    mode="overwrite",  
)
```

usaspending_api/etl/management/commands/load_csv_to_delta.py
sethstoudenmier: Could we instead have a user supply the command with some path to a CSV file that we then move into this location as a parquet file? Thought behind it is that we can then enforce the naming pattern and location of the file powering a Spark table. And will leave the raw CSV untouched in case we need to reference it in the future.

This command doesn't need to account for all possible ways to upload a CSV and can simply handle the case of moving a file from one S3 location to another.

usaspending_api/etl/management/commands/load_csv_to_delta.py
```diff
@@ -0,0 +1,127 @@
...
+        delta_path = f"s3a://{self.spark_s3_bucket}/{CONFIG.DELTA_LAKE_S3_PATH}/{self.csv_metadata.db_name}/{self.csv_metadata.table_name}"
+        csv_path = f"s3a://{self.spark_s3_bucket}/{CONFIG.SPARK_CSV_S3_PATH}/{self.csv_metadata.db_name}/{self.csv_metadata.table_name}"
+        self.spark.sql(f"CREATE DATABASE IF NOT EXISTS {self.csv_metadata.db_name};")
+        df = self.spark.read.csv(self.csv_metadata.path, header=True, schema=self.csv_metadata.schema)
+        created_at = str(datetime.now())
+        df.write.parquet(f"{csv_path}/{self.csv_metadata.table_name}-{created_at}.parquet", mode="overwrite")
+        df.write.saveAsTable(
+            f"{self.csv_metadata.db_name}.{self.csv_metadata.table_name}",
+            mode="overwrite",
+            format="delta",
```
sethstoudenmier, line 115: Realizing now that the ticket had mixed details. In this case we want to create a Spark table in Parquet format, not a Delta table.

usaspending_api/etl/management/commands/load_csv_to_delta.py
```diff
@@ -0,0 +1,124 @@
...
+    schema: StructType
+    source_path: str
+
+
+destination_tables = {
+    "world_cities": CsvTableMetadata(
+        db_name="raw",
+        table_name="world_cities",
+        schema=world_cities_schema,
+        source_path=f"s3a://{CONFIG.SPARK_S3_BUCKET}/{CONFIG.SPARK_CSV_S3_PATH}/worldcities.csv",
```
sethstoudenmier: If I remember correctly, we are removing this from the definition and requiring it as input.

usaspending_api/etl/management/commands/load_csv_to_delta.py
sethstoudenmier: Following up on this:

Previously the plan was to create a Spark table on top of the Parquet file (not an actual Delta table). Since then we have made the decision to load the CSV as a Delta table removing the need to manage the parquet files directly as it will be managed by Delta.

usaspending_api/etl/management/commands/load_csv_to_delta.py
sethstoudenmier: Resolving as this is no longer the case

sethstoudenmier (changes requested):
Small request assuming I am not mistaken

sethstoudenmier (approved):

PR #4517, opened 2025-10-15
Fix Spark Dataframe Difference Test

zachflanders-frb:
One of the spark dataframe difference tests fails in the case when one of the random integers is 0.


usaspending_api/common/tests/integration/test_spark_dataframe_diff.py
```diff
@@ -71,7 +71,7 @@ def test_diff_one_change(spark: SparkSession):
 
     data_right = deepcopy(data_left)
     changed_row_id = data_right[0]["id"]
-    data_right[0]["numeric_val"] = data_right[0]["numeric_val"] * 2
+    data_right[0]["numeric_val"] = data_right[0]["numeric_val"] + 1
```
zachflanders-frb, line 74: In the case that the random integer is zero, this test will fail because multiplying by 2 does not introduce a difference.

PR #4526, opened 2025-10-24
[DEV-13694] add pagination for assistance listing

loreleitrimberger:
## Description:
This adds pagination for assistance listing. Pagination is only there when a code is provided otherwise there's no a good way to go about paginating that. 

## Technical Details:
For endpoint results this is the simplified explanation:
children: filter and no code
no children: code, code and filter, no code and no filter
pagination: code, code and filter
no pagination: no code and no filter, just filter

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [X] Jira Ticket(s)
    1. [DEV-13694](https://federal-spending-transparency.atlassian.net/browse/DEV-13694)

[DEV-13694]: https://federal-spending-transparency.atlassian.net/browse/DEV-13694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/api_contracts/contracts/v2/references/assistance_listing.md
```diff
@@ -12,36 +12,54 @@ includes the full CFDA code and description.
...
+        + `sort`: `sort=description` (optional, enum) `code` and `description` are the only valid inputs. The default is `code`.
+        + `limit`: `limit=15` (optional, number) How many results appear on a page. The default is 10.
+        + `order`: `order=asc` (optional, string) What order results are returned in. The default is descending.
 
 + Response 200 (application/json)
     + Attributes (object)
         + `results` (required, array[AssistanceListingObject], fixed-type)
   
-    + Body
+          + Body
```
sethstoudenmier: I believe this is indented too much

usaspending_api/references/v2/views/filter_tree/assistance_listing.py
```diff
@@ -36,40 +38,81 @@ def _parse_and_validate_request(self, cfda: str | None, requested_data: Request)
...
+            .order_by(*page.robust_order_by_fields)
         )
 
         return results
 
     @cache_response()
     def get(self, request: Request, cfda=None) -> Response:
         requested_data = self._parse_and_validate_request(cfda, request)
-        results = self._business_logic(requested_data["code"], requested_data["filter"])
+        page = PaginationMixin().pagination(request)
```
sethstoudenmier: A mixin is intended to be inherited such as `class SomeClass(PaginationMixin, BaseClass):`. In this case you should be able to define the mixin logic as a new function for the class, or inherit from the mixin. Either way, I wouldn't create an instance of the mixin directly. If this is the only class that will be using this logic then I would lean towards defining it from inside of the class as a function instead of creating a new one for the pagination logic.

zachflanders-frb: It looks like we have a class called `PaginationMixin` defined 3 other places in our code with similar or identical code.  This seems like it might be a good opporunity to consolidate these classes into the `common/` directory.

loreleitrimberger: I went through and added a `PaginationMixin` helper class. The only PaginationMixin class I didn't change was the one defined in disaster_base.py because that one had a lot more going on with it compared to the others that were all relatively the same.

usaspending_api/agency/v2/views/budget_function.py
zachflanders-frb, line 25: Each class that inherits from the `PaginationMixin` will need to define a `self.default_sort_column` and `self.sortable_columns` in its `__init__`.

usaspending_api/references/v2/views/filter_tree/assistance_listing.py
```diff
@@ -42,34 +44,57 @@ def _business_logic(self, cfda_code: int | None, cfda_filter: str | None) -> Que
...
+            .order_by(*self.pagination.robust_order_by_fields)
         )
 
         return results
 
     @cache_response()
     def get(self, request: Request, cfda=None) -> Response:
         requested_data = self._parse_and_validate_request(cfda, request)
+        self.sortable_columns = ["code", "description"]
+        self.default_sort_column = "code"
```
zachflanders-frb: Attributes on `self` should be defined in the `__init__` method. E.g.
```python
        self.default_sort_column = "code"
        self.sortable_columns = ["code", "description"]
```

zachflanders-frb (approved):

PR #4531, opened 2025-10-31
[DEV-12528] Remove Hadoop copy merge step from Spark downloads

aguest-kc:
## Description:
Remove the Hadoop copy merge step from Spark downloads and use the correct number of partitions instead of 1.

## Technical Details:
Updated the number of partitions to be `total records / max records per file` instead of 1. Skipped the Hadoop copy merge step since the previous change results in the correct number of records in each row. Added a method for renaming the part files to the expected format.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-12528](https://federal-spending-transparency.atlassian.net/browse/DEV-12528)

2. API documentation updated (examples listed below)
No API contracts need to be updated for this change.

4. Appropriate Operations ticket(s) created
No operation tickets are needed for this change.

[DEV-12528]: https://federal-spending-transparency.atlassian.net/browse/DEV-12528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/common/etl/spark.py
sethstoudenmier: Can this now be removed?

usaspending_api/common/helpers/s3_helpers.py
```diff
@@ -165,3 +165,21 @@ def delete_s3_objects(
     resp = s3_client.delete_objects(Bucket=bucket_name, Delete={"Objects": object_list})
 
     return len(resp.get("Deleted", []))
+
+
+def rename_s3_object(bucket_name: str, old_key: str, new_key: str, region_name: str = settings.USASPENDING_AWS_REGION):
```
sethstoudenmier, line 172: Do we have to create a copy of the object? Is there any performance benefit to pulling it down onto the Driver then performing the rename of the file? I wonder how much of the time is spent copying the S3 objects for large downloads.

sethstoudenmier (dismissed):
Approving, but left some comments are possible cleanup / improvements.

usaspending_api/common/helpers/s3_helpers.py
aguest-kc: From everything I saw, S3 doesn't support in-place renaming so the only way to rename a file is to copy it to a new file with the desired file name.

We could pull the `part-###.csv` files down locally and then do the renaming. I did the renaming in S3 to keep the workflow the same as before.

usaspending_api/common/etl/spark.py
aguest-kc: It probably can be removed now if we don't plan on using it anywhere else in the future. I forgot to ask on standup if there were any known plans to reuse this function.

I have no issues with removing it.

aguest-kc: Removed unused function.

usaspending_api/common/helpers/s3_helpers.py
aguest-kc: Leaving this as-is for now, but I'll create a follow up ticket to do the renaming locally instead of in S3.

PR #4548, opened 2025-11-20
[DEV-13726] Implement File B downloads in DuckDB

aguest-kc:
## Description:
Add DuckDB as an option for File B Custom Account downloads.

## Technical Details:
DuckDB's Spark implementation doesn't implement every Spark/PySpark function so some substitutions had to be made when using DuckDB. 

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-13726](https://federal-spending-transparency.atlassian.net/browse/DEV-13726)

2. API documentation updated (examples listed below)

[DEV-13726]: https://federal-spending-transparency.atlassian.net/browse/DEV-13726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/common/etl/spark.py
+++ b/usaspending_api/common/etl/spark.py
@@ -760,8 +760,9 @@ def write_csv_file_duckdb(
     """
     start = time.time()
     _pandas_df = df.toPandas()
-    _pandas_df["file_number"] = (_pandas_df.index // max_records_per_file ) + 1
+    _pandas_df["file_number"] = (_pandas_df.index // max_records_per_file) + 1
     df_record_count = len(_pandas_df)
+    rel = duckdb.from_df(_pandas_df)
 
     full_file_paths = []
 


Dockerfile
```diff
@@ -16,11 +16,19 @@ RUN apt update && apt install -y \
...
 COPY . /dockermount
+
 RUN python3 -m pip install -r requirements/requirements.txt && \
     python3 -m pip install -r requirements/requirements-server.txt && \
     python3 -m pip install ansible==2.9.15 awscli==1.34.19
 
-##### Ensure Python STDOUT gets sent to container logs
+# Install DuckDB extensions
+
+RUN mkdir -p /root/.duckdb/extensions/v1.4.1/linux_amd64 && \
```
sethstoudenmier: Probably need another Dockerfile for this because otherwise we are going to add this to a lot of infrastructure where it isn't needed.

usaspending_api/common/etl/spark.py
sethstoudenmier, line 1: Looks like your linter had a field day with this file / GitHub decided to show the entire file as changed. Do you mind summarizing change to this for ease of comparison / review?

usaspending_api/download/delta_downloads/object_class_program_activity.py
sethstoudenmier, line 1: Similar here with GitHub showing the entire file as changed; do you mind summarizing changes?

usaspending_api/download/management/commands/generate_spark_download.py
```diff
@@ -66,20 +69,45 @@ class Command(BaseCommand):
...
         self.download_job = self.get_download_job(options["download_job_id"])
         self.download_json_request = json.loads(self.download_job.json_request)
         self.request_type = self.download_json_request["request_type"]
         self.working_dir_path = Path(settings.CSV_LOCAL_PATH)
         self.columns = self.download_json_request["columns"] if "columns" in self.download_json_request else None
+
+        self.use_duckdb = options["use_duckdb"]
+        if options["use_duckdb"]:
+            self.spark: DuckDBSparkSession = DuckDBSparkSession.builder.getOrCreate()
+            spark_created_by_command = False
```
sethstoudenmier: Should this be `True` so that we run `.stop()`?

usaspending_api/download/management/commands/generate_spark_download.py
```diff
@@ -66,20 +69,45 @@ class Command(BaseCommand):
...
-        self.spark, spark_created_by_command = self.setup_spark_session()
         self.should_cleanup = not options["skip_local_cleanup"]
         self.download_job = self.get_download_job(options["download_job_id"])
         self.download_json_request = json.loads(self.download_job.json_request)
         self.request_type = self.download_json_request["request_type"]
         self.working_dir_path = Path(settings.CSV_LOCAL_PATH)
         self.columns = self.download_json_request["columns"] if "columns" in self.download_json_request else None
+
+        self.use_duckdb = options["use_duckdb"]
+        if options["use_duckdb"]:
```
sethstoudenmier: What are your thoughts on moving this to `setup_spark_session()` method?

sethstoudenmier (changes requested):

Dockerfile
aguest-kc: Do we want to add it to the Spark Dockerfile? Currently, DuckDB is only used in places where Spark is used.

usaspending_api/download/management/commands/generate_spark_download.py
aguest-kc: It _shouldn't_ matter, but I'll update it just to be safe.

usaspending_api/download/management/commands/generate_spark_download.py
aguest-kc: I think that makes sense and makes this easier to read.

usaspending_api/download/management/commands/generate_spark_download.py
aguest-kc: Updated this to `True`

usaspending_api/download/management/commands/generate_spark_download.py
aguest-kc: Moved this to `setup_spark_session()`

usaspending_api/common/etl/spark.py
aguest-kc: Ran `dos2unix` so this should be readable now. I should have noticed every line marked as changed before I committed.

usaspending_api/download/delta_downloads/object_class_program_activity.py
aguest-kc: Ran `dos2unix` so this should be readable now. I should have noticed every line marked as changed before I committed.

Dockerfile
sethstoudenmier: I was thinking of having a new dockerfile that way whatever our final architecture looks like we could install DuckDB on only the resources that need it.

usaspending_api/common/etl/spark.py
```diff
@@ -555,31 +561,122 @@ def _generate_global_view_sql_strings(tables: List[str], jdbc_url: str) -> List[
...
+                )
+            else:
+                # DuckDB will prepend the HTTP or HTTPS so we need to strip it from the AWS endpoint URL
+                endpoint_url = os.getenv("AWS_ENDPOINT_URL", "s3.us-gov-west-1.amazonaws.com")
+                cleaned_endpoint_url = endpoint_url.split("://")[1] if "://" in endpoint_url else endpoint_url
+                spark.sql(
+                    f"""
+                    CREATE OR REPLACE SECRET (
+                        TYPE s3,
+                        REGION 'us-gov-west-1',
```
sethstoudenmier:
```suggestion
                        REGION '{settings.USASPENDING_AWS_REGION}',
```

usaspending_api/common/etl/spark.py
```diff
@@ -555,31 +561,122 @@ def _generate_global_view_sql_strings(tables: List[str], jdbc_url: str) -> List[
...
+                        SECRET '{os.getenv("MINIO_ROOT_PASSWORD", "usaspender")}',
+                        ENDPOINT '{endpoint_url}',
+                        URL_STYLE 'path',
+                        USE_SSL 'false'
+                    );
+                """
+                )
+            else:
+                # DuckDB will prepend the HTTP or HTTPS so we need to strip it from the AWS endpoint URL
+                endpoint_url = os.getenv("AWS_ENDPOINT_URL", "s3.us-gov-west-1.amazonaws.com")
```
sethstoudenmier: Similar to below, if we are going to provide a default then we should use `settings.USASPENDING_AWS_REGION` in the default

usaspending_api/common/etl/spark.py
```diff
@@ -555,31 +561,122 @@ def _generate_global_view_sql_strings(tables: List[str], jdbc_url: str) -> List[
...
+
+            _download_delta_tables = [
+                {"schema": "rpt", "table_name": "account_balances_download"},
+                {"schema": "rpt", "table_name": "object_class_program_activity_download"},
+            ]
+
+            # The DuckDB Delta extension is needed to interact with DeltaLake tables
+            spark.sql("LOAD delta; CREATE SCHEMA IF NOT EXISTS rpt;")
+            for table in _download_delta_tables:
+                s3_path = f"s3://{CONFIG.SPARK_S3_BUCKET}/data/delta/{table['schema']}/{table['table_name']}"
```
sethstoudenmier:
```suggestion
                s3_path = f"s3://{CONFIG.SPARK_S3_BUCKET}/{DELTA_LAKE_S3_PATH}/{table['schema']}/{table['table_name']}"
```

usaspending_api/common/etl/spark.py
```diff
@@ -555,31 +561,122 @@ def _generate_global_view_sql_strings(tables: List[str], jdbc_url: str) -> List[
...
+                    tables=_BROKER_REF_TABLES,
+                    jdbc_url=get_broker_jdbc_url(),
+                )
+                logger.info(
+                    f"Creating the following Broker tables under the global_temp database: {_BROKER_REF_TABLES}"
+                )
+                for sql_statement in broker_sql_strings:
+                    spark.sql(sql_statement)
+        case _:
+            raise RuntimeError(f"Unsupported spark session type: {type(spark)}")
```
sethstoudenmier: Is there a case where we would reach this since this is a `match...case` statement validating on True or False?

sethstoudenmier (changes requested):

usaspending_api/common/etl/spark.py
aguest-kc: Updated to use the `DELTA_LAKE_S3_PATH` variable

usaspending_api/common/etl/spark.py
aguest-kc: Forgot about this setting. Updated.

usaspending_api/common/etl/spark.py
aguest-kc: Forgot about this setting. Updated.

usaspending_api/common/etl/spark.py
aguest-kc: No there isn't and I think this was left from another implementation that I tried earlier. I'll remove it.

usaspending_api/common/etl/spark.py
```diff
@@ -555,31 +561,122 @@ def _generate_global_view_sql_strings(tables: List[str], jdbc_url: str) -> List[
...
-        )
-        logger.info(f"Creating the following Broker tables under the global_temp database: {_BROKER_REF_TABLES}")
-        for sql_statement in broker_sql_strings:
-            spark.sql(sql_statement)
+
+    match isinstance(spark, DuckDBSparkSession):
+        case True:
+            logger.info("Creating ref temp views using DuckDB")
+
+            if IS_LOCAL:
```
sethstoudenmier, line 583: I would look to use the Config more in this conditional block.

For example, you have `CONFIG.AWS_S3_ENDPOINT` which will get you the `endpoint_url` for both the local and deployed environments. And then `CONFIG.AWS_ACCESS_KEY` and `CONFIG.AWS_SECRET_KEY` point to the same as the user and password you have below.

usaspending_api/common/etl/spark.py
```diff
@@ -555,31 +561,122 @@ def _generate_global_view_sql_strings(tables: List[str], jdbc_url: str) -> List[
...
+                try:
+                    spark.sql(
+                        f"""
+                        CREATE OR REPLACE TABLE {table["schema"]}.{table["table_name"]} AS
+                        SELECT * FROM delta_scan('{s3_path}');
+                    """
+                    )
+                    logger.info(f"Successfully created table {table['schema']}.{table['table_name']}")
+                except duckdb.IOException:
+                    logger.error(f"Failed to create table {table['table_name']}")
```
sethstoudenmier: Should this include a `raise` on the line after the log?

sethstoudenmier: Same with some of the other `try...except` blocks below. Do we not want to raise the exceptions? And if they should raise the exceptions, can we change it to be `logger.exception()`?

usaspending_api/common/etl/spark.py
```diff
@@ -635,6 +733,78 @@ def write_csv_file(
...
+    # Convert the Spark DataFrame to a DuckDBPyRelation type to take advantage of the built-in functions
+    # flake8 checks don't see this variable as being used even though it's used in the SQL query below
+    _pandas_df = df.toPandas()  # noqa: F841
+
+    rel = duckdb.sql(
+        f"""
+        SELECT
+            *,
+            CAST((ROW_NUMBER() OVER () - 1) / {max_records_per_file - 1} AS integer) + 1 AS file_number
+        FROM _pandas_df;
```
sethstoudenmier: This syntax of using a variable in the SQL caught me off guard, but I follow now. Does DuckDB support a Dataframe type approach so we can move away from raw SQL more?

Dockerfile
aguest-kc: No issue with creating a new Dockerfile for DuckDB. I can create a new `Dockerfile.duckdb` using the `usaspending-backend` image as the base.

usaspending_api/common/etl/spark.py
aguest-kc: DuckDB has their [relational API](https://duckdb.org/docs/stable/clients/python/relational_api) which is their way of lazily building a query that isn't executed until you fetch the rows.

Unfortunately, I wasn't able to implement this `ROW_NUMBER() OVER ()` logic using that, but I did update this logic to happen in the Pandas dataframe instead of raw SQL.

aguest-kc: I kept the `to_csv()` writing in DuckDB since it's implemented in C++ instead of using Pandas' `to_csv()` which is implemented in Python.

usaspending_api/common/etl/spark.py
aguest-kc: Updated to use the `CONFIG` values

usaspending_api/common/etl/spark.py
aguest-kc: Updated these `try... except` blocks and the `logger` to use exception.

sethstoudenmier (approved):

PR #4555, opened 2025-12-10
[DEV-15050] Threaded downloads

aguest-kc:
## Description:
Use `threading` for the S3 renaming and downloading steps in the Spark download process.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-15050](https://federal-spending-transparency.atlassian.net/browse/DEV-15050)

[DEV-15050]: https://federal-spending-transparency.atlassian.net/browse/DEV-15050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/database_scripts/servers/broker_server.sql
sethstoudenmier, line 1: Assuming this was changed for local testing. Need to undo these changes.

usaspending_api/common/etl/spark.py
```diff
@@ -884,10 +885,81 @@ def rename_part_files(
...
+    failed_count = 0
+    completed_count = 0
+
+    with ThreadPoolExecutor(max_workers=max_threads) as executor:
+        futures = [
+            executor.submit(rename_single_file, idx, part_file) for idx, part_file in enumerate(list_of_part_files)
+        ]
+
+        for future in as_completed(futures):
+            new_path, success = future.result()
```
sethstoudenmier: If the thread encounters an error the exception is raised when you call `.result()` on the future. I would wrap this in a `try...except` so that you can keep track of which fails passed, which succeeded, and then provide that back in the logs while still raising an exception.

usaspending_api/common/etl/spark.py
```diff
@@ -884,10 +885,81 @@ def rename_part_files(
...
+            completed_count += 1
+
+            if success:
+                full_file_paths.append(new_path)
+
+    if failed_count > 0:
```
sethstoudenmier, line 970: Similar to the comment above, it looks like you intended to capture a total success and failure before raising an error. However, the `failed_count` is never incremented and the `try...except` is needed as well.

usaspending_api/common/etl/spark.py
```diff
@@ -884,10 +885,81 @@ def rename_part_files(
...
+        """
+
+        new_key = f"{temp_download_dir_name}/{destination_file_name}_{index + 1:02d}.{file_format}"
+        try:
+            rename_s3_object(bucket_name=bucket_name, old_key=part_file, new_key=new_key)
+            logger.info(f"Successfully renamed {part_file} to {new_key}")
+            return f"s3a://{bucket_name}/{new_key}", True
+        except Exception as e:
+            logger.error(f"Failed to rename {part_file} to {new_key}: {e}", exc_info=True)
+            return None, False
```
sethstoudenmier, line 946: I would let it still error out and log the exception via `logger.exception()` instead of `.error()`, so it appears in the stack trace. But then the `.result()` call can handle the exception below.

sethstoudenmier (changes requested):

usaspending_api/common/etl/spark.py
aguest-kc: Updated to use `logger.exception`

usaspending_api/common/etl/spark.py
aguest-kc: Good catch, that was the intention and I just forgot to increment `failed_count`.

usaspending_api/database_scripts/servers/broker_server.sql
aguest-kc: I think this was an artifact from the DB restore process. Reverted the changes.

sethstoudenmier (approved):

PR #4556, opened 2025-12-10
[DEV-13938] Partition the Custom Account download Delta tables

aguest-kc:
## Description:
Partition the Custom Account download Delta tables

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-13938](https://federal-spending-transparency.atlassian.net/browse/DEV-13938)

[DEV-13938]: https://federal-spending-transparency.atlassian.net/browse/DEV-13938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/download/delta_models/award_financial_download.py
+++ b/usaspending_api/download/delta_models/award_financial_download.py
@@ -334,9 +334,7 @@ def award_financial_df(spark: SparkSession):
 
 def load_award_financial(spark: SparkSession, destination_database: str, destination_table_name: str) -> None:
     df = award_financial_df(spark)
-    df.write.format("delta").mode("overwrite").partitionBy(
-        "reporting_fiscal_year", "funding_toptier_agency_id"
-    ).saveAsTable(f"{destination_database}.{destination_table_name}")
+    df.write.format("delta").mode("overwrite").saveAsTable(f"{destination_database}.{destination_table_name}")
 
 
 def load_award_financial_incremental(
--- a/usaspending_api/download/delta_models/object_class_program_activity_download.py
+++ b/usaspending_api/download/delta_models/object_class_program_activity_download.py
@@ -280,9 +280,7 @@ def load_object_class_program_activity(
     spark: SparkSession, destination_database: str, destination_table_name: str
 ) -> None:
     df = object_class_program_activity_df(spark)
-    df.write.format("delta").mode("overwrite").partitionBy(
-        "reporting_fiscal_year", "funding_toptier_agency_id"
-    ).saveAsTable(f"{destination_database}.{destination_table_name}")
+    df.write.format("delta").mode("overwrite").saveAsTable(f"{destination_database}.{destination_table_name}")
 
 
 def load_object_class_program_activity_incremental(


usaspending_api/download/delta_models/account_balances_download.py
```diff
@@ -170,7 +172,9 @@ def account_balances_df(spark: SparkSession) -> DataFrame:
 
 def load_account_balances(spark: SparkSession, destination_database: str, destination_table_name: str) -> None:
     df = account_balances_df(spark)
-    df.write.format("delta").mode("overwrite").saveAsTable(f"{destination_database}.{destination_table_name}")
+    df.write.format("delta").mode("overwrite").partitionBy(
```
sethstoudenmier: Should verify, but I believe the data should be partitioned upon writing to the Delta table since the Delta table itself is partitioned. When testing locally I was able to define the partitioned at table creation and then run a simple `INSERT INTO` and see data partitioned.

Did you come across anything that needed the partitions defined on the DataFrame Writer?

sethstoudenmier: I didn't test with an overwrite, but I would hope that the partitions in the definition would be protected.

sethstoudenmier (changes requested):
Marking as request changes, but this is more so seeking clarification to make sure a change is necessary.

usaspending_api/download/delta_models/account_balances_download.py
aguest-kc: This appears to working with the `partitionBy()` in the `INSERT` statement. This was leftover from when I was testing why I couldn't see the partitions in Spark locally and I didn't remove it. It's been removed now.

usaspending_api/download/delta_models/object_class_program_activity_download.py
```diff
@@ -279,7 +280,9 @@ def load_object_class_program_activity(
     spark: SparkSession, destination_database: str, destination_table_name: str
 ) -> None:
     df = object_class_program_activity_df(spark)
-    df.write.format("delta").mode("overwrite").saveAsTable(f"{destination_database}.{destination_table_name}")
+    df.write.format("delta").mode("overwrite").partitionBy(
```
sethstoudenmier: Similar comment here for `partitionBy()` since the Delta table itself should be partitioned.

usaspending_api/download/delta_models/award_financial_download.py
```diff
@@ -332,7 +334,9 @@ def award_financial_df(spark: SparkSession):
 
 def load_award_financial(spark: SparkSession, destination_database: str, destination_table_name: str) -> None:
     df = award_financial_df(spark)
-    df.write.format("delta").mode("overwrite").saveAsTable(f"{destination_database}.{destination_table_name}")
+    df.write.format("delta").mode("overwrite").partitionBy(
```
sethstoudenmier: Similar comment here for `partitionBy()` since the Delta table itself should be partitioned.

aguest-kc: I forgot this PR included all three tables. The other two tables have been updated.

PR #4563, opened 2025-12-16
[DEV-14145] Implement file A downloads in DuckDB

aguest-kc:
## Description:
Add the ability to use DuckDB for file A Custom Account downloads.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-14145](https://federal-spending-transparency.atlassian.net/browse/DEV-14145)

[DEV-14145]: https://federal-spending-transparency.atlassian.net/browse/DEV-14145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/download/delta_downloads/account_balances.py
+++ b/usaspending_api/download/delta_downloads/account_balances.py
@@ -273,4 +273,3 @@ class AccountBalancesDownloadFactory(AbstractAccountDownloadFactory):
 
     def create_treasury_account_download(self) -> TreasuryAccountDownload:
         return TreasuryAccountDownload(self.spark, self.filters, self.dynamic_filters)
-


usaspending_api/download/delta_downloads/account_balances.py
```diff
@@ -18,27 +21,39 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if isinstance(self.spark, DuckDBSparkSession):
+            from duckdb.experimental.spark.sql import functions
+        else:
+            from pyspark.sql import functions
+
+        self.sf = functions
```
zachflanders-frb, line 40: I would probably lean towards importing both of these at the top instead of having the imports in a conditional and then setting the right one here.  e.g.
```python
from duckdb.experimental.spark.sql import functions as duckdb_functions
from pyspark.sql import functions as spark_functions
```
```python
            self.sf = duckdb_functions if isinstance(self.spark, DuckDBSparkSession) else spark_functions
```

zachflanders-frb (dismissed):

usaspending_api/download/delta_downloads/account_balances.py
aguest-kc: I thought about doing it this way, but originally went with the conditional import to "not do unnecessary imports". That idea is kind of out the window considering I had to add several other imports for DuckDB anyway.

For this file in particular, I kept this conditional import to keep the same style as the (already merged) file B code.

sethstoudenmier (approved):

PR #4567, opened 2025-12-30
Ftr/dev 14235 update filter tree child count

loreleitrimberger:
## Description:
This updates the count for the tiers in psc filter tree. Now the count follows this pattern:
A - 5
   A1 -2
      A11 - null
      A12 - null
   A2
      A21 - null
B - 2
   B1 - 1
      B11 - null

## Technical Details:
The way I went about this was counting the lowest tier first and working up for the count.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
3. [X] Data validation completed (examples listed below)
5. [X] Jira Ticket(s)
    1. [DEV-14235](https://federal-spending-transparency.atlassian.net/browse/DEV-14235)

[DEV-14235]: https://federal-spending-transparency.atlassian.net/browse/DEV-14235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
+++ b/usaspending_api/api_contracts/contracts/v2/autocomplete/recipient.md
@@ -40,7 +40,11 @@ This route sends a request to the backend to retrieve recipients matching the sp
     + Attributes (object)
         + `limit` (optional, number)
         + `search_text` (required, string)
-        + `recipient_levels` (optional, array[string])
+        + `recipient_levels` (optional, array[enum[string]])
+            + Members
+                + `C`
+                + `P`
+                + `R`
     + Body
 
             {
@@ -58,19 +62,13 @@ This route sends a request to the backend to retrieve recipients matching the sp
                 "results": [
                     {
                         "recipient_name": "LOCKHEED MARTIN",
-                        "recipient_level": "P",
+                        "recipient_level": null,
                         "uei": "PUETYBNR91Z3"
                         "duns": "956973523"
                     },
                     {
                         "recipient_name": "LOCKHEED MARTIN",
-                        "recipient_level": "C",
-                        "uei": "PUETYBNR91Z3",
-                        "duns": "956973523"
-                    },
-                    {
-                        "recipient_name": "LOCKHEED MARTIN",
-                        "recipient_level": "R",
+                        "recipient_level": null,
                         "uei": "null",
                         "duns": "null"
                     }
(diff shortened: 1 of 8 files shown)


usaspending_api/references/v2/views/filter_tree/psc_filter_tree.py
```diff
@@ -250,13 +251,66 @@ def _path_is_valid(self, path: list) -> bool:
                     return False
         return True
 
-    def get_count(self, tiered_keys: list, id) -> int:
+    def get_count(self, tiered_keys: list, id) -> int | None:
```
zachflanders-frb: This function signature is a bit confusing to me.  It seems that items in the `tiered_keys` list are never actually used.  It functions more as a boolean flag to determine whether this is a top level count or a sub tier count.  It is being called with an empty list when it is a top tier count.  I would make the id the first parameter and the second parameter a boolean flag about whether to search the top tier.  Since there is no shared code between the two options, you could even make them two separate functions with a single parameter `id` or `key`.

usaspending_api/references/v2/views/filter_tree/psc_filter_tree.py
```diff
@@ -250,13 +251,66 @@ def _path_is_valid(self, path: list) -> bool:
...
-            ]
-            return PSC.objects.filter(*filters).count()
+            if id.startswith("A"):
+                return self.get_research_and_dev_count(id)
+            elif re.search(r"^[B-Z]", id):
+                return self.get_service_count(id)
+            elif re.search(r"^\d\d", id):
+                return self.get_product_count(id)
+
+    def get_tier_count(self, tier_length, id, is_regex=False, is_r_and_d=False) -> int:
```
zachflanders-frb: This method doesn't use `self` so it could be made a `@staticmethod`
```suggestion
    @staticmethod
    def get_tier_count(tier_length: int, id" str, is_regex: bool = False, is_r_and_d: bool = False) -> int:
```

usaspending_api/references/v2/views/filter_tree/psc_filter_tree.py
zachflanders-frb, line 1: Could we add type hints to the function parameters?

zachflanders-frb, line 1: One drawback of building the count iteratively based on the length of the id is that we are querying the database 1-3 distinct times.  I think we could adjust this to query the database once by adjusting the parameter for what we are filtering out e.g. (`.filter(code_len__gte=tier_length)`)

PR #4570, opened 2026-01-06
[DEV-14110] initial_report_date and last_modified_date changed to DateTimeField

loreleitrimberger:
## Description:
This updates transaction_search to use datetime for initial_report_date and last_modified_date. 

## Technical Details:
Both Delta and Postgres tables have been updated. Postgres uses timestamp with time zone as the type instead of DateTimeField. 

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [X] Jira Ticket(s)
    1. [DEV-14110](https://federal-spending-transparency.atlassian.net/browse/DEV-14110)

[DEV-14110]: https://federal-spending-transparency.atlassian.net/browse/DEV-14110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/download/delta_downloads/account_balances.py
+++ b/usaspending_api/download/delta_downloads/account_balances.py
@@ -1,4 +1,7 @@
-from pyspark.sql import functions as sf, Column, DataFrame, SparkSession
+from duckdb.experimental.spark.sql import SparkSession as DuckDBSparkSession
+from duckdb.experimental.spark.sql.column import Column as DuckDBSparkColumn
+from duckdb.experimental.spark.sql.dataframe import DataFrame as DuckDBSparkDataFrame
+from pyspark.sql import Column, DataFrame, SparkSession
 
 from usaspending_api.common.spark.utils import collect_concat
 from usaspending_api.download.delta_downloads.abstract_downloads.account_download import (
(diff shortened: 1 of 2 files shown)


usaspending_api/awards/v2/data_layer/orm.py
```diff
@@ -70,7 +69,7 @@ def construct_assistance_response(requested_award_dict: dict) -> OrderedDict:
         [
             ("start_date", award["_start_date"]),
             ("end_date", award["_end_date"]),
-            ("last_modified_date", get_date_from_datetime(transaction["_modified_at"])),
+            ("last_modified_date", transaction["_modified_at"].strftime("%Y-%m-%d %H:%M:%S%z")),
```
sethstoudenmier: Why is this changed to a datetime in the API response?

sethstoudenmier: To avoid potential conflict with the frontend I would leave this as is. The main goal of the change in this ticket is targeted at the downloads.

usaspending_api/search/delta_models/dataframes/transaction_search.py
```diff
@@ -92,9 +92,11 @@ def date_cols(self) -> list[Column]:
             sf.to_date(self.transaction_normalized.period_of_performance_current_end_date).alias(
                 "period_of_performance_current_end_date"
             ),
-            sf.coalesce(
-                sf.to_date(self.transaction_fabs.created_at),
-                sf.to_date(self.transaction_fpds.initial_report_date),
+            sf.to_timestamp(
+                sf.coalesce(
+                    sf.to_date(self.transaction_fabs.created_at),
```
sethstoudenmier: `created_at` is a TIMESTAMP so you should be able to remove the `sf.to_date()` and possibly even the `sf.to_timestamp()`.

usaspending_api/awards/v2/data_layer/orm.py
zachflanders-frb: This is a good point and makes me think we should instead cast the new datetime values to date in the responses for award search and transaction search in order to keep the api values consistent and unchanged.

sethstoudenmier (dismissed):

PR #4574, opened 2026-01-08
Fix/uv install order

zachflanders-frb:
## Description:

Moving the extra installs to the second uv sync command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ ENV UV_SYSTEM_PYTHON=1
 RUN --mount=type=cache,target=/root/.cache/uv \
     --mount=type=bind,source=uv.lock,target=uv.lock \
     --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -34,14 +34,14 @@ ENV UV_SYSTEM_PYTHON=1
 RUN --mount=type=cache,target=/root/.cache/uv \
     --mount=type=bind,source=uv.lock,target=uv.lock \
     --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
-    uv sync --extra server --extra ansible --extra awscli --extra spark --locked --no-install-project --no-dev
+    uv sync --locked --no-install-project --no-dev
```
zachflanders-frb: I am wondering if I should add the `--extra`s to both uv sync commands
```suggestion
    uv sync --extra server --extra ansible --extra awscli --extra spark --locked --no-install-project --no-dev
```

PR #4577, opened 2026-01-14
[DEV-14238] - Transaction Loader Refactor

zachflanders-frb:
## Description:
This PR refactors the transaction loader flow to remove the lookup tables.

## Technical Details:
This PR make the following changes to the transaction loader flow:
 - add a hash key to the published_fabs and detached_award_procurement delta tables
 - use the hash key and unique id column (`afa_generated_unique`/`detached_award_proc_unique`) for the merge with transaction fabs/fpds/normalized tables.
     - Using the unique keys allows us to eliminate the use of the transaction lookup tables 
     - Adding the hash allows us to compare the entire table and eliminate the use of the `last_etl_date`.  This ensures the entire tables are kept in sync.
 - Use `WHEN NOT MATCHED BY SOURCE` in the merge statement itself instead of having a separate delete function for transactions
 - Use the `unique_award_key` column to form the relationships between awards and transactions instead of using the award lookup table (eliminates need for the award lookup table)
 - splits the `load_transactions_in_delta` into separate commands
 - Reorganizes code using object-oriented patterns for better maintainability
 
## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)


usaspending_api/etl/management/commands/load_awards_in_delta.py
zachflanders-frb, line 1: I left this step largely unchanged except I used the `unique_award_key/generated_unique_award_id` and `unique_transaction_id` to form the relationships between the awards and normalized transactions.

usaspending_api/common/spark/configs.py
```diff
@@ -42,8 +42,8 @@
     # YARN or Mesos or Kubernetes) cluster manager, only client mode is supported.
     "spark.submit.deployMode": "client",
     # Default of 1g (1GiB) for Driver. Increase here if the Java process is crashing with memory errors
```
aguest-kc:
```suggestion
    # Default of 4g (4GiB) for Driver. Increase here if the Java process is crashing with memory errors
```

This comment should be updated if the memory amount was updated.

usaspending_api/etl/management/commands/load_awards_in_delta.py
```diff
@@ -0,0 +1,301 @@
...
+
+            # Capture earliest last load date of the source tables to update the "last_load_date" after completion
+            next_last_load = get_earliest_load_date(
+                ("source_procurement_transaction", "source_assistance_transaction"), datetime.utcfromtimestamp(0)
+            )
+
+            # Do this check now to avoid uncaught errors later when running queries
+            # Use 'int' because that is what will be targeted for deletes/updates/etc.
+            table_exists = self.spark._jsparkSession.catalog().tableExists(f"int.awards")
+            if not table_exists:
```
aguest-kc, line 58: We could avoid creating a single-use variable here.


```suggestion
            if not self.spark._jsparkSession.catalog().tableExists(f"int.awards"):
```

.github/workflows/test-spark-integration-load-transactions-fabs-fpds.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
         with:
           cov-report-name: 'spark-load-transactions-fabs-fpds-tests'
```
sethstoudenmier, line 63:
```suggestion
          cov-report-name: 'spark-load-transactions-tests'
```
I believe we actually can remove the generation of these reports, however, to be consistent we should change the name for now.

sethstoudenmier: Can we update the filename accordingly as well?

usaspending_api/etl/management/commands/load_awards_in_delta.py
```diff
@@ -0,0 +1,305 @@
...
+        parser.add_argument(
+            "--spark-s3-bucket",
+            type=str,
+            required=False,
+            default=CONFIG.SPARK_S3_BUCKET,
+            help="The destination bucket in S3 for creating the tables.",
+        )
+
+    def handle(self, *args, **options):
+        with self.prepare_spark():
```
sethstoudenmier, line 47: Should this use the defined function in `usaspending_api/etl/transaction_delta_loaders/context_managers.py` instead of the separately defined method below?

usaspending_api/etl/management/commands/load_awards_in_delta.py
```diff
@@ -0,0 +1,305 @@
...
+
+            logger.info(f"Running delete SQL for awards ETL")
+            self.spark.sql(self.delete_records_sql())
+
+            logger.info(f"Running UPSERT SQL for awards ETL")
+            self.update_awards()
+            update_last_load_date("awards", next_last_load)
+
+    @contextmanager
+    def prepare_spark(self):
```
sethstoudenmier, line 69: Similar to my comment above; could this be removed and instead use the function in `usaspending_api/etl/transaction_delta_loaders/context_managers.py`?

usaspending_api/etl/transaction_delta_loaders/loaders.py
```diff
@@ -0,0 +1,443 @@
...
+            target.merge(with_ids, f"t.unique_award_key = s.unique_award_key")
+            .whenMatchedUpdate(set={"t.award_id": "s.award_id"})
+            .execute()
+        )
+
+    def load_transactions(self) -> None:
+        super().load_transactions()
+        self.populate_award_ids()
+        self.populate_transaction_normalized_ids()
+        self.link_transactions_to_normalized()
```
zachflanders-frb, line 322: This involves 4 sequential executions of merge statements.  I need to explore whether I can adjust this to two executions and whether that will be more performant.

usaspending_api/etl/management/commands/load_awards_in_delta.py
```diff
@@ -0,0 +1,305 @@
...
+        if spark_created_by_command:
+            self.spark.stop()
+
+    def delete_records_sql(self):
+        id_col = "generated_unique_award_id"
+        # TODO could do an outer join here to find awards that do not join to transaction fpds or transaction fabs
+        subquery = """
+            SELECT awards.generated_unique_award_id AS id_to_remove
+            FROM int.awards
+            LEFT JOIN int.transaction_normalized on awards.transaction_unique_id = transaction_normalized.transaction_unique_id
```
sethstoudenmier, line 108: This probably isn't too bad performance wise, but it is possible that an EXISTS instead of a JOIN could help with performance some since we don't care about returning any of the values from `int.transaction_normalized` in this query.

I won't add this comment anywhere else, but the same could hold true. Being inside of a subquery may negate benefit of using EXISTS over a JOIN. If performance is looking good, then probably not worth exploring.

usaspending_api/etl/management/commands/load_awards_in_delta.py
```diff
@@ -0,0 +1,305 @@
...
+        )
+
+        yield  # Going to wait for the Django command to complete then stop the spark session if needed
+
+        if spark_created_by_command:
+            self.spark.stop()
+
+    def delete_records_sql(self):
+        id_col = "generated_unique_award_id"
+        # TODO could do an outer join here to find awards that do not join to transaction fpds or transaction fabs
```
sethstoudenmier, line 104: Is this comment saying that you could move away from Transaction Normalized to use FPDS and FABS directly? If so, do you mind clarifying in the comment.

usaspending_api/etl/management/commands/load_awards_in_delta.py
```diff
@@ -0,0 +1,305 @@
...
+                        tn.action_date AS date_signed,
+                        tn.description,
+                        tn.period_of_performance_start_date,
+                        ROW_NUMBER() OVER (
+                            PARTITION BY tn.unique_award_key
+                            /* NOTE: In Postgres, the default sorting order sorts NULLs as larger than all other values.
+                               However, in Spark, the default sorting order sorts NULLs as smaller than all other
+                               values.  In the Postgres transaction loader the default sorting behavior was used, so to
+                               be consistent with the behavior of the previous loader, we need to reverse the default
+                               Spark NULL sorting behavior for any field that can be NULL. */
```
sethstoudenmier, line 148:
```suggestion
                            -- NOTE:  In Spark, the default sorting order sorts NULLs as smaller than all other values.
```
Now that we are so far removed from the Postgres pipeline, I feel that this comment can be updated to include simply how Spark handles NULL values.

usaspending_api/etl/management/commands/load_awards_in_delta.py
```diff
@@ -0,0 +1,305 @@
...
+                    SUM(CAST(fpds.base_exercised_options_val AS NUMERIC(23, 2))) AS base_exercised_options_val,
+                    COUNT(tn.transaction_unique_id) AS transaction_count
+                FROM int.transaction_normalized AS tn
+                LEFT JOIN int.transaction_fpds AS fpds ON tn.transaction_unique_id = fpds.detached_award_proc_unique
+                GROUP BY tn.unique_award_key
+            )
+            SELECT
+                latest.id,
+                latest.unique_award_key,
+                0 AS subaward_count,  -- for consistency with Postgres table
```
sethstoudenmier, line 261:
```suggestion
                0 AS subaward_count,  -- default value that is updated later
```
Similar to other comment suggestion, since we are here and far removed from the Postgres pipeline it would be good to make this more meaningful.

sethstoudenmier (commented):
An initial pass on the changes. Overall, I didn't see anything that I would block over. Will take another pass before approving once testing is done.

zachflanders-frb:
This work and PR #4621 depend on an upgrade to the delta library that we are using.  This work is being done by https://federal-spending-transparency.atlassian.net/browse/OPS-2912.  I am closing in order to remove noise from our sprint with a note to revisit once that ticket is complete.

PR #4617, opened 2026-03-27
[DEV-11932] - Upgrade Psycopg

zachflanders-frb:
## Description:
This PR upgrades psycopg2 -> 3.

## Technical Details:
I removed the archived jobs app. These jobs are still accessible in our commit history, so it does not seem necessary to keep them as an django app that we have to maintain.

## Requirements for PR Merge:
    1. API Contracts
    2. API UI
    3. Comments
5. [x] Jira Ticket(s)
    1. [DEV-11932](https://federal-spending-transparency.atlassian.net/browse/DEV-11932)

[DEV-11932]: https://federal-spending-transparency.atlassian.net/browse/DEV-11932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pyproject.toml
```diff
@@ -50,7 +50,7 @@ dependencies = [
     "openpyxl==3.1.*",
     "pandas==2.2.*",
     "psutil==5.9.*",
-    "psycopg2==2.9.9", # Pinning exact version because this package will drop support for Python versions in patches
+    "psycopg[binary]>=3.3.3",
```
sethstoudenmier: Is there an option to not use `[binary]` that works? We have run into issues in the past with the deployed environments when using `psycopg2-binary`.

sethstoudenmier (changes requested):
Blocking only to make sure that this builds with Sandbox due to past issues when using `psycopg2-binary`.

sethstoudenmier (approved):

PR #4629, opened 2026-04-16
[DEV-15049] Optimize DuckDB download process

aguest-kc:
## Description:
Updates the download logic for DuckDB to only create the download tables that are required for that specific download job. For example, if a Custom Account download only includes File A then we only need to create the File A download table in memory.

## Requirements for PR Merge:

4. [x] Jira Ticket(s)
    1. [DEV-15049](https://federal-spending-transparency.atlassian.net/browse/DEV-15049)

[DEV-15049]: https://federal-spending-transparency.atlassian.net/browse/DEV-15049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -110,12 +123,21 @@ def process_request(self, download_job: DownloadJob, json_request: dict) -> None
             # Eagerly execute the download in this running process
             if self.is_spark_download(json_request):
                 spark_jobs = SparkJobs(LocalStrategy())
+                # Use DuckDBStrategy() for DuckDB downloads to avoid starting an unnecessary Spark session, if the
```
sethstoudenmier: Do we need the commented code to stay?

sethstoudenmier (changes requested):
Changes look good, just a comment around the commented code.

usaspending_api/download/v2/base_download_viewset.py
aguest-kc: I added the commented out code to prevent `LocalStrategy()` from starting a Spark session that wasn't being used when DuckDB is being used instead. I then commented it out because I wasn't sure if everyone had DuckDB setup locally to perform these downloads.

I can remove it, but if we want to mimic the deployed behavior eventually then we should use this logic.

sethstoudenmier (approved):

PR #4634, opened 2026-04-21
[DEV-14570] Implement the spending_by_award filters in Pydantic

aguest-kc:
## Description:
Use Pydantic to validate the `spending_by_award` filters instead of TinyShield.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [x] Jira Ticket(s)
    1. [DEV-14570](https://federal-spending-transparency.atlassian.net/browse/DEV-14570)

[DEV-14570]: https://federal-spending-transparency.atlassian.net/browse/DEV-14570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


anjenkin (dismissed):
I double checked and the FE isn't using 'program_activity' on the spending_by_award endpoint.

usaspending_api/common/pydantic_base_models/codes.py
zachflanders-frb, line 1: I wonder if it makes sense to consolidate these into a single `CodeObject` model?

usaspending_api/common/pydantic_base_models/locations.py
```diff
@@ -0,0 +1,83 @@
...
+    def validate_district(cls, v: str) -> str | None:
+        if v is not None and len(v) != 2:
+            raise ValueError("District code must be exactly 2 characters")
+        return v
+
+    @field_validator('zip')
+    @classmethod
+    def validate_zip(cls, v: str) -> str | None:
+        if v is not None and len(v) != 5:
+            raise ValueError("ZIP code must be exactly 5 digits")
```
zachflanders-frb, line 46: Do we want to check if zip codes are digits as well as length?  Could use `and not v.isdigit()`

usaspending_api/common/validator/request_validations.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    if code is None:
+        return None
+
+    valid_def_codes = list(DisasterEmergencyFundCode.objects.values_list('code', flat=True))
+
+    if isinstance(code, str):
+        code = code.upper()
+        if code not in valid_def_codes:
```
zachflanders-frb, line 24: Instead of getting all the def codes and serializing them to a list it might be faster to do
`DisasterEmergencyFundCode.filter(code=code).exists()` and `DisasterEmergencyFundCode.filter(code__in=code).exists()`

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -138,6 +151,46 @@
...
+    place_of_performance_scope: Literal["domestic", "foreign"] | None = None
+    program_activities: list[ProgramActivityObject] | None = None
+    program_numbers: list[str] | None = None
+    psc_codes: PSCCodeObject | list[str] | None = None
+    recipient_locations: list[StandardLocationObject] | None = None
+    recipient_scope: Literal["domestic", "foreign"] | None = None
+    recipient_search_text: list[str] | None = None
+    recipient_type_names: list[str] | None = None
+    set_aside_type_codes: list[str] | None = None
+    tas_codes: TASCodeObject | None = None
```
zachflanders-frb: If we consolidated the `CodeObject` models (see earlier) we could use that model for the naics codes, psc codes, and tas codes.

usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -138,6 +151,46 @@
...
+    time_period: list[TimePeriod] | None = None
+    treasury_account_components: list[TreasuryAccountComponentsObject] | None = None
+
+
+class SpendingByAwardRequest(BaseModel):
+    fields: list[str]
+    filters: Filters
+    limit: int = 10
+    order: Literal["asc", "desc"] = "desc"
+    page: int | None = None
```
zachflanders-frb: Does page have to be >= 1?  Might be worth a field validation.  I feel like I've made the mistake of asking for page 0 before.

zachflanders-frb (dismissed):
Approving with some comments to consider.

usaspending_api/common/validator/request_validations.py
aguest-kc: That would be more efficient. I did it this way to support the existing error message, which lists all of the possible DEF codes.

usaspending_api/common/pydantic_base_models/codes.py
aguest-kc: I was originally going to combine them, but I kept them separate in case we wanted to do any custom validation for a specific code type in the future.

usaspending_api/search/v2/views/spending_by_award.py
aguest-kc: Added a validation for this value to be greater than or equal to 1, but to still default to `None`.

zachflanders-frb (approved):

PR #4635, opened 2026-04-22
[DEV-14329] refactor table spec

james-at-kc:
## Description:

Changes table spec from a dictionary to a class to enable type checking.

## Technical Details:

* searched for all instances of table spec
* examined parameter data types
* created a class
* changed dictionary-style references ["fieldname"] to class-style references .fieldname

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments

N/A

N/A

5. [ ] Appropriate Operations ticket(s) created

N/A

6. [x] Jira Ticket(s)
    1. [DEV-14329](https://federal-spending-transparency.atlassian.net/browse/DEV-14329)

There is no change to the API, nor to the data validation; no impact to the frontend, no change to expected data, and no performance impact. No operations impact.

[DEV-14329]: https://federal-spending-transparency.atlassian.net/browse/DEV-14329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/common/data_classes.py
```diff
@@ -48,3 +49,36 @@ class TransactionColumn:
...
+
+
+@dataclass(slots=True)
+class TableSpec:
+    """
+    Delta table metadata class
+    """
+
+    destination_database: str
+    delta_table_create_sql: Union[str, Any]   # pyspark.sql.types.StructType
```
sethstoudenmier: > \# pyspark.sql.types.StructType

Why have this as a comment when you can set it as the type?

usaspending_api/common/data_classes.py
```diff
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
-from typing import Optional
+from typing import Any, Callable, List, Optional, Union
```
sethstoudenmier: Won't block over this, but worth calling out that in Python 3.9 they added `list` as a generic type to be used for typing; removing the need to import `List`.

Docs: https://peps.python.org/pep-0585/#implementation

sethstoudenmier: They also removed the need for `Union` and `Optional` (see https://peps.python.org/pep-0604/), but the idea here is that the new and old are both considered acceptable.

usaspending_api/etl/management/commands/load_table_from_delta.py
```diff
@@ -71,7 +70,7 @@ def add_arguments(self, parser: CommandParser) -> None:
             "--alt-delta-name",
             type=str,
             required=False,
-            help="An alternate delta table name to load, overriding the TABLE_SPEC destination_table" "name",
+            help="An alternate delta table name to load, overriding the TABLE_SPEC destination_tablename",
```
sethstoudenmier:
```suggestion
            help="An alternate delta table name to load, overriding the TABLE_SPEC destination_table_name",
```

usaspending_api/etl/tests/integration/test_load_transactions_in_delta_fabs_fpds.py
```diff
@@ -383,14 +389,16 @@ def happy_paths_test_core(
         assert equal_datasets(expected_transaction_id_lookup, delta_data, "")
 
         # Verify country code scalar transformation
-        query = f"SELECT DISTINCT legal_entity_country_code, place_of_perform_country_c FROM int.{self.etl_level} WHERE {self.pk_field} = 4 OR {self.pk_field} = 5"
+        query = f"SELECT DISTINCT legal_entity_country_code, place_of_perform_country_c FROM int.{self.etl_level} "
+        query = query + f"WHERE {self.pk_field} = 4 OR {self.pk_field} = 5"
```
sethstoudenmier: Not blocking over, but a style comment. Could utilize parenthesis to break this string across multiple lines. Or keep it as is and remove redundancy with `query += f"..."` instead of `query = query + f"..."`

sethstoudenmier (changes requested):
A couple of cleanup comments and then additional style related comments that I won't block over, but wanted to call out.

sethstoudenmier (dismissed):

usaspending_api/common/data_classes.py
```diff
@@ -48,3 +50,36 @@ class TransactionColumn:
...
+@dataclass(slots=True)
+class TableSpec:
+    """
+    Delta table metadata class
+    """
+
+    destination_database: str
+    delta_table_create_sql: Union[str, StructType]
+
+    destination_table_name: Optional[str] = None
```
zachflanders-frb: I think since python 3.10 the preferred syntax for union is `str | StructType` instead of `Union[str, StructType]` and optional is `str | None` instead of `Optional[str]`.  Both styles exist in our codebase, but it might be good to start using the newer and more concise version that seems to be a best practice: https://typing.python.org/en/latest/reference/best_practices.html#shorthand-syntax

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -103,190 +104,190 @@
...
+        destination_database="rpt",
+        swap_table="award_search",
+        swap_schema="rpt",
+        partition_column="award_id",
+        partition_column_type="numeric",
+        is_partition_column_unique=True,
+        delta_table_create_sql=award_search_create_sql_string,
+        delta_table_create_options=None,
+        source_schema=AWARD_SEARCH_POSTGRES_GOLD_COLUMNS,
+        custom_schema="recipient_hash STRING, federal_accounts STRING, cfdas ARRAY<STRING>,"
```
zachflanders-frb, line 136: One of the benefits of using a dataclass is that we can set defaults.  I think it would be great if we could only include properties that are different than the defaults and that way we can see what is unique about a particular table spec and also would reduce the amount of boilerplate in our code.  E.g. this one could just be 
```python
 "award_search_gold": TableSpec(
        model=AwardSearch,
        source_query=load_award_search,
        source_query_incremental=load_award_search_incremental,
        destination_database="rpt",
        swap_table="award_search",
        swap_schema="rpt",
        partition_column="award_id",
        partition_column_type="numeric",
        is_partition_column_unique=True,
        delta_table_create_sql=award_search_create_sql_string,
        source_schema=AWARD_SEARCH_POSTGRES_GOLD_COLUMNS,
        custom_schema="recipient_hash STRING, federal_accounts STRING, cfdas ARRAY<STRING>,"
        column_names=list(AWARD_SEARCH_POSTGRES_GOLD_COLUMNS),     
    ),
```

usaspending_api/etl/management/commands/load_query_to_delta.py
```diff
@@ -300,182 +301,182 @@
...
+        source_schema=None,
+        custom_schema=None,
+        column_names=list(),
+        postgres_seq_name=None,
+        tsvectors=None,
+        postgres_partition_spec=None,
+        delta_table_create_partitions=[
             "reporting_fiscal_year",
             "funding_toptier_agency_id",
         ],
```
zachflanders-frb, line 307: e.g. using the defaults this could be shortened quite a bit.
```suggestion
    "award_financial_download": TableSpec(
        source_query=load_award_financial,
        source_query_incremental=load_award_financial_incremental,
        destination_database="rpt",
        partition_column="financial_accounts_by_awards_id",
        partition_column_type="numeric",
        delta_table_create_sql=award_financial_schema,
        delta_table_create_options={"delta.enableChangeDataFeed": True},
        delta_table_create_partitions=["reporting_fiscal_year", "funding_toptier_agency_id"],
    ),
```

README.md
```diff
@@ -61,7 +61,7 @@ Create a `.envrc` file in the repo root, which will be ignored by git. Change cr
 ```shell
 export DATABASE_URL=postgres://usaspending:usaspender@localhost:5432/data_store_api
 export ES_HOSTNAME=http://localhost:9200
-export BROKER_DB=postgres://admin:root@localhost:5435/data_broker
```
zachflanders-frb, line 64: I don't think we want to change this because our local setup typically has the usaspending database at port 5432 and the broker database at port 5435 in so that they don't conflict.

zachflanders-frb (dismissed):
Approving with some non blocking style related comments about type hints and using the dataclass defaults to consider.

zachflanders-frb (approved):
👍

PR #4637, opened 2026-04-22
[DEV-14642] Fix sorting on Award Type and add sorting for loans

AHasan-FRB:
## Description:

Resolved issues with award type not searching correctly due to not using the keyword, rather the text. 
Added Awards to sorting.

## Technical Details:

- Changed award type to use keyword and not text via `type_description.keyword` in places award type is referred to
- Added Award Type to lookup table mappings

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)
    1. [DEV-14642](https://federal-spending-transparency.atlassian.net/browse/DEV-14642)

[DEV-14642]: https://federal-spending-transparency.atlassian.net/browse/DEV-14642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/search/v2/views/spending_by_award.py
```diff
@@ -473,9 +473,9 @@ def get_elastic_sort_by_fields(self) -> list[str]:  # noqa: PLR0912, C901
                     sort_by_fields = [
                         non_loan_assist_mapping[self.pagination["sort_key"]]
                     ]
-
+        if sort_by_fields[0] == "type_description":
+            sort_by_fields[0] = "type_description.keyword"
```
zachflanders-frb: I don't think we need to modify this here since it is updated in the `elasticsearch_lookups`.

usaspending_api/search/v2/views/spending_by_transaction.py
```diff
@@ -202,6 +202,8 @@ def post(self, request: Request) -> Response:  # noqa: PLR0912, PLR0915
                 ]
             case _:
                 sort_by_fields = [TransactionField(payload_sort_key).full_path]
+        if sort_by_fields[0] == "type_description":
+            sort_by_fields[0] = "type_description.keyword"
```
zachflanders-frb: Same here.

zachflanders-frb (changes requested):
I'm requesting changes to see if we can add a test case to sort by `Award Type`/`Contract Award Type`.  There is a test `test_spending_by_award_sort_recipient_location` that you can probably use as a model.

Left a comment about maybe not needing to check for `type_description` in the views, but feel free to ignore.

usaspending_api/search/tests/integration/test_spending_by_award.py
sethstoudenmier, line 1: Some of our tests have a bad habit of testing that an endpoint returns a 200, but don't actually validate the contents to ensure that the 200 isn't a false positive. I would update the tests you have added to validate that the content in the response is correct. You may only need to validate a single object in the response, but it should be enough to validate that the sort is working as expected.

usaspending_api/search/v2/views/spending_by_transaction.py
sethstoudenmier: Bumping Zach's comment here for some cleanup.

sethstoudenmier (dismissed):

usaspending_api/search/tests/integration/test_spending_by_award.py
```diff
@@ -4806,3 +4806,131 @@ def test_spending_by_award_sort_recipient_uei(
...
+
+    resp = client.post(
+        "/api/v2/search/spending_by_award/",
+        content_type="application/json",
+        data=json.dumps(test_payload),
+    )
+
+    assert resp.status_code == status.HTTP_200_OK
+    results = resp.json().get("results")
+    assert len(results) > 0
```
zachflanders-frb: Since this is testing the ordering functionality, can we test that the results are in the correct order instead of just testing that there are results?  We could include a list of expected results and ensure that the return results match or make a new list of objects from the results that are sorted as expected and compare to the original.

usaspending_api/search/tests/integration/test_spending_by_award.py
```diff
@@ -4806,3 +4806,173 @@ def test_spending_by_award_sort_recipient_uei(
...
+    results = resp.json().get("results")
+    assert results is not None
+    assert len(results) > 0, "Expected at least one contract award in test data"
+
+    # Verify all results have the requested fields
+    for result in results:
+        assert "Award ID" in result, "Award ID missing from result"
+        assert "Contract Award Type" in result, "Contract Award Type missing from result"
+        assert "internal_id" in result
+        assert "generated_internal_id" in result
```
zachflanders-frb, line 5031: For this we want to test that they are in the correct order (e.g. sorted by the "Award Type" field.  This field is not currently defined in the pytest fixtures, so they are all none right now.  This field comes from `AwardSearch.type_description`, so well first need to add them to the fixtures like this:
```
baker.make(
        "search.AwardSearch",
        award_id=5159,
        type="03",
        category="contract",
        date_signed="2019-01-01",
        action_date="2019-01-01",
        fain="fain5159",
        display_award_id="award5159",
        generated_unique_award_id="ASST_NON_TESTING_5159",
        recipient_name="ACME INC",
        cfda_number="12",
        cfda_program_title="program1",
        cfdas=['{"cfda_number": "12", "cfda_program_title": "program1"}'],
        type_description="aaa", # <- added this
    )
```

and then in the `asc` test something like this:

```python
expected = [ ... "ccc", "bbb", "aaa"]
actual = [result["Award Type"] for result in results]
assert actual == expected
``` 

For the `desc` test it will be sorted in the opposite order

sethstoudenmier (approved):

zachflanders-frb (approved):

PR #4645, opened 2026-04-30
[DEV-14703] Create NL Search Django Model

AHasan-FRB:
## Description:

Create the Django models and the corresponding migrations, ensuring the correct tables and relationships are created in Postgres, related to NL search assistant design and work.

## Technical Details:

- To load data for fixtures, you can run each separately or alternative run the script to do it automatically with the command `python manage.py load_llm_fixtures`
- New tests created to ensure loading new modelsworks successfully
- Models include __str__ attribute for easy admin debugging
- Default ordering is present for most model fields, for easier indexing and searching

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)
    1. [DEV-14703](https://federal-spending-transparency.atlassian.net/browse/DEV-14703)

[DEV-14703]: https://federal-spending-transparency.atlassian.net/browse/DEV-14703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/llm/fixtures/ai_models.yaml
zachflanders-frb, line 1: Can we add Nova Pro? (`amazon.nova-pro-v1:0`)

usaspending_api/llm/models/db_models.py
```diff
@@ -0,0 +1,107 @@
...
+    provider = models.CharField(max_length=100)
+
+    def __str__(self):
+        return f"{self.name} ({self.provider})"
+
+    class Meta:
+        db_table = "ai_model"
+        ordering = ["-id"]
+
+class Prompts(models.Model):
```
zachflanders-frb, line 17: I wonder about adding a name field to this model in order to have an easier way to get prompts than using the id or the full description?

usaspending_api/llm/models/db_models.py
```diff
@@ -0,0 +1,107 @@
...
+        db_table = "prompts"
+
+
+class Session(models.Model):
+    ai_model = models.ForeignKey(AIModel, on_delete=models.SET_NULL, null=True, related_name="sessions")
+    tools = models.JSONField(default=list)
+    system_prompt = models.ForeignKey(Prompts, on_delete=models.SET_NULL, null=True, related_name="sessions")
+    started_at = models.DateTimeField(auto_now_add=True)
+    ended_at = models.DateTimeField(null=True, blank=True)
+    feedback = models.BooleanField(default=None, null=True, blank=True, help_text="positive=True, negative=False")
```
zachflanders-frb, line 37: Based on the UX mocks it looks like we are going to have a short survey if someone gives feedback.  This makes me think that feedback could be its own model with an `is_positive` field and then potentially a survey  json field to represent survey questions and answers or even survey question and survey response models to fully model out the survey.  OTOH the survey might not use the api at all and use some other tool to collect feedback.

usaspending_api/llm/models/db_models.py
```diff
@@ -0,0 +1,107 @@
...
+
+class Message(models.Model):
+    session = models.ForeignKey(Session, on_delete=models.CASCADE, related_name="messages")
+    role = models.CharField(max_length=100)
+    message = models.TextField()
+    order = models.IntegerField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    input_tokens = models.IntegerField(default=0)
+    output_tokens = models.IntegerField(default=0)
+    total_tokens = models.IntegerField(default=0)
```
zachflanders-frb: In my initial testing `total_tokens` might not be that important to keep track of since input tokens and output tokens are priced differently and this is primarily to keep track of usage and cost.

usaspending_api/llm/models/db_models.py
```diff
@@ -0,0 +1,107 @@
...
+    def __str__(self):
+        preview = self.user_query[:75] + "..." if len(self.user_query) > 75 else self.user_query
+        return f"Query {self.id}: {preview}"
+
+
+    class Meta:
+        db_table = "llm_search_query"
+        indexes = [
+            models.Index(fields=["-created_at"]),
+        ]
```
zachflanders-frb: I found that this model is redundant because the first message in the session will be a user message that includes the user query, so I would recommend that we do not keep this model.

usaspending_api/llm/models/py_models.py
```diff
@@ -0,0 +1,270 @@
...
+    locationDomesticForeign: str = "all"
+    selectedFundingAgencies: dict[str, Any] = Field(default_factory=dict)
+    selectedAwardingAgencies: dict[str, SelectedAgency] = Field(default_factory=dict)
+    selectedRecipients: list[str] = Field(default_factory=list)
+    recipientDomesticForeign: str = "all"
+    recipientType: list[str] = Field(default_factory=list)
+    selectedRecipientLocations: dict[str, Any] = Field(default_factory=dict)
+    awardType: list[str] = Field(default_factory=list)
+    selectedAwardIDs: dict[str, Any] = Field(default_factory=dict)
+    awardAmounts: dict[str, list[int]] = Field(default_factory=dict)
```
zachflanders-frb: I added to the poc branch to give the llm more context and understanding of the awardAmounts field.

```suggestion
    awardAmounts: dict[str, list[int | None]] = Field(
        default_factory=dict,
        description=(
            "Dictionary of award amount ranges for filtering. "
            "Each value is a two-element list: [min_amount, max_amount]. "
            "Use `None` for unbounded ranges.\n\n"
            "TWO MUTUALLY EXCLUSIVE MODES:\n\n"
            "MODE 1 - STANDARD RANGES (can select multiple):\n"
            "- 'range-0': [None, 1000000] - Awards up to $1M\n"
            "- 'range-1': [1000000, 25000000] - Awards $1M to $25M\n"
            "- 'range-2': [25000000, 100000000] - Awards $25M to $100M\n"
            "- 'range-3': [100000000, 500000000] - Awards $100M to $500M\n"
            "- 'range-4': [500000000, None] - Awards over $500M\n\n"
            "MODE 2 - SPECIFIC RANGE (must be alone):\n"
            "- 'specific': [min, max] - Specify exact dollar amounts\n\n"
            "CRITICAL RULES:\n"
            "1. You can use multiple standard ranges together (range-0 through range-4)\n"
            "2. You can use ONE specific range with specific min/max values\n"
            "3. NEVER mix standard ranges with specific range\n"
            "4. When using 'specific', it must be the ONLY key in the dictionary"
        ),
        json_schema_extra={
            "examples": [
                # Example 1: Multiple standard ranges
                {"range-0": [None, 1000000], "range-2": [25000000, 100000000]},
                # Example 2: Single standard range
                {"range-3": [100000000, 500000000]},
                # Example 3: Custom range with both bounds
                {"specific": [5000000, 50000000]},
                # Example 4: Custom range unbounded above
                {"specific": [10000000, None]},
                # Example 5: Custom range unbounded below
                {"specific": [None, 75000000]},
            ]
        },
    )
```

usaspending_api/llm/models/db_models.py
zachflanders-frb, line 1: Can we add the migrations file to this PR?

zachflanders-frb (changes requested):
Looking good!  I am requesting changes to remove the LLMSearchQurey model and to add the migrations file to the commit.

usaspending_api/llm/models/db_models.py
AHasan-FRB: Will maintain it as is for now, if they are handling feedback through us can separate into its own model.

usaspending_api/llm/fixtures/prompts.yaml
```diff
@@ -0,0 +1,11 @@
+- model: llm.prompts
+  pk: 1
+  fields:
+    description: initial system prompt for the search assistant
+    text: >
```
zachflanders-frb, line 6: Can we give this a name like "initial" or similar to match the name field we added to the model?

usaspending_api/llm/tests/integration/test_db_models.py
```diff
@@ -0,0 +1,526 @@
+import pytest
+from django.db import IntegrityError
+from django.utils import timezone
+
+from usaspending_api.llm.models.db_models import AIModel, LLMSearchQuery, Message, Prompts, Session, ToolUse
```
zachflanders-frb: We'll need to remove LLMSearchQuery here too.

usaspending_api/llm/tests/integration/test_db_models.py
```diff
@@ -0,0 +1,526 @@
...
+            session=session
+        )
+        query2 = LLMSearchQuery.objects.create(
+            user_query="Second query",
+            session=session
+        )
+
+        assert session.search_queries.count() == 2
+        assert query1 in session.search_queries.all()
+        assert query2 in session.search_queries.all()
```
zachflanders-frb: These tests should be removed too.

usaspending_api/llm/tests/integration/test_db_models.py
```diff
@@ -0,0 +1,526 @@
...
+            ai_model=ai_model,
+            system_prompt=prompt,
+            tools=["lookup_location", "advanced_search"]
+        )
+
+        # Create search query
+        LLMSearchQuery.objects.create(
+            user_query="Find contracts in Texas",
+            session=session
+        )
```
zachflanders-frb: This can be removed.

usaspending_api/llm/tests/integration/test_db_models.py
```diff
@@ -0,0 +1,526 @@
...
+
+
+@pytest.mark.django_db
+class TestPrompts:
+    def test_create_prompt(self):
+        """Test creating a Prompts instance"""
+        prompt = Prompts.objects.create(
+            description="Test prompt",
+            text="You are a helpful assistant"
+        )
```
zachflanders-frb, line 53: You might need to add `name` here and elsewhere where you are creating a prompt because I don't think this field can accept nulls.

zachflanders-frb (commented):
Just a couple notes to align the fixtures and the tests with the changes to the models.

zachflanders-frb (approved):

PR #4646, opened 2026-05-01
[DEV-15084] - Assistance list - Add alpha Program Numbers

james-at-kc:
## Description:

Change /api/v2/autocomplete/cfda/ to allow non-numeric Program Numbers, 

## Technical Details:

Retain Program Number detection to return more relevant results: if search starts with ##. then search only on program number similar to existing logic.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
    3. Is performance impacted in the changes (e.g., API, pipeline, downloads, etc.)?
    4. Is the expected data returned with the expected format?
5. [x] Appropriate Operations ticket(s) created
6. [x] Jira Ticket(s)
    1. [DEV-15084](https://federal-spending-transparency.atlassian.net/issues?filter=-1&selectedIssue=DEV-15084)

NOTE: for item 3 above (data validation completed) 
a) the frontend requested the change, 
b) there should be no impact to the performance, and 
c) the same data will be returned.

The following strings will trigger a search for the program-id only
- 25, 25.G, 25.HI, 25.HIJ
The following strings will trigger a search on the program name and program popular description
- 25.HIJK, 25., 256, 256.J, 256.JK, 2567, 25HIJK
- other text strings as entered

[DEV-15051]: https://federal-spending-transparency.atlassian.net/browse/DEV-15051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DEV-15084]: https://federal-spending-transparency.atlassian.net/browse/DEV-15084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -30,7 +33,7 @@ def get_request_payload(request):
         try:
             limit = int(json_request.get("limit", 10))
         except ValueError:
-            raise InvalidParameterException("Limit request parameter is not a valid, positive integer")
+            raise InvalidParameterException("Limit request parameter is not a valid, positive integer")  # noqa: B904
```
sethstoudenmier: I would handle `B904` instead of commented it out. It is pretty straightforward and here is the documentation that goes into more detail: https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -300,14 +311,17 @@ class CFDAAutocompleteViewSet(BaseAutocompleteViewSet):
...
         queryset = Cfda.objects.all()
-
-        # Program numbers are 10.4839, 98.2718, etc...
-        if search_text.replace(".", "").isnumeric():
+        # allow 2 digits, or 2 digits with a dot followed by 1-3 digits or letters
+        # INCLUDE 25, 25.G, 25.HI, 25.HIJ
+        # EXCLUDE 25.HIJK, 25., 256, 256.J, 256.JK, 2567, 25HIJK
+        pattern = r'^\d{2}(\.[a-zA-Z0-9]{1,3})?$'
+        if bool(re.match(pattern, search_text)):
+            # Program numbers are 10.4839, 98.2718, 93.HDN, etc... 1-3 Alpha or numeric characters following dot
```
sethstoudenmier:
```suggestion
            # Program numbers are 10.483, 98.271, 93.HDN, etc... 1-3 Alpha or numeric characters following dot
```
Small update to the comment so that the examples reflect the correct format

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -300,14 +311,17 @@ class CFDAAutocompleteViewSet(BaseAutocompleteViewSet):
...
         """Return CFDA matches by number, title, or name"""
         search_text, limit = self.get_request_payload(request)
 
         queryset = Cfda.objects.all()
-
-        # Program numbers are 10.4839, 98.2718, etc...
-        if search_text.replace(".", "").isnumeric():
+        # allow 2 digits, or 2 digits with a dot followed by 1-3 digits or letters
+        # INCLUDE 25, 25.G, 25.HI, 25.HIJ
+        # EXCLUDE 25.HIJK, 25., 256, 256.J, 256.JK, 2567, 25HIJK
```
sethstoudenmier: I would update to also include a single number. Realistically, someone should hit the API with just a single number, but if we are going to allow it then filtering on the `program_number` makes more sense.

sethstoudenmier: I would also update to allow the `25.` (two numbers followed by a dot) case. Reason being that the frontend will commonly start sending back requests to the API after three characters. Meaning a user searching for `25` would need to input `25.` to start their search.

sethstoudenmier (changes requested):
Two things:
1. Update the PR to go into `qat` instead of `master`
2. Couple of comments around the regex

sethstoudenmier (dismissed):

usaspending_api/references/v2/views/autocomplete.py
zachflanders-frb, line 330: Would it make sense to include all the conditions in one queryset?  I am not sure why we are filtering on program number or program_title/popular_name.  Could we do something like:
```python
        queryset = queryset.filter(
            Q(program_number__icontains=search_text)
            | Q(program_title__icontains=search_text)
            | Q(popular_name__icontains=search_text)
        )
```

zachflanders-frb (dismissed):
Approving with a non-blocking comment to consider whether we should just include all filters for the autocomplete and not worry about the regex.

usaspending_api/references/v2/views/autocomplete.py
```diff
@@ -300,14 +311,17 @@ class CFDAAutocompleteViewSet(BaseAutocompleteViewSet):
...
 
         queryset = Cfda.objects.all()
-
-        # Program numbers are 10.4839, 98.2718, etc...
-        if search_text.replace(".", "").isnumeric():
+        # allow 2 digits, or 2 digits with a dot followed by 1-3 digits or letters
+        # INCLUDE 2, 25, 25., 25.G, 25.HI, 25.HIJ
+        # EXCLUDE 25.HIJK, 256, 256.J, 256.JK, 2567, 25HIJK
+        pattern = r'^\d{2}\.?[a-zA-Z0-9]{0,3}$'
+        if bool(re.match(pattern, search_text)):
```
aguest-kc: Is there a reason you added the `bool()` wrapper here? `re.match()` already returns `None` if there's no match which will satisfy the truthy check you're going for here.

aguest-kc (changes requested):
Let's good, but I had just one small question.

usaspending_api/references/v2/views/autocomplete.py
james-at-kc: Yes, we CAN do that. 

I specifically retained the existing logic due to ticket instructions: see https://federal-spending-transparency.atlassian.net/browse/DEV-15084 

> TECHNICAL NOTES: 
> - The current logic attempts to avoid cases where a user is presented with more values than anticipated if searching for a specific program number.
> - Should time box any time spent carrying forward the current logic to search on only the program number.

aguest-kc (changes requested):
You left a `print()` statement that can be removed. Github will not let me comment on the actual line, but it's line 54 in `usaspending_api/references/tests/integration/test_cfda_autocomplete_v2.py`

usaspending_api/references/v2/views/autocomplete.py
sethstoudenmier: I am not certain, but my assumption would be that the logic is intended to prevent results where the description contains a number from coming back when users are searching for the program number.

sethstoudenmier (approved):

aguest-kc (approved):

PR #4651, opened 2026-05-05
[DEV-15092] Implement decorator for LLM endpoints

AHasan-FRB:
## Description:

In the past we have used a pattern that validated a header in an API request before allowing access to some functionality. With the new LLM endpoints we will want to take this one step further than previous implementations. We will need to work with OPS to create a UUID that is stored in AWS and used to access all of the LLM endpoints that we will write.

## Technical Details:

- Added decorator to `api_request_utils.py`
- Uncomment environment variable `LLM_API_SECRET_NAME`

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)
    1. [DEV-15092](https://federal-spending-transparency.atlassian.net/browse/DEV-15092)

[DEV-15092]: https://federal-spending-transparency.atlassian.net/browse/DEV-15092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/common/api_request_utils.py
```diff
@@ -64,15 +74,21 @@ class FilterGenerator:
         "range_intersect": "range_intersect",
     }
 
-    def __init__(self, model, filter_map={}, ignored_parameters=[]):
-        self.filter_map = filter_map
+    def __init__(
+        self,
+        model: Any,
+        filter_map: Optional[Dict[str, str]] = None,
+        ignored_parameters: Optional[List[str]] = None,
```
zachflanders-frb, line 81: Nice job catching these mutable default parameters and updating them to `None`. 

One thing about the type hints: Python updated the syntax for type hints to support the built in types for things like `dict` and `list` and to support the use of the `|` for union.  It is now recommended to use `dict[str, str] | None` for optional. This comment applies to all new type hints in this changeset.  We have examples of both syntax styles in our code base, but it would be good if we could use the newer syntax as we're writing new code and updating for our style checks.  See this for more info: https://typing.python.org/en/latest/reference/best_practices.html#stylistic-practices

@sethstoudenmier @aguest-kc It might be good to implement these rules if we want to enforce this in our code base:
https://docs.astral.sh/ruff/rules/non-pep585-annotation/
https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/
https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/

usaspending_api/common/api_request_utils.py
```diff
@@ -408,11 +429,13 @@ def handle(data_set, body, serializer=None):
...
 
         filter_matched_ids, pk_name = AutoCompleteHandler.get_filter_matched_ids(
-            data_set.all(), body["fields"], body["value"], body.get("mode", "contains"), body.get("limit", 10)
+            data_set.all(), body["fields"], body["value"], body.get("mode", "contains"),
+            body.get("limit", 10)
         )
 
         # Get matching string values, and their counts
-        value_dict, count_dict = AutoCompleteHandler.get_values_and_counts(data_set.all(), filter_matched_ids, pk_name)
+        value_dict, count_dict = AutoCompleteHandler.get_values_and_counts(data_set.all(),
```
zachflanders-frb, line 475: We have an override for allowing 120 characters per line .  It looks like your code formatter is ignoring this override.  It might be worth checking to see if you can get your code formatter to recognize the 120 line character limit because  this adds noise to the PR and the next time someone makes a change to this file all the lines like this will be formatted back to a single line.

usaspending_api/common/api_request_utils.py
```diff
@@ -423,16 +484,143 @@ def handle(data_set, body, serializer=None):
...
+        Example:
+            @LLMAPIKeyHandler.require_api_key
+            @cache_response()
+            @api_view(['GET', 'POST'])
+            def llm_endpoint(request):
+                return Response({"message": "LLM endpoint"})
+        """
+
+        @wraps(function)
+        def wrapper(*args: Any, **kwargs: Any) -> Response:  # noqa: PLR0911, PLR0912
```
zachflanders-frb: Since this is new code I think we can try to meet PLR0911 and PLR0912 either by extracting some of the functionality to smaller functions or consolidating some of the branching.

zachflanders-frb (dismissed):
Functionality look really good 👍 .  I left some comments about code style and hope @sethstoudenmier and @aguest-kc can weigh in on the type hint code rules and whether we should add rules to stick to the newer type hint syntax described here: https://typing.python.org/en/latest/reference/best_practices.html#stylistic-practices

usaspending_api/common/api_request_utils.py
aguest-kc: I like all three of these rules.

sethstoudenmier: Definitely agree on the first two and I find myself going back and forth on the third since `Optional[...]` is sometimes more readable / obvious. With that said, I do like the idea of picking either `Optional[...]` or `... | None` for consistency. Given there is a ruff rule for it here then I vote for it as well.

usaspending_api/common/api_request_utils.py
sethstoudenmier: In response to: https://github.com/fedspendingtransparency/usaspending-api/pull/4651#discussion_r3202429142
>Since this is new code I think we can try to meet PLR0911 and PLR0912 either by extracting some of the functionality to smaller functions or consolidating some of the branching.

Agreed. Should ideally only use `# noqa` for complex checks outside of the code we are touching that just so happens to be in a shared file.

I would go as far as to block the PR to see if we can get this resolved. If there is some reason address these then I am happy to discuss.

sethstoudenmier (changes requested):
Will do a more thorough review but blocking in regard to the comment on the style checks for new code.

usaspending_api/common/api_request_utils.py
AHasan-FRB: I agree this should be addressed, I think these were auto applied, so I will get it adjusted with new changes to review.

usaspending_api/common/api_request_utils.py
```diff
@@ -423,16 +484,202 @@ def handle(data_set, body, serializer=None):
...
+        Returns tuple of (api_key, secret_name) on success or error Response on failure.
+        """
+        if not request:
+            return LLMAPIKeyHandler._error_response(
+                "Request object not found",
+                status.HTTP_500_INTERNAL_SERVER_ERROR
+            )
+
+        llm_api_key = request.headers.get('X-LLM-API-Key')
+        secret_name = os.environ.get('LLM_API_SECRET_NAME')
```
sethstoudenmier, line 584: Won't block on this right now since it wasn't called out for implementation, but it would be good to add this to the default.py config so that it can then be referenced here as `CONFIG.LLM_API_SECRET_NAME`

sethstoudenmier (approved):

PR #4678, opened 2026-06-18
[DEV-12749] Duplicated Rows in Download File

AHasan-FRB:
## Description:

Fixes that caused old/retried downloads to have lingering previous artifacts causing duplicated rows in the download file.

## Technical Details:

- Root Cause: When downloads failed mid-execution, orphaned DownloadJobLookup entries remained in the database. On retry, new entries were created alongside old ones, causing queries to return duplicate rows.
- Solution: Added cleanup_previous_download_attempt() function that runs at the start of each download to remove orphaned lookup entries, incomplete zip files, and temporary directories from failed previous attempts.
- Files Changed: Modified download_generation.py and generate_spark_download.py to call cleanup before processing begins.
- Impact: Ensures idempotent download retries - downloads can now be safely retried without accumulating artifacts or producing duplicate rows.
- Testing: Unit and integration tests verify cleanup removes orphaned data without affecting other jobs and retries to produce correct results.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)
    1. [DEV-12749](https://federal-spending-transparency.atlassian.net/browse/DEV-12749)

[DEV-12749]: https://federal-spending-transparency.atlassian.net/browse/DEV-12749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aguest-kc (dismissed):

usaspending_api/download/filestreaming/download_generation.py
```diff
@@ -68,6 +68,8 @@ def generate_download(download_job: DownloadJob, origination: Optional[str] = No
 
     file_name = start_download(download_job)
 
+    cleanup_previous_download_attempt(download_job)
```
zachflanders-frb: I think this needs to come before `start_download`, because `start_download` will create a new download job by calling `.save` and this comes along afterwards and deletes the newly created download job.

usaspending_api/download/management/commands/generate_spark_download.py
```diff
@@ -274,6 +277,45 @@ def get_download_request(self) -> DownloadRequest:
             type=download_type,
         )
 
+    def cleanup_previous_attempt(self) -> None:
```
zachflanders-frb: This method is duplicated from the `filestreaming/download_generation.py` file.  We should consolidated this code to prevent code duplication, e.g. in the helpers dir.

zachflanders-frb: Additional cleanup logic is in self.cleanup method of this class and in the finally clause of the try/except/finally statement of the `filestreaming/download_generation.py`.  It would be good to get all this code consolidated into a single cleanup method.

zachflanders-frb (changes requested):
Requesting changes to move the cleanup attempt before the start download method in filestreaming/download_generation.py and to consolidate the two duplicated `cleanup_previous_attempt/cleanup_previous_download_attempt` methods.

zachflanders-frb (approved):
I like how you broke down the long download functions into smaller pieces.  LGTM.

PR #4683, opened 2026-07-01
[DEV-15167] implement recipient lookup tool

james-at-kc:
## Description:
Search the recipients index in OpenSearch to find the most relevant results that can then be passed to the Advanced Search query. Cris may have insights from the filter semantic sessions for how to implement this filter.

## Technical Details:
The lookup_recipient tool should look for matches on the following fields: recipient_name, uei, duns

The lookup_recipient tool returns a list of recipient names.

The tool should query the recipient OpenSearch index a single time to find a relevant match

Test cases are created to capture different scenarios for the filter, including the usage of fuzzy matching

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [x] Jira Ticket(s)
    1. [DEV-15167](https://federal-spending-transparency.atlassian.net/browse/DEV-15167)

[DEV-15167]: https://federal-spending-transparency.atlassian.net/browse/DEV-15167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/llm/tools/lookup_recipient.py
```diff
@@ -0,0 +1,362 @@
+import logging
+import re
+from typing import Any, Optional
+
+from django.db.models import Q
+from elasticsearch_dsl import Q as ES_Q
```
zachflanders-frb: We should be using the `opensearchpy` library instead of `elasticsearch_dsl`
```suggestion
from opensearchpy.helpers.query import Q as ES_Q
```

usaspending_api/llm/tests/helper.py
zachflanders-frb, line 1: It seems that these helper functions are only being used in the tests.  Can we move this file into a helper.py file in ../tests/?

usaspending_api/llm/models/py_models.py
```diff
@@ -15,6 +15,39 @@ class AITool(BaseModel):
...
+    """Model for a selected entity"""
+
+    identifier: str = Field(
+        description=(
+            "Unique identifier for the recipient. "
+            "Format varies by type: UEI for entities, DUNS for legacy, or internal ID"
+        )
+    )
+    filter: RecipientFilter
+    display: RecipientDisplay
```
zachflanders-frb:
```suggestion
```
The selectedRecipient filter does not follow the same pattern as the location. It instead is just a simple list of stirngs.  This is already adequated modeled on the `Filters` model with `selectedRecipients: list[str] = Field(default_factory=list)` (line 389)

usaspending_api/llm/tools/lookup_recipient.py
```diff
@@ -0,0 +1,362 @@
...
+        if duns:
+            retval = f"DUNS_{duns}"
+        if recipient_hash and recipient_level and recipient_level != "subcontractor":
+            retval = f"{recipient_hash}-{recipient_level}"
+        if retval is not None:
+            normalized_name = re.sub(r"[^A-Z0-9]+", "_", (recipient_name or "").upper()).strip("_")
+            retval = f"NAME_{normalized_name or 'UNKNOWN'}"
+        return retval
+
+    def _build_display(
```
zachflanders-frb: Because this should just return a list of strings, we can remove this.

usaspending_api/llm/tools/lookup_recipient.py
```diff
@@ -0,0 +1,362 @@
...
+        should_queries_dict = [q.to_dict() for q in should_queries]
+
+        return (
+            RecipientSearch()
+            .query("bool", should=should_queries_dict, minimum_should_match=1)
+            .source(list(self.RECIPIENT_SOURCE_FIELDS))
+            .sort({"_score": {"order": "desc"}})[:top_k]
+        )
+
+    def _transform_results(self, response: Any) -> list[dict[str, dict[str, Any]]]:
```
zachflanders-frb: The search should return a list of strings (selectedRecipients expects a list of strings, unlike selectedLocations), so this can be simplified and the _build_identifier, _build_display, etc. methods can be removed.

usaspending_api/llm/tools/lookup_recipient.py
```diff
@@ -0,0 +1,362 @@
...
+        results = []
+        for subcontractor in self._get_subcontractors(uei=uei, duns=duns, recipient_name=recipient_name):
+            selected_recipient, score = self._transform_to_selected_recipient(
+                recipient_name=subcontractor["recipient_name"],
+                uei=subcontractor.get("uei"),
+                duns=subcontractor.get("duns"),
+                recipient_level="subcontractor",
+                recipient_hash=None,
+                score=None,
+            )
```
zachflanders-frb: I'm not sure about this design.  The match that best meets the user's intent is often not the top match, so I'm not sure it makes sense to use the top result to populate the subcontractors here.  As I use this I think it may make sense to have the subcontractors be a separate tool that can be called after this tool.  Or we could populate subcontractors for each result, but that could result in quite a few queries against a pretty big table.  My preference woudl be to not include the subcontractors lookup as part of this ticket and instead make it a followup item, probably after the core tools are implemented.

zachflanders-frb (changes requested):
I pulled this code down and tried it locally.  The results seem good!  I am requesting changes in order to 
 - use opensearchpy instead of elasticsearch_dsl library
 - move the functions in recipient_retrieval into a helpers file in the tests directory
 - Update the return value of the lookup tool to be a list of strings and remove the pydantic models from py_models (this one does not follow the pattern of the selectedLocations)
 
Additionally, it is my preference that we punt on the subcontractors lookup until later and potentially make it a separate tool, but would love others to weigh in on this too.

zachflanders-frb (approved):

PR #4686, opened 2026-07-06
[OPS-2987] - Fix Hierarchical Filter

zachflanders-frb:
## Description:
This fixes an issue wherea lack of max limits on code paths and recursive parent/child lookups can result in too many nodes being created.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments

### DIFF
--- a/usaspending_api/search/filters/shared/abstract_filter.py
+++ b/usaspending_api/search/filters/shared/abstract_filter.py
@@ -8,9 +8,9 @@ CodePath = list[str]
 class BaseNode(ABC):
 
     code: str
-    ancestors: list
+    ancestors: list[str]
     positive: bool
-    children: list
+    children: list["BaseNode"]
 
     def __init__(
         self, code: CodePath, positive: bool, positive_codes: list[CodePath], negative_codes: list[CodePath]
@@ -65,7 +65,6 @@ class BaseHierarchicalFilter(ABC):
 
     MAX_TOTAL_CODES = 100
     MAX_TREE_DEPTH = 10
-    MAX_CHAIN_LENGTH = 15
 
     @classmethod
     def _validate_complexity(cls, require: list[list[str]], exclude: list[list[str]]) -> None:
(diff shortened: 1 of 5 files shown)


usaspending_api/search/filters/shared/abstract_filter.py
```diff
@@ -0,0 +1,145 @@
...
+    """Shared logic for hierarchical code filtering (TAS, PSC, NAICS)"""
+
+    MAX_TOTAL_CODES = 100
+    MAX_TREE_DEPTH = 10
+
+    @classmethod
+    def _validate_complexity(cls, require: list[list[str]], exclude: list[list[str]]) -> None:
+        total_codes = len(require) + len(exclude)
+        if total_codes > cls.MAX_TOTAL_CODES:
+            raise ValueError(f"Total codes ({total_codes}) exceeds limit ({cls.MAX_TOTAL_CODES})")
```
aguest-kc: Can we handle this `ValueError` gracefully? Right now this will raise a `500 Internal Server Error` and the user won't know why. Returning a `422 Unprocessable Entity` with this error message would be better.

This could be handled in the code that calls this class method with a `try/except` block.
```python

try:
    cls._validate_complexity(require, exclude)
except ValueError as e:
    raise UnprocessableEntityException(str(e)) from e
```

usaspending_api/search/filters/shared/abstract_filter.py
```diff
@@ -0,0 +1,145 @@
...
+
+        all_codes = require + exclude
+        if not all_codes:
+            return
+
+        max_depth = max(len(code) for code in all_codes)
+        if max_depth > cls.MAX_TREE_DEPTH:
+            raise ValueError(f"Code depth ({max_depth}) exceeds limit ({cls.MAX_TREE_DEPTH})")
```
aguest-kc: Same as above. This will cause a `500 Internal Server Error` response to the user.

usaspending_api/search/filters/shared/abstract_filter.py
```diff
@@ -0,0 +1,145 @@
...
+class BaseHierarchicalFilter(ABC):
+    """Shared logic for hierarchical code filtering (TAS, PSC, NAICS)"""
+
+    MAX_TOTAL_CODES = 100
+    MAX_TREE_DEPTH = 10
+
+    @classmethod
+    def _validate_complexity(cls, require: list[list[str]], exclude: list[list[str]]) -> None:
+        total_codes = len(require) + len(exclude)
+        if total_codes > cls.MAX_TOTAL_CODES:
```
aguest-kc, line 124: `cls.MAX_TOTAL_CODES` should be 200 not 100. Since `require` and `exclude` can each have a max length of 100 this check can fail while still passing the TinyShield check.

aguest-kc (changes requested):
Overall it looks good, but I left a couple of questions on small tweaks.

aguest-kc (dismissed):

aguest-kc (approved):

PR #4689, opened 2026-07-09
DEV-14675 - LLM 'filter-search/' API endpoint

gregrholden:
## Description:
[DEV-14675] This is the initial implementation for the `/api/v2/llm/filter-search/` endpoint.
NOTES:
- This is not ready to be merged until the BE team has reviewed/tested it and agrees with the assumptions listed under `Technical Details`.
- Most of the heavy lifting for the LLM interaction is handled by another branch that is currently in PR (filter_search_assistant).
- This PR contains 2 other branches that are in PR and so needs to be handled in the appropriate order.

## Technical Details:
Branch is based off `dev-14679-implement-search-assistant` (currently in PR).
Branch `ftr/DEV-15167-implement-recipient-lookup-tool` was merged into this branch for additional tooling (that branch is also currently in PR).

Code Updates:
- `llm/v2/urls.py` was updated with the route for the endpoint.
- `llm/v2/views/filter_search.py` was created, defining the endpoint’s baseline behavior.
- `llm/v2/views/llm_base.py` was updated with high-level methods that can be shared by future LLM views.
- `llm/tests/integration/test_filter_search.py` was created to provide test coverage for the endpoint’s baseline behaviors.
- `llm/fixtures/prompts.yaml` had a typo fixed (line 7: “additonal” to “additional”).
- `llm/assistants/filter_search.py` was updated to provide more robust error handling, logging, and performance.

Assumptions:
- `LLMAPIKeyHandler.require_api_key` is complete and ready for use as a decorator on the post() method of the endpoint.
- Using the LLMBase class to provide shareable functions is an acceptable update to the codebase on this task.
- The default LLM model is “nova micro”, set on an `LLMBase` variable that can be overridden by an env variable.
- The user query should be validated against TinyShield, as was done in `agency_base.py` (line 76), before proceeding with prompting the LLM.
- CHECK: The return of a 422 error on TinyShield validation failure is acceptable, despite the API Contract specifying all errors should return a 403.
- The `filter_search_assistant.py` formulation is as complete as needed for this endpoint’s initial behavior with the LLM/tools.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)

[DEV-14675]: https://federal-spending-transparency.atlassian.net/browse/DEV-14675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/llm/v2/views/llm_base.py
```diff
@@ -1,10 +1,89 @@
...
+
+        Args:
+            message: Error message to return to the client.
+            search_id: Optional session/search ID. If None, generates a temporary UUID.
+
+        Returns:
+            StreamingHttpResponse with error event.
+        """
+        error_event = {
+            "search_id": search_id if search_id else str(uuid.uuid4()),
```
zachflanders-frb: If we don't have a search_id we should just return `None` instead of created a random uuid.

usaspending_api/llm/v2/views/filter_search.py
```diff
@@ -0,0 +1,93 @@
...
+                ai_model=ai_model,
+                tools=[tool.description.name for tool in tools],
+            )
+
+            # Add the above to the filter search assistant.
+            assistant = FilterSearchAssistant(
+                model=ai_model,
+                tools=tools,
+                session=session,
+            )
```
zachflanders-frb: I think we want to give the Assistant a system message as well.

usaspending_api/llm/v2/views/filter_search.py
```diff
@@ -0,0 +1,93 @@
...
+    The response is delivered as a series of JSON chunks, allowing real-time updates on search progress,
+    tool execution, and results.
+    """
+
+    endpoint_doc = "usaspending_api/api_contracts/contracts/v2/llm/filter_search.md"
+
+    # Define a list of allowed AI tools to pass to the assistant.
+    tools = [
+        lookup_location_tool,
+        lookup_recipient_tool,
```
zachflanders-frb, line 31: I'm realizing that we need to implement the `execute_filter` tool before this endpoint can actually work correcltly.

usaspending_api/llm/v2/views/filter_search.py
```diff
@@ -0,0 +1,93 @@
...
+
+            # Craft Response stream.
+            response = StreamingHttpResponse(
+                event_stream(),
+                content_type="application/x-ndjson"
+            )
+            # Disable webserver caching/buffering to enable pass-through behavior of chunks in the stream.
+            response["Cache-Control"] = "no-cache"
+            response["X-Accel-Buffering"] = "no"
+
```
zachflanders-frb, line 102: I think we want to capture the session end time for the `ended_at` property unless we want to remove this property.

zachflanders-frb (changes requested):
This is looking good, I especially like the improvements around getting the text messages from response content.  I have a few comments for consideration:

usaspending_api/llm/v2/views/filter_search.py
gregrholden: I added a finally block to capture the end time

usaspending_api/common/api_request_utils.py
zachflanders-frb: Can we update this example to show the class based view approach?

usaspending_api/common/api_request_utils.py
```diff
@@ -533,9 +533,11 @@ def llm_endpoint(request):
...
 
         @wraps(function)
-        def wrapper(*args: Any, **kwargs: Any) -> Response:
-            # Extract request from args
-            request = args[0] if args else kwargs.get('request')
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            # Extract request from kwargs
+            request = kwargs.get('request')
+            if not request and args:
+                request = next((arg for arg in args if hasattr(arg, 'headers')), None)
```
zachflanders-frb: This doesn't quite work either becuase it will grab the options request that has the headers attribute, but doesn't have the feature header that we are looking for.  I found that I had to use:
```suggestion
            required_header = 'X-LLM-API-Key'
            request = kwargs.get('request')

            if not request and args:
                request = next((
                    arg
                    for arg in args
                    if (
                        hasattr(arg, 'headers')
                        and required_header.lower() in [header.lower() for header in arg.headers]
                )), None)
```

zachflanders-frb (changes requested):

usaspending_api/common/api_request_utils.py
```diff
@@ -525,17 +525,23 @@ def require_api_key(function: Any) -> Any:
...
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            # Extract request from kwargs
+            request = kwargs.get('request')
+
+            if not request and args:
+                request = next((
+                    arg
+                    for arg in args
+                    if hasattr(arg, 'headers')
+                ), None)
```
zachflanders-frb: I found that this was selecting the `OPTIONS` request which had headers, but not the llm header we are loking for.  I think we'll need to update to something like 
```python
            required_header = 'X-LLM-API-Key'
            request = kwargs.get('request')

            if not request and args:
                request = next((
                    arg
                    for arg in args
                    if (
                        hasattr(arg, 'headers')
                        and required_header.lower() in [header.lower() for header in arg.headers]
                )), None)
```

zachflanders-frb (changes requested):
I have one small change on the decorator that I discovered while testing this locally.

usaspending_api/common/api_request_utils.py
gregrholden: Your code works when run with our search, but it's breaking the unit tests for the decorator because an empty header is returning as `None` instead of the 403 it expects from the llm validator logic. It's why I was trying to separate the request selection from the request validation. I'll keep working on this to see if I can get a way to keep your code and get it passing the unit tests.

gregrholden: I made a few tweaks, but this isn't the best solution either. It gets unit tests passing, but I still think a separation of purpose is needed. I'd like to try to separate this out so it 1: selects the correct Request object, 2. filters out OPTIONS requests, 3. validates the headers (which may be passable to the later validator method), and does these things in separate steps. Will continue working it for now.

zachflanders-frb (approved):

PR #4690, opened 2026-07-10
[DEV-15460] mitigate python stack trace returned to download user

james-at-kc:
## Description:

If an exception is encountered during the download generation process the /api/v2/download/status endpoint will return the Python stack trace in the API response’s message field. This can expose information about the API deployment environment.

## Technical Details:

Log these error message for future reference, but remove from the API response.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)
    1. [DEV-15460](https://federal-spending-transparency.atlassian.net/browse/DEV-15460)

N/A for number 2 - no change to API - generic message returned instead of stack trace
N/A for number 3 - no change to frontend, performance nor expected data
N/A for number 4 - No operations tickets required - straight merge to qat then default promotion to prod.

[DEV-15460]: https://federal-spending-transparency.atlassian.net/browse/DEV-15460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/download/v2/download_status.py
+++ b/usaspending_api/download/v2/download_status.py
@@ -1,5 +1,8 @@
+import logging
+
 from django.db import connections
 from rest_framework.exceptions import NotFound
+from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
@@ -7,6 +10,11 @@ from usaspending_api.common.exceptions import InvalidParameterException
 from usaspending_api.download.models import DownloadJob
 from usaspending_api.download.v2.base_download_viewset import get_file_path
 from usaspending_api.routers.replicas import ReadReplicaRouter
+
+logger = logging.getLogger(__name__)
+FAILED_DOWNLOAD_MESSAGE = (
+    "An error occurred while generating the download. Please try again"
+)
 
 
 class DownloadStatusViewSet(APIView):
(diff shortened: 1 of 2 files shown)


usaspending_api/download/v2/download_status.py
```diff
@@ -52,6 +60,18 @@ def get_download_status_response(self, file_name: str):
...
+        if download_job.job_status.name != "failed":
+            return None
+
+        if download_job.error_message:
+            logger.error(
+                "Download job %s failed: %s",
+                download_job.file_name,
+                download_job.error_message,
+            )
+        return FAILED_DOWNLOAD_MESSAGE
```
zachflanders-frb: Could there be a scenario where we have an error message, but the download job doesn't fail?  In the original code if returns the  `download_job.error_messsage` regardless of the `download_job.job_status`.  We could do something similar to the original code where we display a generic message if there is an error and log the whole message?
```suggestion
        if download_job.error_message:
            logger.error(f"Download job {download_job.file_name} failed: {download_job.error_message}")
            return "An error occurred."
        return None
```

zachflanders-frb (dismissed):
Looks good, I left one non blocking comment for consideration.

zachflanders-frb (approved):

PR #4692, opened 2026-07-14
[DEV-15496] date range when not new awards only filter

james-at-kc:
## Description:

Fix award-level new_awards_only so it narrows results relative to the default time period filter instead of replacing it. QA showed /api/v2/search/spending_by_award_count/ returning HIGHER values WITH new_awards_only than without it for the same FY time window.

## Technical Details:

"New Awards only" is intended to be an additional constraint on the selected time period. Tab counts with NAO checked must be less or equal to counts for the same window without the NAO filter.

Default Filter:
- action_date >= start
- date_signed <= end

Intended NAO semantics:
- default (union) date_signed >= end

Discovery:
QA compared two spending_by_award requests for FY 2026 (2025-10-01 to 2026-09-30)
- defailt time period (no date_type)
- same window with "date__type": "new_awards_only"

Several categories increased counts under NAO.

An early theory blamed the asymmetric default (gte=action_date, lte=date_signed) and proposed changing both bounds to action_date. That would break the intended overlap and would NOT fix the subset behavior. An examination of the NewAwardsOnlyTimePeriod showed the real defect:

- Default mode: action_date >= start and date_signed <= end
- Award NAO (BEFORE): date_signed >= start and date_signed <= end
- Transaction NAO: keeps action_date and signed_date on both sides

Award NAO dropped the action_date >= start so awards with date_sogned in-range but action_date before start were counted under NAO and excluded from the default filter.

Mitigation:
In NewAwardsOnlyTimePeriod for QueryType.AWARDS, added action_date >= start to the get filters

Award NAO is now
- date_signed >= start
- date_signed <= end
- action_date >= start

TESTS:
- Unit test test_time_period.py added a check that both date_signed and action_date are both included for Award NAO filter
- Integration test test_spending_by_award_count_new_awards_only_is_subset_of_default() include test data with one award with date_signed in window but action_date before start (which would be excluded form both) and one valid award (included in both). Asserts that NAO contract count <= default count.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [x] Jira Ticket(s)
    1. [DEV-15496](https://federal-spending-transparency.atlassian.net/browse/DEV-15496)

Item 2 above is N/A because no API change was implemented. Same documentation applies.
Item 3 above is N/A because no data change was made. No frontend change is needed to accommodate any new data.
Item 4 above is N/A because no ops involvement is needed to promote this change.

[DEV-15496]: https://federal-spending-transparency.atlassian.net/browse/DEV-15496


james-at-kc:
Here are the tests I ran against the QAT and PROD endpoints with and without the new_awards_only filter.
[2026-07-14-new-awareds-only-filter-tests.txt](https://github.com/user-attachments/files/30093784/2026-07-14-new-awareds-only-filter-tests.txt)

zachflanders-frb (approved):

PR #4693, opened 2026-07-15
[DEV-15427] Fixed Incorrect Category Field Modification

DavidMikolaKC:
## Description:

Category is a mutable dataclass (spending_by_category.py:34) instantiated as a class attribute on each concrete view (spending_by_locations.py:189). Line 128 mutates self.category.agg_key in place, and since no per-request self.category = ... exists anywhere except DEFCViewSet's replace() at spending_by_industry_codes.py:143, the mutation hits the class-level singleton shared by all requests on the worker. The mapper is one-way (sub_* keys map to themselves via the default), so one unauthenticated spending_level=subawards POST to /api/v2/search/spending_by_category/county (urls_spending_by_category.py:32) permanently swaps the agg_key, after which transaction/award branches at lines 133-141/205/265 aggregate on the wrong ES field and @cache_response() can persist the bad output.

## Technical Details:

Replaced instances of self.category.* fields in the AbstractSpendingByCategoryViewSet to instead create a new instance of the dataclass with the updated fields.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)
    1. [DEV-15427](https://federal-spending-transparency.atlassian.net/browse/DEV-15427)

[DEV-15427]: https://federal-spending-transparency.atlassian.net/browse/DEV-15427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
```diff
@@ -68,6 +68,10 @@ class AbstractSpendingByCategoryViewSet(APIView, metaclass=ABCMeta):
     def post(self, request: Request, *args, **kwargs) -> Response:
         original_filters = request.data.get("filters")
 
+        # Creates an instance copy for the category for each request
+        # Prevents modifications from previous requests affecting current requests
+        self.category = copy.deepcopy(self.category)
```
zachflanders-frb: I think we need two properties `self._cateogry` as the immutable class template and `self.category` as the class property that is used. Otherwise we run the risk of mutating `self.category` in the `replace` function below and then the next time we copy the mutated `self.category`.  Instead we should have a `self._cateogry` that we copy from and never mutate so that it remains the original `Category` each time this code is accessed.

```suggestion
        self.category = copy.deepcopy(self._category)
```

usaspending_api/search/v2/views/spending_by_category_views/spending_by_category.py
zachflanders-frb:
```suggestion
    _category: Category
```
Prefix this with an underscore to show that it is a class template and should not be mutated.  We then copy from this unchanged property each time a post is made.  We will need the update all the classes that inherit from this to ensure that they set the `self._category` property and not the `self.category`

zachflanders-frb (changes requested):
Requesting changes to add a `_category` property to the abstract class.  We will need to update all the classes that inherit from this class so ensure that they set the `_category` property instead of the `category` property.

zachflanders-frb (approved):

PR #4699, opened 2026-07-21
[DEV-15421] Update default values for dictionaries

aguest-kc:
## Description:
Update the dictionary logic when setting a default value.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-15421](https://federal-spending-transparency.atlassian.net/browse/DEV-15421)

[DEV-15421]: https://federal-spending-transparency.atlassian.net/browse/DEV-15421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/common/validator/helpers.py
```diff
@@ -151,9 +151,9 @@ def validate_float(rule):
     return rule["value"]
 
 
-def validate_integer(rule):
-    rule["min"] = rule.get("min") or MIN_INT
-    rule["max"] = rule.get("max") or MAX_INT
+def validate_integer(rule: dict) -> int:
+    rule["min"] = rule["min"] if rule.get("min") else MIN_INT
+    rule["max"] = rule["max"] if rule.get("max") else MAX_INT
```
zachflanders-frb, line 156: Can we just use a default for the `.get` methods? e.g.
```
>>> rule ={"min": 0}
>>> rule
{'min': 0}
>>> MIN_INT = -999
>>> rule.get("min") or MIN_INT
-999
>>> rule.get("min", MIN_INT)
0
```

zachflanders-frb (changes requested):
I think supplying a default for the `.get` method is the more straightforward solution here unless I am missing something.

usaspending_api/common/validator/helpers.py
aguest-kc: The `min` and `max` keys are set to a default value of `None`, so if we use `.get("min", MIN_INT)` the default `None` value will always be returned. That's why I did it this, admittedly clunky, way.

This is definitely excessive for this logic, but it should address the code defect that was identified.

aguest-kc: I wanted to change as little code as required for this patch, so I did not make all of the required changes to make `.get()` work in this file.

zachflanders-frb (approved):

PR #4703, opened 2026-07-23
DEV-15456: Added validation methods for group and filter fields to prevent regex injection

gregrholden:
## Description:
**DEV-15456: Sanitize Postgres input [f112]**
Fixing a vulnerability that allowed unauthenticated attackers to inject arbitrary Django ORM lookup suffixes (particularly __regex/__iregex) into aggregate group parameters and filter field paths, enabling regex DoS attacks and schema enumeration.
AC: Implement model-based validation that validates field paths against Django model schemas.

## Technical Details:
Changes:
- Aggregate Group Validation (`common/mixins.py`):
    - Added `_resolve_field_path()` method to validate group parameter fields segment-by-segment.
    - Validates fields against model schema using Django's `_meta.get_field()` API.
    - Supports legitimate FK traversal (e.g., `treasury_account__federal_account_id`).
    - Allows safe transform suffixes (e.g., `__fy`) defined in `_sql_function_transformations` (line 78).
    - Raises `InvalidParameterException` for invalid field paths.
    - NOTE: Also needed to update return value definitions across the file, and reorder imports, to pass lint tests.
- Filter Field Validation (`common/api_request_utils.py`):
    - Added `_validate_filter_field()` and `_resolve_field_path()` methods to `FilterGenerator`.
    - Validates POST filter field parameters before creating Q objects.
    - Validates GET query parameter keys before use in filters.
    - Added `_get_allowed_lookups()` that derives safe Django lookups from operators dictionary.
    - Explicitly blocks `__regex` and `__iregex` lookups to prevent ReDoS attacks.
    - Supports compound lookups (e.g., `__len__gt` for ArrayField length checks).
    - Handles special cases: field arrays for range_intersect and search operations.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)


usaspending_api/common/api_request_utils.py
```diff
@@ -260,10 +268,91 @@ def _validate_single_filter(self, filt: Dict[str, Any]) -> None:
...
+            if resolved_field.is_relation and not is_last:
+                current_model = resolved_field.related_model
+            elif not is_last:
+                # Non-relation field with more segments. Check if remaining segments are all valid lookups.
+                remaining_segments = segments[index + 1:]
+                allowed_lookups = self._get_allowed_lookups()
+                if all(seg in allowed_lookups for seg in remaining_segments):
+                    # All remaining segments are valid lookups, validation complete.
+                    return
+                raise InvalidParameterException("Invalid field: {}".format(field_path))
```
sethstoudenmier: Small thing not worth blocking over, but I would try to use f-strings where you can. We have slowly been trying to use them more and more but do still have many cases where they still appear. An exception is when you are spreading a dictionary to fill in multiple values.

sethstoudenmier (dismissed):
Overall LGTM! Ideally, we will move away from the v1 endpoints and replace with v2, but this is a great middle ground for now.

usaspending_api/common/api_request_utils.py
gregrholden: Thanks for the feedback. I've updated the output to use f-strings.

usaspending_api/common/api_request_utils.py
```diff
@@ -260,10 +268,91 @@ def _validate_single_filter(self, filt: Dict[str, Any]) -> None:
...
+            if resolved_field.is_relation and not is_last:
+                current_model = resolved_field.related_model
+            elif not is_last:
+                # Non-relation field with more segments. Check if remaining segments are all valid lookups.
+                remaining_segments = segments[index + 1:]
+                allowed_lookups = self._get_allowed_lookups()
+                if all(seg in allowed_lookups for seg in remaining_segments):
+                    # All remaining segments are valid lookups, validation complete.
+                    return
+                raise InvalidParameterException(f"Invalid field: {field_path}")
```
james-at-kc, line 319: Now uses f"" strings throughout per suggestion.

james-at-kc (approved):
LGTM. improved code by using f-strings as suggested in previous approval.

PR #4707, opened 2026-07-27
DEV-15453: Updated ES sanitizing functions to handle sanitization byp…

gregrholden:
## Description:
DEV-15453: Sanitize OpenSearch Input
- Vulnerability scan finding f022 stated that the es_minimal_sanitize length-comparison logic returns unsanitized input when the removals equal escapes.
- The fix was to ensure the sanitizing function returns the sanitized input regardless of length comparison and then apply that sanitization to the keyword search class.
- Unit tests were added to cover the reported vulnerability parameters/vectors. 

## Technical Details:

1. In `search/v2/es_sanitization.py` (lines 38-54): Changed function to always return the `processed_string` instead of the `keyword` variable that got assigned the `processed_string` value only if the lengths were not equal, which eliminates the length-equality bypass vulnerability and ensures the return value is always a sanitized version of the input.
2. In `common/query_with_filters.py` (lines 197-202): Applied the above sanitization to the `_KeywordSearch` class before passing values to the `ES_Q` query string.
3. Added more test cases to `search/tests/unit/test_elasticsearch_helpers.py` to validate changes made.
4. Fixed missing return types and import ordering to pass lint tests.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    6. Comments
8. [ ] Jira Ticket(s)


james-at-kc (approved):
LGTM. instead of returning original string. now returns sanitized string regardless if len is same or not. Ready for merge.

PR #4708, opened 2026-07-27
DEV-15466: Replaced cURL command with PUT Request in es_configure.py

gregrholden:
## Description:
In vulnerability finding f222, it was reported that ES_HOSTNAME (with embedded basic-auth) passed through the shell curl command line. DEV-15466 notes:
- ES_USER and ES_PASSWORD values can potentially be added to the cURL request made in es_configure.py, but that it only happens when someone is using environment variables to interact with the API and so would not happen in production.
- Initial solution was to remove the ability to use env vars for passing this user/pass combo to secure it.
- Another proposed solution was to remove the cURL command entirely in favor of a PUT Request.
    - This option was implemented.

## Technical Details:
Solution implemented:
1. Removed subprocess dependency in exchange for the use of the Requests library.
2. Added secure HTTP handling with `urllib.parse`.
3. Refactored `run_curl_cmd` method to handle the PUT request using the Request library instead of a cURL command.
4. Small code refactoring to fix lint findings (e.g., missing data types and return types).

**NOTE**: There is an assumption made here that this will continue working in production with this change, but this needs to be tested in an environment (not a local) to be certain.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    6. Comments
8. [ ] Jira Ticket(s)


james-at-kc (approved):
LGTM. cURL construction was fully replaced by PUT request. Ready for merge.

PR #4710, opened 2026-07-28
Update db router to ensure llm models read and write from same db

zachflanders-frb:
Update django db router to ensure that llm models always read and write from the same db and not replicas.


james-at-kc (approved):
Looks good to me! OK for merging.

PR #4713, opened 2026-07-30
[DEV-15635] Update download regex and add test case

sethstoudenmier:
## Description:

Fix some of the download columns swapping their values due to similar names between the generated CSVs and the database columns.

## Technical Details:

Django changed how the queryset queries are represented which broke the regex that is used to determine which columns are a direct select vs annotations. This PR updates that regex, removes some checks that were previously added, and then adds a test case to hopefully catch any future potential changes.

A decent amount of the changes came from running ruff format on the file.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [x] Jira Ticket(s)
    1. [DEV-15635](https://federal-spending-transparency.atlassian.net/browse/DEV-15635)

[DEV-15635]: https://federal-spending-transparency.atlassian.net/browse/DEV-15635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/download/tests/integration/test_download_generation.py
+++ b/usaspending_api/download/tests/integration/test_download_generation.py
@@ -88,7 +88,7 @@ def test_apply_annotations_to_sql_joined_table():
         'LEFT OUTER JOIN "transaction_search" ON '
         '("award_search"."earliest_transaction_search_id" = "transaction_search"."transaction_id") '
         f'LEFT OUTER JOIN "transaction_search" {table_alias} '
-        'ON ("award_search"."latest_transaction_search_id" = T4."transaction_id") '
+        f'ON ("award_search"."latest_transaction_search_id" = {table_alias}."transaction_id") '
         "ORDER BY 1 ASC LIMIT 1"
     )
     assert result_query == expected_query
--- a/usaspending_api/download/tests/integration/test_download_generation.py
+++ b/usaspending_api/download/tests/integration/test_download_generation.py
@@ -1,6 +1,5 @@
 import pytest
 from django.db.models import F, Value
-from model_bakery import baker
 
 from usaspending_api.common.helpers.orm_helpers import ConcatAll, generate_raw_quoted_query
 from usaspending_api.download.filestreaming.download_generation import apply_annotations_to_sql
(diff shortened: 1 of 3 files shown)


usaspending_api/download/tests/integration/test_download_generation.py
```diff
@@ -0,0 +1,94 @@
...
+        "CONCAT("
+        f'{table_alias}."period_of_performance_start_date", '
+        "' - ', "
+        f'{table_alias}."period_of_performance_current_end_date"'
+        ') AS "date_range" '
+        'FROM "award_search" '
+        'LEFT OUTER JOIN "transaction_search" ON '
+        '("award_search"."earliest_transaction_search_id" = "transaction_search"."transaction_id") '
+        f'LEFT OUTER JOIN "transaction_search" {table_alias} '
+        'ON ("award_search"."latest_transaction_search_id" = T4."transaction_id") '
```
james-at-kc: I think there is a bug in this SQL statement: the T4. appears to be a hardcoded table alias. Based on the rest of the clause, it appears that table alias should be the variable {table_alias} (here set to the value foobar).

```
SELECT 
    "award_search"."award_id" AS "award_id",
    "transaction_search"."solicitation_date" AS "earliest_transaction_search__solicitation_date",
    foobar."action_date" AS "latest_transaction_search__action_date",
    "award_search"."generated_unique_award_id" AS "assistance_award_unique_key",
    "award_search"."fain" AS "award_id_fain",
    CONCAT(
        foobar."period_of_performance_start_date",
        ' - ',
        foobar."period_of_performance_current_end_date"
    ) AS "date_range"
FROM 
    "award_search"
LEFT OUTER JOIN "transaction_search" 
    ON ("award_search"."earliest_transaction_search_id" = "transaction_search"."transaction_id")
LEFT OUTER JOIN "transaction_search" foobar 
    ON ("award_search"."latest_transaction_search_id" = T4."transaction_id") <<<<<<<<<<<<<<<<<<<<<
ORDER BY 1 ASC
LIMIT 1
```

sethstoudenmier: Good catch! That is a typo and it should have been the table alias that is fetched via a regex for comparison

james-at-kc (approved):
LGTM

PR #4717, opened 2026-07-31
[DEV-15461] - Mark failed downloads

zachflanders-frb:
## Description:
This PR fixes a logic error that attempted to cast a json string as a int.  I added a pydantic to better document the structure of the download job message and fixed the logic error to correctly get the download job id.

5. [x] Jira Ticket(s)
    1. [DEV-15461](https://federal-spending-transparency.atlassian.net/browse/DEV-15461)

[DEV-15461]: https://federal-spending-transparency.atlassian.net/browse/DEV-15461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/download/v2/base_download_viewset.py
+++ b/usaspending_api/download/v2/base_download_viewset.py
@@ -4,6 +4,7 @@ from typing import List, Optional, Type
 
 from django.conf import settings
 from django.db.models import Max
+from pydantic import BaseModel, Field
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.views import APIView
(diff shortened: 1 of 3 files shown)


usaspending_api/download/v2/base_download_viewset.py
```diff
@@ -33,6 +34,16 @@
 from usaspending_api.submissions.models import DABSSubmissionWindowSchedule
 
 
+class DownloadJobMessage(BaseModel):
+    download_job_id: int = Field(..., gt=0)
```
sethstoudenmier: Any reason to leave the ellipses?

sethstoudenmier (dismissed):
Approving with one small comment for cleanup

usaspending_api/download/v2/base_download_viewset.py
zachflanders-frb: No, I saw this in a lot of examples, but pydantic actually discourages this in the docs: https://pydantic.dev/docs/validation/latest/concepts/fields/ 
> Note

> Even though name is assigned a value, it is still required and has no default value. If you want to emphasize on the fact that a value must be provided, you can use the [ellipsis](https://docs.python.org/3/library/constants.html#Ellipsis):
```
class Model(BaseModel):
    name: str = Field(..., frozen=True)
```
> However, its usage is discouraged as it doesn’t play well with static type checkers.

sethstoudenmier (approved):

PR #4718, opened 2026-08-03
[DEV-15536] Set explicit ALLOWED_HOSTS value

DavidMikolaKC:
## Description:

The API currently has ALLOWED_HOSTS set to an environment variable or * (if the variable is not present). We should limit this to a list of our domain names.

## Technical Details:

Defined a list of allowed_hosts in settings.py.  Additional or missing hosts may be added as needed.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)
    1. [DEV-15536](https://federal-spending-transparency.atlassian.net/browse/DEV-15536)

[DEV-15536]: https://federal-spending-transparency.atlassian.net/browse/DEV-15536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/settings.py
+++ b/usaspending_api/settings.py
@@ -57,7 +57,7 @@ SECRET_KEY = get_random_secret_key()
 DEBUG = os.environ.get("DJANGO_DEBUG", "").lower() in ["true", "1", "yes"]
 
 HOST = os.environ.get("HOST", "localhost:3000")
-ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "[]").split(",")
+ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "").split(",")
 
 # Define local flag to affect location of downloads
 IS_LOCAL = os.environ.get("IS_LOCAL", "").lower() in ["true", "1", "yes"]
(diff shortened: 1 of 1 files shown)


DavidMikolaKC:
I'm unsure if the list of approved hosts I gave is fully conclusive.  If there are additional hosts that may be missing please note them here

usaspending_api/settings.py
```diff
@@ -57,7 +57,7 @@
 DEBUG = os.environ.get("DJANGO_DEBUG", "").lower() in ["true", "1", "yes"]
 
 HOST = os.environ.get("HOST", "localhost:3000")
-ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "*").split(",")
+ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", [".usaspending.gov", ".localhost", "127.0.0.1", "[::1]"])
```
zachflanders-frb: I would suggest that we default this to `None` since it will use localhost when it is `None` and we should be setting debug to True in local dev. 

> When [DEBUG](https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-DEBUG) is True and ALLOWED_HOSTS is empty, the host is validated against ['.localhost', '127.0.0.1', '[::1]'].

https://docs.djangoproject.com/en/6.0/ref/settings/#allowed-hosts

```suggestion
ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS")
```

zachflanders-frb (changes requested):
Meant to select request changes for the above comment instead of "comment"

zachflanders-frb (approved):

PR #4737, opened 2026-08-14
[DEV-15929]: LLM inference configs

gregrholden:
## Description:
Add inference controls to the AI Models in use in natural language search. This includes updating the AI Model, AI Model fixtures, and adding management commands that can modify the inference configs on the fly.

## Technical Details:
This branch does a few things:
1. Updates the AI Model to include inference configs as a field.
2. Creates a migration to add the field to the AI Model.
3. Updates the `ai_models.yaml` fixtures to include default inference configs.
4. Creates a set of management commands that allow devs to update the inference configs on stored AI Models on the fly (or reset them).
5. Updates the AI Assistant code to instantiate Converse with default inference configs if none are present on the model being called.
6. Adds broad integration test coverage to check how these changes might be evaluated.
7. Adds a markdown file documenting how to use the management commands (`managing_inference_configs.md`).
8. Updates some code formatting via ruff in the affected/changed files.

## Requirements for PR Merge:

9. [ ] API documentation updated (examples listed below)
    1. API Contracts
    2. API UI
    10. Comments
11. [ ] Appropriate Operations ticket(s) created
12. [x] Jira Ticket(s)
    1. [DEV-15929](https://federal-spending-transparency.atlassian.net/browse/DEV-15929)

[DEV-15929]: https://federal-spending-transparency.atlassian.net/browse/DEV-15929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/llm/assistants/filter_search.py
```diff
@@ -21,7 +20,7 @@ def __init__(
         tools: list[AITool],
         session: Session,
         system_message: str = (
-            "You are USASpending search assistant. " "Help the user select filters to search for federal spending"
+            "You are USASpending search assistant. Help the user select filters to search for federal spending"
```
sethstoudenmier:
```suggestion
            "You are USAspending search assistant. Help the user select filters to search for federal spending"
```
Super small thing I won't block over because it was like that before, but would be good to update that

usaspending_api/llm/management/commands/load_llm_fixtures.py
```diff
@@ -0,0 +1,15 @@
...
+    help = "Load LLM fixture data for AI models and prompts"
+
+    def handle(self, *args, **options) -> None:
+        self.stdout.write("Loading AI models...")
+        call_command("loaddata", "usaspending_api/llm/fixtures/ai_models.yaml")
+
+        self.stdout.write("Loading prompts...")
+        call_command("loaddata", "usaspending_api/llm/fixtures/prompts.yaml")
+
+        self.stdout.write(self.style.SUCCESS("Successfully loaded LLM fixtures"))
```
sethstoudenmier: Can you update the output to use the logging module (e.g., `logger = logging.getLogger(__name__)`)? You can see the pattern on some other jobs, but high level is that you can define as a global and then reference it below. There is some middleware that defines the format of the logger for consistency across the codebase.

If the decision is to update `self.stdout` to use `logger.info` then it would be good to update all of the new commands.

sethstoudenmier: Looking into the use of `self.stdout` more after writing this comment I am curious your experience with it.

sethstoudenmier (changes requested):
Marking as requested changes to use the logger instead of self.stdout in the management commands, but I am also open to changing this to approval. The LLM jobs are the only ones with `self.stdout` instead of the logger so wanted to check if there is a reason to make the change.

usaspending_api/llm/assistants/filter_search.py
gregrholden: I'll get this updated soon.

usaspending_api/llm/management/commands/load_llm_fixtures.py
gregrholden: I didn't write this management command. I just moved it into the commands sub-folder after creating my other management commands for the inference controls.

gregrholden: Went ahead and updated this (and my own management commands) to use logger after realizing I copied the structure from this one into my own 😄

usaspending_api/llm/models/db_models.py
```diff
@@ -5,6 +5,11 @@ class AIModel(models.Model):
     name = models.CharField(max_length=100)
     model_id = models.CharField(max_length=100)
     provider = models.CharField(max_length=100)
+    inference_config = models.JSONField(
+        default=dict,
+        blank=True,
+        help_text="Inference configuration (e.g., temperature, topP, maxTokens)",
+    )
```
zachflanders-frb, line 12: I'm not sure that it makes sense to make the inference config part of the ai_model model.  We can run different models with different inference configs and will probably want to adjust the inference config independently of the model.

usaspending_api/llm/assistants/filter_search.py
```diff
@@ -95,8 +94,29 @@ def tool_config(self) -> dict[str, list[dict]]:
...
+            Dictionary with inference parameters (temperature, topP, maxTokens, stopSequences).
+        """
+        if self.model.inference_config:
+            return self.model.inference_config
+
+        # Default configuration for deterministic output.
+        return {
+            "temperature": 0.0,
+            "topP": 1.0,
+            "maxTokens": 2048,
```
zachflanders-frb: 2048 tokens seems like it might be low.  The default is the maximum allowed for the model, which is 5k for the nova models.  The filter objects can get pretty long and I would hate for it to fail because it gets cut off.

zachflanders-frb (commented):
I really like the idea of having a django management command that allows us to make changes to the inference configs because this would allow us to update the assistant without having to do a redeploy.  However, I am not sure that it makes sense to have the inference as a property of the ai_model.  This PR makes me think we should introduce a new assistant model that includes the properties ai_model, system_prompt and inference_config along with a management command to update these on the fly to give us the ability to pull different levers and adjust the performance of the endpoint without having to do a deploy each time.

sethstoudenmier (dismissed):

usaspending_api/llm/models/db_models.py
gregrholden: Will be resolved in the next iteration.

sethstoudenmier (approved):

PR #4742, opened 2026-08-18
[DEV-15853]: Initial update to FABS loader mapping values for GSDM 1.2

gregrholden:
## Description:
DEV-15853: Update the FABS loader to map action types
First pass at mapping the action types for the FABS loader. This PR only adds the mappings, it does not handle the one-time backfill script that will be built out after this gets merged to the GSDM branch.
UPDATE (8/21): Added in a first pass at the backfill script, with some integration tests for mocking the load to/from delta workflow for the backfill script.

## Technical Details:
1. It's unclear if the `gsdm_maps.py` file is used in the project, but it contained a list of mappings, including those being updated. As such, it was updated to reflect the incoming changes.
2. The `transactions/delta_models/transaction_fabs.py` code contained the column data/mapping for the FABS loader being called by `etl/management/commands/load_transactions_in_delta.py`. The updates pass a scalar transformation over the input values for the action_type column so that those values are updated appropriately according to the new mappings for GSDM 1.2. The column info and normalized column info columns were updated with the new mappings.
3. UPDATED: The `etl/management/commands/backfill_fabs_action_types.py` file adds in SQL/Spark updates to run one-time commands that update the respective Postgres/Delta tables for the FABS data.
4. UPDATED: The `etl/tests/integration/test_backfill_fabs_action_types.py` file creates some mocked Spark tests imitates some of the other tests in that folder performing similar tests against Delta/Spark in order to check that the backfill script works correctly without local data to test it on.

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    6. Comments
5. [ ] Data validation completed (examples listed below)
8. [x] Jira Ticket(s)
    1. [DEV-15853](https://federal-spending-transparency.atlassian.net/browse/DEV-15853)

[DEV-15853]: https://federal-spending-transparency.atlassian.net/browse/DEV-15853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/usaspending_api/etl/management/commands/backfill_fabs_action_types.py
+++ b/usaspending_api/etl/management/commands/backfill_fabs_action_types.py
@@ -43,20 +43,6 @@ class Command(BaseCommand):
             WHEN action_type = 'C' THEN 'Other Action, Financial'
             WHEN action_type = 'D' THEN 'Other Action, Financial'
             WHEN action_type = 'E' THEN 'Mixed Aggregate'
-            WHEN action_type = 'A1' THEN 'New Award'
-            WHEN action_type = 'A2' THEN 'Renewal Award'
-            WHEN action_type = 'B1' THEN 'Continuation'
-            WHEN action_type = 'C1' THEN 'Termination Initiated: Material Failure to Comply'
-            WHEN action_type = 'C2' THEN 'Termination Initiated: Mutual Consent'
-            WHEN action_type = 'C3' THEN 'Termination Initiated: Recipient-Initiated'
-            WHEN action_type = 'C4' THEN (
-                'Termination Initiated: No Longer Effectuates Program Goals or Agency Priorities'
-            )
-            WHEN action_type = 'D1' THEN 'Closeout'
-            WHEN action_type = 'E1' THEN 'Recipient Change'
-            WHEN action_type = 'EX' THEN 'Other Action, Non-Financial'
-            WHEN action_type = 'FX' THEN 'Other Action, Financial'
-            WHEN action_type = 'G1' THEN 'Mixed Aggregate'
             ELSE action_type_description
         END
     """
@@ -84,18 +70,11 @@ class Command(BaseCommand):
                 int.transaction_normalized, rpt.transaction_search)
             """,
         )
-        parser.add_argument(
-            "--spark-master",
-            type=str,
-            default=None,
-            help="Spark master URL (e.g., spark://spark-master:7077). If not provided, uses local mode.",
-        )
 
     def handle(self, *args, **options) -> None:
         dry_run = options["dry_run"]
         postgres_only = options["postgres_only"]
         delta_only = options["delta_only"]
-        spark_master = options.get("spark_master")
 
         if postgres_only and delta_only:
             raise ValueError("Cannot specify both --postgres-only and --delta-only")
@@ -113,9 +92,7 @@ class Command(BaseCommand):
         # Update Delta tables.
         if not postgres_only:
             logger.info("Updating Delta tables...")
-            if spark_master:
-                logger.info(f"Connecting to Spark cluster at: {spark_master}")
-            self.update_delta_tables(dry_run, spark_master=spark_master)
+            self.update_delta_tables(dry_run)
 
         logger.info("FABS action type backfill complete!")
 
@@ -192,13 +169,11 @@ class Command(BaseCommand):
                 cursor.execute(self._generate_update_sql("source_assistance_transaction"))
                 logger.info(f"Updated {cursor.rowcount} records in source_assistance_transaction")
 
-    def update_delta_tables(self, dry_run: bool, spark_master: str = None) -> None:
+    def update_delta_tables(self, dry_run: bool) -> None:
         """Update action_type and action_type_description in Delta tables.
 
         Args:
-            dry_run: If True, only count records without making changes
-            spark_master: Spark master URL (e.g., spark://spark-master:7077).
-                         If None, runs in local mode (requires Java).
+            dry_run: If True, only count records without making changes.
         """
 
         # Initialize Spark session.
@@ -216,7 +191,7 @@ class Command(BaseCommand):
                 "spark.sql.jsonGenerator.ignoreNullFields": "false",  # keep nulls in our json
             }
             # Pass master parameter to configure_spark_session
-            self.spark = configure_spark_session(**extra_conf, master=spark_master, spark_context=self.spark)
+            self.spark = configure_spark_session(**extra_conf, spark_context=self.spark)
 
         try:
             # Update raw.published_fabs.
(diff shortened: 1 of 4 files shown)


usaspending_api/transactions/delta_models/transaction_fabs.py
```diff
@@ -2,8 +2,54 @@
...
 TRANSACTION_FABS_COLUMN_INFO = [
     TransactionColumn("action_date", "action_date", "STRING", "string_datetime_remove_timestamp"),
-    TransactionColumn("action_type", "action_type", "STRING"),
-    TransactionColumn("action_type_description", "action_type_description", "STRING"),
+    TransactionColumn(
+        "action_type",
+        "action_type",
+        "STRING",
+        scalar_transformation="""CASE
+            WHEN {input} IN ('0', '') OR {input} IS NULL THEN 'A1'
```
zachflanders-frb: > a NULL code under the Old column means that the Action Type is new and won’t need to be mapped

Does this mean that nulls should stay nulls?

gregrholden: That's a good question and I think it's confused by what's meant by "NULL code under the `Old` column". I read it as referring to the chart itself, meaning don't create a 1-to-1 mapping for those rows, which I haven't done. But a NULL or 0 `value` in the data itself is something else. This mapping only impacts new, incoming transactions and since there's a note saying to handle 0 and NULL I assumed, as they're new entries, they would get placed in A1. But definitely need input from @sethstoudenmier or the Broker team on this to clarify.

sethstoudenmier: Terrible wording on my part for the ticket and @gregrholden you are correct in your assumption that it refers to the table above in the ticket. This is referencing the table above that contains an "Old" column. For something such as "A2" there would be no need to perform a mapping because it is a "New" value. The use of `ELSE {input}` handles the intent of that bullet.

sethstoudenmier: >This mapping only impacts new, incoming transactions and since there's a note saying to handle 0 and NULL I assumed, as they're new entries, they would get placed in A1

This piece is not correct however; we would want only `A` to be mapped to `A1`. And to @zachflanders-frb point the null values should continue to be null via the `ELSE`.

sethstoudenmier: > NOTE: 0 and NULL should both be handle

This line in the AC is referring to how you handle the `EX` case.

usaspending_api/etl/management/commands/backfill_fabs_action_types.py
```diff
@@ -0,0 +1,301 @@
...
+                Only update Delta tables (raw.published_fabs, int.transaction_fabs,
+                int.transaction_normalized, rpt.transaction_search)
+            """,
+        )
+        parser.add_argument(
+            "--spark-master",
+            type=str,
+            default=None,
+            help="Spark master URL (e.g., spark://spark-master:7077). If not provided, uses local mode.",
+        )
```
sethstoudenmier: Did you need this for anything? With how we run Spark jobs we wouldn't exactly have this detail on hand. The Spark jobs should be able to run without specifying the master.

usaspending_api/etl/management/commands/backfill_fabs_action_types.py
```diff
@@ -0,0 +1,301 @@
...
+        if additional_where:
+            where_clause = f"{additional_where} AND ({where_clause})"
+
+        return f"""
+            UPDATE {table_name}
+            SET
+                action_type = {self.ACTION_TYPE_MAPPING},
+                action_type_description = {self.ACTION_TYPE_DESCRIPTION_MAPPING}
+            WHERE {where_clause}
+        """
```
sethstoudenmier, line 119: We will want to test this on `transaction_search` in QAT. The table is extremely large and may result in this taking a really long time. Another option that is more involved, but probably more efficient is to update `transaction_search` in Delta and then copy it down to Postgres, replacing the current Postgres table entirely.

usaspending_api/etl/tests/integration/test_backfill_fabs_action_types.py
```diff
@@ -0,0 +1,141 @@
...
+    load_delta_table_from_postgres("published_fabs", s3_unittest_data_bucket)
+
+    call_command(
+        "create_delta_table", f"--spark-s3-bucket={s3_unittest_data_bucket}", "--destination-table=transaction_fabs"
+    )
+    call_command(
+        "create_delta_table",
+        f"--spark-s3-bucket={s3_unittest_data_bucket}",
+        "--destination-table=transaction_normalized",
+    )
```
sethstoudenmier: Assuming this is here because the backfill script handles all Delta tables? Not going to block over it, but it would be nice to only call out to Spark if it is needed to reduce time spent on tests.

usaspending_api/transactions/delta_models/transaction_fabs.py
```diff
@@ -2,8 +2,52 @@
...
+            WHEN ucase(trim(action_type)) = 'C2' THEN 'Termination Initiated: Mutual Consent'
+            WHEN ucase(trim(action_type)) = 'C3' THEN 'Termination Initiated: Recipient-Initiated'
+            WHEN ucase(trim(action_type)) = 'C4' THEN (
+                'Termination Initiated: No Longer Effectuates Program Goals or Agency Priorities'
+            )
+            WHEN ucase(trim(action_type)) = 'D1' THEN 'Closeout'
+            WHEN ucase(trim(action_type)) = 'E1' THEN 'Recipient Change'
+            WHEN ucase(trim(action_type)) = 'EX' THEN 'Other Action, Non-Financial'
+            WHEN ucase(trim(action_type)) = 'FX' THEN 'Other Action, Financial'
+            WHEN ucase(trim(action_type)) = 'G1' THEN 'Mixed Aggregate'
```
sethstoudenmier: Similar to the mapping above, I would have it be either a mapping of an old description to a new or you simply use the new via the `else`. That way we don't need to make additional changes should a description change and we decided not to backfill.

sethstoudenmier (changes requested):

usaspending_api/etl/management/commands/backfill_fabs_action_types.py
gregrholden: No. Was playing with Spark options/configs locally and forgot to remove from the PR. Removing in next push.

usaspending_api/etl/management/commands/backfill_fabs_action_types.py
gregrholden: Not sure I understand this one. Should I remove the call to update this table in PG? Or do you mean to rewrite this update to be a Delta update instead?

usaspending_api/etl/tests/integration/test_backfill_fabs_action_types.py
gregrholden: Removing in next push.

usaspending_api/transactions/delta_models/transaction_fabs.py
gregrholden: Fixing in next push.

sethstoudenmier (approved):

PR #4736, opened 2026-08-13
[DEV-15246] - Adding pgvector migration

zachflanders-frb:
## Description:
This PR adds pgvector extension to our database via a django migration.

## Technical Details:
I added a dockerfile for our usaspending-db service to install pgvector

## Requirements for PR Merge:
4. [ ] Appropriate Operations ticket(s) created
5. [x] Jira Ticket(s)
    1. [DEV-15246](https://federal-spending-transparency.atlassian.net/browse/DEV-15246)

### Explain N/A in above checklist:

[DEV-15246]: https://federal-spending-transparency.atlassian.net/browse/DEV-15246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4743, opened 2026-08-18
[DEV-15246] - Add Embedding Mixin

zachflanders-frb:
## Description:
This is PR 2/3 for this ticket.  The [first PR](https://github.com/fedspendingtransparency/usaspending-api/pull/4736) included a migration and changes to the local db setup to support pgvector. This PR adds the EmbeddingMixin and updates the models to inherit from this Mixin.  It includes migrations for adding the embedding and embedding_generated_at fields to these models.  The third PR will include a django management command for populating these embeddings.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-15246](https://federal-spending-transparency.atlassian.net/browse/DEV-15246)

[DEV-15246]: https://federal-spending-transparency.atlassian.net/browse/DEV-15246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/common/mixins.py
```diff
@@ -283,3 +290,60 @@ def build_response(self, request: Request, *args, **kwargs) -> Any:
...
+            logger.debug(f"Embedding already exists for: {self.__class__.__name__} {self.pk}")
+        else:
+            text = self.get_embedding_text()
+
+            if not text or not text.strip():
+                logger.warning(f"Embedding text is empty for: {self.__class__.__name__} {self.pk}")
+            else:
+                try:
+                    generator = self.get_embedding_generator()
+                    self.embedding = generator.generate_embedding(text)
```
james-at-kc, line 321: The EmbeddingGenerator .generate_embedding() routine 
- catches errors and returns None,
- returns None if there is no "embedding" key in the response

In this case, the values will be:
- self.embedding = None, 
- self.embedding_generated_at = now,
- did_generate = True,

I think this needs to accommodate a return value of None from the EmbeddingGenerator .generate_embedding() function

james-at-kc (changes requested):

usaspending_api/common/mixins.py
zachflanders-frb: Good catch!

usaspending_api/common/mixins.py
```diff
@@ -283,3 +290,60 @@ def build_response(self, request: Request, *args, **kwargs) -> Any:
...
+
+    def save(
+        self,
+        auto_generate_embedding: bool = True,
+        force_generate_embedding: bool = False,
+        verbose: bool = False,
+        *args,
+        **kwargs,
+    ) -> None:
+        if auto_generate_embedding and not self.has_embedding or force_generate_embedding:
```
aguest-kc: For clarity can we add parentheses here to separate the checks?


```suggestion
        if (auto_generate_embedding and not self.has_embedding) or force_generate_embedding:
```

usaspending_api/common/mixins.py
zachflanders-frb: good catch!

james-at-kc (dismissed):
LGTM!

aguest-kc (approved):

PR #4746, opened 2026-08-18
[DEV-15246] - add generate embeddings command

zachflanders-frb:
## Description:
This is PR 3/3 for this feature. [The first PR](https://github.com/fedspendingtransparency/usaspending-api/pull/4736) added the pgvector extension to the app, [the second PR](https://github.com/fedspendingtransparency/usaspending-api/pull/4743) adds the EmbeddingMixin and the fields on the models that should have embeddings generated.  This PR adds the django management command to generate these embeddings. 

## Technical Details:
These embeddings will potentially take several hours to generate, but will only need to be generated once.  The embeddings are auto generated on object save or creation. The existing loaders (e.g. load_naics) will generate embeddings for new records created with get_or_create, but updated or bulk_update records bypass the save method and will need to be manually regenerated.  If the embedding fails the save will continue without an exception being raised, so it will still work in the legacy aws environment if bedrock is not available.

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-15246](https://federal-spending-transparency.atlassian.net/browse/DEV-15246)

[DEV-15246]: https://federal-spending-transparency.atlassian.net/browse/DEV-15246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/llm/management/commands/generate_embeddings.py
```diff
@@ -0,0 +1,205 @@
...
+                embedding_models.append(model)
+        return embedding_models
+
+    def process_model(
+        self,
+        model_class: Type[Model],
+        batch_size: int,
+        force: bool,
+        dry_run: bool,
+        limit: int = None,
```
aguest-kc: Can you add a type of `None` to this type hint since the default value is `None`?


```suggestion
        limit: int | None = None,
```

usaspending_api/llm/management/commands/generate_embeddings.py
```diff
@@ -0,0 +1,205 @@
...
+from django.apps import apps
+from django.core.management.base import BaseCommand, CommandError
+from django.db.models import Model, QuerySet
+
+from usaspending_api.common.mixins import EmbeddingMixin
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
```
aguest-kc, line 14: To be consistent with other management commands, can you add a `help` property to the command that explains what it's doing?

usaspending_api/llm/management/commands/generate_embeddings.py
```diff
@@ -0,0 +1,205 @@
...
+        self.stdout.write(f"Processing: {model_name} (dimensions: {dimensions})")
+
+        queryset = self._build_queryset(model_class, force, limit)
+        total = queryset.count()
+
+        if total == 0:
+            self.stdout.write(self.style.SUCCESS(f"No {model_name} records need embeddings"))
+            return 0, 0
+
+        pks_to_process = list(queryset.values_list("pk", flat=True))
```
aguest-kc: Not a big issue, but this block runs two queries and they could be consolidated into one query.


```suggestion
        pks_to_process = list(queryset.values_list("pk", flat=True))
        total = len(pks_to_process)

        if total == 0:
            self.stdout.write(self.style.SUCCESS(f"No {model_name} records need embeddings"))
            return 0, 0
```

aguest-kc (changes requested):
Overall looks good! I requested two changes and left one improvement suggestion.

aguest-kc (dismissed):

aguest-kc (approved):

PR #4754, opened 2026-08-27
Update ai_models.yaml to increase titan max tokens

zachflanders-frb:
## Description:
Amazon titan has a max tokens of 8192.  The embedding generator and the models'  `get_embedding_text` methods were built with this assumption.  5000 may truncate the text and result in worse performance.


gregrholden (approved):
The update is in line with the Anthropic model maxToken values. Looks good to me.

PR #4757, opened 2026-09-04
[DEV-16075]: Create Management Command for Assistant Table Records

gregrholden:
## Description:
[DEV-16075](https://federal-spending-transparency.atlassian.net/browse/DEV-16075) Create management command for assistant table records
This PR creates a management command for making updates to the newly created AI Assistant table records with a robust collection of options/flags available to it. It also updates a variety or related elements of the `llm` application to allow the management commands to function as intended. Comprehensive set of change notes are below under **Technical details**.

## Technical Details:
File-by-file changelog notes:
1. `llm/assistants/filter-search.py`: Updated how the assistant's inference configs get instantiated from the Assistant instead of the AI Model.
2. `llm/fixtures/ai_models.yaml` and `llm/fixtures/assistants.yaml`: both updated to reflect the changes to their respective models.
3. `llm/management/commands/manage_llm_assistant.py`: Creates a management command with a robust collection of options/flags, the details of which can be found in a new file in the `llm` application named `managing_llm_assistants.md`.
4. The AI Model and Assistant models were both updated (`llm/models/db_models.py`). Migrations were updated to reflect the incoming changes (`0003_assistant.py` and `0004_aimodel_unique_model_id.py`). The 0004 migration requires that the AIModel.model_id be unique. The old migration that added the inference config object to the AI Model was removed.
5. `llm/models/py_models.py`: Several tweaks were made to the Pydantic model for inference config validation.
6. Unit and Integration tests were updated to cover these changes and additions.
7. `llm/v2/views/filter_search.py`: Instantiate the assistant with the new Assistant table object instead of an object containing inference configs.

## Requirements for PR Merge:

8. [ ] API documentation updated (examples listed below)
    1. API Contracts
    2. API UI
    9. Comments
10. [ ] Appropriate Operations ticket(s) created
11. [x] Jira Ticket(s)
    1. [DEV-16075](https://federal-spending-transparency.atlassian.net/browse/DEV-16075)

[DEV-16075]: https://federal-spending-transparency.atlassian.net/browse/DEV-16075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-16075]: https://federal-spending-transparency.atlassian.net/browse/DEV-16075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+            ),
+        )
+        parser.add_argument(
+            "--inference-config-json",
+            type=str,
+            help=(
+                "Full inference config as JSON string "
+                '(e.g., \'{"temperature": 0.5, "topP": 0.8, "maxTokens": 5000, "stopSequences": []}\')'
+            ),
+        )
```
sethstoudenmier: Won't block on this, but given the number of arguments this job currently has it would be a good candidate for using argument groups (https://docs.python.org/3/library/argparse.html#argument-groups). For example, the rows highlighted in this comment would be part of an inference config group.

sethstoudenmier: Or a "helper" group that includes flags such as "--list" and "--list-with-prompts". I would leave up to you the groups that make sense.

usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+            "description",
+        )
+        flag_options = (
+            "create_new",
+            "clear_system_prompt",
+            "combine_prompts",
+            "clear_inference_config",
+            "is_active",
+            "is_inactive",
+        )
```
sethstoudenmier, line 184: Not a blocking comment, more just a thought. I go back and forth on failing if someone provides more variables vs logging that it will ignore all values when `--list` or `--list-with-prompts` is provided. My concern with this approach is that it creates a long list of values that we have to keep up to date; but ideally, we won't change this much.

usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+        if any(options.get(option) is not None for option in value_options) or any(
+            options.get(option) for option in flag_options
+        ):
+            raise CommandError("--list or --list-with-prompts must be used alone.")
+
+    @staticmethod
+    def _validate_options(options: dict) -> None:
+        if options.get("is_active") and options.get("is_inactive"):
```
sethstoudenmier: Anywhere you are using the `action="store_true"` or some default value I would update it to be a direct dictionary key instead of the `.get()` method (e.g., `options["is_active"]`). That way if someone makes a breaking change in the future it fails the lookup.

usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+            raise CommandError("Cannot specify both --is-active and --is-inactive.")
+        if options.get("combine_prompts") and not (
+            options.get("system_prompt_id") is not None or options.get("new_system_prompt") is not None
+        ):
```
sethstoudenmier: Can you remove the double negatives from the conditionals? It may just be me, but my brain took a while to visualize the double negatives.

`X & NOT (Y != None | Z != None)` ->`X & Y == None & Z == None`

I had to draw it out, but I believe the above should get what you are looking for with at least the first conditional that checks "combine_prompts".

usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+        try:
+            if model_id is not None:
+                model = AIModel.objects.get(model_id=model_id)
+            else:
+                model = AIModel.objects.get(name=model_name)
+            return model.pk
+        except AIModel.DoesNotExist:
+            raise CommandError(f"Model not found: {model_id or model_name}.") from None
+        except AIModel.MultipleObjectsReturned:
+            raise CommandError(f"Multiple models found for: {model_id or model_name}.") from None
```
sethstoudenmier:
```suggestion
        except AIModel.DoesNotExist as exc:
            raise CommandError(f"Model not found: {model_id or model_name}.") from exc
        except AIModel.MultipleObjectsReturned as exc:
            raise CommandError(f"Multiple models found for: {model_id or model_name}.") from exc
```

usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+            active_state = self._update_active_state(options)
+        else:
+            active_state = False
+        assistant = Assistant.objects.create(
+            name=options.get("name"),
+            ai_model_id=model_pk,
+            system_prompt_id=None,
+            inference_config={},
+            is_active=active_state,
+            description=options.get("description") or "",
```
sethstoudenmier:
```suggestion
            description=options.get("description", ""),
```

usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+        new_prompt: str | None,
+        new_prompt_name: str | None,
+    ) -> int:
+        """Handles combining system prompts as specified by command options."""
+        # Get the text of the prompt currently in use.
+        if current_prompt_pk > 0:
+            try:
+                current_prompt = Prompts.objects.get(pk=current_prompt_pk)
+            except Prompts.DoesNotExist:
+                raise CommandError(f"Current system prompt with pk '{current_prompt_pk}' not found.") from None
```
sethstoudenmier:
```suggestion
            except Prompts.DoesNotExist as exc:
                raise CommandError(f"Current system prompt with pk '{current_prompt_pk}' not found.") from exc
```

usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+        else:
+            current_prompt_text = ""
+        # Get the text of the new prompt being created via flag.
+        new_prompt_text = new_prompt or ""
+        # Combine the prompts.
+        if existing_prompt_pk is not None:
+            try:
+                existing_prompt = Prompts.objects.get(pk=existing_prompt_pk)
+            except Prompts.DoesNotExist:
+                raise CommandError(f"System prompt with pk '{existing_prompt_pk}' not found.") from None
```
sethstoudenmier:
```suggestion
            except Prompts.DoesNotExist as exc:
                raise CommandError(f"System prompt with pk '{existing_prompt_pk}' not found.") from exc
```

usaspending_api/llm/management/commands/manage_llm_assistant.py
```diff
@@ -0,0 +1,597 @@
...
+        # Craft the new Prompt to be created in the DB.
+        try:
+            with transaction.atomic():
+                prompt = Prompts.objects.create(
+                    name=prompt_name,
+                    description="Custom prompt created by management command: 'manage_llm_assistant'",
+                    text="\n".join(filtered_text),
+                )
+        except IntegrityError:
+            raise CommandError(f"A prompt named '{prompt_name}' already exists.") from None
```
sethstoudenmier:
```suggestion
        except IntegrityError as exc:
            raise CommandError(f"A prompt named '{prompt_name}' already exists.") from exc
```

usaspending_api/llm/migrations/0003_aimodel_inference_config.py
sethstoudenmier, line 1: Don't delete a migration file. Instead, in the next migration you need to remove the `inference_config` field created by this migration. Once a migration has been pushed up / deployed to production we only delete them for squashing migrations.

usaspending_api/llm/migrations/0004_assistant.py
sethstoudenmier, line 1: Same concept as the above migration, you will need to leave this migration as-is and instead create a new migration that builds on it.

usaspending_api/llm/migrations/0006_aimodel_unique_model_id.py
sethstoudenmier, line 1: This will need to capture in a new 0005 migration that includes additional changes to accommodate changes requested in the other migration files.

usaspending_api/llm/models/py_models.py
```diff
@@ -4,12 +4,27 @@
...
+    topP: float | None = Field(
+        default=1.0, ge=0.0, le=1.0, description="Top P sampling (0.0-1.0). Set to null to use model default."
+    )
+    maxTokens: int | None = Field(
+        default=5000, gt=0, description="Maximum tokens to generate. Set to null to use model default."
+    )
+    stopSequences: list[str] | None = Field(
+        default_factory=list,
+        description="Stop sequences ([] for none, null for model default).",
+    )
```
sethstoudenmier, line 27: Won't block on it because it is outside of the scope of your work, but if you have time it would be great to update this to include an `alias_generator` (https://pydantic.dev/docs/validation/latest/concepts/alias/#using-an-aliasgenerator) or field alias (https://pydantic.dev/docs/validation/latest/concepts/fields/#field-aliases) so that we can keep the variable itself as snakecase.

sethstoudenmier (changes requested):
Overall it is looking really good! Not all comments are blocking; I tried to go through and mention whether or not they are blocking. Some of the comments are just cleanup.

usaspending_api/llm/assistants/filter_search.py
```diff
@@ -23,12 +23,14 @@ def __init__(
         system_message: str = (
             "You are USAspending search assistant. Help the user select filters to search for federal spending"
         ),
+        inference_config: dict | None = None,
```
zachflanders-frb: Can we replace mode, system_message and inference_config with assistant?
```suggestion
assistant: Assistant
tools: list[AITool],
session: Session,
```

usaspending_api/llm/assistants/filter_search.py
```diff
@@ -23,12 +23,14 @@ def __init__(
...
     ) -> None:
         self.model = model
```
zachflanders-frb:
```suggestion
        self.model = assistant.ai_model
```

usaspending_api/llm/assistants/filter_search.py
```diff
@@ -23,12 +23,14 @@ def __init__(
...
         self.tools = tools
         self.tools_by_name = {tool.description.name: tool for tool in tools}
         self.session = session
         self.system_message = system_message
+        self._inference_config = inference_config
```
zachflanders-frb:
```suggestion
        self.system_message = assistant.system_prompt
        self._inference_config = assistant.inference_config
```

zachflanders-frb (changes requested):
Overall look good. I think we can reduce the number of parameters for the filter search assistant class by using the assistant model instead of individual parameters for model, prompt, and inference config.

usaspending_api/llm/management/commands/manage_llm_assistant.py
gregrholden: I grouped each flag into relevant labels.

usaspending_api/llm/management/commands/manage_llm_assistant.py
gregrholden: Done.

usaspending_api/llm/management/commands/manage_llm_assistant.py
gregrholden: I had to work through this myself a bit. Should have resolved all of the overly negative conditionals into more positive versions, using helper variables to make readability a bit easier.

usaspending_api/llm/management/commands/manage_llm_assistant.py
gregrholden: This played a bit of havoc with my setup. The description field gets tested in a few ways and teasing out whether null or empty is allowed required a few iterations. Should be resolved now with this fix and updated configs.

usaspending_api/llm/migrations/0003_aimodel_inference_config.py
gregrholden: Migrations re-added and new ones incremented as described.

usaspending_api/llm/models/py_models.py
gregrholden: I can look into this for a future iteration. Thanks!

usaspending_api/llm/assistants/filter_search.py
gregrholden: Updated the assistant to actually use the assistant. Good catch.

zachflanders-frb (approved):
👍

sethstoudenmier (approved):

PR #4758, opened 2026-09-08
[DEV-15169] - Add Lookup Code Tool

zachflanders-frb:
## Description:
This PR adds the code lookup tool.  It also adds an expand query tool for use by this tool and other tools in the future.

## Technical Details:
 - The tool first looks for an exact match or a prefix match on the code field.  
 - If it does not find a match it will proceed to a hybrid text similarity search using trigram and semantic search using text embeddings.  
 - It will use query fan out to produce 3 additional query terms.
     - the expand query tool is called by the tool in order to produce the additional terms. 
 - It constructs a hierarchy for nested codes so that the LLM can understand the relationships  

## Requirements for PR Merge:

5. [x] Jira Ticket(s)
    1. [DEV-15169](https://federal-spending-transparency.atlassian.net/browse/DEV-15169)

[DEV-15169]: https://federal-spending-transparency.atlassian.net/browse/DEV-15169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/llm/tests/integration/test_lookup_code.py
```diff
@@ -0,0 +1,412 @@
...
+    CodeTypeConfig,
+    SearchResultNode,
+    lookup_code_tool,
+)
+from usaspending_api.references.models import ToptierAgency
+from usaspending_api.references.models.cfda import Cfda
+from usaspending_api.references.models.naics import NAICS
+from usaspending_api.references.models.psc import PSC
+
+pytestmark = pytest.mark.django_db
```
sethstoudenmier, line 20: Didn't know that existed and wishing I had

usaspending_api/llm/tools/expand_query.py
```diff
@@ -0,0 +1,95 @@
...
+
+from usaspending_api.llm.models.db_models import AIModel
+
+logger = logging.getLogger(__name__)
+
+
+def expand_query(query: str, num_variations: int = 3, model: AIModel = None) -> list[str]:
+    """Generate related search queries using Amazon Bedrock converse API with tool"""
+    if model is None:
+        model = AIModel.objects.get(name="nova micro")
```
sethstoudenmier: To avoid accidently using the wrong model I am wondering if we should make the model required.

sethstoudenmier (dismissed):
LGTM! One small comment around if the model should be required or not but not going to block on it.

usaspending_api/llm/tests/integration/test_lookup_code.py
zachflanders-frb: you can thank gitlab duo.

sethstoudenmier (dismissed):

sethstoudenmier (approved):

PR #4765, opened 2026-09-10
Bug/dev 15666 fix submission reveal date query logic

DavidMikolaKC:
## Description:

The SQL queries in the files listed below return funding results, but these results are not filtered to be before the submission_reveal_date. This logic should be updated to not included submissions that have not been revealed yet.

## Technical Details:

Implemented join logic to filter out entries when it has not yet reached their submission reveal date where relevant.

## Requirements for PR Merge:

2. [ ] Jira Ticket(s)
    1. [DEV-15666](https://federal-spending-transparency.atlassian.net/browse/DEV-15666)

[DEV-15666]: https://federal-spending-transparency.atlassian.net/browse/DEV-15666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


usaspending_api/awards/v2/views/count/federal_accounts_count.py
```diff
@@ -36,9 +37,19 @@ def _business_logic(self, request_data: dict) -> list:
...
+            FinancialAccountsByAwards.objects.annotate(
+                reveal_date=Subquery(
+                    SubmissionAttributes.objects.filter(submission_id=OuterRef("submission_id")).values(
+                        "submission_window__submission_reveal_date"
+                    )
+                )
+            )
+            .filter(award_id=award.id, reveal_date__lte=now())
+            .count()
+        )
```
sethstoudenmier: Any reason to use a subquery here instead of a join?

usaspending_api/awards/v2/views/accounts.py
```diff
@@ -54,6 +55,11 @@
             fa.id = taa.federal_account_id
         left outer join agency a on a.id = gfaba.funding_agency_id
         left outer join toptier_agency ta on ta.toptier_agency_id = a.toptier_agency_id
+        inner join submission_attributes sa on
+            gfaba.submission_id = sa.submission_id
+        inner join dabs_submission_window_schedule dabs on
+            sa.submission_window_id = dabs.id and
+            dabs.submission_reveal_date <= now()
```
sethstoudenmier, line 62: Won't block on this because it isn't always the case, but in a lot of cases using an `EXISTS( ... )` statement in the `WHERE` can prove to be more performant than adding to the `JOIN` in the case of a query like this where the contents of the `JOIN` isn't used in the select and is only used to determine what records need to be referenced.

usaspending_api/submissions/helpers.py
```diff
@@ -268,26 +269,30 @@ def _get_latest_submission_ids_for_each_fiscal_quarter(
 
 
 def get_latest_submission_ids_for_each_fiscal_quarter_file_a(
-    fiscal_years: List[int] = [], federal_account_id: int = None
-):
+    fiscal_years: List[int] = None, federal_account_id: int = None
```
sethstoudenmier:
```suggestion
    fiscal_years: list[int] | None = None, federal_account_id: int | None = None
```

usaspending_api/submissions/helpers.py
```diff
@@ -268,26 +269,30 @@ def _get_latest_submission_ids_for_each_fiscal_quarter(
...
+        fiscal_years = []
     return _get_latest_submission_ids_for_each_fiscal_quarter(
         "appropriationaccountbalances", fiscal_years, federal_account_id
     )
 
 
 def get_latest_submission_ids_for_each_fiscal_quarter_file_b(
-    fiscal_years: List[int] = [], federal_account_id: int = None
-):
+    fiscal_years: List[int] = None, federal_account_id: int = None
```
sethstoudenmier:
```suggestion
    fiscal_years: list[int] | None = None, federal_account_id: int | None = None
```

sethstoudenmier (changes requested):
Left a question about subquery usage and then some comments for cleanup. The comment about JOIN vs EXISTS is more a thought and I won't block on it.

usaspending_api/awards/v2/views/count/federal_accounts_count.py
DavidMikolaKC: No.  Updated to just use the filter call on FinancialAccountsByAwards instead

sethstoudenmier (approved):

PR #4766, opened 2026-09-11
Ftr/dev 16050 subaward ratio

james-at-kc:
## Description:

The /api/v2/search/spending_by_subaward_grouped endpoint accepts some filters and returns a list of prime awards and some data on the aggregate of the subawards underneath. Currently the endpoint response resembles the following:

 {
   "limit": ...,
   "page_metadata": ...,
   "messages": ...,
   "results": [
     {
       "award_id": ...,
       "subaward_count": ...,
       "subaward_obligation": ...,
       "award_generated_internal_id": ...
     },
     ...
   ]
 }
To support further work with the endpoint we want to add a new field that captures the ratio of subaward amount to prime award amount.

## Technical Details:
This may involve adding the ratio into the OpenSearch index depending on how something such as a bucket sort on the ratio of two values may work. If this needs to be added, we will need to add a new _sort field as well to account for the floating point precision issue.

Acceptance Criteria
The results list should contain a new field called subaward_to_award_ratio that captures subaward_obligation / <OBLIGATIONS>

<OBLIGATION> refers to  total_subsidy_cost for loans and total_obligation for all other types

subaward_obligation refers to the value currently returned in the response

The results list should contain a new field called award_obligation that captures total_subsidy_cost for loans and total_obligation for all other types

Pagination should be supported for both the new fields (subaward_to_award_ratio and award_obligation ), falling back to the award_id as a default in the case of two values matching

The API contract should be updated to include these new fields

## Requirements for PR Merge:

    1. API Contracts
    2. API UI
    3. Comments
5. [ ] Jira Ticket(s)


usaspending_api/api_docs/markdown/endpoints.md
sethstoudenmier, line 1: Did line endings get updated in this file? Seeing a lot more in the changeset than I would have thought.

usaspending_api/etl/es_award_template.json
```diff
@@ -847,6 +847,10 @@
         "type": "scaled_float",
         "scaling_factor": 100
       },
+      "subaward_to_award_ratio": {
+        "type": "scaled_float",
+        "scaling_factor": 100000000
+      },
```
sethstoudenmier, line 853: Did you find that you needed this logic for subawards that are much smaller than the award amount?

sethstoudenmier (changes requested):
I left some questions to just make sure of some logic before approving. Also, would like to have some tests added / updated to include the new field.

usaspending_api/api_docs/markdown/endpoints.md
james-at-kc: I disavow this monstrosity. 

I made a change to one line: 198

I committed.

Either the linter or some other process alters tabs / spaces / line endings / whitespace following table rows.

It has happened previously when I changed this or similar .md files

I do not know what to tell you. If needed, I can try altering the file using a different editor: notepad, np++, vim...

usaspending_api/etl/es_award_template.json
james-at-kc: Yes, I did. 

For an award of tens of millions of dollars, a subaward of $1000 would go to zero if we scaled to 100 (like other $ amounts in this file). 

The API does not use this scaled_float; pagination uses the _sort long. We can switch this mapping to a normal float with no API impact.

sethstoudenmier: If you saw no issues with using a regular float, then I would recommend moving away from this larger scaled float and instead using a regular float. In this case, the ratio isn't a dollar amount and doesn't need to be scaled by 100.

usaspending_api/database_scripts/etl/award_delta_view.sql
```diff
@@ -147,6 +152,12 @@ SELECT
...
   CAST("total_iija_outlay" * 100 AS BIGINT) AS total_iija_outlay_sort,
   CAST("generated_pragmatic_obligation" * 100 AS BIGINT) AS generated_pragmatic_obligation_sort,
-  CAST("total_subaward_amount" * 100 AS BIGINT) AS total_subaward_amount_sort
+  CAST("total_subaward_amount" * 100 AS BIGINT) AS total_subaward_amount_sort,
+  /*
+   Ratio sort uses a larger scale than dollar amounts so close floating-point ratios stay distinguishable.
+   */
+  CAST(
+    COALESCE("total_subaward_amount", 0) / NULLIF("award_amount", 0) * 100000000 AS BIGINT
+  ) as subaward_to_award_ratio_sort
```
sethstoudenmier: If no longer relying on scaled_float above we should be good to remove the _sort field here. This is intended as a workaround for sorting on a scaled float.

usaspending_api/etl/es_award_template.json
james-at-kc: done!

sethstoudenmier (approved):

usaspending_api/database_scripts/etl/award_delta_view.sql
sethstoudenmier: Didn't mean to keep this comment in the review; closing it as a result.

